PR #34, opened 2025-02-17
 Simplify launch arguments

jonathanKingston:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208825639115711
Tech Design URL:
CC:

### Description

Simplifies launch arguments logic so user preferences can be used instead of environment variables.

### Testing Steps

1. Ensure you can skip dialogs by enabling the environment variable
2. Ensure dialogs are present when they're not set.

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


iOS/DuckDuckGoTests/LaunchOptionsHandlerTests.swift
```diff
@@ -35,37 +35,36 @@ final class LaunchOptionsHandlerTests: XCTestCase {
         try super.tearDownWithError()
     }
 
-    // MARK: - isUITesting
+    // MARK: - isOnboardingCompleted
 
-    func testShouldReturnTrueWhenIsUITestingIsCalledAndLaunchArgumentContainsIsUITesting() {
+    func testShouldReturnTrueWhenIsOnboardingCompletedAndEnvironmentsOnboardingIsTrue() {
```
alessandroboron:
```suggestion
    func testShouldReturnTrueWhenIsOnboardingCompletedAndEnvironmentsOnboardingIsFalse() {
```

ONBOARDING var in this case should be false for sut.isOnboardingCompleted to be true

iOS/DuckDuckGoTests/LaunchOptionsHandlerTests.swift
```diff
@@ -35,37 +35,36 @@ final class LaunchOptionsHandlerTests: XCTestCase {
...
         // WHEN
-        let result = sut.isUITesting
+        let result = sut.isOnboardingCompleted
 
         // THEN
         XCTAssertTrue(result)
     }
 
-    func testShouldReturnFalseWhenIsUITestingIsCalledAndLaunchArgumentsDoesNotContainIsUITesting() {
+    func testShouldReturnFalseWhenIsOnboardingCompletedAndEnvironmentsOnboardingIsFalse() {
```
alessandroboron:
```suggestion
    func testShouldReturnFalseWhenIsOnboardingCompletedAndEnvironmentsOnboardingIsTrue() {
```

We should flip the “ONBOARDING”. similar to the above test

iOS/DuckDuckGoTests/DaxDialogTests.swift
```diff
@@ -80,6 +80,18 @@ final class DaxDialog: XCTestCase {
         XCTAssertEqual(onboarding.nextHomeScreenMessageNew(), .addFavorite)
         XCTAssertTrue(onboarding.isAddFavoriteFlow)
     }
+    
+    func testWhenLaunchOptionsHandlerSkipsOnboardingThenDialogsAreNotEnabled() {
+        let launchOptionsHandler = LaunchOptionsHandler()
```
alessandroboron, line 87: Test is failing because settings.isDismissed in onboarding.isEnabled returns false

We need to control that dependency so we should update the test with the below:

```
func testWhenLaunchOptionsHandlerSkipsOnboardingThenDialogsAreNotEnabled() {
        let settings = MockDaxDialogsSettings()
        settings.isDismissed = false
        let launchOptionsHandlerOnboardingEnabled = LaunchOptionsHandler(environment: ["ONBOARDING": "true"])
        var onboarding = DaxDialogs(settings: settings,
                                    entityProviding: MockEntityProvider(),
                                    launchOptionsHandler: launchOptionsHandlerOnboardingEnabled)
        XCTAssertTrue(onboarding.isEnabled)

        let launchOptionsHandlerOnboardingDisabled = LaunchOptionsHandler(environment: ["ONBOARDING": "false"])
        onboarding = DaxDialogs(settings: settings,
                                entityProviding: MockEntityProvider(),
                                launchOptionsHandler: launchOptionsHandlerOnboardingDisabled)
        XCTAssertFalse(onboarding.isEnabled)
    }
```

jonathanKingston:
@alessandroboron agreed if we want the force 'on' then tri-state enum makes the most sense. I think it's a little unfortunate to have that go everywhere but I'd like to get my other work unblocked that needs this change.

I am however a little concerned that the .maestro files needed to add isOnboardingCompleted or run  `onboarding.yaml`

jonathanKingston:
> I am however a little concerned that the .maestro files needed to add isOnboardingCompleted or run `onboarding.yaml`

We've dropped the runflow in setup.yml that caused that. Should we just add that back and drop the other changes?

alessandroboron (approved):

alessandroboron:
> @alessandroboron agreed if we want the force 'on' then tri-state enum makes the most sense. I think it's a little unfortunate to have that go everywhere but I'd like to get my other work unblocked that needs this change.
> 
> I am however a little concerned that the .maestro files needed to add isOnboardingCompleted or run `onboarding.yaml`

As mentioned in MM there are tests that kill and relaunch the App again. In that scenario the onboarding is presented again because the launch argument is not passed to the main application. This causes the tests to fail. The solution is to pass ${ONBOARDING_COMPLETED} in these tests.

BEFORE:
```
- stopApp
- launchApp
```

AFTER:
```
- stopApp
- launchApp:
    arguments:
      isOnboardingCompleted: ${ONBOARDING_COMPLETED}
```

My first thought was to replace `- launchApp` with 

```
runWorkflow:
   file: ../shared/setup.yaml
```

But that workflow reset the state of the app. If the test had a webview open calling setup.yaml would clear everything.

jonathanKingston:
Took me a bit of time to figure out why we'd removed the run flow at the end of setup.yml but in summary it means we can rely on `isOnboardingCompleted: true` to do the work for us. The onboarding tests are the inverse so they needed that added back in. Being more explicit what is happening is the result of this change.

Thanks for changing those bits again whilst I was out :)

PR #70, opened 2025-02-20
cleanup debug screen

brindy:
Task/Issue URL: https://app.asana.com/0/42792087274227/1209261804076504
Tech Design URL:
CC:

### Description
Clean up the debug menu a bit.

Leaves in place the old debug menu accessible via "legacy" item.

Allows you to pin debug screens/actions for quick access. 

Allows the menu to be filtered by typing text.

### Testing Steps
1. Access the debug menu (via settings -> All debug options)
2. Quick check each page renders
3. Check the legacy menu renders 
4. Validate on iPad and iPhone
5. Check pinning works - pin an item or two and restart the app, after accessing settings should still be pinned
6. Check filtering for actions / views works - swipe down to show the filter field and type

### Impact and Risks
None: internal tooling only

#### What could go wrong?
* Users might accidentally get access to this menu
* Useful parts of the menu stop working

### Quality Considerations
* Check for memory leaks

### Notes to Reviewer
n/a


iOS/DuckDuckGo/DebugScreensViewModel+Screens.swift
```diff
@@ -0,0 +1,187 @@
...
+        return [
+            // MARK: Actions
+            .action(title: "Reset Autoconsent Prompt", { _ in
+                AppUserDefaults().clearAutoconsentUserSetting()
+            }),
+            .action(title: "Reset Sync Promos", { d in
+                let syncPromoPresenter = SyncPromoManager(syncService: d.syncService)
+                syncPromoPresenter.resetPromos()
+            }),
+            .action(title: "Reset TipTik", { d in
```
samsymons:
```suggestion
            .action(title: "Reset TipKit", { d in
```

iOS/DuckDuckGo/RootDebugViewController.swift
```diff
@@ -93,6 +94,17 @@ class RootDebugViewController: UITableViewController {
...
     }
 
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // This might be annoying but I want to flush out any missing functionality before removing it.
+        let controller = UIAlertController(title: "Something missing?", message: "Something you need in here that isn't in the other screen?  Let us know by dropping a task in the iOS App Development project.", preferredStyle: .actionSheet)
+        controller.addAction(UIAlertAction(title: "OK!", style: .default))
+
+        present(controller: controller, fromView: self.view)
```
samsymons, line 104: That's a really great touch, nice thinking 👏

iOS/DuckDuckGo/DebugScreensViewModel+Screens.swift
```diff
@@ -0,0 +1,187 @@
...
+                                            bookmarksDatabase: d.bookmarksDatabase)
+                }
+            }),
+            .controller(title: "Configuration Refresh Info", { _ in
+                let storyboard = UIStoryboard(name: "Debug", bundle: nil)
+                return storyboard.instantiateViewController(identifier: "ConfigurationDebugViewController") { coder in
+                    ConfigurationDebugViewController(coder: coder)
+                }
+            }),
+            .controller(title: "NetP", { _ in
```
samsymons: We could rename this to `VPN` now while we're making changes, but definitely not a blocker.

samsymons (approved):
LGTM from a run through of each of the debug screens, nothing seems obviously missing/broken.

The Keychain debug menu screen seems to be a bit too inset for whatever reason, so its top header gets clipped a little bit (screenshot attached showing as far upwards as it'll scroll, and the header is still getting truncated).

Otherwise though, everything looks great!

Simulator Screenshot - iPhone 16 Pro - 2025-02-23 at 13 52 53

iOS/DuckDuckGo/DebugScreensViewModel+Screens.swift
brindy: Thanks! Would have used your suggestion but the GitHub UI is broke and I couldn't see the button.

PR #73, opened 2025-02-21
make allowList thread safe

SabrinaTardio:
Task/Issue URL: https://app.asana.com/0/1204186595873227/1209454436197381

### Testing Steps

1. Ci is green

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


BrowserServicesKit/Sources/BrowserServicesKit/PropertyWrappers/Locked.swift
```diff
@@ -0,0 +1,42 @@
...
+            defer { lock.unlock() }
+            return value
+        }
+        set {
+            lock.lock()
+            value = newValue
+            lock.unlock()
+        }
+    }
+}
```
bwaresiak, line 53: While this pattern works in the scenario you are fixing, I'm a bit hesitant to actually add it. :) Is there a guarantee that everyone using this ill understand the possible issues that arise by mutating the property value, be it collection  - with append, or scalar with `+=` ? :) What do you think?

SabrinaTardio: Just to clarify… 
are you concerned that people using the Locked class for other things might not understand its limitations?

I see your point. We could add documentation for it to make it clearer… I like that it does save some boilerplate code and makes sure that the lock is as close to the change as possible
Otherwise I am happy to move the lock into the AdClickAttributionFeature class directly

bwaresiak: I think documenting the dangers of mutating underlying value or ensuring state consistency would be a step into the right direction. Better to add it now with correct assumptions than to risk someone adding it without these :)

SabrinaTardio: @bwaresiak I have added some documentation.
However if you still think it’s not a good idea I am happy to change approach.

bwaresiak (approved):
LGTM, good job @SabrinaTardio :)

PR #90, opened 2025-02-24
Post app delegate work - cleanup of and old code and make auto clear visually smoother

jaceklyp:
Task/Issue URL: https://app.asana.com/0/0/1208832732122405
CC: @dus7 

### Description
1. Remove old migration code
2. Make autoclearing smoother, by removing overlay if there is no work to be done

### Testing Steps
1. Enable Data Clearing in Settings after 5 minutes
2. Change the code inside AutoClear 
```
        case .delay5min:
            return elapsedTime > 5 * 60
```
to:
```
        case .delay5min:
            return elapsedTime > 10
```
3. Open any site and go to background for at least 10 seconds.
4. In App Switcher notice you cannot see app content, because of BlankSnapshotVC
5. Reopen app and notice the content is gone.
6. Open any site again and go to background for less than 10 secs (so autoclearing doesn't happen)
7. Again, in App Switcher notice you cannot see app content, because of BlankSnapshotVC
8. Reopen app and notice the content loads almost immediately (there is no glitch, with some intermediary screen blinking)

### Impact and Risks
Medium

#### What could go wrong?
1/ Make sure the autoclearing still works (especially the one that doesn't require app restart)


iOS/DuckDuckGo/AutoClear.swift
```diff
@@ -102,17 +103,18 @@ final class AutoClear: AutoClearing {
...
-            let timestamp = timestamp,
-            shouldClearData(elapsedTime: baseTimeInterval - timestamp) else { return }
+    var shouldClearDataIfTimeExpired: Bool {
+        guard isClearingEnabled, let timestamp = timestamp else { return false }
+        return shouldClearData(elapsedTime: Date().timeIntervalSince1970 - timestamp)
+    }
 
-        self.timestamp = nil
+    @MainActor
+    func clearDataDueToTimeExpired(applicationState: DataStoreWarmup.ApplicationState) async {
```
bwaresiak, line 112: I'm not sure I like this - the logic of using timestamp was encapsulated in previous implementation, now it leaks outside of the class. It is also possible to start clearing at wrong time. I'm thinking of leaving old `clearDataIfEnabledAndTimeExpired` as is, and just have this new method to check if time has expired.

dus7 (approved):

PR #94, opened 2025-02-25
AuthV2 / iOS app Auth V1 + V2

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1209170372758738/f
CC: @miasma13 

### Description

This PR implements the use of AuthV2 in the IOS app.

The good:
-  AuthV2 is fully integrated into the iOS app.
- The authentication still uses V1 but V2 is configured and added behind an informal feature flag. (see `isAuthV2Enabled`)
- NetworkProtection doesn't import Subscription anymore and instead uses shared protocols like `BSK/Common/SubscriptionTokenHandling.swift`

The bad
- We now have some code duplication, especially in the settings UI, all will be removed alongside AuthV1
- I introduced "bridge" protocols (see `SubscriptionAuthV1toV2Bridge`) for aligning AuthV1 and AuthV2 behaviours, all will be removed alongside AuthV1
- The code is not very clean, I prioritised risk reduction and changes minimisation over code cleanness, all will be improved when we remove AuthV1
- There are some ugly force unwrap for optional vars that shouldn't really be optional like `let subscriptionManagerV1: (any SubscriptionManager)?` and `let subscriptionManagerV2: (any SubscriptionManagerV2)?`, all will be removed with AuthV1 but for now they are needed for allowing the coexistence of Auth v1 and v2

### Testing Steps
1. Subscription should work as usual, no feature changes are expected.

### Impact and Risks

High impact, we can break Subscription + VPN

---
###### Internal references:


iOS/DuckDuckGo/AppLifecycle/AppStates/Foreground.swift
```diff
@@ -176,10 +176,6 @@ extension Foreground {
 
         let appDependencies: AppDependencies
 
-        init(appDependencies: AppDependencies) {
```
federicocappelli, line 177: Removed by SwiftLint --fix

iOS/DuckDuckGo/NetworkProtectionTunnelController.swift
```diff
@@ -285,11 +284,14 @@ final class NetworkProtectionTunnelController: TunnelController, TunnelSessionPr
         }
 
         options["activationAttemptId"] = UUID().uuidString as NSString
+
         do {
-            options["authToken"] = try tokenStore.fetchToken() as NSString?
+            let token =  try await tokenHandler.getToken()
+            options["authToken"] = NSString(string: token)
```
miasma13, line 284: Why is NSString required here?

BrowserServicesKit/Sources/Subscription/SubscriptionAuthenticationState.swift
```diff
@@ -0,0 +1,23 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public protocol SubscriptionAuthenticationStateProvider {
```
miasma13: Filename and header are not matching protocol name.

iOS/DuckDuckGo/VPNRedditSessionWorkaround.swift
```diff
@@ -0,0 +1,114 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import NetworkProtection
+import Subscription
+import WebKit
+import Core
+
+final class VPNRedditSessionWorkaround {
```
miasma13: Something seems off. There is no usage of reddit workaround in the code anywhere and also the file is not even in the project.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -332,3 +333,302 @@ final class SubscriptionSettingsViewModel: ObservableObject {
...
+            async let reloadedEmail = await self.fetchAndUpdateAccountEmail(cachePolicy: .reloadIgnoringLocalCacheData,
+                                                                            loadingIndicator: !hasLoadedEmailFromCache)
+            async let reloadedSubscription = await self.fetchAndUpdateSubscriptionDetails(cachePolicy: .reloadIgnoringLocalCacheData,
+                                                                                          loadingIndicator: !hasLoadedSubscriptionFromCache)
+            let (hasReloadedEmail, hasReloadedSubscription) = await (reloadedEmail, reloadedSubscription)
+
+            // In case any fetch fails show an error
+            if !hasReloadedEmail || !hasReloadedSubscription {
+                self.showConnectionError(true)
+            }
```
miasma13: This is a regression, from the old implementation this part was removed. See: https://github.com/duckduckgo/iOS/pull/3893/files#diff-864fea16a7bc9aa536aacbc970ced652ec00f6e0368a77a59f95b207ed2a0df8

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -332,3 +333,302 @@ final class SubscriptionSettingsViewModel: ObservableObject {
         signOutObserver = nil
     }
 }
+
+final class SubscriptionSettingsViewModelV2: ObservableObject {
+
```
miasma13: Question, wouldn't it be possible to skip duplication and having to create both `SubscriptionSettingsViewV2` and `SubscriptionSettingsViewModelV2` by only adapting the `SubscriptionSettingsViewModel` to use the `Bridge` class instead?

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -332,3 +333,302 @@ final class SubscriptionSettingsViewModel: ObservableObject {
...
+        case .returnCacheDataElseLoad:
+            tokensPolicy = .localValid
+        case .returnCacheDataDontLoad:
+            tokensPolicy = .local
+        }
+
+        if let tokenContainer = try? await subscriptionManager.getTokenContainer(policy: tokensPolicy) {
+            Task { @MainActor in
+                self.state.subscriptionEmail = tokenContainer.decodedAccessToken.email
+                if loadingIndicator { self.displayEmailLoader(true) }
```
miasma13: Please diff against v1 as this does not seem to correctly mimic older implementation: the loader was suppsed to be set before fetching the email and disabled when complete. Here it is the opposite - it sets the email and sets the indicator.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -332,3 +333,302 @@ final class SubscriptionSettingsViewModel: ObservableObject {
...
+    private func setupNotificationObservers() {
+        signOutObserver = NotificationCenter.default.addObserver(forName: .accountDidSignOut, object: nil, queue: .main) { [weak self] _ in
+            DispatchQueue.main.async {
+                self?.state.shouldDismissView = true
+            }
+        }
+    }
+
+    @MainActor
+    private func updateSubscriptionsStatusMessage(status: PrivacyProSubscription.Status, date: Date, product: String, billingPeriod: PrivacyProSubscription.BillingPeriod) {
```
miasma13: Please diff against v1 as this does not seem to correctly mimic older implementation: previous implementation used `subscription.hasActiveTrialOffer` to determine variants of the string for the trial subscription.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -332,3 +333,302 @@ final class SubscriptionSettingsViewModel: ObservableObject {
...
+            DispatchQueue.main.async {
+                self.state.isShowingConnectionError = value
+            }
+        }
+    }
+
+    @MainActor
+    func showTermsOfService() {
+        Logger.subscription.log("Show terms of service")
+        self.openURL(SettingsSubscriptionView.ViewConstants.privacyPolicyURL)
```
miasma13: Please diff against v1 as this does not seem to correctly mimic older implementation: changed to a quicklink URL.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -533,19 +571,545 @@ private extension SubscriptionPagesUseSubscriptionFeature {
...
+    }
+
+    weak var broker: UserScriptMessageBroker?
+
+    var featureName = SubscriptionPagesUseSubscriptionFeatureConstants.featureName
+
+    var messageOriginPolicy: MessageOriginPolicy = .only(rules: [
+        .exact(hostname: OriginDomains.duckduckgo),
+        .exact(hostname: OriginDomains.abrown)
+    ])
```
miasma13: Regression as this has changed, please diff V1 and V2.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -533,19 +571,545 @@ private extension SubscriptionPagesUseSubscriptionFeature {
...
+            onFeatureSelected?(.networkProtection)
+        case .dataBrokerProtection:
+            onFeatureSelected?(.dataBrokerProtection)
+        case .identityTheftRestoration:
+            onFeatureSelected?(.identityTheftRestoration)
+        case .identityTheftRestorationGlobal:
+            onFeatureSelected?(.identityTheftRestorationGlobal)
+        case .unknown:
+            break
+        }
```
miasma13, line 904: Why not just `onFeatureSelected?(featureSelection.productFeature)` ?

miasma13 (commented):
Please find couple of comments for things I found. 

1. The most apparent one were regressions where V2 counterparts were missing some of the changes ported from V1. I would suggest diffing V1 and V2 in a separate file on a side to make sure we are not missing anything crucial.

2. Another one is a thought if maybe there would be a change and we could limit the size of changes by avoiding the V1/V2 duplication for `Views` and `ViewModels` and instead adapt them to use the `SubscriptionAuthV1toV2Bridge`. I am afraid that by this duplication we may easily introduce some regressions.

iOS/DuckDuckGo/NetworkProtectionTunnelController.swift
federicocappelli: I think because "option" is sent via IPC that is an Objc API, I just maintained NSString, see old `as NSString?`

BrowserServicesKit/Sources/Subscription/SubscriptionAuthenticationState.swift
federicocappelli: fixed

iOS/DuckDuckGo/VPNRedditSessionWorkaround.swift
federicocappelli: correct, bad merge from my old branch, this file is unused, removed

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: The main problem here is `self.subscriptionManager.accountManager.fetchAccountDetails(with: token)` that exists only in v1 and not v2, making a common interface would require a lot of changes, both in the bridge and the current code. In this case I prioritised minimising changes and risks over clean code.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: Well spotted, it was a copy and paste mistake probably.
I moved SubscriptionSettingsViewModelV2 in its own file and did a diff, all discrepancies have been fixed.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: done in https://github.com/duckduckgo/apple-browsers/pull/94#discussion_r1975471748

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: done in https://github.com/duckduckgo/apple-browsers/pull/94#discussion_r1975471748

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
miasma13: Actually my main concern is not aiming for a clean code at this temporary stage but actually minimise the changes. As you can see from the comments by having duplicate V2 `Views` and `ViewModels` there is huge risk of introducing regression (e.g. something is not carried over form V1 or some change is missed). Maybe having the `fetchAccountDetails` method in the bridge that would be no-op for V2 would be good enough sacrifice to reduce the amount of changes and duplication in this PR? Again I want to underline that my only goal here is to reduce the surface area of the changes that may result in introducing regression.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
federicocappelli: done as above, diffed and fixed

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
federicocappelli: Unfortunately, the name `productFeature` is misleading, they are 2 different enums with the same values, all will disappear when v1 is removed.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: correct, I diffed the files and removed this already.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: I understand, I should have pointed out that `SubscriptionSettingsViewModelV2` and `SubscriptionPagesUseSubscriptionFeature` are the only 2 files with substantial changes that I brought from my old branch, all other duplications are done starting from the most recent code and modifying it manually, so the risk should be very limited. The other point about not modifying the current code too much is that all these "adaptations" will need to be reverted in 1-2 months, so I'm trying to keep them at a minimum.

iOS/DuckDuckGo/NetworkProtectionTunnelController.swift
diegoreymendez: It's been a while since we wrote this but I do recall this was an explicit choice, so keeping it is likely a good idea.

federicocappelli: I have probably tried to use a String and the beginning and that caused a nasty and silent "crash" where the NetExt simple disappeared without traces or crash reports

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
miasma13: I think that "adaptations" argument is invalid, as regardless if it is for "adaptations" or V1/V2 there will be some cleanup in the future after the full rollout. My main concern here is if we can approach it in any more safe manner that minimises the risks, as manually diffing V1 and V2 if we have matching implementations is not feasible for both code review and development, and this very code review uncovered couple of such discrepancies.

I am just highlighting it here, and leave the final call to you.

federicocappelli: What you are proposing is the approach I used in other places and what I tried to use here too, the final result was a file with a lot of if authV2IsEnabled, changes to bridge used only in this file and in general a file that was not comparable with the original v1 in any meaningful way. Now that I fixed the merge problems by doing a diff of all the files, I'm confident this is the best and cleaner approach.

P.S. GH is still misbehaving, so the tests are not yet green.

miasma13: Fair enough, I trust your judgement on picking that approach then.

miasma13 (approved):

PR #106, opened 2025-02-26
Implement deleting history by range in History View

ayoy:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209328755192083

### Description
This change allows to delete history items by range in History View. A native dialog is added to confirm history
deletion. The deletion defaults to burning, which additionally clears website data, but it can be adjusted to only
delete history entries.

### Testing Steps

1. Launch the app from Xcode, set internal user flag and override `historyView` flag in Debug Menu so that it's enabled. Open main menu one more time to allow cmd+y shortcut to load.
2. Open history view (cmd+y) and verify that it's empty.
3. Visit a few websites, go back to the history view and refresh it.
4. Verify that you see the "Today" range apart from "All".
5. Click the trash icon next to the "Today" range and verify that it displays the delete dialog.
6. Cancel the dialog and verify that history is not deleted.
7. Repeat step 5 and delete history. Verify that history entries are gone.
8. Visit a website where you can log in to an account. Go to history view and delete history with burning. Verify that you're logged out of the account.
9. Populate fake history via Debug Menu -> History -> Add 10 visits each day. Refresh history view.
10. Try deleting various ranges and verify that data is properly counted in the dialog[0] and then properly deleted.\
11. Repeat step 9 as needed and use the "Delete All" button (note that "Delete All" while filtering the view by a search phrase is not implemented yet). Verify that it works the same as deleting the currently active range.

[0] – please note that multiple visits to the same URL on a single day are conflated into 1 entry on the screen and only visible items are counted and reported to the user in the delete dialog, while upon deletion, all visits are deleted whatsoever.

### Impact and Risks

Low: new feature, and the mechanisms for deleting and burning (including fireproofing) are reused from existing code so it's mainly the new UI wrapping.

#### What could go wrong?

Nothing outstanding comes to my mind.

### Quality Considerations

This is a WIP feature, so Pixels will come later.
I've checked performance using console logging but it was blazingly fast (usually in order of fractions of milliseconds to delete history, with burning obviously taking longer).
This change seems large, but almost half of it is the unit tests (including overdue tests for the data provider).

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


macOS/DuckDuckGo/History/Services/HistoryViewDataProvider.swift
ayoy, line 1: This file appears as deleted, but in fact it was moved and partially reworked.

macOS/DuckDuckGo/History/View/HistoryViewCoordinator.swift
```diff
@@ -26,7 +27,7 @@ final class HistoryViewCoordinator {
     let actionsManager: HistoryViewActionsManager
 
     init(
-        historyCoordinator: HistoryGroupingDataSource,
+        historyCoordinator: HistoryGroupingDataSource & HistoryDeleting,
```
ayoy, line 30: This will be replaced with a single protocol in an upcoming pull request.

macOS/LocalPackages/HistoryView/Sources/HistoryView/DataClient.swift
```diff
@@ -120,6 +125,21 @@ public final class DataClient: HistoryViewUserScriptClient {
...
+    @MainActor
+    private func entriesMenu(params: Any, original: WKScriptMessage) async throws -> Encodable? {
+        return nil
+    }
+
+    @MainActor
+    private func entriesDelete(params: Any, original: WKScriptMessage) async throws -> Encodable? {
+        return nil
+    }
+
```
ayoy, line 142: These are to be implemented (coming up in the second pull request).

BrowserServicesKit/Sources/Common/Extensions/CalendarExtension.swift
```diff
@@ -24,4 +24,12 @@ extension Calendar {
 		return numberOfDays.day
 	}
 
+    public func firstWeekday(_ weekday: Int, before referenceDate: Date) -> Date {
```
jotaemepereira, line 37: 💡 Given that this function is accessible to both macOS and iOS we should add a coment explaining how it works, maybe with some example.

Also, I would be cautious about `weekday`, because it seems it should be in the range from 1 to 7. Could we have a problem if somebody uses 0 or a number greater than 7? Should we use an enum here?

macOS/DuckDuckGo/History/View/HistoryViewDataProvider.swift
```diff
@@ -0,0 +1,271 @@
...
+
+    init(fireproofDomains: DomainFireproofStatusProviding = FireproofDomains.shared, fire: (() async -> Fire)? = nil) {
+        self.fireproofDomains = fireproofDomains
+        self.fire = fire ?? { @MainActor in FireCoordinator.fireViewModel.fire }
+    }
+
+    func burn(_ visits: [Visit], animated: Bool) async {
+        await withCheckedContinuation { continuation in
+            Task { @MainActor in
+                await fire().burnVisits(visits, except: fireproofDomains, isToday: animated) {
```
jotaemepereira, line 40: Could there be a problem with the `burnVisits` call here? From the stacktrace, the `animates` flag is true when the range is today or all, if it is all, could it lead to a problem?

jotaemepereira (changes requested):
@ayoy I’m requesting changes because I’m seeing the follow bug when deleting the Today range, it shows the fire animation and restart the application, is that expected?


https://github.com/user-attachments/assets/77283fb7-7b22-4551-aa14-31946f925519

BrowserServicesKit/Sources/Common/Extensions/CalendarExtension.swift
ayoy: Yes, absolutely, let me add docs here.

I didn't want to use an enum – `weekday` is a property of DateComponents and it's assumed to be between 1 and 7. I will add an assertion to enforce that range in development and I'll also cap the value between 1 and 7 in case this API is ever used with an arbitrary input:

```swift
public func firstWeekday(_ weekday: Int, before referenceDate: Date) -> Date {
    assert((1...7).contains(weekday), "Weekday must be between 1 and 7 (as per DateComponents API)")

    // ensure weekday is between 1 and 7
    let adjustedWeekday: Int = min(max(weekday, 1), 7)
```

macOS/DuckDuckGo/History/View/HistoryViewDataProvider.swift
ayoy: This isn't a bug actually :) The behavior follows history clearing behavior from the main menu – clearing data for all history needs to close all tabs. The same for today (because we're clearing data for the tabs opened today, i.e. tabs open right now, therefore they must be closed). Data cleared for any previous day doesn't close tabs.

We will update the copy of the burn dialog to reflect this (see [Copy Review discussion](https://app.asana.com/0/38424471409662/1209167289674669) for reference) - the implementation is [already in place](https://github.com/duckduckgo/apple-browsers/commit/31704cb5d238166b45c00629d920468aec18632e), but due to changes to history delete dialog, the commit is added on another branch.

jotaemepereira (approved):
@ayoy Ah, okay, that makes sense. I’m going to approve this so I can focus on the second part. Great job.

PR #116, opened 2025-02-28
scam protection

SabrinaTardio:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209164066387913

### Testing Steps

1. Go to the debug menu -> FeatureFlags and check the malicious site protection is on and scam protection is off
2. Test that https://privacy-test-pages.site/security/badware/scam.html
https://privacy-test-pages.site/security/badware/scam-local-check.html are not blocked
3. From the debug menu -> Feature Flags Turn the Scam protection flag on 
4. Test both Test that https://privacy-test-pages.site/security/badware/scam.html
https://privacy-test-pages.site/security/badware/scam-local-check.html are not blocked 
5. Test leaving the page, bypassing the error and seeing the help page.

### Impact and Risks
<!— 

—>

#### What could go wrong?

### Quality Considerations
<!— 
- What edge cases exist?
—>

### Notes to Reviewer

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

—
###### Internal references:


BrowserServicesKit/Sources/MaliciousSiteProtection/MaliciousSiteDetector.swift
```diff
@@ -42,16 +42,18 @@ public final class MaliciousSiteDetector: MaliciousSiteDetecting {
...
     private let eventMapping: EventMapping<Event>
 
-    public convenience init(apiEnvironment: APIClientEnvironment, service: APIService = DefaultAPIService(urlSession: .shared), dataManager: DataManager, eventMapping: EventMapping<Event>) {
-        self.init(apiClient: APIClient(environment: apiEnvironment, service: service), dataManager: dataManager, eventMapping: eventMapping)
+    public convenience init(apiEnvironment: APIClientEnvironment, service: APIService = DefaultAPIService(urlSession: .shared), dataManager: DataManager, featureFlagger: MaliciousSiteProtectionFeatureFlagger, eventMapping: EventMapping<Event>) {
+        self.init(apiClient: APIClient(environment: apiEnvironment, service: service), dataManager: dataManager, featureFlagger: featureFlagger, eventMapping: eventMapping)
     }
 
-    init(apiClient: APIClient.Mockable, dataManager: DataManaging, eventMapping: EventMapping<Event>) {
+    init(apiClient: APIClient.Mockable, dataManager: DataManaging, featureFlagger: MaliciousSiteProtectionFeatureFlagger, eventMapping: EventMapping<Event>) {
```
alessandroboron: I wonder if it would be cleaner to inject a parameter with default implementation such as `supportedThreats: [ThreatKind] = ThreatKind.allCases`. Then on the client we would set to [.phishing, .malware] if `isScamProtectionEnabled` is disabled. 

Reasons are: 
- `isMaliciousSiteProtectionEnabled` feature flag is checked on the client side while `isScamProtectionEnabled` is checked here. I know they are two different use case but we could leave the library unaware of the feature flag.
- Simplified tests as we don’t have to inject mocks.
- If for example we add another layer of protection such as `fakeNews` sites protection and we put it under another feature flag then this code
`for threatKind in supportedThreats` would stay the same while this
```
if !isScamProtectionEnabled && threatKind == .scam {
  return .none
}
```
would just be `if !supportedThreats.contains(threatKind) { return .none }` and stay the same for any type of unsupported threats.

But this may be a personal preference, nothing wrong with your approach. Just food for thought.

BrowserServicesKit/Sources/MaliciousSiteProtection/MaliciousSiteDetector.swift
```diff
@@ -124,6 +128,9 @@ public final class MaliciousSiteDetector: MaliciousSiteDetecting {
         if let match {
             let threatKind = match.category.flatMap(ThreatKind.init) ?? hashPrefixMatchingThreatKinds[0]
             await fireErrorPageShown(threatKind: threatKind, clientSideHit: false)
+            if !isScamProtectionEnabled && threatKind == .scam {
```
alessandroboron: Just double checking, should we send the pixel in this case? We’re not really showing the error page but at the same time we hit the backend.

BrowserServicesKit/Tests/MaliciousSiteProtectionTests/MaliciousSiteProtectionUpdateManagerTests.swift
```diff
@@ -34,6 +34,7 @@ class MaliciousSiteProtectionUpdateManagerTests: XCTestCase {
     var clock: TestClock<Duration>!
     var willSleep: ((TimeInterval) -> Void)?
     var updateTask: Task<Void, Error>?
+    var featureFlagger: MockMaliciousSiteProtectionFeatureFlags!
```
alessandroboron: don’t forget to nil it in `tearDown` :)

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/UpdateManager.swift
```diff
@@ -57,20 +58,26 @@ public struct UpdateManager: InternalUpdateManaging {
...
+    public init(apiEnvironment: APIClientEnvironment, service: APIService = DefaultAPIService(urlSession: .shared), dataManager: DataManager, featureFlags: MaliciousSiteProtectionFeatureFlagger, eventMapping: EventMapping<Event>, updateIntervalProvider: @escaping UpdateIntervalProvider) {
+        self.init(apiClient: APIClient(environment: apiEnvironment, service: service), dataManager: dataManager, featureFlags: featureFlags, eventMapping: eventMapping, updateIntervalProvider: updateIntervalProvider)
     }
 
-    init(apiClient: APIClient.Mockable, dataManager: DataManaging, eventMapping: EventMapping<Event>, sleeper: Sleeper = .default, updateInfoStorage: MaliciousSiteProtectioUpdateManagerInfoStorage = UpdateManagerInfoStore(), updateIntervalProvider: @escaping UpdateIntervalProvider) {
+    init(apiClient: APIClient.Mockable, dataManager: DataManaging, featureFlags: MaliciousSiteProtectionFeatureFlagger, eventMapping: EventMapping<Event>, sleeper: Sleeper = .default, updateInfoStorage: MaliciousSiteProtectioUpdateManagerInfoStorage = UpdateManagerInfoStore(), updateIntervalProvider: @escaping UpdateIntervalProvider) {
```
alessandroboron: Similar to https://github.com/duckduckgo/apple-browsers/pull/116/files#r1978723390 but nothing wrong with the approach

macOS/IntegrationTests/MaliciousSiteProtection/MaliciousSiteProtectionIntegrationTests.swift
```diff
@@ -302,6 +304,53 @@ class MaliciousSiteProtectionIntegrationTests: XCTestCase {
...
+        try await loadUrl(url1)
+        XCTAssertEqual(tabViewModel.tab.error as NSError? as? MaliciousSiteError, MaliciousSiteError(code: .scam, failingUrl: url1))
+
+        let url2 = URL(string: "http://broken.third-party.site/")!
+        try await loadUrl(url2)
+        XCTAssertNil(tabViewModel.tab.error)
+    }
+
+    @MainActor
+    func testScamDetectedThenDDGLoaded_tabIsNotMarkedMalware() async throws {
```
alessandroboron:
```suggestion
    func testScamDetectedThenDDGLoaded_tabIsNotMarkedScam() async throws {
```

macOS/UnitTests/MaliciousSiteProtection/MaliciousSiteProtectionTests.swift
```diff
@@ -92,6 +99,18 @@ final class MaliciousSiteProtectionTests: XCTestCase {
         XCTAssertEqual(isMalicious, .malware)
     }
 
+    func testWhenScamDetected_scamThreatReturned() async {
+        MaliciousSiteProtectionPreferences.shared.isEnabled = true
+        let isMalicious = await phishingDetection.evaluate(URL(string: "https://scam.com")!)
+        XCTAssertEqual(isMalicious, .scam)
+    }
+
+    func testWhenScamDetected_andScamProtectionOff_nosThreatReturned() async {
```
alessandroboron:
```suggestion
    func testWhenScamDetected_andScamProtectionOff_noThreatReturned() async {
```

BrowserServicesKit/Sources/MaliciousSiteProtection/MaliciousSiteDetector.swift
SabrinaTardio: That is a good question… Thinking more about it since the pixel is “Error Page Shown” I think we should not send it

BrowserServicesKit/Sources/MaliciousSiteProtection/MaliciousSiteDetector.swift
SabrinaTardio: The reasons I did not use this approach were:
- I think that since they are all strictly tied together and are effectively one feature this flag is only temporary 
- If I passed a parameter it would not be reactive to feature flag changes (we would need to restart the app for the change to take effect).
- The framework is already aware of feature flag because it is already passed in other places

Maybe we can pass a function but I am not sure if very clean, I’ll think a bit about it

SabrinaTardio: I am injecting a function in the end! I think it looks better now.

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/UpdateManager.swift
SabrinaTardio: Replied above

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/UpdateManager.swift
```diff
@@ -144,7 +153,9 @@ public struct UpdateManager: InternalUpdateManaging {
         Task {
             // run update jobs in background for every data type
             await withTaskGroup(of: Bool.self) { group in
-                for dataType in DataManager.StoredDataType.dataTypes(for: datasetType) {
```
alessandroboron, line 147: @SabrinaTardio we’re ignoring the parameter `datasetType` here and loop through all the supported datatypes.

The reason we pass the `datasetType` and we don’t loop in iOS is because we have two background tasks, one is used to fetch all the hashPrefixes and one is used to fetch all the filterSets.

The reason we used two type of background tasks is that they have different update times, 20 minutes and 12 hours respectively.

With this loop we will download everything every time call this function.

So we can either change `dataTypes(for kind: DataManager.StoredDataType.Kind) -> [DataManager.StoredDataType]` to accept a threatKind parameter and remove the internal loop, or once we call `DataManager.StoredDataType.dataTypes(for: datasetType)` we can filter out the values. I guess the former is more efficient as it requres only one loop.

iOS/DuckDuckGo/MaliciousSiteProtection/MaliciousSiteProtectionManager.swift
```diff
@@ -45,7 +46,7 @@ final class MaliciousSiteProtectionManager {
             apiEnvironment: api.environment,
             service: api.service,
             dataManager: dataManager,
-            eventMapping: MaliciousSiteProtectionEventMapper.debugEvents
+            eventMapping: MaliciousSiteProtectionEventMapper.debugEvents, supportedThreatsProvider: supportedThreatsProvider
```
alessandroboron: nip: supportedThreatsProvider: supportedThreatsProvider on next line for consistency

macOS/DuckDuckGo/Tab/ViewModel/TabViewModel.swift
```diff
@@ -414,14 +414,9 @@ final class TabViewModel {
...
             case is URLError where tab.error?.isServerCertificateUntrusted == true:
-                title = UserText.sslErrorPageTabTitle
-            case .some(let error as MaliciousSiteError):
-                switch error.code {
-                case .phishing:
-                    title = UserText.phishingErrorPageTabTitle
-                case .malware:
-                    title = UserText.malwareErrorPageTabTitle
-                }
+                return
```
alessandroboron: just to confirm, we don’t show any title for SSL expired certificate?

macOS/IntegrationTests/MaliciousSiteProtection/MaliciousSiteProtectionIntegrationTests.swift
```diff
@@ -323,7 +365,7 @@ class MaliciousSiteProtectionIntegrationTests: XCTestCase {
         }
         defer { task.cancel() }
 
-        let result = await waiter.fulfillment(of: [loadingExpectation], timeout: 2)
+        let result = await waiter.fulfillment(of: [loadingExpectation], timeout: 5)
```
alessandroboron: just checking, are we hitting the backend in this integration tests?

alessandroboron (commented):
@Sabrina, I tested it and works as expected.

There’s only one thing we need to address related to iOS datasets fetch. I left a comment in https://github.com/duckduckgo/apple-browsers/pull/116#discussion_r1986464846

macOS/DuckDuckGo/Tab/ViewModel/TabViewModel.swift
SabrinaTardio: No I did not mean to change it reset it to use a value

macOS/IntegrationTests/MaliciousSiteProtection/MaliciousSiteProtectionIntegrationTests.swift
SabrinaTardio: We do but I was testing locally I did not mean to push this change

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/UpdateManager.swift
SabrinaTardio: You are right, I did not realise it was different from macOS!

alessandroboron (approved):
LGTM! 💪

macOS/DuckDuckGo/Tab/TabExtensions/NavigationProtectionTabExtension.swift
SabrinaTardio, line 1: need to remove this

BrowserServicesKit/Sources/BrowserServicesKit/GPC/GPCRequestFactory.swift
```diff
@@ -50,43 +50,56 @@ public struct GPCRequestFactory {
         return false
     }
 
+    /// Returns the header fields to be applied for GPC.
+    /// If no header should be applied, returns an empty dictionary.
+    public func headersForGPC(basedOn incomingRequest: URLRequest,
```
mallexxx: this duplicates the `computedGPCHeaders`, maybe use it directly instead?

BrowserServicesKit/Sources/BrowserServicesKit/GPC/GPCRequestFactory.swift
```diff
@@ -50,43 +50,56 @@ public struct GPCRequestFactory {
...
+                              gpcEnabled: Bool) -> [String: String] {
+        return computedGPCHeaders(for: incomingRequest, config: config, gpcEnabled: gpcEnabled)
+    }
+
+    /// Returns a modified URLRequest with the GPC header added or removed if needed.
+    /// Returns nil if no modifications are necessary.
     public func requestForGPC(basedOn incomingRequest: URLRequest,
                               config: PrivacyConfiguration,
                               gpcEnabled: Bool) -> URLRequest? {
+        let computedHeaders = computedGPCHeaders(for: incomingRequest, config: config, gpcEnabled: gpcEnabled)
```
mallexxx: we can make it `gpcHeader(for:..) -> (key: String, value: String)?` instead, we don‘t need the dictionary for the single header

BrowserServicesKit/Sources/BrowserServicesKit/GPC/GPCRequestFactory.swift
```diff
@@ -50,43 +50,56 @@ public struct GPCRequestFactory {
...
+        let currentHeader = incomingRequest.value(forHTTPHeaderField: Constants.secGPCHeader)
 
-        func removingHeader(fromRequest incomingRequest: URLRequest) -> URLRequest? {
-            var request = incomingRequest
-            if let headers = request.allHTTPHeaderFields, headers.firstIndex(where: { $0.key == Constants.secGPCHeader }) != nil {
+        // If computed headers are empty and the header is currently present, remove it.
+        if computedHeaders.isEmpty {
```
mallexxx: this and the previous comment could be simplified to `if let header = gpcHeader(for:…), currentHeader == nil {…} else if currentHeader != nil { remove }`

macOS/DuckDuckGo/Tab/TabExtensions/NavigationProtectionTabExtension.swift
```diff
@@ -117,15 +117,22 @@ extension NavigationProtectionTabExtension: NavigationResponder {
...
-                                                              gpcEnabled: isGPCEnabled) {
-            request = newRequest
-        }
 
-        if request != navigationAction.request {
-            return .redirectInvalidatingBackItemIfNeeded(navigationAction) {
-                $0.load(request)
+        // If custom headers are supported, update the preferences rather than the request.
+        if NavigationPreferences.customHeadersSupported {
+            let headers = GPCRequestFactory().headersForGPC(basedOn: request, config: contentBlocking.privacyConfigurationManager.privacyConfig, gpcEnabled: isGPCEnabled)
```
mallexxx: I think this 2 `if`-s should be inline, checking for the nullable single GPC header and if the returned header is nil we should fall through to the `else` part that will remove the header (opt-out support gets broken)

mallexxx: upd: it actually works as intended when testing on `https://globalprivacycontrol.org/` with protections disabled - the request gets no GPC header on even if requested before with the header set; So I guess we don‘t need the header removal logics.

macOS/DuckDuckGo/Tab/TabExtensions/NavigationProtectionTabExtension.swift
```diff
@@ -117,15 +117,22 @@ extension NavigationProtectionTabExtension: NavigationResponder {
         }
 
         let isGPCEnabled = WebTrackingProtectionPreferences.shared.isGPCEnabled
-        if let newRequest = GPCRequestFactory().requestForGPC(basedOn: request,
-                                                              config: contentBlocking.privacyConfigurationManager.privacyConfig,
-                                                              gpcEnabled: isGPCEnabled) {
-            request = newRequest
-        }
 
-        if request != navigationAction.request {
```
mallexxx: we‘re losing this check at this line if GPC header is added

SabrinaTardio: This was already changed after the commit you saw, but doesn’t matter anyway since I to revert these changes

macOS/DuckDuckGo/Tab/TabExtensions/NavigationProtectionTabExtension.swift
SabrinaTardio: as discussed I had to revert the changes since the header was not actually applied

BrowserServicesKit/Sources/BrowserServicesKit/GPC/GPCRequestFactory.swift
SabrinaTardio: as discussed I had to revert the changes since the header was not actually applied

BrowserServicesKit/Sources/BrowserServicesKit/GPC/GPCRequestFactory.swift
SabrinaTardio: as discussed I had to revert the changes since the header was not actually applied

BrowserServicesKit/Sources/BrowserServicesKit/GPC/GPCRequestFactory.swift
SabrinaTardio: as discussed I had to revert the changes since the header was not actually applied

PR #122, opened 2025-02-28
Add History View onboarding popover

ayoy:
Task/Issue URL: https://app.asana.com/0/1201899738287924/1209538877789229

### Description
This change adds history view onboarding popover to be displayed from the more options menu button.
Onboarding popover is displayed once only for old users (more than 1 week since installation).

### Testing Steps

1. Clear app data `./clean_app.sh debug`, run the app from Xcode, enable historyView feature flag via Debug Menu and quit the app.
2. Run the app and verify that you **do not see** the history onboarding popover.
3. Quit the app.
4. Run `defaults delete com.duckduckgo.macos.browser.debug first.app.launch.date` to mark yourself as old user.
5. Run the app again and verify that you see the history onboarding popover. Accept the popover and verify that you're taken to the history view.
6. Open a new tab (and/or new window) and verify that you don't see the history onboarding popover.
7. You can reset the dialog or force showing it via Debug Menu -> History.

Disabled feature flag:
1. Clear app data and run the app (**do not set** historyView feature flag).
2. Verify that you **do not see** the history onboarding popover.
3. Open a new tab and verify the same.
4. Quit the app, and run `defaults delete com.duckduckgo.macos.browser.debug first.app.launch.date` to mark yourself as old user.
5. Open the app and verify that you don't see the popover.
6. Open a new tab/window and verify that you still don't see the popover :)

### Impact and Risks

Impact is overall low, or none if we verify that the popover doesn't show up for users while the flag is disabled.

#### What could go wrong?

It could interfere with some other popovers but we are willing to accept this for now.

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


macOS/DuckDuckGo/Common/Localizables/UserText.swift
```diff
@@ -208,6 +208,10 @@ struct UserText {
     static let mainMenuHelpDuckDuckGoHelp = NSLocalizedString("DuckDuckGo Help", comment: "Main Menu Help item")
 
     // MARK: - History
+    static let historyViewOnboardingTitle = NSLocalizedString("history.view.onboarding.title", value: "Manage Your History", comment: "Title for the history view onboarding popover")
+    static let historyViewOnboardingMessage = NSLocalizedString("history.view.onboarding.message", value: "We’ve added a dedicated History page, making it easier for you to for you to see, revisit, and clear the sites you’ve previously visited.", comment: "Message for the history view onboarding popover")
```
jotaemepereira: ⚠️ This error message has some typoes, it repeats the for you. It reads “for you to for you to see”

macOS/DuckDuckGo/History/View/Onboarding/HistoryViewOnboardingPopover.swift
```diff
@@ -0,0 +1,55 @@
...
+        self.viewModel = HistoryViewOnboardingViewModel(ctaCallback: ctaCallback)
+        let view = HistoryViewOnboardingView(model: viewModel)
+        super.init(rootView: view)
+    }
+
+    @MainActor @preconcurrency required dynamic init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    private let viewModel: HistoryViewOnboardingViewModel
```
jotaemepereira, line 59: ❓ I saw this in the previous file. What’s the reason for moving the constants, and variables definition after the class initialization? 

For example, the popover has it at the top; is it because it is internal?

It doesn’t seem to be consistent with the current codebase.

macOS/DuckDuckGo/NavigationBar/View/NavigationBarViewController.swift
```diff
@@ -236,6 +263,31 @@ final class NavigationBarViewController: NSViewController {
...
+     * Presents History View onboarding.
+     *
+     * This is gater by the decider that takes into account whether the user is new,
+     * whether they've seen the popover already and whether the feature flag is enabled.
+     *
+     * > `force` parameter is only used by `HistoryDebugMenu`.
+     */
+    func presentHistoryViewOnboardingIfNeeded(force: Bool = false) {
+        Task { @MainActor in
+            guard force || HistoryViewOnboardingDecider().shouldPresentOnboarding, view.window?.isKeyWindow == true else {
```
jotaemepereira: 💡 Should we check also that the current window is not a fire window? I wonder how users will react if they see something related to navigation history when they are in a non-storing session.

jotaemepereira (changes requested):
@ayoy I’m requesting changes because I’m seeing some problems with the popover.

The first one is this bug, debugging the code. The `.onAppear` in SwiftUI is not being called all the time, causing the popover to get shown multiple times.

https://github.com/user-attachments/assets/6b836f72-68c7-423a-bf5e-b1c45e2a4459

The popover when in full screen it gets cut (I do not know if is related to my computer or something):

Screenshot 2025-03-02 at 10 24 18 PM

The third one is related to fire windows, with this code it seems possible to show the popover in a fire window. I believe we should avoid that.

macOS/DuckDuckGo/Common/Localizables/UserText.swift
ayoy: Thanks, fixed :)

macOS/DuckDuckGo/NavigationBar/View/NavigationBarViewController.swift
ayoy: Good point, it shouldn't happen overall because Fire Windows aren't persisted across sessions, but in case the flag is remotely enabled while the app was running, it could happen for a Fire Window too. I've fixed it.

macOS/DuckDuckGo/History/View/Onboarding/HistoryViewOnboardingPopover.swift
ayoy: Yes, it's on the bottom because it's private. We don't have rules defined for that but there is plenty of files following this convention.

ayoy:
@jotaemepereira thanks a lot for the thorough review as always! I think I've fixed all the issues. The popover will now be marked as shown from the View Controller, which seems more reliable. Let me know what you think. Thanks!

jotaemepereira (approved):
Tested it again and it looks great 🎉

PR #124, opened 2025-02-28
Fix issue with omnibar animation

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1209483539091359

### Description
Fix issue with omnibar animation.

The issue was happening because the animation was being cancelled during state changes, but it shouldn’t be.
There’s no reason to stop this animation since the icons visibility are already handled by the state

### Testing Steps
1. Open a new tab and quickly type something. I couldn’t do this on a real device, use the simulator

### Impact and Risks
Low: small bug fixes

#### What could go wrong?
Can’t think of anything besides the animation being visible when it shouldn’t, which is unlikely since the icons are handled by the state.


dus7 (approved):
👌

PR #126, opened 2025-02-28
Fix macOS VPN location picker issues

samsymons:
Task/Issue URL: https://app.asana.com/0/1203708860857015/1209538586919556
Task/Issue URL: https://app.asana.com/0/1199230911884351/1209344742578598
Tech Design URL:
CC:

### Description

This PR fixes three issues:
1. When selecting a city with the location picker, we were trying to reuse that same selected city with other countries, causing the city picker to default to blank - this is now fixed by passing `.nearest` to countries that do not have a selected city
2. The loading state of the view was blank, a `ProgressView` has been added
3. When trying to select a different city while the `locations` endpoint was not reachable, the selection wouldn't update - this has been fixed by keeping a cache of the last loaded cities, and updating which one was selected from that as a fallback

### Testing Steps

1. Check that location loading works as expected
2. Select a city for a location, and then scroll down to verify that the other countries with selectable cities are listed as "Nearest"
3. With the location picker open, kill your network connection entirely, then wait a minute or so - after that, try to select another city, and verify that the UI updates correctly. Then, re-enable your connection and connect to the VPN to verify that the new location is respected

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNLocation/VPNLocationViewModel.swift
```diff
@@ -93,10 +94,14 @@ final class VPNLocationViewModel: ObservableObject {
 
     @MainActor
     private func reloadList() async {
-        guard let locations = try? await locationListRepository.fetchLocationList().sortedByName() else { return }
+        guard let locations = (try? await locationListRepository.fetchLocationList().sortedByName()) ?? Self.cachedLocations else {
```
samsymons, line 97: This line was the cause of the unresponsive UI. When the locations endpoint is unreachable, this hits an early return, so the list of cities doesn't get properly refreshed. After the change, it will try to refresh, but fall back to the last fetched values otherwise.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNLocation/VPNLocationViewModel.swift
```diff
@@ -113,7 +118,12 @@ final class VPNLocationViewModel: ObservableObject {
                 cityPickerItems = currentLocation.cities.map { currentCity in
                     return VPNCityItemModel(cityName: currentCity.name)
                 }
-                selectedCityItem = location.city.flatMap(VPNCityItemModel.init(cityName:)) ?? .nearest
+
+                if isCountrySelected {
+                    selectedCityItem = location.city.flatMap(VPNCityItemModel.init(cityName:)) ?? .nearest
+                } else {
+                    selectedCityItem = .nearest
+                }
```
samsymons, line 126: Previously we were passing the selected city to all locations, since `location.city.flatMap(VPNCityItemModel.init(cityName:))` was always non-nil. So, we could be passing `London` to the `United States` country, for example. Now we make sure that the country is actually selected, otherwise pass `.nearest`.

diegoreymendez: Nice catch!

diegoreymendez:
Adding some feedback here: I'm seeing some trouble where the list of locations is blank for me after a brief loading animation.

Now I can't start the VPN due to this error:


If I recall correctly @graeme had fixed an issue like this in the past where there was an error with how we were encoding locations to pass to the VPN.

diegoreymendez (approved):
Seems my problems were due to some backend issues in Staging, apologies for the noise.

Works as described and changes look good!

PR #140, opened 2025-03-04
[WIP] Align suggestions scoring

mallexxx:
Task/Issue URL: https://app.asana.com/0/0/1209533698125405/f
Tech Design URL:
CC:

### Testing Steps

1.
2.

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


BrowserServicesKit/Sources/Suggestions/SuggestionProcessing.swift
```diff
@@ -28,139 +28,237 @@ public enum Platform {
...
-            guard case .historyEntry = suggestion else {
-                expanded.append(item)
-                continue
+    /// Determines if a suggestion should be included in top hits, following Windows algorithm rules
+    private func isTopHit(_ scoredSuggestion: ScoredSuggestion, _ suggestionKinds: Set<SuggestionType.Kind>) -> Bool {
+        let suggestion = scoredSuggestion.suggestion
+        
+        // Check if the suggestion is for website, favorite or history
+        // Otherwise the suggestion should not be part of top hits
+        guard [.website, .favorite, .historyEntry].contains(suggestion.kind) else { return false }
```
mikamikem: I think one issue might be here.  On Windows we intersect this list with `suggestionKinds`.  In the case of a history or website and open tab entry, the open tab would take priority over the history/website and be set on `suggestion.kind`, so the suggestion would just get dropped from the top hits section here instead of being processed and fixed up later in `handleTopHitsOpenTabCase`.

BrowserServicesKit/Sources/Suggestions/SuggestionProcessing.swift
```diff
@@ -28,139 +28,237 @@ public enum Platform {
 /// Class encapsulates the whole ordering and filtering algorithm
```
mikamikem, line 28: (Adding to a random line so we can reply to the comment and track it easier) It seems like there's diffs in the way `subtitle` is calculated, specifically:

- Search results in history on Windows still use "Search DuckDuckGo" as the subtitle, but macOS uses the url.
- It seems to inconsistently trim the "www." from the start of the url.

mikamikem, line 28: The fire window tabs test seems to be failing on deterministic ordering of the same scored value.  Can you step through the test run and see where the order changes?

mikamikem, line 28: The Windows side is also populating tabId in the case where a suggestion is both an open tab and another type.  We could strip this from the test data and I can fix the test generator to remove these values.  It's unused data on the Windows side, so it doesn't need to be there and is definitely also causing a diff in some of the tests.

mikamikem, line 28: It's hard to tell if the other things flagged are causing the pinned tab issue as well or if there's an additional issue with the pinned tabs.  I think in cases where a pinned tab and a non-pinned tab are equally matched in score, there are likely ordering differences between the two platforms.  I'm happy to fix up something on my side as well since your pinned tab feature is actually working and Windows doesn't have it yet.

Yhan0808 (commented):
No

PR #147, opened 2025-03-05
(7.160.0) VPN Entitlement check fix

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1201011656765697/1208864390316201
CC: @Bunn @samsymons @miasma13 

### Description

Fix for a bug introduced in https://app.asana.com/0/1205842942115003/1209170372758738 where a failure to fetch the subscription entitlements was treated as the user not having the entitlements at all.

### Testing Steps
1. Install 7.160.0
2. Subscribe to Privacy pro and enable VPN
3. Install this fix on top
4. Check that subscription is still available and the VPN is still enabled

1. Enable VPN
2. Click on the vpn widget for disabling the VPN
3. All should work as expected

1. Disable VPN
2. Click on the vpn widget for enabling the VPN
3. All should work as expected

### Impact and Risks

High, we could log out the Subscription Pro users

#### What could go wrong?

- User can be logged out from Subscription Pro
- The VPN could be disconnected
- We could present the wrong VPN related settings

---
###### Internal references:

### DIFF
--- a/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionAuthV1toV2Bridge.swift
+++ b/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionAuthV1toV2Bridge.swift
@@ -67,7 +67,7 @@ extension DefaultSubscriptionManager: SubscriptionAuthV1toV2Bridge {
 extension DefaultSubscriptionManagerV2: SubscriptionAuthV1toV2Bridge {
 
     public func isEnabled(feature: Entitlement.ProductName) async throws -> Bool {
-        switch feature { // TODO Check what happens if fails to fetch current features
+        switch feature { // todo Check what happens if fails to fetch current features
         case .networkProtection:
             return await isFeatureAvailableForUser(.networkProtection)
         case .dataBrokerProtection:
--- a/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionAuthV1toV2Bridge.swift
+++ b/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionAuthV1toV2Bridge.swift
@@ -31,7 +31,7 @@ public protocol SubscriptionAuthV1toV2Bridge: SubscriptionTokenProvider, Subscri
 
 extension SubscriptionAuthV1toV2Bridge {
     public func isEnabled(feature: Entitlement.ProductName) async throws -> Bool {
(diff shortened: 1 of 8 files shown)


BrowserServicesKit/Sources/Subscription/Managers/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -64,8 +72,8 @@ extension DefaultSubscriptionManager: SubscriptionAuthV1toV2Bridge {
 
 extension DefaultSubscriptionManagerV2: SubscriptionAuthV1toV2Bridge {
 
-    public func isEnabled(feature: Entitlement.ProductName) async -> Bool {
-        switch feature {
+    public func isEnabled(feature: Entitlement.ProductName, cachePolicy: APICachePolicy) async throws -> Bool {
+        switch feature { // todo Check what happens if fails to fetch current features
```
miasma13, line 76: Is this TODO actionable in any way now? Maybe a follow up Asana task would make it such.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -1691,7 +1691,9 @@ class MainViewController: UIViewController {
     private func onEntitlementsChange(_ notification: Notification) {
         Task {
             let subscriptionManager = AppDependencyProvider.shared.subscriptionAuthV1toV2Bridge
-            guard await subscriptionManager.isEnabled(feature: .networkProtection) else { return }
+            guard let hasEntitlement = try? await subscriptionManager.isEnabled(feature: .networkProtection),
+                      hasEntitlement == false
+            else { return }
```
miasma13, line 1696: Good catch, this logic got inverted in previous PR.

BrowserServicesKit/Sources/Subscription/Managers/SubscriptionAuthV1toV2Bridge.swift
federicocappelli: I'm on it, but the AuthV2 code is not used yet, I'll create a new PR (with less urgency) against main if any change is needed.

miasma13 (approved):
I have been able to reproduce the problem with VPN when forced the entitlement cache to a very short period. The issue was one of the entitlement checks that by mistake had inverted logic and triggered VPN stop and teardown (which additionally prompted the notification about subscription expiring) when the VPN entitlement was present (instead of missing).

PR #152, opened 2025-03-06
Break up `runScanOperation` function into smaller sub-functions

samsymons:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209576214758610
Tech Design URL:
CC:

### Description

This PR breaks up the `runOperation` function of the operation manager into smaller functions, which are more clearly marked with what they're doing under the hood.

### Testing Steps

1. Check that the E2E tests succeed
2. Smoke test PIR overall

### Impact and Risks

The risk here is high, since the operation manager is core to PIR's functionality. However, this PR is deliberately not changing **any** of the functionality - instead it is moving things into smaller functions and calling them from the top level scan function.

I checked the code coverage and overall tests themselves, these changes _look_ to be adequately covered but I'd like to take a fresh look at that on Monday before thinking about merging this.

#### What could go wrong?

PIR could break, which is why I'm avoiding functionality changes. This PR is overall just cosmetic.

### Optional E2E tests**:
- [x] Run macOS PIR E2E tests

---
###### Internal references:


macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/UIWeb/UIMapper.swift
```diff
@@ -238,6 +238,7 @@ struct MapperToUI {
         do {
             let encoder = JSONEncoder()
             encoder.outputFormatting = .prettyPrinted
+            encoder.outputFormatting = .sortedKeys
```
samsymons, line 241: I was having a hard time following all the metadata log entries - using `sortedKeys` means that the logged JSON blob isn't constantly shifting around.

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/macOS/UserNotifications/DataBrokerProtectionUserNotificationService.swift
```diff
@@ -110,9 +110,9 @@ public class DefaultDataBrokerProtectionUserNotificationService: NSObject, DataB
         userNotificationCenter.add(request) { error in
             if error == nil {
                 if days != nil {
-                    Logger.dataBrokerProtection.log("Notification scheduled")
+                    Logger.dataBrokerProtection.log("PIR user notification scheduled")
```
samsymons, line 113: Just making some log statements a little clearer for my own benefit, since "notification" is a very generic term.

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Pixels/DataBrokerProtectionEventPixels.swift
```diff
@@ -75,7 +75,7 @@ final class DataBrokerProtectionEventPixels {
         handler.fire(.scanningEventNewMatch)
     }
 
-    func fireReAppereanceEventPixel() {
+    func fireReappeareanceEventPixel() {
```
samsymons, line 78: Sorry I normally don't care about things like this but I had to fix the typo 😅

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerOperation.swift
```diff
@@ -64,7 +64,7 @@ class DataBrokerOperation: Operation, @unchecked Sendable {
     private var _isFinished = false
 
     deinit {
-        Logger.dataBrokerProtection.log("Deinit operation: \(String(describing: self.id.uuidString), privacy: .public)")
+        Logger.dataBrokerProtection.log("Deinit DataBrokerOperation: \(String(describing: self.id.uuidString), privacy: .public)")
```
samsymons, line 67: Similar to the notification comment, updating some logs to make it clearer to me what's going on.

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -101,7 +101,8 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
         }
     }
 
-    // swiftlint:disable cyclomatic_complexity
```
samsymons, line 104: This can be deleted now 🫡

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -113,127 +114,85 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
...
-                                                                          isImmediateOperation: isManual)
+        let stageCalculator = DataBrokerProtectionStageDurationCalculator(
+            dataBroker: brokerProfileQueryData.dataBroker.name,
+            dataBrokerVersion: brokerProfileQueryData.dataBroker.version,
+            handler: pixelHandler,
+            isImmediateOperation: isManual
+        )
 
         do {
+            // 3. Record the start of the scan job:
```
samsymons, line 140: I added comments each step of the way, I don't know that these will be necessary once this class is broken up in a more considered way, but I find them helpful for the time being.

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -113,127 +114,85 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
...
-                                                          sevenDaysConfirmationPixelFired: false,
-                                                          fourteenDaysConfirmationPixelFired: false,
-                                                          twentyOneDaysConfirmationPixelFired: false)
-
-                        try database.saveOptOutJob(optOut: optOutJobData, extractedProfile: extractedProfile)
-
-                        Logger.dataBrokerProtection.log("Creating new opt-out operation data for: \(String(describing: extractedProfile.name))")
-                    }
-                }
+            Logger.dataBrokerProtection.log("OperationManager found profiles: \(profilesFoundDuringCurrentScanJob)")
```
quanganhdo:
```suggestion
            Logger.dataBrokerProtection.log("OperationManager found profiles: \(profilesFoundDuringCurrentScanJob, privacy: .public)")
```

Nit

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/UIWeb/UIMapper.swift
THISISDINOSAUR: Hurray! Did you look at any of the rest of UIMapper? It's not my favorite class 😅

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -113,127 +114,85 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
                                    shouldRunNextStep: @escaping () -> Bool) async throws {
         Logger.dataBrokerProtection.log("Running scan operation: \(brokerProfileQueryData.dataBroker.name, privacy: .public)")
 
-        guard let brokerId = brokerProfileQueryData.dataBroker.id, let profileQueryId = brokerProfileQueryData.profileQuery.id else {
+        // 1. Validate that the broker and profile query data objects each have an ID:
```
THISISDINOSAUR, line 117: I'm curious to know when this wouldn't be the case. I'd potentially be up for removing this depending on the answer (although would likely require further changes).
I guess what unsettles me is, how did we get to this stage with that being the case? 

It's possibly that we never do. I'm perhaps overly alert to it because the codebase has a lot of other instances of "defensive" programming, some that were a much bigger deal, so it's perhaps worth validating. 

At any rate, if this should happen, I think we should handle it differently. If it shouldn't happen, I think we should handle it differently (e.g. at least an assert)

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -113,127 +114,85 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
...
 
         defer {
             try? database.updateLastRunDate(Date(), brokerId: brokerId, profileQueryId: profileQueryId)
-            Logger.dataBrokerProtection.log("Finished scan operation: \(brokerProfileQueryData.dataBroker.name, privacy: .public)")
             notificationCenter.post(name: DataBrokerProtectionNotifications.didFinishScan, object: brokerProfileQueryData.dataBroker.name)
+            Logger.dataBrokerProtection.log("Finished scan operation: \(brokerProfileQueryData.dataBroker.name, privacy: .public)")
         }
 
+        // 2. Set up dependencies used to report the status of the scan job:
         let eventPixels = DataBrokerProtectionEventPixels(database: database, handler: pixelHandler)
```
THISISDINOSAUR, line 131: Not directly relevant to this PR, but I think part of the problem is some of these classes have confusing names. E.g. EventPixels is not remotely what I would have expected from the name

StageDurationCalculator too I think is a bit weird, since that describes one of the things it does to send the pixels it actually cares about. I.e. it's not named after the the reason it exists/utility it actually provides. We don't seem to use the stage duration from this anywhere else. And what's worse, in at least one place we seem to calculate it again separately

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -256,12 +215,150 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
             throw error
         }
     }
-    // swiftlint:enable cyclomatic_complexity
 
-    private func sendProfileRemovedNotificationIfNecessary(userNotificationService: DataBrokerProtectionUserNotificationService, database: DataBrokerProtectionRepository) {
+    private func scheduleExtractedProfilesForOptOut(extractedProfiles: [ExtractedProfile],
```
THISISDINOSAUR: I find this name slightly confusing to read but it might just be me? I think something like `scheduleOptOutsForExtractedProfiles(_ extractedProfiles...` might be better

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
```diff
@@ -256,12 +215,150 @@ struct DataBrokerProfileQueryOperationManager: OperationsManager {
...
+                                                    eventPixels: DataBrokerProtectionEventPixels,
+                                                    stageCalculator: DataBrokerProtectionStageDurationCalculator) throws {
+        stageCalculator.fireScanSuccess(matchesFound: extractedProfiles.count)
+
+        let event = HistoryEvent(
+            brokerId: brokerId,
+            profileQueryId: profileQueryId,
+            type: .matchesFound(count: extractedProfiles.count)
+        )
+        try database.add(event)
```
THISISDINOSAUR: I'd be inclined to move this and the pixel firing above out of this function since it's an unrelated/unexpected side effect

THISISDINOSAUR (approved):
Testing LGTM! 
I've added a few comments. Only a couple are actionable, the rest are just thoughts on codebase

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/UIWeb/UIMapper.swift
samsymons: Not in too much depth yet but I do want to poke around in there 😄

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
samsymons: Yeah this is something I want to change as part of the larger refactoring, I don't know that this function necessarily needs to care about this particular case, it should really be an up front prerequisite to _all_ jobs and considered a high priority issue if we ever detect it in a pixel, based on my understanding that these should always have IDs.

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
samsymons: Yeah good call, this is definitely a bit backwards on a subsequent reading - lemme tweak this

samsymons: Renamed ✅

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Shared/Operations/DataBrokerProfileQueryOperationManager.swift
samsymons: Yeah very good call, I think we'll end up doing that a lot as part of the TD too but this is a good change to make now ahead of time, lemme do that

samsymons: Okay, done - updated the comments as well

PR #183, opened 2025-03-10
Authv2 subscription recovery

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1209568624984667
CC: @miasma13 

### Description

Major improvements:
- OAuthClient is now an actor, this mitigates an issue where multiple get/refresh token attempts were made in parallel and caused useless API calls and creation of multiple tokens.
- Full implementation of `DeadTokenRecoverer`, class in charge of automatically recovering a subscription in case of expired refresh token.
- Solved an issue where, in case of no connection at the app startup, StorePurchaseManagerV2 / availableProducts were not updated and every subsequent purchase attempt failed. Now the products are updated every time requested, if not previously available.
- `currentSubscriptionFeatures()` and `subscriptionManager.isFeatureAvailableForUser(...)` can now propagate an error correctly, this improved all entitlements check allowing the SubscriptionManager to detect fetch error and distinguish between a real lack of entitlements or a simple network failure (example: `macOS/DuckDuckGoVPN/DuckDuckGoVPNAppDelegate.swift`)

### Testing Steps

No-connection tests:
- subscribe to privacy pro
- kill the app
- disable connection (airplane mode on)
- open the app
- re-enable connection
- All should work as expected, a part [this bug](https://app.asana.com/0/1205842942115003/1209608044788440)

get Subscription failures
- subscribe to privacy pro
- while in the app, disable connection
- navigate the app settings and the subscription sub pages, all subscription entitlements shouldn't change and the subscription status shouldn't change

---
###### Internal references:


BrowserServicesKit/Tests/CommonTests/Extensions/DateExtensionTest.swift
```diff
@@ -79,19 +79,19 @@ final class DateExtensionTests: XCTestCase {
         let startOfDayTomorrow = Date.startOfDayTomorrow
         let tomorrow = Calendar.current.date(byAdding: .day, value: 1, to: Date())!
 
-        XCTAssertEqual(startOfDayTomorrow, Calendar.current.startOfDay(for: tomorrow))
+        XCTAssertEqual(startOfDayTomorrow.timeIntervalSince1970, Calendar.current.startOfDay(for: tomorrow).timeIntervalSince1970, accuracy: 5)
```
federicocappelli, line 82: Fixes for random test failures

BrowserServicesKit/Tests/NetworkingTests/Auth/OAuthClientTests.swift
```diff
@@ -47,18 +47,22 @@ final class OAuthClientTests: XCTestCase {
     // MARK: -
 
     func testUserNotAuthenticated() async throws {
-        XCTAssertFalse(oAuthClient.isUserAuthenticated)
+        let authenticated = await oAuthClient.isUserAuthenticated
```
federicocappelli, line 50: changes due to the transition of oAuthClient to Actor

miasma13:
@federicocappelli Can you please clarify scope of this PR and changes for testing, is it iOS only (title and branch name suggests so) or both platforms?

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -590,7 +590,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
         Logger.networkProtection.log("Load auth token")
         switch options.authToken {
         case .set(let newAuthToken):
-            Logger.networkProtection.log("Set new token: \(newAuthToken)")
+            Logger.networkProtection.log("Set new token")
```
miasma13, line 593: Can you take a look at the `no 'async' operations occur within 'await' expression` warnings below. They are not introduced by changes in this PR but an earlier one. Seems that the closure is `async` for macOS and not on iOS.

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -20,28 +20,42 @@ import Foundation
...
-            break
-        case .failure:
-            throw SubscriptionManagerError.tokenUnRefreshable
-        }
+    private static func reportFailure() throws {
+        recoveryAttemptCount = 0
+        throw SubscriptionManagerError.tokenUnRefreshable
+    }
+
+    public static func attemptRecoveryFromPastPurchase(subscriptionManager: any SubscriptionManagerV2) async throws {
```
miasma13: I find this function confusing. Since there is no option to even attempt recovery let's name it appropriately. Also do we need a separate function for that as I see logic above that checks that.

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -20,28 +20,42 @@ import Foundation
...
 
-        switch await restoreFlow.restoreAccountFromPastPurchase() {
-        case .success:
-            break
-        case .failure:
-            throw SubscriptionManagerError.tokenUnRefreshable
-        }
+    private static func reportFailure() throws {
+        recoveryAttemptCount = 0
+        throw SubscriptionManagerError.tokenUnRefreshable
```
miasma13: Is this a correct error to be thrown here? This is being called when the `restoreFlow.restoreAccountFromPastPurchase` fails but can do so to multiple reasons.

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -20,28 +20,42 @@ import Foundation
...
+        switch subscriptionManager.currentEnvironment.purchasePlatform {
+        case .appStore:
+            switch await restoreFlow.restoreAccountFromPastPurchase() {
+            case .success:
+                break
+            case .failure:
+                try reportFailure()
+            }
+        case .stripe:
+            Logger.subscription.debug("Subscription purchased via Stripe can't be restored automatically, notifying the user...")
```
miasma13, line 44: Shouldn't we also sign out user here?

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -1624,6 +1626,13 @@ class MainViewController: UIViewController {
             }
             .store(in: &vpnCancellables)
 
+        NotificationCenter.default.publisher(for: .expiredRefreshTokenDetected)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] notification in
+                self?.onExpiredRefreshTokenDetected(notification)
+            }
+            .store(in: &vpnCancellables)
```
miasma13, line 1656: Not sure this is the right place to put it since it is `subscribeToNetworkProtectionEvents`

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -291,6 +291,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
 
+            // Expired refresh token recovery
+            if #available(iOS 15.0, macOS 12.0, *) {
+                let restoreFlow = DefaultAppStoreRestoreFlowV2(subscriptionManager: subscriptionManager, storePurchaseManager: subscriptionManager.storePurchaseManager())
+                subscriptionManager.tokenRecoveryHandler = {
+                    try await DeadTokenRecoverer.attemptRecoveryFromPastPurchase(subscriptionManager: subscriptionManager, restoreFlow: restoreFlow)
+                }
+            } else {
+                subscriptionManager.tokenRecoveryHandler = {
+                    try await DeadTokenRecoverer.attemptRecoveryFromPastPurchase(subscriptionManager: subscriptionManager)
```
miasma13: Do we need this call here, for the non supported OS (which is now only macOS under 12.0) this will be a no-op.

BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -367,18 +377,23 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
...
     func attemptTokenRecovery() async throws -> TokenContainer {
         Logger.subscription.log("The refresh token is expired, attempting subscription recovery...")
+        pixelHandler(.deadToken)
         await signOut(notifyUI: false)
         do {
-            try await autoRecoveryHandler()
+            try await tokenRecoveryHandler?()
             return try await getTokenContainer(policy: .local)
         } catch {
             throw SubscriptionManagerError.tokenUnRefreshable
```
miasma13: Related to earlier comment, I see more `SubscriptionManagerError.tokenUnRefreshable` to be `SubscriptionManagerError.tokenRefreshFailed`. As it not necessary means that it unrefreshable but the recovery failed due to one of the reasons.

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -20,28 +20,42 @@ import Foundation
...
-        guard subscription.platform == .apple else {
-            throw SubscriptionManagerError.tokenUnRefreshable
+        switch subscriptionManager.currentEnvironment.purchasePlatform {
+        case .appStore:
+            switch await restoreFlow.restoreAccountFromPastPurchase() {
+            case .success:
+                break
+            case .failure:
+                try reportFailure()
+            }
```
miasma13: **Important:** I am not sure this is all inherent logic of this call is something we expect here. For example:
- `restoreFlow.restoreAccountFromPastPurchase` only succeeds when the currently restored subscription is active
- in case of successful restore but with an inactive subscription the current user is silently signed out from the subscription

miasma13 (commented):
Please find some comments about the implementation, we can sync for clarification. I still need to do the round of testing.

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: function renamed and improved

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: This is tricky, the recovery process is a best effort and done only once after we detect an expired refresh token, before executing it the user is logged out so there's no second attempt, if the recovery fail once, for any reason, there's no recovery and the suer sees the subscription disappear.
So we don't really care which error is causing the failure, but for clarity reasons I added an underlying error to SubscriptionManagerError.tokenUnRefreshable

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: The sign out is done in the handler caller, I removed the left over here

iOS/DuckDuckGo/MainViewController.swift
federicocappelli: entitlementsDidChange and accountDidSignOut are also in the function, and they are subscription and authentication notifications, I could create a new subscription only function, but I don't think really matters

macOS/DuckDuckGo/Application/AppDelegate.swift
federicocappelli: the call is just for logging the issue and sending the notification, nothing too bad

BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: Good point, error renamed

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: I see what you mean, my initial thought was that restoring an expired subscription was not very useful, but it could be good to show the expired sub UI to the user.... I've modified the call for using a custom restore subscription that doesn't check if the subscription is expired and doesn't signs off the user

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: The macOS function is asynchronous and the iOS one is synchronous, I've duplicated it and fixed it.

BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
miasma13: My understanding is that the purpose of `DeadTokenRecoverer` is to attempt to re-authenticate when we detect in rare case that the refresh token is expired. In that sense I would disconnect it from making any assumptions of subscription state, in the end even if it expired we want to keep the user signed in with the expired state.

Another idea, instead of using `restoreFlow` (which has its own logic for the restore scenario), just implement it in place in `DeadTokenRecoverer` the steps to fetch JWT and use it to authenticate?

federicocappelli: I agree, I've implemented a custom function that ignores the subscription state and succeeds anyway, even if the subscription is expired.
I did it in the flow for dependency injection and for mockability reasons, moving that function in DeadTokenRecoverer would create a cascade of other changes

miasma13 (approved):

PR #208, opened 2025-03-12
Per-window Pinned Tabs

tomasstrba:
Task/Issue URL: https://app.asana.com/0/1148564399326804/1208198115490419
Tech Design URL: https://app.asana.com/0/1148564399326804/1209368276431904

### Description
We are enhancing the pinned tabs mechanism, allowing them to be specific to individual windows in addition to being shared across all application windows
The behavior of pinned tabs can be adjusted in General Settings
New users will have per-window pinned tabs as the default setting.
Users currently using shared pinned tabs will keep this behavior. However, after pinning another tab, they will see a discovery dialog introducing the option to switch to per-window pinned tabs

### Testing Steps

**Test new users have per-window pinned tabs set as default Setting**
1. Execute `./macOS/clean-app.sh debug`
2. Run the app and finish the onboarding process
3. Pin a couple of tabs in the current window
5. Open a new window and verify there are no pinned tabs
6. Test pinning a different set of tabs in the new window
7. Go to Settings -> General and verify the pinned tab setting is set to "Separate in each window"

**Test pinned tabs are merged correctly when switching to shared pinned tabs**
1. With the state from the previous test with multiple windows containing different set of pinned tabs, go to Settings -> General
2. Switch the pinned tabs setting to "Shared across all windows"
3. Verify you received an alert (warning) about pinned tabs being merged
4. Confirm the change in the alert
5. Verify the current window contains all pinned tabs merged
6. Verify the other window contains the same pinned tabs
7. Make some changes to pinned tabs and make sure they are applied to all windows

**Test switching from shared pinned tabs to per-window pinned tabs**
1. With the state from the previous test, go to Settings -> General
2. Switch the pinned tabs setting to "Separete in each window"
3. Verify the current window contains the same set of pinned tabs.
4. Verify the other window has no pinned tabs

**Test restoration of per window pinned tabs when the session restoration is turned off**
1. In Settings -> General, set "Open a new window" for the session restoration setting and "Separate in each window" for pinned tabs setting
2. Open a bunch of windows with a couple of tabs and set some pinned tabs in each window.
3. Quit the app
4. Open the app and verify all pinned tabs have been restored and no regular tabs have been restored

**Test restoration of per window pinned tabs when the session restoration is turned on**
1. In Settings -> General, set "Reopen all windows from last session" for the session restoration setting and "Separate in each window" for pinned tabs setting
2. Open a bunch of windows with a couple of tabs and set some pinned tabs in each window.
3. Quit the app
4. Open the app and verify all windows with regular and pinned tabs have been restored.

**Test restoration of shared pinned tabs with the various session restoration settings**
1. In Settings -> General, set "Open a new window" for the session restoration setting and "Shared across all windows" for pinned tabs setting
2. Open a bunch of windows with a couple of tabs and create some shared pinned tabs.
3. Quit the app
4. Open the app and verify shared pinned tabs have been restored correctly
5. Repeat the test with the "Reopen all windows from last session" session restoration setting and verify windows are restored with both regular and pinned tabs correctly

**Test 'Reset Pinned Tabs' when per-window pinned tabs are enabled**
1. Set the "Separate in each window" pinned tabs setting
2. Open multiple windows with some pinned tabs in each window
3. Go to Debug -> Reset Data -> Reset Pinned Tabs
4. Make sure all pinned tabs have been cleared

**Test 'Reopen last closed tab' when per-window pinned tabs are enabled**
1. Set the "Separate in each window" pinned tabs setting
2. Open multiple windows with some pinned tabs in each window
3. Close a pinned tab
4. Go to History -> Reopen Last Closed Tab
5. Verify the pinned tab has been added back
6. Close a pinned tab again
7. Switch to the other window
8. Go to History -> Reopen Last Closed Tab
9. Verify the pinned tab has been added to the new window

**Test Fire button when per-window pinned tabs are enabled**
1. Set the "Separate in each window" pinned tabs setting
2. Open a window and pin some tabs
3. Burn a pinned tab
4. Make sure the tab is still present in the window and it has been reloaded
5. Burn the current window
6. Make sure the set of pinned tabs has been preserved after burning
7. Open a couple of more windows and pin some tabs
8. Burn all data
9. Make sure all pinned tabs sets have been preserved. It means, all tabs have been closed except pinned tabs. (In case a window didn't contain any pinned tab, it's gone)

**Test migration of shared pinned tabs for existing users**
1. Execute `./macOS/clean-app.sh debug`
2. Switch to the `main` branch
4. Run the app and pin a couple of tabs
5. Quit the application
6. Switch back to the `tom/per-window-pinned-tabs` branch
7. Run the app
8. Make sure the same pinned tabs are present and they are shared across browser windows
9. Optionally, verify the pinned tabs setting is set to "Shared across all windows"

**Test the Discovery dialog**
1. With the state from the end of the previous test, pin another tab
10. Verify the app presented a discovery dialog - a pop up dialog pointing to the new pinned tabs allowing user to switch to per-window pinned tabs
11. Switch to per-window pinned tabs using the dialog
12. Verify the pinned tabs are now separate in each window

**Test back/forward list is not preserved in session state data**
1. In a pinned tab, navigate to couple of URLs to create back/forward list of websites
2. Quit the app
3. Start the app again and verify the back/forward list is empty for the pinned tab

**Test pixels**
1. Test `userPinnedTab` is fired when user pins a tab
2. Test `userUnpinnedTab` is fired when user unpins a tab
3. Test  `userSwitchedToSharedPinnedTabs` is fired when user switches to shared pinned tabs
4. Test  `userSwitchedToPerWindowPinnedTabs` is fired when user switches to per-window pinned tabs
5. Make sure `userPinnedTab` is not fired when migration is performed from shared to per-window and back

### Impact and Risks
High

### Optional E2E tests**:
- [ ] Run macOS PIR E2E tests

---
###### Internal references:


macOS/DuckDuckGo/PinnedTabs/Model/PinnedTabsManagerProvider.swift
```diff
@@ -0,0 +1,197 @@
...
+    private func getSharedPinnedTabManager(shouldMigrate: Bool) -> PinnedTabsManager {
+        if shouldMigrate {
+            migrateAllPerWindowPinnedTabsToShared()
+        }
+        return sharedPinnedTabsManager
+    }
+
+    @MainActor
+    private func migrateAllPerWindowPinnedTabsToShared() {
+        let allTabs = perWindowPinnedTabsManagers.flatMap { $0.tabCollection.tabs }.uniqued()
```
ayoy: @tomasstrba I wanted to bring your attention to the uniquing mechanism here. I have a few problems with it:

This calls `Array(Set(self))` under the hood, which randomizes the collection. I’ve merged 2 windows and went from this state:
Screenshot 2025-03-20 at 22 38 36
to this state:
Screenshot 2025-03-20 at 22 38 55
Notice that R goes before TP and there’s youtube in between them.

Secondly, I suppose that uniquing isn’t really needed here – for Tab instance it wouldn’t remove “duplicates” because there are no 2 identical tab instances (assuming they’re hashed using `ObjectIdentifier` via `Identifiable` protocol). Have you considered flattening the array without calling `uniqued()`?

tomasstrba: @ayoy Great catch! The use of Array(Set(self)) on Tabs came from an AI-assisted refactor, and I didn’t realize it would both randomize the collection and be unnecessary for the Tab class. Previously, the deduplication was focused on URLs.

I agree with you. Let’s skip deduplication entirely here, as the likelihood of duplicate URLs across different windows is low. I'll refactor this later today

macOS/DuckDuckGo/PinnedTabs/Model/PinnedTabsManagerProvider.swift
```diff
@@ -0,0 +1,197 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Combine
+
+protocol PinnedTabsManagerProviding {
+
+    /// True if per-window pinned tabs are enabled in Settings
+    var arePerWindowPinnedTabsEnabled: Bool { get }
```
ayoy: I noticed that this property is used externally only with a negation, which may be tricky to quickly understand for the reader. Have you considered using a 2-case enum to make it clear what is it about?

E.g:
```swift
enum PinnedTabsMode {
    case perWindow, applicationWide
}

protocol PinnedTabsManagerProviding {
    var pinnedTabsMode: PinnedTabsMode { get }
    // …
}
```

macOS/DuckDuckGo/PinnedTabs/Model/PinnedTabsManagerProvider.swift
```diff
@@ -0,0 +1,197 @@
...
+final class PinnedTabsManagerProvider: @preconcurrency PinnedTabsManagerProviding {
+
+    private let tabsPreferences: TabsPreferences
+    private var closedWindowPinnedTabCache: TabCollection?
+
+    var settingChangedPublisher: AnyPublisher<Void, Never>
+
+    init(tabsPreferences: TabsPreferences = TabsPreferences.shared) {
+        self.tabsPreferences = tabsPreferences
+        self.settingChangedPublisher = tabsPreferences.$pinnedTabsMode
```
ayoy, line 62: In case you really are interested in “changed” events, then add `.dropFirst()` after the publisher. Otherwise a subscription to a published property will emit the initial value too. While this doesn’t have to be buggy, it’s sometimes unnecessary.

macOS/DuckDuckGo/PinnedTabs/View/PinnedTabsDiscoveryPopUpView.swift
```diff
@@ -0,0 +1,81 @@
...
+
+                createButton(title: UserText.pinnedTabsDiscoveryPopoverSeparate,
+                             style: StandardButtonStyle()) {
+                    setPerWindowPinnedTabs()
+                    callback?(true)
+                }
+            }.frame(width: Constants.panelWidth - 2 * Constants.verticalSpacing)
+        }
+        .padding()
+        .frame(width: Constants.panelWidth, height: Constants.panelHeight)
```
ayoy: I realize this is copied over from the AI Chat popup, but I’d have a few suggestions for you - they may be needed when you receive translations and see that the view is off:
1. We shouldn’t use fixed width and height here (or at least the height). Instead, try using `minWidth` and `minHeight`. For HStack and VStack, instead of setting frame you can set `padding(.vertical, Constants.verticalSpacing)`
2. Even if we keep width and height fixed, text views may break when text is longer than expected (see below). To fix them, use `.fixMultilineScrollableText()` (the name is misleading and should be changed, but it does the job). I couldn’t quickly figure out how to go about resizing the buttons to show the entire text. Considering that translations for button titles may be lengthy (in French or Polish, for instance), we might consider laying out the buttons vertically. Just a food for thought though, maybe you could bring this up with the designer.
Screenshot 2025-03-21 at 14 58 40

macOS/DuckDuckGo/RecentlyClosed/Model/RecentlyClosedCoordinator.swift
```diff
@@ -91,6 +102,15 @@ final class RecentlyClosedCoordinator: RecentlyClosedCoordinating {
             .store(in: &cancellables)
     }
 
+    private func subscribeToPinnedTabsSettingChanged() {
+        pinnedTabsManagerProvider.settingChangedPublisher
+            .dropFirst()
```
ayoy: Ah :D

macOS/DuckDuckGo/StateRestoration/AppStateRestorationManager.swift
```diff
@@ -88,7 +88,7 @@ final class AppStateRestorationManager: NSObject {
     // Cleans all stored snapshots except snapshots listed in the state
     func cleanTabSnapshots(state: WindowManagerStateRestoration? = nil) {
         let tabs = state?.windows.flatMap { $0.model.tabCollection.tabs } ?? []
-        let pinnedTabs = state?.pinnedTabs?.tabs ?? []
+        let pinnedTabs = state?.applicationPinnedTabs?.tabs ?? []
```
ayoy: So here we’re excluding app-wide pinned tabs from removing snapshots. Does this need any special handling for per-window pinned tabs?

macOS/DuckDuckGo/StateRestoration/WindowManager+StateRestoration.swift
```diff
@@ -136,12 +161,14 @@ final class WindowRestorationItem: NSObject, NSSecureCoding {
         static let frame = "frame"
         static let model = "model"
         static let isMiniaturized = "isMiniaturized"
+        static let pinnedTabs = "pinned_tabs"
```
ayoy: Nit: should we call it `pinnedTabs` to follow `isMiniaturized`?

macOS/DuckDuckGo/Tab/Model/Tab.swift
```diff
@@ -629,6 +629,12 @@ protocol NewWindowPolicyDecisionMaker {
     }
 
     func getActualInteractionStateData() -> Data? {
+        if let pinnedTabsManager = pinnedTabsManagerProvider.pinnedTabsManager(for: self),
+           pinnedTabsManager.isTabPinned(self) {
+            // To optimize the performance, don't save interaction state data for pinned tabs
```
ayoy, line 638: 💪

ayoy (commented):
@tomasstrba I’m halfway through the review, so far the testing goes well and I decided to look at the code. I’m leaving a bunch of comments, but what’s really the most important is the onboarding dialog UI. Otherwise looking great so far! I’ll finish it off Monday morning.

macOS/DuckDuckGo/StateRestoration/WindowManager+StateRestoration.swift
tomasstrba: It's already bad. There is `key_idx` in WindowManagerStateRestoration and I copied `pinned_tabs` from there. When aligned now, there are both `pinned_tabs` and `pinnedTabs`. I think a good solution doesn't exist here. :D (Alignment of all would cause trouble for current users)

macOS/DuckDuckGo/PinnedTabs/View/PinnedTabsDiscoveryPopUpView.swift
tomasstrba: Good spot! I usually verify this when translations are ready. Thanks for all tips, I'll post a comment here when I have translations and the issue is resolved

macOS/DuckDuckGo/StateRestoration/AppStateRestorationManager.swift
tomasstrba: I overlooked this. It's resolved now, per-window pinned tabs are included in the exceptions too 👍

macOS/DuckDuckGo/Tab/Model/Tab.swift
tomasstrba: 💪

macOS/DuckDuckGo/PinnedTabs/Model/PinnedTabsManagerProvider.swift
tomasstrba: Good idea 👍

macOS/DuckDuckGo/PinnedTabs/Model/PinnedTabsManagerProvider.swift
tomasstrba: Yeah I think you are right. I reused `PinnedTabsMode` from `TabsPreferences`

macOS/DuckDuckGo/StateRestoration/WindowManager+StateRestoration.swift
ayoy: Right :) let’s keep it as is then :)

macOS/DuckDuckGo/Preferences/View/PreferencesGeneralView.swift
```diff
@@ -43,6 +43,14 @@ extension Preferences {
         @State private var showWarningAlert = false
         @State private var pendingSelection: PinnedTabsMode?
 
+        private func firePinnedTabsPixel(_ newMode: PinnedTabsMode) {
+            if newMode == .shared {
+                PixelKit.fire(GeneralPixel.userSwitchedToSharedPinnedTabs)
```
ayoy: 2 ideas for pixels:
1. I can’t find the relevant Apple Team proposal now, but here’s [one for the Windows Team](https://app.asana.com/0/1203533013062555/1208795557996921/f) – the pattern I’ve been following recently is to use daily + standard mode for all newly added pixels. This way you get 2 metrics: the number of occurrences of an event, and the number of users triggering that event daily. [Here’s](https://app.asana.com/0/1201899738287924/1209364382402737/f) the recent discussion about History View pixels where I noted that all pixels are daily + standard.

```suggestion
                PixelKit.fire(GeneralPixel.userSwitchedToSharedPinnedTabs, frequency: .dailyAndStandard)
```

2. This is optional, but I think helps to keep our Pixels in order. We’ve had [another proposal](https://app.asana.com/0/1201011656765697/1208001594278408/f) (that, notably, haven’t left the draft phase yet but Fernando promised to finish it later today :D) about keeping pixels related to a specific domain/feature in separate file. I think Pinned tabs could use it. Have a look at all `*Pixel.swift` files inside Statistics directory (e.g. ContextualOnboardingPixel.swift). It’s quite straightforward to add a new enum with a bunch of pixels and then use it instead of GeneralPixel.

ayoy (commented):
@tomasstrba the updates are looking great, thank you!

Just leaving a comment/suggestion about pixels for now. What’s left for me is to do a bit more testing and I’d consider it good to go. Will get back to it today evening or tomorrow the latest if that’s fine. Thanks!

macOS/DuckDuckGo/Preferences/View/PreferencesGeneralView.swift
tomasstrba: Done ✔️

macOS/DuckDuckGo/PinnedTabs/View/PinnedTabsDiscoveryPopUpView.swift
tomasstrba: I'm still waiting for translations but decided to resolve this to speed things up. 
The solution focuses on setting the panel width based on the real width of buttons. It works well for any size of buttons and texts.

Screenshot 2025-03-25 at 11 50 28
Screenshot 2025-03-25 at 11 51 30

tomasstrba:
@ayoy, I've fixed the remaining dialog for more languages and translations should be ready today. It would be great to finalize this today, so please let me know if there is anything else you'd like to improve.

I'm now focusing on fixing the UI tests, but that shouldn't touch the code of solution at all

macOS/DuckDuckGo/PinnedTabs/View/PinnedTabsDiscoveryPopUpView.swift
ayoy: This looks great, thanks a lot!

ayoy (approved):
@tomasstrba amazing job on this, I’ve tested it thoroughly and it worked well every time :) thanks!

PR #120, opened 2025-02-28
AuthV2 \ macOS auth v1 and v2

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1209170372758739/f
CC: @miasma13 @diegoreymendez 

### Description

This PR implements the use of AuthV2 in the macOS app.

**The good:**

- AuthV2 is fully integrated into the iOS app.
- The authentication still uses V1 but V2 is configured and added behind an informal feature flag. (see `isAuthV2Enabled`)

**The bad**

- We now have some code duplication, especially in the settings UI, all will be removed alongside AuthV1
- "bridging" protocols (see SubscriptionAuthV1toV2Bridge) used where authv1 and authv2 need to cohexist
- The code is not very clean, I prioritised risk reduction and changes minimisation over code cleanness, all will be improved when we remove AuthV1

### Testing Steps

All tests need to be executed with App Store build and DMG build, both using App Store and Stripe purchase flows

1. Purchase subscription
2. Enable VPN
3. Disable VPN
4. Enable VPN
5. Remove subscription
6. Restore subscription
7. Kill app and enable/disable VPN from the system menu

In general, Subscription should work as usual, no user-facing changes are expected.

### Note to the reviewer

During development and testing, I had a lot of issues with the VPN not being able to restart immediately after being stopped, I worked 3 days on that problem until @diegoreymendez tested my code on his machine and found no issues. If the problem can be replicated during testing (see steps 2,3,4) then we'll need to fix it, if not, then it's just my machine misbehaving.

### Quality Considerations

Most of this code is temporary and will be removed around the 24th of April when AuthV1 will be removed

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #218, opened 2025-03-12
vpn issue fix attempt

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1209654847406840/f
CC: @diegoreymendez  @THISISDINOSAUR @samsymons @miasma13 

### Description

Fix for the token issue introduced in https://github.com/duckduckgo/apple-browsers/pull/120

What happened:
Different parts of the app use different ways to store and retrieve the access token. 
The `DefaultAccountManager` is used in the main app but not in the SysExt where `NetworkProtectionKeychainTokenStore` is used directly. The two classes use different keys for storing the accessToken.
In [a recent PR](https://github.com/duckduckgo/apple-browsers/pull/120) I mistakenly swapped the classes that were initialised side by side and implemented the same protocol (`SubscriptionTokenHandling`) so no errror was detected at build time.

### Testing Steps
1. Subscribe to privacy pro
2. Start VPN
3. Start DBP

---
###### Internal references:


THISISDINOSAUR:
What was the actual change the broke the e2e tests?
And is the use of autoreleasepool necessary?

federicocappelli:
> What was the actual change the broke the e2e tests? And is the use of autoreleasepool necessary?

Sorry I thought you saw @samsymons task: https://app.asana.com/0/1193060753475688/1209661386167901 
The autorelease pool fixes a big memory issue where the test was reaching 20Gb+ of used memory, which creates issues in the GH runner and delays in the test. Now the test ends in 18-19 minutes instead of 28

The real issue for the pir e2e test was in how `DataBrokerProtectionAuthenticationManaging` implements `var isUserAuthenticated`, before it was just checking the token presence (sync) and recently I modified using the SubscriptionManager isUserAuthenticated. This creates a problem because the test token is overwritten locally and the SubscriptionManager is not aware of it. So I reverted the change, but now getToken is async, so I had to use a semaphore for making the call sync.

macOS/DBPE2ETests/DBPEndToEndTests.swift
```diff
@@ -125,12 +125,16 @@ final class DBPEndToEndTests: XCTestCase {
         await awaitFulfillment(of: profileSavedExpectation,
                                withTimeout: 3,
                                whenCondition: {
-            try! database.fetchProfile() != nil
+            autoreleasepool {
```
THISISDINOSAUR: Can we add a comment explaining why this is necessary? Otherwise it's prone to getting deleted by someone in the future.

Side note, I have a vague memory that there was a better way to implement this using Swift Testing, but at the time it wasn't released yet. So in future we might be able to remove the hack

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Authentication/DataBrokerProtectionAuthenticationManaging.swift
```diff
@@ -36,7 +36,15 @@ public final class DataBrokerProtectionAuthenticationManager: DataBrokerProtecti
...
-        subscriptionManager.isUserAuthenticated
+        var token: String?
+        // extremely ugly hack, will be removed as soon auth v1 is removed
+        let semaphore = DispatchSemaphore(value: 0)
+        Task {
+            token = await accessToken()
+            semaphore.signal()
+        }
+        semaphore.wait()
+        return token != nil
```
THISISDINOSAUR, line 47: I think this hack concerns me a little bit. How long is it likely to be before we can remove it?

macOS/DBPE2ETests/DBPEndToEndTests.swift
federicocappelli: comment added

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Authentication/DataBrokerProtectionAuthenticationManaging.swift
federicocappelli: As soon AuthV2 is released and we remove AuthV1 code, this is the task: https://app.asana.com/0/72649045549333/1209279739041315 by current estimation this should be done around the 14/04

diegoreymendez:
> I moved DefaultAccountManager+SubscriptionTokenHandling into the iOS app, where it is correctly used. This will prevent swapping NetworkProtectionKeychainTokenStore with DefaultAccountManager in future.

Is this change pushed?  I'm not seeing it in the diff.

diegoreymendez:
@federicocappelli  - The VPN connects but there's the second issue [I reported here](https://app.asana.com/1/137249556945/task/1209654847406840/comment/1209666419198790?focus=true) that I'd like to be addressed too - we shouldn't have assertion failures on known and "accepted" error conditions.

federicocappelli:
@diegoreymendez Assert removed and error handling modified accordingly

diegoreymendez (approved):
Looks good from the VPN side.

PR #222, opened 2025-03-12
Fix E2E test memory usage for PIR

samsymons:
Task/Issue URL: https://app.asana.com/0/1199333091098016/1209661386167901
Tech Design URL:
CC:

### Description

This PR reduces PIR E2E test memory usage, as it's likely causing issues with the runner images. The issue is related to the `awaitFulfillment` blocks in the suite, which run their block repeatedly until they either get a success or time out.

The fix is to wrap these checks in an `autoreleasepool` to force the memory to be released. After this change, the suite never consumes more than 60MB for me at its peak, down from multiple gigabytes.

### Testing Steps

1. Check that CI is green - note that I'm not sure that we don't also have other issues here with Auth v2 so green might not be possible yet, but this change should at least help overall

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/macOS/DBPE2ETests/DBPEndToEndTests.swift
+++ b/macOS/DBPE2ETests/DBPEndToEndTests.swift
@@ -125,12 +125,16 @@ final class DBPEndToEndTests: XCTestCase {
         await awaitFulfillment(of: profileSavedExpectation,
                                withTimeout: 3,
                                whenCondition: {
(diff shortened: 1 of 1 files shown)


samsymons:
Closing this since the fix has been added to Federico's branch, and this change isn't enough to fix PIR tests.

PR #225, opened 2025-03-13
Card like webview

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201392122292466/task/1209485028986793
CC: @dus7 

### Description
1. Apply rounded rect corners to WebView to make it look like a card
2. Fix glitch when swiping through tabs

### Testing Steps
1. Make sure nothing is changed when experimentalBrowserTheming is disabled
3. Turn on experimentalBrowserTheming in debug menu and restart the app
4. Notice the webview has rounded corners and validate against designs (e.g. top bar, bottom bar, keyboard open, keyboard hidden, portrait vs landscape etc.)
5. Notice the bug has been fixed when swiping tabs when omnibar is at the bottom. (previously the container was wrongly calculated so the snapshot image was unnecessarily enlarged)
6. Make sure pull to refresh works as before when experimentalBrowserTheming is disabled
7. Make sure to test thoroughly new pull to refresh behavior when experimentalBrowserTheming is enabled - horizontally, vertically, drag from different places, drag and do not let go, drag on sites that load slowly multiple times in a row, etc.

### Impact and Risks
<!-- 
Low: Corners should not be visible for users, but perhaps the swiping could be broken in some scenario which I don't know of


brindy (approved):
LGTM - tested it with flags on/off etc.  Didn't re-test after the name change, but that looks good to me assuming the build goes green.

PR #228, opened 2025-03-13
Fix duck.ai downloads

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201011656765697/task/1209638221458142

### Description
Fix chat download button not working in the custom duck.ai view

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


iOS/LocalPackages/AIChat/Sources/AIChat/Views/ToastView.swift
```diff
@@ -0,0 +1,55 @@
+//
+//  ToastView.swift
```
Bunn, line 2: I considered using `ActionMessageView` for this, but given this is a separated package, and `ActionMessageView` is still using .xibs, I’ve decided to create a very simple toast that works only for duck.ai needs. We might even expand this in the future so keeping them separated seems good to me.

iOS/DuckDuckGo/DownloadsDirectoryHandling.swift
```diff
@@ -0,0 +1,67 @@
+//
```
Bunn, line 1: I’ve extracted directory handling functions from `DownloadManager` into this new struct so it can be reused.

iOS/DuckDuckGo/DownloadManager.swift
```diff
@@ -259,3 +225,9 @@ extension NSNotification.Name {
     static let downloadFinished: NSNotification.Name = Notification.Name(rawValue: "com.duckduckgo.notification.downloadFinished")
     static let downloadsDirectoryChanged: NSNotification.Name = Notification.Name(rawValue: "com.duckduckgo.notification.downloadsDirectoryChanged")
 }
+
+
+extension DownloadManager {
```
Bunn, line 230: Forwarding expected properties from `DownloadManager` now that `DownloadsDirectoryHandling` is its own protocol

iOS/LocalPackages/AIChat/Sources/AIChat/AIChatWebViewController.swift
```diff
@@ -57,10 +60,26 @@ final class AIChatWebViewController: UIViewController {
...
+    }
+
+    private func setupDownloadHandler() {
+        downloadHandler.onDownloadComplete = { [weak self] result in
+            guard let self = self else { return }
+
+            switch result {
+            case .success(let filename):
+                self.view.showDownloadCompletionToast(for: filename) {
+                    self.delegate?.aiChatWebViewController(self, didFinishDownload: filename)
```
miasma13: I find the naming of the delegates a bit misleading here:
1. This line is actually an `onButtonTapped` closure of the toast call
2. It means that the `self.delegate?.aiChatWebViewController(self, didFinishDownload: filename)` is not being called when the download has finished but when user tapped "Show" on the toast after the download has finished.

My suggestion would be to rename delegate methods (all in the chain) to denote that it is a request to open downloads.

Bunn: Good point, doing the change now

miasma13 (approved):
Worked well for me both for the Duck.ai chats and regular downloads. Have checked the migration scenario and the Downloads contents was not affected. Code looks solid with one slight suggestion to adjust the delegate function naming.

PR #258, opened 2025-03-18
Malicious Site Protection - Attempt to prevent crashes

alessandroboron:
Task/Issue URL: https://app.asana.com/0/1206329551987282/1209638992004951/f

### Description

This PR attempts to mitigate crashes happening for Malicious Site Protection by:

1. Use `fatalError` if ApplicationSupportDirectory is unavailable when the DataManager is initialised. This is a slight improvement as the `fatalError` message should be available in the crash logs. [fc3b11f](https://github.com/duckduckgo/apple-browsers/commit/fc3b11ff388d280c2c656a58f70c4abd5744c3c9)

2. Ensure Background Tasks identifiers are registered only once. [9914e8a](https://github.com/duckduckgo/apple-browsers/commit/9914e8a23ce2c8dfbbd8e8206833223e357619ad)

3. Prevent starting new dataset updates if there are already requests in flight. [603ec57](https://github.com/duckduckgo/apple-browsers/commit/603ec576f3f6a7dc0589ba0b70df0826928e8c64)

4. Prevent starting/stopping update tasks multiple times if the user continuously switches the toggle flag in the settings by adding a debounce. [74a20d3](https://github.com/duckduckgo/apple-browsers/commit/74a20d3391064122f658e1569cd3d041d9d36027)

5. Serialise the update of datasets when `updateData` is called [514b260](https://github.com/duckduckgo/apple-browsers/pull/258/commits/514b260b2ce82ff338101483b16664238651fd35) 

### Testing Steps
1. Ensure CI is green
2. Smoke test Malicious Site Protection feature:
2.1 Delete the App.
2.2 Ensure datasets are fetched when the app launches
2.3 Ensure [phishing](http://privacy-test-pages.site/security/badware/phishing.html), [malware](http://privacy-test-pages.site/security/badware/malware.html), and [scam](http://privacy-test-pages.site/security/badware/scam.html) (if the feature is enabled)websites are intercepted and blocked

### Impact and Risks
High:  Cold affect or break a core flow

### Quality Considerations
While the new approach reduces the risk of data races by keeping dataset modifications within the actor's isolation, it introduces a small performance tradeoff. Indeed, the actor remains locked during dataset changes, potentially reducing concurrent access in scenarios where large number of objects need to be inserted, updated, or deleted.

### Notes to Reviewer
I was unable to reproduce the crashes. These changes attempt to address potential root causes, such asaces in dataset operations and system-triggered terminations, such as when possible data or background tasks with the same identifiers are registered more than once.

The crashes appear in Sentry with a SIGKILL 0 signature, indicating that the OS terminated the process. Example crash report:

```
Application Specific Information:
SIGKILL 0
```
These reports lack the termination reason that would typically help identify the underlying cause, as described in [Apple's documentation](https://developer.apple.com/documentation/xcode/sigkill). I’m unsure if this is an issue with Apple's reporting or with how we're integrating with Sentry.
 
@mallexxx I’ve implemented the changes we discussed to make the dataset changes as single operation.
@bwaresiak If you could please take a look at the PR that would be great.
@SabrinaTardio could you just quickly check I didn’t affect your scam work since I had some merge conflicts?  

---
###### Internal references:

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
@@ -21,7 +21,7 @@ import os
 
 protocol DataManaging {
     func dataSet<DataKey: MaliciousSiteDataKey>(for key: DataKey) async -> DataKey.DataSet
-    func store<DataKey: MaliciousSiteDataKey>(_ dataSet: DataKey.DataSet, for key: DataKey) async throws
+    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, changeSet: APIClient.ChangeSetResponse<DataKey.DataSet.Element>) async throws
     func preloadData(for threatKinds: [ThreatKind]) async
 }
 
@@ -98,7 +98,10 @@ public actor DataManager: DataManaging {
         return storedDataSet
     }
 
-    func store<DataKey: MaliciousSiteDataKey>(_ dataSet: DataKey.DataSet, for key: DataKey) throws {
+    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, changeSet: APIClient.ChangeSetResponse<DataKey.DataSet.Element>) async throws {
+        var dataSet = self.dataSet(for: key)
+        dataSet.apply(changeSet)
+
         let dataType = key.dataType
         let fileName = fileNameProvider(dataType)
         self.store[dataType] = dataSet
--- a/SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
@@ -98,7 +98,7 @@ public actor DataManager: DataManaging {
         return storedDataSet
     }
 
-    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, changeSet: APIClient.ChangeSetResponse<DataKey.DataSet.Element>) async throws {
+    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, changeSet: APIClient.ChangeSetResponse<DataKey.DataSet.Element>) throws {
         var dataSet = self.dataSet(for: key)
         dataSet.apply(changeSet)
 
--- a/iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
+++ b/iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
@@ -23,6 +23,7 @@ import BackgroundTasks
 import Core
 import Combine
 import BrowserServicesKit
+import CombineSchedulers
 
 protocol MaliciousSiteProtectionDatasetsFetching {
     @MainActor
@@ -37,6 +38,7 @@ final class MaliciousSiteProtectionDatasetsFetcher {
     private let updateManager: MaliciousSiteUpdateManaging
     private let backgroundTaskScheduler: BGTaskScheduling
     private let application: BackgroundRefreshCapable
+    private let preferencesScheduler: AnySchedulerOf<DispatchQueue>
 
     // Avoid multiple background tasks registrations.
     private static var registeredTaskIdentifiers: Set<String> = []
@@ -71,7 +73,8 @@ final class MaliciousSiteProtectionDatasetsFetcher {
         userPreferencesManager: MaliciousSiteProtectionPreferencesProvider,
         dateProvider: @escaping () -> Date = Date.init,
         backgroundTaskScheduler: BGTaskScheduling = BGTaskScheduler.shared,
-        application: BackgroundRefreshCapable = UIApplication.shared
+        application: BackgroundRefreshCapable = UIApplication.shared,
+        preferencesScheduler: AnySchedulerOf<DispatchQueue> = DispatchQueue.main.eraseToAnyScheduler()
     ) {
         self.updateManager = updateManager
         self.featureFlagger = featureFlagger
@@ -79,6 +82,7 @@ final class MaliciousSiteProtectionDatasetsFetcher {
         self.dateProvider = dateProvider
         self.backgroundTaskScheduler = backgroundTaskScheduler
         self.application = application
+        self.preferencesScheduler = preferencesScheduler
     }
 }
 
@@ -140,6 +144,7 @@ private extension MaliciousSiteProtectionDatasetsFetcher {
 
         preferencesManagerCancellable = userPreferencesManager
             .isMaliciousSiteProtectionOnPublisher
+            .debounce(for: 0.5, scheduler: preferencesScheduler)
             .scan((previous: isMaliciousSiteProtectionOn, current: isMaliciousSiteProtectionOn)) { accumulated, newValue in // Get the previous value of the publisher
                 (accumulated.current, newValue)
             }
--- a/iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
+++ b/iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
@@ -25,10 +25,12 @@ import Combine
 import BrowserServicesKit
 
 protocol MaliciousSiteProtectionDatasetsFetching {
-    func startFetching()
+    @MainActor
+    @discardableResult
+    func startFetching() -> Task<Void, Error>
 }
 
-final class MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDatasetsFetching {
+final class MaliciousSiteProtectionDatasetsFetcher {
     private let featureFlagger: MaliciousSiteProtectionFeatureFlagger & MaliciousSiteProtectionFeatureFlagsSettingsProvider
     private let userPreferencesManager: MaliciousSiteProtectionPreferencesProvider
     private let dateProvider: () -> Date
@@ -39,16 +41,23 @@ final class MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDatas
     // Avoid multiple background tasks registrations.
     private static var registeredTaskIdentifiers: Set<String> = []
 
+    @MainActor
+    private(set) var inFlyUpdateTasks: [DataManager.StoredDataType.Kind: Task<Void, any Error>] = [:]
+
     private var preferencesManagerCancellable: AnyCancellable?
     private var featureFlagOverrideCancellable: AnyCancellable?
 
+    @MainActor
     private var shouldUpdateHashPrefixSets: Bool {
         // Absolute interval to avoid never updating the dataset if the `lastHashPrefixSetUpdateDate` is mistakenly set in the far future
+        inFlyUpdateTasks[.hashPrefixSet] == nil &&
         abs(dateProvider().timeIntervalSince(updateManager.lastHashPrefixSetUpdateDate)) > .minutes(featureFlagger.hashPrefixUpdateFrequency)
     }
 
+    @MainActor
     private var shouldUpdateFilterSets: Bool {
         // Absolute interval to avoid never updating the dataset if the `lastFilterSetUpdateDate` is mistakenly set in the far future
+        inFlyUpdateTasks[.filterSet] == nil &&
         abs(dateProvider().timeIntervalSince(updateManager.lastFilterSetUpdateDate)) > .minutes(featureFlagger.filterSetUpdateFrequency)
     }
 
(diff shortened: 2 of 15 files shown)


BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
```diff
@@ -40,6 +54,12 @@ public actor DataManager: DataManaging {
         self.fileNameProvider = fileNameProvider
     }
 
+    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, job: (inout DataKey.DataSet) async throws -> Bool) async throws {
+        var dataSet = dataSet(for: key)
+        guard try await job(&dataSet) else { return }
+        try store(dataSet, for: key)
```
bwaresiak: I think this is ok in current implementation, but I wouldn't really call this an atomic approach, as any await can yield execution to another queue. So:
 - can this be called multiple times in parallel?
 - are we worried about job call influencing order of execution e.g. 
    - call Stack A: invoke `updateDataSet`
    - call Stack A: yield on await job
    - call Stack B: invoke `updateDataSet`
    - call Stack B: yield on await job
    - call Stack B (job finished): store
    - call Stack A (job finished): store

bwaresiak: Next: question about `store` function implementation: are we concerned about `self.store` member and underlying file going out of sync in case of write failure?

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
```diff
@@ -60,36 +73,60 @@ final class MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDatas
...
 
         // If hashPrefix Sets need to be updated fetch them
         if shouldUpdateHashPrefixSets {
             Logger.MaliciousSiteProtection.datasetsFetcher.debug("Downloading HashPrefixSets")
-            _ = updateManager.updateData(datasetType: .hashPrefixSet)
+            let task = updateManager.updateData(datasetType: .hashPrefixSet)
+            inFlyUpdateTasks[.hashPrefixSet] = task
+            Task {
+                try? await task.value
+                inFlyUpdateTasks[.hashPrefixSet] = nil
```
bwaresiak: Personal opinion: considering the juggling needed here to make sure asynchronous code actually is reliable, I find it hard to believe we should be adopting it (async/await) widely, as simple queue here would just do the trick in a way that would be far more clear to reason about.

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
```diff
@@ -40,6 +54,12 @@ public actor DataManager: DataManaging {
         self.fileNameProvider = fileNameProvider
     }
 
+    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, job: (inout DataKey.DataSet) async throws -> Bool) async throws {
```
bwaresiak: Is this now the preferred public interface to use? I'm trying to validate if we should update the public/private access modifiers of other methods to make it more clear how to use this, or at least better document that.

bwaresiak (commented):
@alessandroboron I did not test the code, just went though it with some validation in mind, I think we should wait for Alex to review this, and generally include that in next week's internal.

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
alessandroboron: Initially, I marked `func store<DataKey: MaliciousSiteDataKey>(_ dataSet: DataKey.DataSet, for key: DataKey)` private to enforce it. Then I noticed that tests use it to set up the preconditions.

We have the following interface that we inject into the `UpdateManager`. The `UpdateManger can only interact with the DataManager via that interface so I think it should be fine. Also, the clients can’t call the methods of this class.

```
protocol DataManaging {
   func dataSet<DataKey: MaliciousSiteDataKey>(for key: DataKey) async -> DataKey.DataSet
   func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, job: (inout DataKey.DataSet) async throws -> Bool) async throws
}
```

BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
alessandroboron: You’re right, it is not correct to call this approach atomic as the actor will handle other operations when `await job(&dataSet)` is called. But this approach ensure reading, updating and writing the datasets should be done within the actor isolated context. 

Replies to your questions:
**- can this be called multiple times in parallel?**
Yes for each supported protection (phishing, malware, scam) we create a parallel task to update the datasets for that protection. 

**- are we worried about job call influencing order of execution**
It should not, each task would read, modify and store the dataset in different files without affecting each other.

**- are we concerned about self.store member and underlying file going out of sync in case of write failure?**
You referring to the fact that we set` self.store[dataType] = dataSet` before writing to disk right? Indeed, If writing to disk fails the data will be out of sync. I don’t think it causes issues but it makes sense to set the store after writing.

mallexxx: I agree with @bwaresiak here that the operation is not atomic here and using a queue may make more sense -reducing concurrent large data chunks copying.

If we add an extra async job step to the test and set a symbolic breakpoint to `copy()` method disclosing where the actual Copy-On-Write happens - it would reveal that we're still copying the data from the `store` on an asynchronous context - which was the only concern I had related to the code safety (here on the screenshot you can see the 2 threads operating concurrently on the expected-to-be `actor` context):
Screenshot 2025-03-24 at 16 21 47

While the old code still looks _safe_ to me from the data isolation standpoint, the updated one seems not changing much. 
We‘re still copying the value from the dictionary (now using the extra `updateDataSet` function) and updating it from another thread after yielding control to the data fetcher.
Frankly I‘m struggling to suggest an approach that would “fix” it, probably the things get broken in some other place leading to the data corruption while the data is being copied here.

mallexxx: maybe we could make it better mutating the dictionary value directly avoiding the copy being made

something like
```
let oldRevision = await dataManager.dataSet(for: key).revision
let changeSet = try await apiClient.load(request)
await dataManager.updateDataSet(for: key) { $0.apply(changeSet }
```
while the updateDataSet shouldn‘t be async but mutating the value directly like:
```
func updateDataSet(job) {
job(&store[key])
}
```
using the store directly in the non-escaping function shouldn‘t cause the data being copied (unless held somewhere else)

…still it‘s a "manual optimization" not really fixing the code that isn‘t broken, unless it‘s broken somewhere in the Swift async machinery guts

alessandroboron: @mallexxx given the crashes manifest on iOS but not on macOS I would exclude that these crashes are due to something broken in the Swift concurrency machinery guts. I know macOS has less users but if that was the case we would see crashes in macOS as well.

Maybe we can take an iterative approach. In this PR, I made a change to prevent update tasks from starting if others are already in progress. This helps mitigate scenarios where multiple threads attempt to update the same dataset simultaneously. We could release this change and monitor its impact on crash numbers without modifying the BSK code. Based on the new crash log data, we can then decide the next steps

@bwaresiak what do you also think?

alessandroboron: @bwaresiak @mallexxx I reverted the changes on BSK and instead serialised the tasks to update the datasets.

mallexxx (commented):
Looking at it again, I still can‘t find anything suspicious that could cause the crashes.
I tend to agree with @bwaresiak about the Operation Queue would be better to use here if we want to streamline the update tasks (as it was done with the Sync Queue by @ayoy that did fix sync problems with Swift concurrency).
BUT, that would require more effort to rethink the whole update manager part to get rid of Swift concurrency entirely in here.
I‘ve run the test steps and the update went well. 
Still there‘s one part that could be improved here: the hash prefixes and filter sets are still applied in parallel; Could we make those queued as well potentially only allowing one instance of the `startFetching()` task to run at once, i.e. replace the `inFlyUpdateTasks` with a single `isFetching` flag and return early if fetching is in progress.
And inside the `startFetching()` we could have the single `Task` returned that first applies the hash prefixes and then fetches the filter sets so the operations aren‘t run in parallel?

alessandroboron:
> Looking at it again, I still can‘t find anything suspicious that could cause the crashes. I tend to agree with @bwaresiak about the Operation Queue would be better to use here if we want to streamline the update tasks (as it was done with the Sync Queue by @ayoy that did fix sync problems with Swift concurrency). BUT, that would require more effort to rethink the whole update manager part to get rid of Swift concurrency entirely in here. I‘ve run the test steps and the update went well. Still there‘s one part that could be improved here: the hash prefixes and filter sets are still applied in parallel; Could we make those queued as well potentially only allowing one instance of the `startFetching()` task to run at once, i.e. replace the `inFlyUpdateTasks` with a single `isFetching` flag and return early if fetching is in progress. And inside the `startFetching()` we could have the single `Task` returned that first applies the hash prefixes and then fetches the filter sets so the operations aren‘t run in parallel?

@mallexxx I update the PR to download the datasets serially. I also moved the datasets update within the actor as per @bwaresiak  suggestion.

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
```diff
@@ -60,36 +71,51 @@ final class MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDatas
         userPreferencesManager: MaliciousSiteProtectionPreferencesProvider,
         dateProvider: @escaping () -> Date = Date.init,
         backgroundTaskScheduler: BGTaskScheduling = BGTaskScheduler.shared,
-        application: BackgroundRefreshCapable = UIApplication.shared
+        application: BackgroundRefreshCapable = UIApplication.shared,
+        preferencesScheduler: AnySchedulerOf<DispatchQueue> = DispatchQueue.main.eraseToAnyScheduler()
```
bwaresiak, line 75: We add new dependency to use it in one place for a single function? Isn't there a way to write this differently so we don't need to pull in another package?

alessandroboron: The main reason for adding this package is that I needed a way to control the scheduler in tests. I've added a debounce to the user preference publisher so we don’t start multiple datasets fetch if the users turn on / off the feature in sequence.   

I understand the concern about adding a new dependency, and I usually don’t like to add 3rd party dependencies unless it’s necessary. In this case, I don’t have a way to set an expectation or wait for the Task to finish to assert on the condition. I also thought that adding this package could be useful for other team members to test their Combine code. If we think we won’t be using much Combine in the future I’m fine to remove this dependency. The only con in this case s is that I won’t be able to test this behaviour.

bwaresiak: @alessandroboron thanks for explanation - gh collapsed the diff for me and I didn't find the Test case you've mentioned on the first two passes, but now I see it. It makes sense in that context, thanks. :)

SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
```diff
@@ -90,7 +90,10 @@ public actor DataManager: DataManaging {
         return storedDataSet
     }
 
-    func store<DataKey: MaliciousSiteDataKey>(_ dataSet: DataKey.DataSet, for key: DataKey) throws {
+    func updateDataSet<DataKey: MaliciousSiteDataKey>(with key: DataKey, changeSet: APIClient.ChangeSetResponse<DataKey.DataSet.Element>) async throws {
```
mallexxx: why should it be `async`?

alessandroboron: I copied the signature from the `DataManaging` protocol which `DataManager` conforms to. In the protocol we need to explicitly define the methods `async`. If we remove that Xcode would complain with `Actor-isolated instance method 'updateDataSet(with:changeSet:)' cannot be used to satisfy nonisolated protocol requirement` error.

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
```diff
@@ -60,36 +71,51 @@ final class MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDatas
...
-            Logger.MaliciousSiteProtection.datasetsFetcher.debug("Downloading HashPrefixSets")
-            _ = updateManager.updateData(datasetType: .hashPrefixSet)
-        }
+        return Task {
+            // If hashPrefix Sets need to be updated fetch them
+            if shouldUpdateHashPrefixSets {
+                Logger.MaliciousSiteProtection.datasetsFetcher.debug("Downloading HashPrefixSets")
+                await updateManager.updateData(datasetType: .hashPrefixSet).value
+                // Reset the flag to false only if FilterSets don't have to be fetched.
+                isDatasetsFetchInProgress = shouldUpdateFilterSets
```
mallexxx: this probably could be placed as `defer { isDatasetsFetchInProgress = false }` for clarity and to avoid mistakes in future

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
mallexxx: @alessandroboron, I‘m not sure if the debounce is actually needed there, switching preference toggle is not something user would do on a daily basis and it isn‘t easy to turn it accidentally, and even if it is, would 0.5 second be enough?
If we launch a `Task` it won‘t be cancelled and still proceed in the background (out of scope but we may want to consider implementing cooperative cancellation)
Maybe it would be easier to just drop the dependency and the debouncer?

SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/DataManager.swift
mallexxx: no, xcode won‘t complain as `DataManager` is an `actor` and all of its methods are `async` by default, so we should be fine dropping the keyword from the func definition, not from the protocol

mallexxx (approved):
LGTM!
Consider addressing the comments inline

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
alessandroboron: yeah it’s actually cleaner :)

PR #261, opened 2025-03-18
AuthV2 - Feature flag for iOS and macOS

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1209170372758740/f
CC: @miasma13 @diegoreymendez 

### Description

- All Subscription Auth V1 and V2 initialisations are now behind a disabled development feature flag that can be enabled via the debug menu
- Migration from V1 to V2 and rollback from V2 to V1 supported and tested

### Testing Steps

For both iOS and macOS

1. Double check the feature flag is disabled (Debug > Feature Flag Overrides > privacyProAuthV2)
2. Subscribe to Privacy Pro
3. Enable VPN
4. Enable `privacyProAuthV2` feature flag (it enables AuthV2)
5. Kill the app
6. Restart the app, The VPN should be disconnected but can be restarted straight away.
7. Disable `privacyProAuthV2` feature flag (it disables AuthV2 and roll backs to V1)
8. Kill the app
9. Restart the app, The VPN should be disconnected but can be restarted straight away.

### Impact and Risks

High: we could break the VPN and the subscription UX

#### What could go wrong?
- The VPN could not restart automatically after a migration from V1 to V2 or vice versa, that's ok, but if started manually everything should work as usual

---
###### Internal references:


federicocappelli:
@diegoreymendez can I ask for your review for the VPN configuration with the new isAuthV2Enabled option? (mostly VPNSettings and PacketTunnelProvider)

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -273,7 +271,8 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
     private lazy var keyStore = NetworkProtectionKeychainKeyStore(keychainType: keychainType,
                                                                   errorEvents: debugEvents)
 
-    private let tokenHandler: any SubscriptionTokenHandling
+    private let tokenHandlerProvider: () -> any SubscriptionTokenHandling
```
federicocappelli, line 274: This change is due to the VPN not having any kind of communication with the main app before the init, so every kind of init choice between auth v1 and v2 can't be done in the VPN init but needs to be delegated after the VPN have loaded the startup options, so I'm forced to init both auth v1 and v2 classes and then swap between them at runtime

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -694,6 +706,8 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
     open override func startTunnel(options: [String: NSObject]? = nil) async throws {
         Logger.networkProtection.log("🚀 Starting tunnel")
 
+        observeSettingChanges()
```
federicocappelli: @diegoreymendez this one is tricky, I experienced a lot of issues testing the VPN, having the XPC connection timing out after 3 (seconds? attempts? I don't know) and after a few tests I noticed that the `XPacketTunnelProvider` init took a lot of time, so I tried to move as much work as possible outside the inits. Not sure if this is correct, the VPN works as expected.

BrowserServicesKit/Sources/NetworkProtection/VPNAuthTokenBuilder.swift
```diff
@@ -26,14 +26,14 @@ public struct VPNAuthTokenBuilder {
 
     public static func getVPNAuthToken(from tokenProvider: any SubscriptionTokenHandling) async throws -> String {
         let token = try await tokenProvider.getToken()
-        if token.hasPrefix(tokenPrefix) {
-            // In AuthV1 adding the token prefix is managed by the classes storing the token, in AuthV2 the token is stored as the original TokenContainer and the prefix is added by this builder
-            return token
-        }
-        return tokenPrefix + token
+        return getVPNAuthToken(from: token)
```
federicocappelli, line 29: here I have just made the function safer and removed some code duplication

iOS/PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -31,7 +31,6 @@ import WidgetKit
 import WireGuard
 import BrowserServicesKit
 
-// Initial implementation for initial Network Protection tests. Will be fleshed out with https://app.asana.com/0/1203137811378537/1204630829332227/f
```
federicocappelli, line 34: I asked graham if I could remove this

macOS/DuckDuckGo/NetworkProtection/NetworkExtensionTargets/NetworkExtensionTargets/MacPacketTunnelProvider.swift
```diff
@@ -428,97 +429,95 @@ final class MacPacketTunnelProvider: PacketTunnelProvider {
...
-            let authEndpointService = DefaultAuthEndpointService(currentServiceEnvironment: subscriptionEnvironment.serviceEnvironment)
-            let accountManager = DefaultAccountManager(accessTokenStorage: tokenStore,
-                                                       entitlementsCache: entitlementsCache,
-                                                       subscriptionEndpointService: subscriptionEndpointService,
-                                                       authEndpointService: authEndpointService)
-
-            entitlementsCheck = {
-                Logger.networkProtection.log("Subscription Entitlements check...")
-                return await accountManager.hasEntitlement(forProductName: .networkProtection, cachePolicy: .reloadIgnoringLocalCacheData)
+        // V1
```
federicocappelli: Major change here, we are forced to init both Auth classes versions here and decide which one to use at runtime.

macOS/DuckDuckGo/Subscription/SubscriptionManager+StandardConfiguration.swift
```diff
@@ -125,7 +125,13 @@ extension DefaultSubscriptionManagerV2 {
         let authClient = DefaultOAuthClient(tokensStorage: tokenStorage,
                                             legacyTokenStorage: legacyAccountStorage,
                                             authService: authService)
-        apiService.authorizationRefresherCallback = { _ in
+        apiService.authorizationRefresherCallback = { request in
+
+            guard request.url?.absoluteString.contains("api/auth/v2") == false else {
```
federicocappelli, line 137: this prevents attempts to re-authenticate an auth API request in the case of 401s and therefore will prevent authentication loops or useless retries

macOS/DuckDuckGoDBPBackgroundAgent/DuckDuckGoDBPBackgroundAgentAppDelegate.swift
```diff
@@ -107,8 +86,24 @@ final class DuckDuckGoDBPBackgroundAgentAppDelegate: NSObject, NSApplicationDele
     private let subscriptionManager: any SubscriptionAuthV1toV2Bridge
     private var manager: DataBrokerProtectionAgentManager?
 
-    init(subscriptionManager: any SubscriptionAuthV1toV2Bridge) {
-        self.subscriptionManager = subscriptionManager
+    override init() {
+
+        // MARK: - Configure Subscription
```
federicocappelli: Moved the Subscription configuration from `DuckDuckGoDBPBackgroundAgentApplication` to `DuckDuckGoDBPBackgroundAgentAppDelegate` because it is here were the auth settings are available.

macOS/LocalPackages/DataBrokerProtection/Sources/DataBrokerProtection/Authentication/DataBrokerProtectionSubscriptionManaging.swift
```diff
@@ -29,15 +29,23 @@ public protocol DataBrokerProtectionSubscriptionManaging {
...
-        if dbpSettings.storedRunType == .integrationTests,
-           let token = ProcessInfo.processInfo.environment["PRIVACYPRO_STAGING_TOKEN"] { // todo auth V1 token??
-            return token
+        // the tests locally
+        if dataBrokerProtectionSettings.storedRunType == .integrationTests {
+            var tokenKey: String
+            if !dataBrokerProtectionSettings.isAuthV2Enabled {
+                tokenKey = "PRIVACYPRO_STAGING_TOKEN"
+            } else {
+                tokenKey = "PRIVACYPRO_STAGING_TOKEN_V2"
```
federicocappelli: @THISISDINOSAUR : As anticipated I've modified the code for accepting auth V2 tokens override for Auth V2. The PIR e2e test will fail as soon as the feature flag is enabled by default, I'll create a task for that.

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -433,17 +432,17 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
                 keychainType: KeychainType,
-                tokenHandler: any SubscriptionTokenHandling,
+                tokenHandlerProvider: @escaping () -> any SubscriptionTokenHandling,
                 debugEvents: EventMapping<NetworkProtectionError>,
                 providerEvents: EventMapping<Event>,
                 settings: VPNSettings,
                 defaults: UserDefaults,
                 entitlementCheck: (() async -> Result<Bool, Error>)?) {
-        Logger.networkProtectionMemory.debug("[+] PacketTunnelProvider")
+        Logger.networkProtectionMemory.log("[+] PacketTunnelProvider")
```
diegoreymendez, line 441: This change means these entries are logged by default, even in production builds.  I think these should stay debug-build only and should not be included in production logs.

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -457,8 +456,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
         super.init()
 
-        observeSettingChanges()
-        Logger.networkProtectionMemory.debug("[+] PacketTunnelProvider initialized")
+        Logger.networkProtectionMemory.log("[+] PacketTunnelProvider initialized")
```
diegoreymendez, line 459: Same as above.

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -586,51 +585,64 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+        switch options.isAuthV2Enabled {
+        case .set(let newAuthVersion):
+            Logger.networkProtection.log("Set new isAuthV2Enabled")
+            settings.isAuthV2Enabled = newAuthVersion
+        case .useExisting:
+            Logger.networkProtection.log("Use existing isAuthV2Enabled")
+        case .reset:
+            settings.isAuthV2Enabled = false
+        }
+        PacketTunnelProvider.isAuthV2Enabled = settings.isAuthV2Enabled
```
diegoreymendez, line 600: Why are we storing this in two separate places?  A single source of truth would likely simplify things - all of the code that needs access to this can access `settings` directly, so the static var could go away easily.

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: This has been working well for a really long time now and if it suddenly started causing trouble I think it's be important to understand why.

This should be moved back and if there's an issue we should try to understand what changed.

BrowserServicesKit/Sources/NetworkProtection/Settings/Extensions/UserDefaults+isAuthV2Enabled.swift
```diff
@@ -0,0 +1,47 @@
...
+    private var isAuthV2EnabledKey: String {
+        "networkProtectionSettingIsAuthV2Enabled"
+    }
+
+    static let isAuthV2EnabledDefaultValue = false
+
+    @objc
+    dynamic var networkProtectionSettingIsAuthV2Enabled: Bool {
+        get {
+            value(forKey: isAuthV2EnabledKey) as? Bool ?? Self.isAuthV2EnabledDefaultValue
```
diegoreymendez: Nit: this is fine but on some newer code I'm using `#keypath` which avoids the need for a separate `isAuthV2EnabledKey`.  See `UserDefaults+isInternalUser` for an example.

BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -286,21 +286,31 @@ final public actor DefaultOAuthClient: @preconcurrency OAuthClient {
...
-              let legacyToken = legacyTokenStorage.token else {
+        guard !isUserAuthenticated else {
+            Logger.OAuthClient.log("Migration not needed, user is already authenticated")
             return nil
         }
 
-        Logger.OAuthClient.log("Migrating legacy token")
+        guard let legacyTokenStorage else {
+            Logger.OAuthClient.fault("Auth migration attempted without a LegacyTokenStorage")
+            assertionFailure("Auth migration attempted without a LegacyTokenStorage")
```
diegoreymendez: This assertion seems rather harsh - it will crash the caller on debug sessions if this scenario is hit, which is easy to miss if the caller is the VPN extension.

If the `legacyTokenStorage` is needed when this method is called, it could be passed as an input parameter.  If it's not needed, we should consider removing the assertion.

Either way avoiding this inconsistency seems quite trivial which makes me think crashing the app can and should be avoided.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
```diff
@@ -70,7 +70,7 @@ final class NetworkProtectionTunnelController: TunnelController, TunnelSessionPr
 
     private let accessTokenStorage: SubscriptionTokenKeychainStorage
     private let subscriptionManagerV2: any SubscriptionManagerV2
-    private let isAuthV2Enable: Bool
+//    private let isAuthV2Enable: Bool
```
diegoreymendez: Can this be removed?

diegoreymendez (commented):
Added some initial feedback.  Still testing.

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: I changed this because I need them to be persistent and show up in the system logs when exported via Terminal because sometimes the SysExt is not initialised at all.
Additionally, the debug logs disappear after seconds from my console and make debugging very difficult.

BrowserServicesKit/Sources/NetworkProtection/Settings/Extensions/UserDefaults+isAuthV2Enabled.swift
federicocappelli: Updated, thanks

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: the problem is `let tokenHandlerProvider: () -> any SubscriptionTokenHandling = {` that is initialised in MacPacketTunnelProvider before self (and settings) is available, can't do without a static var.

BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
federicocappelli: Correct, legacyTokenStorage is not really optional anymore, this is a leftover, removed.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
federicocappelli: removed

BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
federicocappelli: Nice catch, thanks. The assertion is a leftover and is wrong. Removed.

BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -286,21 +286,28 @@ final public actor DefaultOAuthClient: @preconcurrency OAuthClient {
...
+        Logger.OAuthClient.log("Migrating legacy token...")
         do {
             let tokenContainer = try await exchange(accessTokenV1: legacyToken)
-            Logger.OAuthClient.log("Tokens migrated successfully, removing legacy token")
+            Logger.OAuthClient.log("Tokens migrated successfully")
 
             // NOTE: We don't remove the old token to allow roll back to Auth V1
 
-            // Store new tokens
-            tokenStorage.tokenContainer = tokenContainer
```
miasma13, line 303: Why is storing of the new `tokenContainer` removed?

federicocappelli: exchange(accessTokenV1:..) already stores the tokens, that line was useless.

macOS/DuckDuckGoDBPBackgroundAgent/DuckDuckGoDBPBackgroundAgentAppDelegate.swift
```diff
@@ -108,8 +87,24 @@ final class DuckDuckGoDBPBackgroundAgentAppDelegate: NSObject, NSApplicationDele
...
+            let subscriptionAppGroup = Bundle.main.appGroup(bundle: .subs)
+            let subscriptionUserDefaults = UserDefaults(suiteName: subscriptionAppGroup)!
+            let subscriptionEnvironment = DefaultSubscriptionManager.getSavedOrDefaultEnvironment(userDefaults: subscriptionUserDefaults)
+            subscriptionManager = DefaultSubscriptionManagerV2(keychainType: .dataProtection(.named(subscriptionAppGroup)),
+                                                               environment: subscriptionEnvironment,
+                                                               userDefaults: subscriptionUserDefaults,
+                                                               canPerformAuthMigration: false,
+                                                               canHandlePixels: false)
+        }
+        // MARK: -
```
miasma13: Can be removed

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -349,8 +352,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         vpnSettings.alignTo(subscriptionEnvironment: subscriptionAuthV1toV2Bridge.currentEnvironment)
 
         // Update DBP environment and match the Subscription environment
-        let dbpSettings = DataBrokerProtectionSettings(defaults: .dbp)
+        let dbpSettings = DataBrokerProtectionSettings(defaults: .netP)
```
miasma13: Why is DBP to use netP's user defaults?

federicocappelli: I think GH is misbehaving O_o I don't have that! checking

macOS/DuckDuckGoDBPBackgroundAgent/DuckDuckGoDBPBackgroundAgentAppDelegate.swift
federicocappelli: removed alongside all other mark: -

macOS/DuckDuckGo/Application/AppDelegate.swift
federicocappelli: I think the GH disruptions are having an effect here, this is fixed and pushed from a while

federicocappelli: Now looks ok, closing

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: I'm not saying we shouldn't change these for debugging, I'm saying we shouldn't change these in production.  I think it's fine to change locally for testing but we should roll back before merging.

The unified logging system can drop logs if necessary and by marking the least important ones as `debug` they take precedence in being dropped.

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: But `isAuthV2Enabled` could be passed as a parameter to `tokenHandlerProvider`.  If you want callers not to have to provide the parameter, a wrapper method in `PacketTunnelProvider` would do.

```swift
func gimmeTokenHandlerProvider() -> any SubscriptionTokenHandling {
    tokenHandlerProvider(settings.isAuthV2Enabled)
}
```

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: There are  a couple of facts that convinced me to change these log levels:
1. Debug logs are dropped too early and not included in the system logs exported by the users in case of issues.
2. The SysExt behaviour is weird and in some cases I see the extension disappearing without any log except the [+] init and [-] deinit of the various classes involved
So if I'll need to debug an issue and the only available source of information is a System logs export I want to see the lifecycle of these classes. I'm happy to revert back to Debug these logs (and more) after AuthV2 is fully adopted and the risk is lowered.

BrowserServicesKit/Sources/NetworkProtection/KeyManagement/NetworkProtectionTokenStore.swift
```diff
@@ -149,3 +152,98 @@ public final class NetworkProtectionKeychainTokenStore: NetworkProtectionTokenSt
...
+    private let errorEvents: EventMapping<NetworkProtectionError>?
+
+    public struct Defaults {
+        static let tokenStoreEntryLabel = "DuckDuckGo Network Protection Auth Token Container"
+        public static let tokenStoreService = "com.duckduckgo.networkprotection.authTokenContainer"
+        static let tokenStoreName = "com.duckduckgo.networkprotection.tokenContainer"
+    }
+
+    /// - isSubscriptionEnabled: Controls whether the subscription access token is used to authenticate with the NetP backend
+    /// - accessTokenProvider: Defines how to actually retrieve the subscription access token
```
diegoreymendez, line 178: This won't work as-is: on one end because the params are outdated, but mostly pointing out that the documentation isn't properly coded to be parsed by Xcode's doc system.  Alt+clicking on the init method will show no documentation.

Consider replacing with something like this for proper doc parsing.

```suggestion
    /// Default initializer
    ///
    /// - Parameters:
    ///     - keychainType: Type of keychain used for storing the tokens.
    ///     - serviceName: The name of the service which will be used for the tokens.
    ///     - errorEvents: A mapper for error events.
    ///
```

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: Ok, understood.  Only request is let's please try to roll back less important logs to debug once done.  I've already gotten logs from users where a lot of important info is being dropped by the system so I want to tell the system which ones to drop first.

federicocappelli: Ok clear, I've created a task for this: https://app.asana.com/0/0/1209768980795746

BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: I already went down this rabbit hole, `entitlementCheck` and `tokenHandlerProvider` would need the parameter, then the problem is that `tokenHandlerProvider()` is used in a series of lazy vars like `serverSelectionResolver`, `deviceManager` and `serverStatusMonitor`, for the lazy vars we can't be sure that they are initialised after we read the startup options, so I tried to make them not lazy and move them in the init but then another can of worms opened... 
Additionally `NetworkProtectionEntitlementMonitor` uses `tokenHandlerProvider()` but doesn't know the authv2 flag state... so I should pass that too using another closure :/
I COULD do what you suggested but the amount of changes and the variables involved make me sweat and get worried about the amount of re-testing I would need to do.
If we can skip this complication I would like to skip it, otherwise I'll allocate some time to it.

BrowserServicesKit/Sources/NetworkProtection/KeyManagement/NetworkProtectionTokenStore.swift
federicocappelli: Fixed with the xcode auto-generated template

diegoreymendez (approved):
Ok seems to be working well on my end, let's move forward!

I'll try to monitor next week for any issues and will ping you if there's anything, but I'm hoping it'll be fine.

PR #272, opened 2025-03-19
Stabilize LocalBookmarkManager unit tests

ayoy:
Task/Issue URL: https://app.asana.com/0/1205237866452338/1209660514434895/f

### Description:
This change adds completion handler to some of the BookmarkManager APIs in order to allow for more reliable unit testing.

### Testing Steps
Verify that CI is green.

### Impact and Risks
None (tests code + some API changes but made in a way that is transparent to the existing code).

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo-macOS.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/macOS/DuckDuckGo-macOS.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -41,8 +41,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/duckduckgo/duckduckgo-autofill.git",
       "state" : {
-        "revision" : "b5976277a68e18f93edb0d853889e461e06557fa",
-        "version" : "16.2.2"
+        "revision" : "2bc93bec9cbe6d916e84b42346b925c6c057bfa5",
+        "version" : "17.0.0"
       }
     },
     {
@@ -104,7 +104,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/duckduckgo/privacy-dashboard",
       "state" : {
-        "revision" : "fd0a479bdc9db4a9d4ae3fb691e9057f7a67c578",
+        "revision" : "b370e1e1afa322a48a8a37c6c954a444c15489e4",
         "version" : "9.0.0"
       }
     },
--- a/macOS/DuckDuckGo/Bookmarks/Model/BookmarkManager.swift
+++ b/macOS/DuckDuckGo/Bookmarks/Model/BookmarkManager.swift
@@ -33,7 +33,7 @@ protocol BookmarkManager: AnyObject {
     func getBookmark(forUrl url: String) -> Bookmark?
     func getBookmark(forVariantUrl variantURL: URL) -> Bookmark?
     func getBookmarkFolder(withId id: String) -> BookmarkFolder?
-    @discardableResult func makeBookmark(for url: URL, title: String, isFavorite: Bool, index: Int?, parent: BookmarkFolder?) -> Bookmark?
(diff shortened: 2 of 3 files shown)


macOS/DuckDuckGo-macOS.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -41,8 +41,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/duckduckgo/duckduckgo-autofill.git",
       "state" : {
-        "revision" : "b5976277a68e18f93edb0d853889e461e06557fa",
-        "version" : "16.2.2"
+        "revision" : "2bc93bec9cbe6d916e84b42346b925c6c057bfa5",
```
jotaemepereira, line 44: Are the changes here intended?

macOS/DuckDuckGo/Bookmarks/Model/BookmarkManager.swift
```diff
@@ -203,10 +203,10 @@ final class LocalBookmarkManager: BookmarkManager {
     }
 
     @discardableResult func makeBookmark(for url: URL, title: String, isFavorite: Bool) -> Bookmark? {
-        makeBookmark(for: url, title: title, isFavorite: isFavorite, index: nil, parent: nil)
+        makeBookmark(for: url, title: title, isFavorite: isFavorite, index: nil, parent: nil, completion: { _ in })
```
jotaemepereira: ❓ Do we need to add the empty completion handler here? We already define the default argument in the makeBookmark function.

macOS/DuckDuckGo/Bookmarks/Model/BookmarkManager.swift
```diff
@@ -203,10 +203,10 @@ final class LocalBookmarkManager: BookmarkManager {
...
     }
 
-    @discardableResult func makeBookmark(for url: URL, title: String, isFavorite: Bool, index: Int? = nil, parent: BookmarkFolder? = nil) -> Bookmark? {
+    @discardableResult func makeBookmark(for url: URL, title: String, isFavorite: Bool, index: Int? = nil, parent: BookmarkFolder? = nil, completion: @escaping (Error?) -> Void) -> Bookmark? {
```
jotaemepereira, line 209: I know that the idea of this is to make tests more reliable, but the completion is confusing. We do not call it when it is complete; we only call it when an error saving to the database happens.

I would clarify that it is for error only by making the Error non-nullable.

jotaemepereira (approved):
@ayoy I’ve added some comments, but otherwise looks good 👍🏼

macOS/DuckDuckGo-macOS.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
ayoy: Yes, these changes is macOS project catching up with the workspace.

macOS/DuckDuckGo/Bookmarks/Model/BookmarkManager.swift
ayoy: Good catch, it’s not needed :) I’ve removed it.

macOS/DuckDuckGo/Bookmarks/Model/BookmarkManager.swift
ayoy: We’re calling the completion after save is complete, regardless of whether the error happened. The error parameter is optional, but the completion handler is called in all cases.

See the place where save happens (LocalBookmarkStore) and completion is called both on didSave and on error:
Screenshot 2025-03-19 at 15 31 22

ayoy: After syncing on MM I was made aware that I forgot to add the completion handler for the error case. The fix is pushed now (we’re interested in calling the completion handler in all cases of course) 💪

PR #276, opened 2025-03-19
ensure removing tabs removes previews and improve clean up code

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1209699433442062/f
Tech Design URL:
CC: @alessandroboron 

### Description

The driving change is the 1 line `replaceTab` which should fix the spike in unremoved previews.

Also decided to do some refactoring here too:
* Simplify the cleanup code
* Test that cleanup code gets called

### Testing Steps
To test replace tab change:
1. Validate https://github.com/duckduckgo/iOS/pull/3901

To test the refactoring:
1. Make `func removePreview(forTab tab: Tab)` of `TabPreviewsSource` return immediately so that the preview is not deleted
2. Add a breakpoint to `removePreviewsWithIdNotIn` in `TabPreviewsSource`
3. Open a 3-4 tabs and make sure previews are generated
4. Close the tabs one by one (not all at once)
5. Background / kill the app
6. Launch the app - the breakpoint should hit
7. Wait a few seconds
8. Background / kill the app again
9. Launch the app again - the breakpoint should not hit
11. Reset the code and smoke test browsing, opening tabs, tab switching and swiping between tabs

### Impact and Risks
Low: Small bug fixes, improvement to existing features

#### What could go wrong?
* Previews could not be deleted

### Quality Considerations
* Ensure that previews are still working

### Notes to Reviewer
Nothing not specified above.


iOS/DuckDuckGo/TabPreviewsSource.swift
```diff
@@ -176,46 +188,17 @@ class TabPreviewsSource {
...
-                for previewURL in previews where previewURL.lastPathComponent.hasSuffix(".png") {
-                    let previewID = previewURL.lastPathComponent.dropLast(4)
-                    
-                    if !validIDs.contains(String(previewID)) {
-                        try? FileManager.default.removeItem(at: previewURL)
-                    }
-                }
+    func removePreviewsWithIdNotIn(_ ids: Set<String>) async {
+        guard let directory = previewStoreDir else { return }
+        let contents = try? FileManager.default.contentsOfDirectory(atPath: directory.path)
```
afterxleep, line 194: This loads the entire contents of the directory in memory.   What do you think about using a FileEnumerator to stream them instead?

```
guard let enumerator = FileManager.default.enumerator(
    at: directory,
    includingPropertiesForKeys: nil,
    options: [.skipsHiddenFiles]
) else { return }

for case let url as URL in enumerator {
// Remove it...
}

```

afterxleep (approved):
Nice cleanup.   Works as expected.   

One suggestion about file access but given actual usage is not life or death.  Up to you to implement.

iOS/DuckDuckGo/TabPreviewsSource.swift
brindy: Yeah, good idea

brindy: Actually, I thought about this and a couple of things. I believe that enumerator takes a snapshot of the structure at that point, which means it's using the memory anyway.  If that's not true then I'd be worried about the structure changing under it.  

Given that, I'm going to leave the implementation I already did.

PR #284, opened 2025-03-20
Reopen window in the same position after burning

ayoy:
Task/Issue URL: https://app.asana.com/0/1199178362774117/1209703704629721/f

### Description:
This change remembers burned window frame in order to reopen a new window with the same exact frame
(instead of cascading from that frame that resulted in shifting the window to the right and down).

### Testing Steps
1. Launch the app and, to simplify testing, put the window in the upper-left corner of the screen.
2. Burn all data. Verify that the new window is placed in the exact same frame in the upper-left corner of the screen.
3. Repeat the same with burning “current window”.
4. Repeat the same with burning “current tab” when you only have 1 tab.
5. Open multiple windows and burn all data. Verify that the new window is placed in the exact same frame as the window you’ve initiated burning from.

### Impact and Risks
Low

#### What could go wrong?
Nothing really – we’re assigning an optional dropping point to new window function. If non-nil, it represents previous window frame, if nil it does nothing (like currently).

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/Fire/Model/Fire.swift
+++ b/macOS/DuckDuckGo/Fire/Model/Fire.swift
@@ -307,23 +307,30 @@ final class Fire {
     @MainActor
     private func closeWindows(entity: BurningEntity) {
 
-        func closeWindow(of tabCollectionViewModel: TabCollectionViewModel) {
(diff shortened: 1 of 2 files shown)


jaceklyp (approved):
LGTM!

PR #285, opened 2025-03-20
PixelKit: Move platform prefixing to avoid sometimes double prefixing pixels

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1209736748944376/f
Tech Design URL:
CC:

### Description
In some cases the platform prefix would be added twice, so this moves where it happens to fix that
Also makes the dbpshared pixels handler actually use its injected instance of pixel kit

### Testing Steps

1. Make sure pixel logging is on
2. Attach the debugger to the background agent as well as the main app
3. Go through the PIR flow and let it run. Make sure scans finish and it gets to opting out
4. Check both the app and the background agent logs to make sure all pixels fired have the m_mac prefix and that none of them have it twice

### Notes to Reviewer

---
###### Internal references:


BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -426,21 +423,22 @@ public final class PixelKit {
         }
 
     // Only set up for macOS and for Experiments
```
THISISDINOSAUR, line 425: @samsymons I don't really understand this comment so think I need to talk about this more before we could merge this PR.
I think the prefixing has to live here for obvious reasons, but I don't understand the flow of what happens on iOS

More generally, I don't like what we're doing here and the fact that I've added more to it, so if you have any context on how things ended up this way that might help me come up with an alternative

samsymons: From digging into this further, it doesn't look like a bug that is actively affecting things... yet. This is definitely at timebomb waiting to happen though. The reason it doesn't affect us today is because the only iOS pixels being sent through here are experiment ones, which have custom handling up front that returns before the `else` case can occur.

There's a project to bring PixelKit to iOS, but I'm inclined to do something about this before then so that it doesn't bite us. Overall I'd say it's out of scope of what you're doing here though.

samsymons (approved):
This worked well in my testing, thanks - nice catch 😄

BrowserServicesKit/Sources/PixelKit/PixelKit.swift
THISISDINOSAUR: I'll certainly need it for PIR on iOS, so I'll have to do something imminently

PR #286, opened 2025-03-20
PIR shared code extraction: further cleanup

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1209737141547780/f
Tech Design URL:
CC:

### Description
Some small changes to further clean up the shared code extraction. Includes: 
- Rename the macOS pixel handler
- Splits up the execution config into shared and platform specific bits
- Removes the local freemium dependency from dbp shared

### Testing Steps

1. Make sure pixel logging is turned on
2. Connect the debugger to both the main app and the background agent
3. Start a PIR scan and let it finish
4. Go through both logs and check PIR pixels are still firing correctly

### Notes to Reviewer

---
###### Internal references:


aataraxiaa (approved):
@THISISDINOSAUR I observed pixels in Xcode via filtering on `Subsystem: com.duckduckgo.PixelKit`. During a scan and subsequent optout pixels appeared to be firing as expected (assuming that if we see a large amount of PIR pixels than they all should be working, as I don’t believe these changes would negatively impact a **specific** set of pixels). 

So, LGTM.

PR #288, opened 2025-03-20
Add unit tests for fallBackToSmaller in FaviconsManager

ayoy:
Task/Issue URL: https://app.asana.com/0/1201899738287924/1209678667397515/f

### Description
This change adds unit tests for FaviconsManager APIs related to looking up smaller size categories
if favicon of requested size category was not found.
Enabling unit testing required to refactor FaviconsManager slightly to allow mocking image cache and
reference cache, and then adding the actual mocks, so the change here ends up quite big.

### Testing Steps
1. Verify that CI is green
2. Smoke test favicons by opening multiple tabs (e.g. from bookmarks or from history view) and verifying that favicons load and there are no crashes.

### Impact and Risks
Medium: The change moves around some code related to favicons underlying logic. I took extra care to make sure interfaces are replicated 1:1, and I did exhaustive smoke-testing so all should work fine.

---
###### Internal references:


macOS/UnitTests/Favicons/Mocks/CapturingFaviconImageCache.swift
```diff
@@ -0,0 +1,78 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Common
+@testable import DuckDuckGo_Privacy_Browser
+
+@MainActor
+final class CapturingFaviconImageCache: FaviconImageCaching {
```
tomasstrba, line 23: Would it be better if the class name contains "Mock" for better clarity?

tomasstrba (approved):
This PR looks good to me 👍 There are no crashes and favicons load as expected.

I've discovered one issue - a visual glitch when navigating to a new URL, but it is present in the original branch too, so let's focus on it in the other PR.

macOS/UnitTests/Favicons/Mocks/CapturingFaviconImageCache.swift
ayoy: I’m personally fine with both, but I noticed the distinction between “Capturing” and “Mock” classes in our codebase where mocks just provide dummy implementation while capturing classes allow for customizing function implementations and count function calls.

I could change it, but it won’t help much because there’s a lot of Capturing* classes in the codebase already :D so perhaps we would benefit from some proposal unifying the naming.

PR #304, opened 2025-03-21
Create `SharedPackages` directory

samsymons:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1209745503606409/f
Tech Design URL:
CC:

### Description

This PR moves BrowserServicesKit to a new `SharedPackages` directory.

### Testing Steps

1. Check that CI is green

### Notes to Reviewer

---
###### Internal references:


fastlane/Fastfile
```diff
@@ -2,7 +2,7 @@ import("../iOS/fastlane/Fastfile")
 import("../macOS/fastlane/Fastfile")
 
 lane :update_dependencies do
-  Dir.chdir("../BrowserServicesKit") do
+  Dir.chdir("../SharedPackages/BrowserServicesKit") do
```
samsymons, line 5: @ayoy I've fixed this path, but the approach won't scale to new packages being added. I was thinking about iterating through every directory in `SharedPackages` instead and doing the update, what do you think?

ayoy: @samsymons we could do it, but I’d defer it until we know we would be opening any other shared package as a standalone Xcode project. I’m just not sure yet if it’s a valid use case for us.

Your idea sounds good to me, but if we could get away without adding complexity, I’d wait :)

ayoy (approved):
Looks great, thanks @samsymons !

PR #311, opened 2025-03-23
Fix scrolling when moving to a not visible tab

jotaemepereira:
Task/Issue URL: https://app.asana.com/0/1204006570077678/1209751086393041/f
Tech Design URL:
CC:

### Description
This fixes an issue with scrolling when tabs are in the overflow state. For example, if you have

### Testing Steps
1. Go to the Debug Menu -> UI Triggers -> Append Tabs -> 50 tabs to open 50 tabs
2. Select the last tab, and use CMD + 1 to select the first tab
3. Close the first tab
4. **The last tab should be selected and visible to the user**

### Impact and Risks

I’m putting medium because I’m changing the order of some methods being called in the `didRemoveTab`.

#### What could go wrong?
As I mentioned above, I changed the order of how the two methods were called. I believe we had a bug in the past, but I want to ensure that I’m not missing anything.

### Quality Considerations
I’ve added a UI test for this case to avoid regressions.

### Notes to Reviewer
I want to understand why, in the past, those two functions had the wrong order.

—
###### Internal references:


macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
```diff
@@ -497,8 +497,8 @@ final class TabCollectionViewModel: NSObject {
             newSelectionIndex = selectionIndex.sanitized(for: self)
         }
 
-        notifyDelegate()
         select(at: newSelectionIndex, forceChange: forced)
+        notifyDelegate()
```
jotaemepereira, line 501: @tomasstrba @ayoy The `notifyDelegate()` was being called before the select (this was before I added any change to the recently closed functionality).

Which seemed to be a bug, given the `notifyDelegate` was the following:

```swift
func notifyDelegate() {
    if index.isUnpinnedTab {
        let newSelectionIndex = self.selectionIndex?.isUnpinnedTab == true ? self.selectionIndex?.item : nil
        delegate?.tabCollectionViewModel(self, didRemoveTabAt: index.item, andSelectTabAt: newSelectionIndex)
    }
}
```

Where the index is the index of the removed tab, and the `newSelectionIndex` is the index of the newly selected tab that is calculated with the `selectionIndex`.

The problem is `select(at: newSelectionIndex, forceChange: forced)` happened **after** the notify delegate. The select method is responsible for changing the value of `selectionIndex`, which made the new selection index in the delegate to be wrong.

I assume it was this way because before, we mainly selected the next or previous tab or, in some cases, the parent tab.

I want to ensure I’m not missing anything here that could cause issues. All tests are running, and I added one more UI test to verify this behavior.

tomasstrba: Agree with you @jotaemepereira, it was a bug before. Thanks for the fix 👍

tomasstrba (approved):
LGTM! 👍 Tab bar scrolls to the active tab like a charm 🥇

macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
ayoy: Yep, nicely spotted! 🙏

ayoy (approved):
Thanks for the prompt fix @jotaemepereira, it works great!

PR #323, opened 2025-03-25
Revert "Fix scrolling when moving to a not visible tab (#311)"

jotaemepereira:
Task/Issue URL: https://app.asana.com/0/1204006570077678/1209751086393041/f
Tech Design URL:
CC:

### Description
This reverts commit 7e15ec6e4f6f83c084e71cd052fcef41249109de.

I was testing the new internal release, and I realized that if you have a pinned tab and try to close the only unpinned tab, the application would be in a weird state where the unpinned tab would not disappear, and new tabs would not appear.

### Testing Steps
1. Open a tab and pin it
2. Open an unpinned tab
4. Close the unpinned tab
5. The unpinned tab should disappear and adding new tabs should work as normal

### Impact and Risks
Low: we are removing a change that introduced a bug.

#### What could go wrong?
Nothing


samsymons (approved):
Works great, thanks!

PR #333, opened 2025-03-25
Fix logic when moving to previously active tab

jotaemepereira:
Task/Issue URL: https://app.asana.com/0/1204006570077678/1209785407607138/f
Tech Design URL:
CC:

### Description
This change fixes two things:
- With the previous change, I introduced a bug in which if the closed tab was not active, we applied the logic as if it was active, which was wrong and unintentional.
- The rule of returning to the previous active tab was missing some logic. On Windows, the previously active tab is only remembered until the user moves to an existing tab or creates new tabs.

### Testing Steps

**Testing steps of the first bug**
1. Open some site like Wikipedia, and open some links from there using CMD + Tap
2. Open a new tab and keep that tab selected
3. Right tap one of the child tabs and close it
4. The active tab should be the new tab and no other sibiling or parent of the closed tab should be selected

**Testing steps of the new rule logic when returning to the previous active tab**
1. Open five tabs (no related)
2. Select the first tab
3. Use CMD + T or tapping the + in the tab bar to open a new tab
4. Close that tab
5. The first tab should be selected

1. Open five tabs (no related)
2. Select the first tab
3. Use CMD + T or tapping the + in the tab bar to open a new tab
4. Select tab number 5 (the one before the recently opened)
5. Go back to the last tab and close it
6. Tab number 5 should be selected instead of the first tab

### Impact and Risks

#### What could go wrong?
The scenario disrupts other flows when closing tabs.

### Quality Considerations
I’ve added unit tests and UI tests to cover this new scenarios.

### Notes to Reviewer
Do as much test as you can for your user flows.

### DIFF
--- a/macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
+++ b/macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
@@ -124,6 +124,10 @@ final class TabCollectionViewModel: NSObject {
         return homePage
     }
 
+    /// This property logic will be true when the user appends a new tab
+    /// it will be set to false when the user moves selects an existing tab
+    private var shouldReturnToPreviousActiveTab: Bool = false
+
     init(
         tabCollection: TabCollection,
         selectionIndex: TabIndex = .unpinned(0),
@@ -226,6 +230,7 @@ final class TabCollectionViewModel: NSObject {
     // MARK: - Selection
 
     @discardableResult func select(at index: TabIndex, forceChange: Bool = false) -> Bool {
+        shouldReturnToPreviousActiveTab = false
         switch index {
         case .unpinned(let i):
             return selectUnpinnedTab(at: i, forceChange: forceChange)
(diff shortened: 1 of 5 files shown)


macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
```diff
@@ -124,6 +124,10 @@ final class TabCollectionViewModel: NSObject {
         return homePage
     }
 
+    /// This property logic will be true when the user appends a new tab
+    /// it will be set to false when the user moves selects an existing tab
```
ayoy: Just making sure the docs are clear

```suggestion
    /// it will be set to false when the user selects an existing tab
```

macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
```diff
@@ -501,7 +511,11 @@ final class TabCollectionViewModel: NSObject {
         select(at: newSelectionIndex, forceChange: forced)
     }
 
-    func getLastSelectedTab() -> TabIndex? {
+    func getPreviosulyActiveTab() -> TabIndex? {
```
ayoy: Just fixing a typo
```suggestion
    func getPreviouslyActiveTab() -> TabIndex? {
```

ayoy (approved):
@jotaemepereira this looks solid to me, I tried breaking it in multiple ways and all flows worked well for me. Great job as always! And thanks for updating UI tests 🫡

PR #341, opened 2025-03-26
Sabrina/add breackage metrics

SabrinaTardio:
Task/Issue URL:
Tech Design URL:
CC:

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


BrowserServicesKit/Sources/BrowserServicesKit/PrivacyConfig/Features/PrivacyFeature.swift
```diff
@@ -69,6 +70,7 @@ public enum PrivacyFeature: String {
     case webViewStateRestoration
     case experimentalBrowserTheming
     case setAsDefaultAndAddToDock
+    case contentScopeExperiments
```
jonathanKingston: Shouldn't this be under the feature fingerprintingCanvas as Marcos is asking in Asana?

SabrinaTardio: Unfortunately we can’t do that.
On apple all sub features of a parent feature that is disabled are disabled, which means any experiment under fingerprintingCanvas would be disabled by default. So we need to have separate enabled parent feature.

Not important for this context but: When we’ll run the project to decide a broader approach for these tests, this btw could be an approach to distinguish feature for which we need to do default cohort assignment before passing the config to C-S-S. Which would work with your proposal.

A lot of the code you see here was part of another PR that has been merged just this has not been rebase. And this branch it’s just my working branch for adding the info to the breakage report def not ready for review I use draft PR to help visualise my changes sometimes.

SabrinaTardio:
Moved to https://github.com/duckduckgo/apple-browsers/pull/372

PR #357, opened 2025-03-26
Refactor FaviconManager API and remove singleton

ayoy:
Task/Issue URL: https://app.asana.com/0/1201899738287924/1209769490544821/f

### Description
This change reworks FaviconManaging, FaviconImageCaching and FaviconReferenceCaching APIs
to move MainActor requirement from protocols and into individual functions/properties, to allow
for initializing in non-isolated contexts.
It also updates some APIs to async and fixes MainActor requirements across the codebase, limiting
the number of warnings.
FaviconManager.shared was removed and replaced with an instance owned by App Delegate
and instantiated there. Nonisolated and MainActor(unsafe) directives were removed where possible.

### Testing Steps
Smoke test favicons:
* Open multiple tabs at once (e.g. from the bookmarks bar or bookmarks manager) and verify that all tabs and their favicons load fine and that no crashing happens.
* Verify that favicons are properly cached and retrieved from cache.
* Verify that favicons are correctly restored during state restoration.

### Impact and Risks
Medium: Favicons refactoring may break user flows and lead to occasional crashes.

### Quality Considerations
Favicons operations were all happening on the main thread already before this change so there should be no
performance impact here.

### Notes to Reviewer
I’ve been testing this change for almost a week now and am feeling pretty confident about it, but I’d appreciate a critical look. Worst case we won’t merge it, but hopefully the changes are reasonable.

---
###### Internal references:


BrowserServicesKit/Sources/Bookmarks/FaviconsFetcher/FaviconsFetchOperation.swift
```diff
@@ -100,21 +100,27 @@ final class FaviconsFetchOperation: Operation, @unchecked Sendable {
     }
 
     func fetchFavicons() async throws {
-        var idsToProcess = try stateStore.getBookmarkIDs()
+        var (bookmarkDomains, idsToProcess) = try await MainActor.run { () -> (BookmarkDomains?, Set<String>) in
```
ayoy: Filtering BookmarkDomains needs to happen on main actor because of the requirement for `hasFavicon`.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -442,9 +449,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // https://app.asana.com/0/1177771139624306/1207024603216659/f
         LottieConfiguration.shared.renderingEngine = .mainThread
 
-        if case .normal = AppVersion.runType {
-            FaviconManager.shared.loadFavicons()
```
ayoy, line 449: I moved this to `FaviconManager` initializer because since it’s not a singleton anymore, we have full control over the timing of the initialization. FaviconManager is now initialized in AppDelegate init after Database.shared is already set up.

macOS/DuckDuckGo/Favicons/Model/FaviconReferenceCache.swift
```diff
@@ -255,26 +223,27 @@ final class FaviconReferenceCache: FaviconReferenceCaching {
...
             }
             return baseDomains.contains(host) && !bookmarkedHosts.contains(host) && !logins.contains(host) && !history.contains(host)
-        }).value
+        }
     }
 
     // MARK: - Private
 
+    @MainActor
     private func insertToHostCache(faviconUrls: (smallFaviconUrl: URL?, mediumFaviconUrl: URL?), host: String, documentUrl: URL) {
```
ayoy, line 243: For these functions that spawn tasks I assumed we do it to improve performance as these functions don’t need to wait for updating the store, so I kept the tasks as is (adding `detached` everywhere because only some of them used Task.detached). Otherwise we could make functions async and await these tasks.

BrowserServicesKit/Sources/Bookmarks/FaviconsFetcher/FaviconsFetchOperation.swift
mallexxx: This seems suboptimal as bookmark IDs fetching involving `missingIDsFile` reading and parsing and domain filtering happens on the main thread; 
Thinking out loud, looking at the async code of the favicon caches - it looks kinda fragile although the protocols are now covered with the `@MainActor` attributes the classes themselves are performing async operations…
e.g. `awaitFaviconsLoaded()` is called from a background thread and is accessing `faviconsLoaded` with a possible data race. 
Although actual data flow and storage access should be fine, I guess, it might be better to convert those to `actor`-s - this should improve data safety and the code effectiveness. But since you‘re saying it was already happening on the main thread - it‘s probably fine to live with it for now, maybe adding a backlog task for `async` refactoring.

SharedPackages/BrowserServicesKit/Sources/Bookmarks/FaviconsFetcher/FaviconsFetchOperation.swift
```diff
@@ -100,21 +100,27 @@ final class FaviconsFetchOperation: Operation, @unchecked Sendable {
...
-            return
+            guard !idsToProcess.isEmpty else {
+                Logger.bookmarks.debug("No new Favicons to fetch")
+                return (nil, [])
+            }
+            Logger.bookmarks.debug("Favicons Fetch Operation started")
+
+            var bookmarkDomains = mapBookmarkDomainsToUUIDs(for: idsToProcess)
+            bookmarkDomains.filterDomains { [weak self] domain in
+                self?.faviconStore.hasFavicon(for: domain) == false
```
mallexxx, line 114: it seems we only need to call `filterDomains`+`hasFavicon` on the main thread, the rest might run in the background?

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -388,6 +389,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         privacyStats = PrivacyStats(databaseProvider: PrivacyStatsDatabase())
 #endif
         PixelKit.configureExperimentKit(featureFlagger: featureFlagger, eventTracker: ExperimentEventTracker(store: UserDefaults.appConfiguration))
+
+#if DEBUG
+        faviconManager = FaviconManager(cacheType: AppVersion.runType == .normal ? .standard : .inMemory)
```
mallexxx: should we check for `requiresEnvironment` instead to cover Int/UI Tests as well or `.normal`+`.uiTests`+`.uiTestsOnboarding` here?

BrowserServicesKit/Sources/Bookmarks/FaviconsFetcher/FaviconsFetchOperation.swift
ayoy: You’re of course right, my initial idea was to implement an actor here, but that would be a heavier refactoring because it would require all `getFavicon*` functions to become async, while they’re actually all used on the Main Actor anyway (in the UI context). I agree that using an actor would be safest, and we could scope it for later.

macOS/DuckDuckGo/Common/View/SwiftUI/FaviconView.swift
```diff
@@ -44,14 +44,13 @@ struct FaviconView: View {
...
 
-    @MainActor(unsafe)
     func refreshImage() {
         if let duckPlayerImage = DuckPlayer.shared.image(for: self) {
             image = duckPlayerImage
             return
         }
 
-        if faviconManagement.areFaviconsLoaded, let url = url {
+        if faviconManagement.isFaviconsLoaded, let url = url {
```
mallexxx: The old way looked more natural😕
Maybe should we consider `isCacheLoaded`?

macOS/DuckDuckGo/Favicons/Model/FaviconImageCache.swift
```diff
@@ -22,52 +22,39 @@ import Common
...
-@MainActor
 protocol FaviconImageCaching {
 
     init(faviconStoring: FaviconStoring)
 
+    @MainActor
     var loaded: Bool { get }
 
+    @MainActor
     func load() async throws
```
mallexxx, line 32: it seems we don‘t need the `load()` method to be fully `@MainActor` covered: `FaviconStore.loadFavicons()` might be run in the background (while `loaded` and `entries` should be covered with `@MainActor` on the class level in this architecture)

SharedPackages/BrowserServicesKit/Sources/Bookmarks/FaviconsFetcher/FaviconsFetchOperation.swift
ayoy: You’re correct! But filterDomains is mutating and I also had errors about sending bookmarkDomains between asynchronous contexts. Happy to take suggestions on how to rewrite it so that it makes sense and only filters on the main actor. I suppose we’d have to rewrite BookmarksDomains somehow which seemed a larger change so I decided to defer it for now.

mallexxx (approved):
The changes look reasonable overall, testing steps worked fine for me, although I think there‘s still room for improvement
I‘m not sure whether it‘s related to the change or to some other browser parts that were already suboptimal but there‘s a considerable lag on the app launch even with few bookmarks.
Consider adding an extra data race protection for the flag I mentioned inline and feel free to merge

macOS/DuckDuckGo/Application/AppDelegate.swift
ayoy: Great point, let me update to check for `requiresEnvironment` 👍

macOS/DuckDuckGo/Common/View/SwiftUI/FaviconView.swift
ayoy: The old school[0] of naming APIs was deliberate about using `is` and `has` regardless of singular/plural, resulting in broken English, but easier variable lookup, so I decided to rename this. That said, I like your idea about `isCacheLoaded`! I’ve updated the name.

[0] – citation needed, I can’t even locate the link, but I read it maybe 10 years ago.

macOS/DuckDuckGo/Favicons/Model/FaviconImageCache.swift
ayoy: I’ve updated this function to not require main actor, and only populate `entries` on the main actor 🫡

PR #376, opened 2025-03-27
Fix NTP User Script leak

samsymons:
Task/Issue URL: https://app.asana.com/0/1199333091098016/1209821590760695/f
Tech Design URL:
CC:

### Description

This PR fixes (in a not so ideal way) a leak with the NTP user scripts. I'll elaborate on why it's not ideal in the comments.

### Testing Steps

1. Open and close a bunch of windows
2. Open the memory graph debugger
3. Filter for `NewTab` and check that you don't see a large number of user content controllers, user scripts, etc.
4. Test that the repro steps in the [original Asana task](https://app.asana.com/0/1201037661562251/1209767015721459/f) are fixed too

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
+++ b/macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
@@ -192,6 +192,7 @@ final class BrowserTabViewController: NSViewController {
     @objc
     private func windowWillClose(_ notification: NSNotification) {
         self.removeWebViewFromHierarchy()
+        self.newTabPageWebViewModel.removeUserScripts()
     }
 
     @objc
--- a/macOS/DuckDuckGo/NewTabPage/NewTabPageWebViewModel.swift
+++ b/macOS/DuckDuckGo/NewTabPage/NewTabPageWebViewModel.swift
@@ -72,6 +72,13 @@ final class NewTabPageWebViewModel: NSObject {
             }
             .store(in: &cancellables)
     }
+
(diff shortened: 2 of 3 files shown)


macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
```diff
@@ -192,6 +192,7 @@ final class BrowserTabViewController: NSViewController {
     @objc
     private func windowWillClose(_ notification: NSNotification) {
         self.removeWebViewFromHierarchy()
+        self.newTabPageWebViewModel.removeUserScripts()
```
samsymons, line 195: I don't like this approach overall, the view model shouldn't necessarily have to care about breaking the retain cycle. However the issue is laid out in [this post](https://stackoverflow.com/a/26383032), where the content controller strongly retains its message handler. We have applied the same fix described in that post elsewhere in the codebase (look at `WeakScriptMessageHandlerBox`) but applying that here wasn't straightforward.

We can't put this step into the view model's `deinit` either due to Swift Concurrency isolation issues, it's not until Swift 6.2 that we have better control there.

Overall this works and isn't actively harmful, but it's a bit of a bandaid rather than a fix at a lower level. Open to suggestions or strong feelings here. :)

ayoy (approved):
Thanks for the fix, @samsymons! Confirming that NewTabPageUserScript and content controller are correctly cleaned up now :sunglasses: 👏

PR #387, opened 2025-03-28
Fix macOS permission popover leak

samsymons:
Task/Issue URL: https://app.asana.com/0/1201037661562251/1209767015721459/f
Tech Design URL:
CC:

### Description

This PR fixes an issue with the permission popover leaking when you close a browser tab while it's still visible.

I ran out time to track down where the actual retain cycle is coming from here, so instead added a workaround where the Address Bar view controller makes sure to close the permission popover when it's disappearing.

### Testing Steps

1. Check the testing steps in the task to ensure that they don't happen any more (for real this time)
2. Smoke test the permission feature otherwise - navigate tabs etc while it's active, that sort of thing

I noticed an issue where if you open the Privacy Dashboard while the permission popover is active, it closes and then immediately reopens - this happens on `main` too so I considered it out of scope here.

### Notes to Reviewer

---
###### Internal references:


macOS/DuckDuckGo/Permissions/Model/PermissionModel.swift
```diff
@@ -81,7 +81,9 @@ final class PermissionModel {
 
     private func subscribe(to permissionManager: PermissionManagerProtocol) {
         permissionManager.permissionPublisher.sink { [weak self] value in
-            self?.permissionManager(permissionManager,
```
samsymons, line 84: This was quietly strongly referencing the `permissionManager`, but I don't know that it was really causing any issues, happy to revert this change.

ayoy: That change looks good to me, I would personally even use `[weak self, weak permissionManager]` and then  `guard let permissionManager` to make it clearer that we’re not capturing it strongly.

samsymons: Damn, thanks - good point about that syntax. Let me change it.

ayoy (approved):
This fix does the job, so let’s merge it. Thanks a lot @samsymons!

PR #394, opened 2025-03-30
Update get_tasks_in_last_internal_release.sh to handle Asana V1 URL

ayoy:
Task/Issue URL: https://app.asana.com/0/1201899738287924/1209664522647706/f

### Description
This change adds support for Asana V1 task URLs to get_tasks_in_last_internal_release.sh which
is used to populate the list of tasks in the release task comment by Dax (“Added in this release:”).
macOS script contained a bunch of dead code that has since been moved to the fastlane plugin so
it was replaced by its newer iOS counterpart.

### Testing Steps
I’ve tested it the following way (you’d need Asana access token in ASANA_ACCESS_TOKEN env variable):
1. Enter iOS directory
2. Run the following command and verify that it returns some V1 URLs. Note down task IDs for these URLs.
```
git log 7.162.0+ios..HEAD ./ ../SharedPackages | grep -A 1 'Task.*URL' | sed -nE 's|.*(https://app\.asana\.com.*)|\1|p' | uniq
```
3. Update line 62 in `./scripts/get_tasks_in_last_internal_release.sh` to say `last_release_tag=“7.162.0+ios"`
4. Run `./scripts/get_tasks_in_last_internal_release.sh` and verify that the output contains task IDs from V1 URLs from step 2.
5. Repeat the same from step 1 but with macOS directory (you can use the same iOS tag)

### Impact and Risks
None

---
###### Internal references:


macOS/scripts/get_tasks_in_last_internal_release.sh
ayoy, line 1: Overall this file is a tech debt and we should move it to the fastlane plugin one day, but we decided to keep it as is to speed up the iOS release automation project.

PR #410, opened 2025-03-31
macOS: Add Duck.ai to omnibar

Bunn:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209441152832762/f

### Testing Steps
Before testing it, set the app environment like so:

Debug → Remote Configuration
https://www.jsonblob.com/api/1354855059191685120

Debug → AIChat → Web Communication
 https://use-devtesting12.duckduckgo.com/

1. Make sure you see the omnibar duck.ai button visible by default
2. Go to settings, turn it on/off, see if it's being updated as expected
3. Click the button without any text in the address bar. Duck.ai should open with it blank
4. Click the button with some text in the address bar. Duck.ai should open and auto submit the query
5. Command click the duck.ai button. It should open in a new tab
6. Write something in the omnibar. Press Shift + Enter. It should open duck.ai and auto submit the text
7. Do the same without the duck.ai button visible. It should open the search query, and not duck.ai

----
The toolbar duck.ai button and the popover with it's onboarding was removed as part of this feature:
1. Open https://duck.ai twice. You should NOT see a popover about duck.ai
2. Right click the toolbar where we have the downloads/passwords/etc button. There should NOT be an option for duck.ai

### Impact and Risks
Low: Minor visual changes

#### What could go wrong?
Duck.ai button not working inside the omnibar 
Visual glitches with the bookmarks and/or close button


SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/PrivacyConfig/Features/PrivacyFeature.swift
```diff
@@ -131,12 +131,6 @@ public enum AIChatSubfeature: String, Equatable, PrivacySubfeature {
         .aiChat
     }
 
-    /// Displays the settings item for showing a shortcut in the macOS application menu.
```
Bunn, line 134: Following with the [removal of FF](https://app.asana.com/0/1200194497630846/1209313616180294/f) approach. I'm removing some old AI Chat feature flags here

macOS/DuckDuckGo/AIChat/AIChatAddressBarQueryExtractor.swift
```diff
@@ -0,0 +1,51 @@
...
+    /// Extracts a query string from the given value.
+    ///
+    /// - Parameter value: The value from which to extract the query string.
+    /// - Returns: A query string if it can be extracted from the value, otherwise `nil`.
+    func queryForValue(_ value: ValueType) -> String?
+}
+
+/// A struct that implements the `QueryExtractable` protocol for extracting query strings
+/// from values of type `AddressBarTextField.Value`.
+struct AIChatAddressBarQueryExtractor: QueryExtractable {
```
Bunn: Decided to extract this into a struct so it's easier to test and maintain. Not 100% happy with the naming of it but it's the best I could came up with 😅  open to suggestions =)

macOS/DuckDuckGo/AIChat/AIChatMenuVisibilityConfigurable.swift
```diff
@@ -21,50 +21,30 @@ import BrowserServicesKit
 
 protocol AIChatMenuVisibilityConfigurable {
 
-    /// This property validates remote feature flags and user settings to determine if the shortcut
+    /// This property validates user settings to determine if the shortcut
     /// should be presented to the user.
     ///
-    /// - Returns: `true` if the application menu shortcut should be displayed; otherwise, `false`.
```
Bunn, line 27: Related to the feature flags removal

macOS/DuckDuckGo/AIChat/AIChatPreferencesStorage.swift
```diff
@@ -38,20 +37,8 @@ struct DefaultAIChatPreferencesStorage: AIChatPreferencesStorage {
         userDefaults.showAIChatShortcutInApplicationMenuPublisher
     }
 
-    var shouldDisplayToolbarShortcutPublisher: AnyPublisher<Bool, Never> {
```
Bunn, line 41: Removing lots of code related to the toolbar

macOS/DuckDuckGo/AIChat/UserScript/AIChatMessageHandler.swift
```diff
@@ -0,0 +1,91 @@
...
+//  limitations under the License.
+//
+
+enum AIChatMessageType {
+    case nativeConfigValues
+    case nativeHandoffData
+    case nativePrompt
+}
+
+protocol AIChatMessageHandling {
```
Bunn, line 25: I'm planning to move this to a shared AIChat package with iOS. It will be done as part of the AOR, didn't want to scope creep the project now.

macOS/DuckDuckGo/Common/Localizables/UserText.swift
```diff
@@ -434,15 +434,7 @@ struct UserText {
 
     // Misc
 
-    // AI Chat
-    static let aiChatOnboardingPopoverTitle = NSLocalizedString("duckai.onboarding.popover.title", value: "Launch Duck.ai directly from your toolbar", comment: "AI Chat onboarding popover title")
-    static let aiChatOnboardingPopoverMessageMarkdown = NSLocalizedString("duck.ai.onboarding.popover.message-markdown", value: "You can adjust this and other Duck.ai features in **Settings** > **Duck.ai**.", comment: "AI Chat onboarding popover message, make sure to keep Settings and AI Chat inside ** **")
-    static let aiChatOnboardingPopoverMessageFallback = NSLocalizedString("ai-chat.onboarding.popover.message-fallback", value: "You can adjust this and other AI Chat features in Settings > AI Chat", comment: "AI Chat onboarding popover message continuation")
```
Bunn, line 440: lots of toolbar onboarding stuff that was removed

macOS/DuckDuckGo/NavigationBar/View/NavigationBar.storyboard
```diff
@@ -742,6 +713,43 @@
                                     <action selector="clearButtonAction:" target="mcr-nj-GmC" id="m9h-NY-Ujx"/>
                                 </connections>
                             </button>
+                            <stackView distribution="fill" orientation="horizontal" alignment="top" horizontalStackHuggingPriority="249.99998474121094" verticalStackHuggingPriority="249.99998474121094" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="L9C-04-I02">
```
Bunn: Xib magic 😅

* Added the duck.ai button inside a stack view for future proofing and to easily handle the case where it’s not visible.
* Changed the constraints for the bookmark and clear buttons to include horizontal spacing in the new stack view.

macOS/DuckDuckGo/NavigationBar/View/NavigationBarPopovers.swift
```diff
@@ -229,29 +228,13 @@ final class NavigationBarPopovers: NSObject, PopoverPresenter {
...
-        }
-
         if autofillOnboardingPopover?.isShown ?? false {
             autofillOnboardingPopover?.close()
         }
 
         return true
     }
 
-    func showAIChatOnboardingPopover(from button: MouseOverButton,
```
Bunn, line 243: Removed the old toolbar onboarding

macOS/DuckDuckGo/NavigationBar/View/NavigationBarViewController.swift
```diff
@@ -1039,60 +1028,6 @@ final class NavigationBarViewController: NSViewController {
             .store(in: &navigationButtonsCancellables)
     }
 
-    // MARK: - AI Chat
-
-    private func subscribeToAIChatOnboarding() {
```
Bunn, line 1060: Old duck.ai toolbar related functions

macOS/UnitTests/Menus/MoreOptionsMenuTests.swift
```diff
@@ -171,25 +171,26 @@ final class MoreOptionsMenuTests: XCTestCase {
...
-        XCTAssertFalse(moreOptionsMenu.items[14].hasSubmenu)
-        XCTAssertTrue(moreOptionsMenu.items[15].isSeparatorItem)
-        XCTAssertEqual(moreOptionsMenu.items[16].title, UserText.fireproofSite)
-        XCTAssertEqual(moreOptionsMenu.items[17].title, UserText.findInPageMenuItem)
-        XCTAssertEqual(moreOptionsMenu.items[18].title, UserText.shareMenuItem)
-        XCTAssertEqual(moreOptionsMenu.items[19].title, UserText.printMenuItem)
-        XCTAssertTrue(moreOptionsMenu.items[20].isSeparatorItem)
-        XCTAssertEqual(moreOptionsMenu.items[21].title, UserText.mainMenuHelp)
-        XCTAssertEqual(moreOptionsMenu.items[22].title, UserText.settings)
+        XCTAssertEqual(moreOptionsMenu.items[5].title, UserText.newAIChatMenuItem)
```
Bunn, line 174: The tests had the feature flag off for the duck.ai button. Now that I have removed the feature flag, it needs to be considered as well. This essentially offsets all indexes by +1 after the duck.ai button

ayoy:
@Bunn this is looking great, and I’m excited about better integration of Duck AI with the macOS browser!

I’ll have a look at the code tomorrow, but for now wanted to report 2 things:

1. This looks like a bug to me. When I’m on Duck AI tab, typing a phrase in the address bar and pressing Shift+Enter has no effect. I think I'd expect it to open a new Duck AI tab.
2. Now that we have Duck AI button inside the address bar, we should probably update the address bar context menu with the Duck AI button visibility control. Perhaps that’s something you could suggest (unless it was deliberately omitted).
Screenshot 2025-04-03 at 22 45 06

macOS/DuckDuckGo/AIChat/AIChatAddressBarQueryExtractor.swift
```diff
@@ -0,0 +1,51 @@
...
+    /// Extracts a query string from the given `AddressBarTextField.Value`.
+    ///
+    /// - Parameter value: The `AddressBarTextField.Value` from which to extract the query string.
+    /// - Returns: A query string if it can be extracted from the value, otherwise `nil`.
+    func queryForValue(_ value: ValueType) -> String? {
+        switch value {
+        case let .text(text, _):
+            return text
+        case let .url(_, url, _):
+            print("URL \(url)")
```
ayoy: Please remove 👮‍♂️

macOS/DuckDuckGo/AIChat/AIChatAddressBarQueryExtractor.swift
ayoy: Aren’t we actually extracting a _prompt_ here? That would be my only suggestion. If that works, I’d call the protocol `AIChatPromptExtracting` (rather than Extractable, because the protocol describes an action rather than a capability) and `AIChatAddressBarPromptExtractor`.

macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
```diff
@@ -19,30 +19,39 @@
 import UserScript
 
 protocol AIChatUserScriptHandling {
-    func handleGetUserValues(params: Any, message: UserScriptMessage) -> Encodable?
-    func openSettings(params: Any, message: UserScriptMessage) async -> Encodable?
+    func openAIChatSettings(params: Any, message: UserScriptMessage) async -> Encodable?
+    func getAIChatNativeConfigValues(params: Any, message: UserScriptMessage) -> Encodable?
+    func closeAIChat(params: Any, message: UserScriptMessage) -> Encodable?
```
ayoy: `Subfeature.Handler` is an `async throws` closure, so you can define all your methods in the protocol as such (I do it in my user scripts anyway), and then you’d avoid using `Task` in `closeAIChat`.

macOS/DuckDuckGo/MainWindow/MainViewController.swift
```diff
@@ -611,7 +610,11 @@ extension MainViewController {
         case kVK_Return  where navigationBarViewController.addressBarViewController?
                 .addressBarTextField.isFirstResponder == true:
 
-            navigationBarViewController.addressBarViewController?.addressBarTextField.addressBarEnterPressed()
+            if flags.contains(.shift) && aiChatMenuConfig.shouldDisplayAddressBarShortcut {
```
ayoy, line 613: smooth :sunglasses:

macOS/DuckDuckGo/Menus/MainMenuActions.swift
```diff
@@ -68,7 +68,7 @@ extension AppDelegate {
 
     @objc func newAIChat(_ sender: Any?) {
         DispatchQueue.main.async {
-            AIChatTabOpener.openAIChatTab()
+            AIChatTabOpener.shared.openAIChatTab()
```
ayoy: Let’s not add new singletons. Instead, if you need an app-wide instance of an object, declare it in App Delegate. If you think you’ll bloat App Delegate too much by adding several new properties, consider creating `AIChatCoordinator` that would initialize them all and keep that one in App Delegate (similar to `newTabPageCoordinator` that instantiates NTP actions manager).

These singletons are now small and easy, but in case they grow and we start facing problems due to lack of control over their initialization time, we’ll get into trouble :)

macOS/DuckDuckGo/NavigationBar/View/AddressBarButtonsViewController.swift
```diff
@@ -281,6 +290,16 @@ final class AddressBarButtonsViewController: NSViewController {
         openPrivacyDashboardPopover()
     }
 
+    @IBAction func aiChatButtonAction(_ sender: Any) {
+        let shouldOpenInNewTab = NSEvent.modifierFlags.contains(.command)
```
ayoy: I understand that this is a product decision, but perhaps worth raising that cmd+click in other places of the app opens a new tab in background, not foreground. See `NewTabPageLinkOpener.open()` (available in main) for the full logic (also including shift and option key modifiers).

macOS/DuckDuckGo/NavigationBar/View/NavigationBar.storyboard
ayoy: That sounds good :D thanks for taking time to explain the changes :)

macOS/DuckDuckGo/NavigationBar/View/NavigationBarViewController.swift
ayoy: Great to see the cleanups 🫡

ayoy (commented):
The code is great too, thanks a lot @Bunn!

I’m leaving a few comments with suggestions for improvements. I’d mainly be interested in handling the singleton, and the rest is optional. Let me know what you think, and when you make changes, ping me to have another look :)

macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
Bunn: Good catch, thanks!

macOS/DuckDuckGo/Menus/MainMenuActions.swift
Bunn: Good call. Thanks for the suggestion!

macOS/DuckDuckGo/NavigationBar/View/AddressBarButtonsViewController.swift
Bunn: I’ll do it. I think it makes sense. Thanks

ayoy (approved):
Changes are looking great, thank you very much @Bunn!

PR #412, opened 2025-03-31
Consolidate xcode-version into a single file

samsymons:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1209550681759862/f
Tech Design URL:
CC:

### Description

This PR consolidates all `xcode-version` files into a single one at the root of the repo. It also adds a new action that can set the version for you automatically, factoring in the `working-directory` setting, while also allowing the version to be overridden.

### Testing Steps

1. Check that CI is green
2. Check that [this iOS ad-hoc workflow](https://github.com/duckduckgo/apple-browsers/actions/runs/14181797736) succeeded
3. Check that [this macOS notarized build workflow](https://github.com/duckduckgo/apple-browsers/actions/runs/14181801533) succeeded
4. Check that there are no remaining `xcode-version` files locally
5. Check that upgrading or downgrading the root copy of the file does the right thing (it should warn you if your version is too new, and fail the build if it's too old)

#### What could go wrong?

The only risk here is at the CI level, this will break workflows if they aren't updated correctly.

### Notes to Reviewer

---
###### Internal references:


.github/actions/select-xcode-version/action.yml
```diff
@@ -0,0 +1,45 @@
...
+  xcode-version:
+    description: 'The selected Xcode version'
+    value: ${{ steps.select-xcode.outputs.xcode-version }}
+
+runs:
+  using: 'composite'
+  steps:
+    - id: select-xcode
+      shell: bash
+      run: |
```
samsymons, line 19: I would be happy to move this into a dedicated script so that ShellCheck can look at it, no strong opinions.

ayoy: ShellCheck isn’t configured to look at actions scripts at the moment anyway. I would put it in a separate bash file just for better ergonomics (syntax highlighting, etc.).

ayoy (approved):
Looks great, thanks @samsymons! If you have a moment, you could put the script into a dedicated file, but it’s simple enough (and works and won’t be changing much) that we can keep it directly in the action as well.

PR #421, opened 2025-04-01
Show Add Bookmark dialog for Bookmark This Page in Tab Context Menu

mallexxx:
Task/Issue URL: https://app.asana.com/0/1202406491309510/1206759607104110/f

### Description
- Show Add Bookmark dialog when "Bookmark This Page" is chosen in a Tab Context Menu

### Testing Steps

1. Open couple of tabs with non-bookmarked websites
2. Right-click inactive tab, choose Bookmark this page, validate Add Bookmark dialog is shown; click Add/Cancel, validate bookmark is created/dialog is hidden
3. Validate the same works for an active tab
4. Validate the scenario for a pinned tab
5. Activate another app; Right click inactive tab in DuckDuckGo while its window is displayed under the current app‘s window (non-active) and choose Bookmark this page, validate the scenario works as expected

### Notes to Reviewer

---
###### Internal references:


ayoy (approved):
LGTM!

One thing that I noticed is if the app is this scenario:
1. Have a pinned tab that is bookmarked
2. Activate another app
3. Delete the bookmark from the pinned tab via context menu (the app won’t activate)
4. Trigger context menu for the pinned tab again
5. It still shows “Delete Bookmark” even though the bookmark was deleted. The action then does nothing.

This is, however, present in the current production build and this change doesn’t affect it so we can handle it separately.

PR #422, opened 2025-04-01
macOS VPN: Exclude push.apple.com traffic from our proxy

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/1207603085593419/1209681033289804/f

### Description

Exclude push.apple.com traffic from our proxy, so that it will always be routed by the system routing tables / VPN.  This should resolve issues we were seeing where the system would drop phone calls from handoff on connect.

### Testing Steps

1. Connect to VPN
2. Make sure you're using an Apple ID that you're also using on a phone.
3. Call your phone.
4. macOS should show the call too - answer the call on macOS
5. The call should connect.

### Impact and Risks

Low: excluding traffic from the proxy just means users can't exclude said traffic from the VPN.

#### What could go wrong?

Not much - maybe users could try to exclude apple.com and notice it's not working for some of the traffic, but seems like a very edge case and unlikely t be noticed.

### Quality Considerations

None

### Notes to Reviewer

None

---
###### Internal references:


samsymons (approved):
LGTM, thanks!

PR #426, opened 2025-04-01
Increase navigation bar size and change paddings for new visual refresh

jotaemepereira:
Task/Issue URL: https://app.asana.com/0/1204006570077678/1209350564013383/f
Tech Design URL:
CC:

### Description
This adds the new height for the navigation bar; now, it is `52` for both the default and home page. The pop-up remains unchanged.

It also changes the paddings, and now we have `6` for both the top and bottom.

This change introduces how we will use the feature flag to apply the different changes.

### Testing Steps
1. Select the feature flag from `Debug -> Feature Flag Overrides -> visualRefresh`
2. Open a new window
3. The navigation bar should be taller compared to the other one
4. Turn off the override
5. Open another window
6. The navigation bar size should be back to production values

### Impact and Risks

#### What could go wrong?
I moved some logic around in the `AddressBarSizeClass`. Now, part of that logic for calculating the address bar height and padding is part of the `VisualStyleConfigurable`.

### Quality Considerations
Nothing to consider.

### Notes to Reviewer
It’s a simple change. The Dax logo will be larger when the address bar height increases, but we will move the logo to inside the address bar in the next PR, so I’m not fixing that now.

—
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
+++ b/macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
@@ -21,18 +21,88 @@ import BrowserServicesKit
 import FeatureFlags
 
 protocol VisualStyleConfigurable {
-    var toolbarHeight: CGFloat { get }
+    func addressBarHeight(for type: AddressBarSizeClass) -> CGFloat
+    func addressBarTopPaddig(for type: AddressBarSizeClass) -> CGFloat
+    func addressBarBottomPaddig(for type: AddressBarSizeClass) -> CGFloat
+}
+
+enum AddressBarSizeClass {
+    case `default`
+    case homePage
+    case popUpWindow
+
+    var logoWidth: CGFloat {
+        switch self {
+        case .homePage: 44
+        case .popUpWindow, .default: 0
+        }
+    }
+
+    var isLogoVisible: Bool {
+        switch self {
+        case .homePage: true
+        case .popUpWindow, .default: false
+        }
+    }
 }
 
 struct VisualStyle {
-    let toolbarHeight: CGFloat
+    private let addressBarHeightForDefault: CGFloat
+    private let addressBarHeightForHomePage: CGFloat
+    private let addressBarHeightForPopUpWindow: CGFloat
+    private let addressBarTopPaddingForDefault: CGFloat
+    private let addressBarTopPaddingForHomePage: CGFloat
+    private let addressBarTopPaddingForPopUpWindow: CGFloat
+    private let addressBarBottomPaddingForDefault: CGFloat
+    private let addressBarBottomPaddingForHomePage: CGFloat
+    private let addressBarBottomPaddingForPopUpWindow: CGFloat
+
+    func addressBarHeight(for type: AddressBarSizeClass) -> CGFloat {
+        switch type {
+        case .default: return addressBarHeightForDefault
+        case .homePage: return addressBarHeightForHomePage
+        case .popUpWindow: return addressBarHeightForPopUpWindow
+        }
+    }
+
+    func addressBarTopPaddig(for type: AddressBarSizeClass) -> CGFloat {
+        switch type {
+        case .default: return addressBarTopPaddingForDefault
+        case .homePage: return addressBarTopPaddingForHomePage
+        case .popUpWindow: return addressBarTopPaddingForPopUpWindow
+        }
+    }
+
+    func addressBarBottomPaddig(for type: AddressBarSizeClass) -> CGFloat {
+        switch type {
+        case .default: return addressBarBottomPaddingForDefault
+        case .homePage: return addressBarBottomPaddingForHomePage
+        case .popUpWindow: return addressBarBottomPaddingForPopUpWindow
+        }
+    }
 
     static var old: VisualStyle {
-        return VisualStyle(toolbarHeight: 44)
+        return VisualStyle(addressBarHeightForDefault: 48,
+                           addressBarHeightForHomePage: 52,
+                           addressBarHeightForPopUpWindow: 42,
+                           addressBarTopPaddingForDefault: 6,
+                           addressBarTopPaddingForHomePage: 10,
+                           addressBarTopPaddingForPopUpWindow: 0,
+                           addressBarBottomPaddingForDefault: 6,
+                           addressBarBottomPaddingForHomePage: 8,
+                           addressBarBottomPaddingForPopUpWindow: 0)
     }
 
     static var new: VisualStyle {
-        return VisualStyle(toolbarHeight: 64)
+        return VisualStyle(addressBarHeightForDefault: 52,
+                           addressBarHeightForHomePage: 52,
+                           addressBarHeightForPopUpWindow: 52,
+                           addressBarTopPaddingForDefault: 6,
+                           addressBarTopPaddingForHomePage: 6,
+                           addressBarTopPaddingForPopUpWindow: 6,
+                           addressBarBottomPaddingForDefault: 6,
+                           addressBarBottomPaddingForHomePage: 6,
+                           addressBarBottomPaddingForPopUpWindow: 6)
     }
 }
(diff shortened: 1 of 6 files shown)


macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
```diff
@@ -21,18 +21,88 @@ import BrowserServicesKit
 import FeatureFlags
 
 protocol VisualStyleConfigurable {
```
ayoy: Sorry for not raising it in your previous pull request, but I’d like to revise the naming here. The name `VisualStyleConfigurable` indicates a trait of a view (e.g. you could make NSButton “visual style configurable” by conforming to that protocol). Meanwhile, this protocol actually defines the API for `VisualStyleManager`, which only provides styling values for views.

What would you say if we did the following:
* rename `VisualStyleConfigurable` to `VisualStyleProviding`,
* rename `VisualStyleManager` to `VisualStyleProvider`.
* rename `visualStyleConfigurable` properties all over the codebase as `visualStyleProvider` or just `visualStyle`.

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
```diff
@@ -21,18 +21,88 @@ import BrowserServicesKit
...
-    var toolbarHeight: CGFloat { get }
+    func addressBarHeight(for type: AddressBarSizeClass) -> CGFloat
+    func addressBarTopPaddig(for type: AddressBarSizeClass) -> CGFloat
+    func addressBarBottomPaddig(for type: AddressBarSizeClass) -> CGFloat
```
ayoy: Typo in _padding_ :)

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
```diff
@@ -21,18 +21,88 @@ import BrowserServicesKit
...
+
+    func addressBarBottomPaddig(for type: AddressBarSizeClass) -> CGFloat {
+        switch type {
+        case .default: return addressBarBottomPaddingForDefault
+        case .homePage: return addressBarBottomPaddingForHomePage
+        case .popUpWindow: return addressBarBottomPaddingForPopUpWindow
+        }
+    }
 
     static var old: VisualStyle {
```
ayoy: Given that “new” will be the new “current” :D I would consider renaming them as “legacy” and “current”, and then renaming “currentStyle” as “activeStyle”, but I’m not strong on that.

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
```diff
@@ -67,7 +149,7 @@ final class VisualStyleManager: VisualStyleConfigurable {
         overridesHandler.flagDidChangePublisher
             .filter { $0.0 == .visualRefresh }
             .sink { (_, enabled) in
-                /// Here I need to apply the visual changes. Should I restart the app?
+                /// Here I need to apply the visual changes. The easier way should be to restart the app.
```
ayoy: This (feature flag overriding) is an internal-only feature so it’s fine if it requires app restart. If you want, you can show an (unlocalized) alert from here stating that the user needs to restart the app for the changes to make effect. Perhaps add a “don’t show this again” checkbox. I suppose it may streamline the Ship Review, by reducing confusion :)

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
```diff
@@ -41,18 +111,30 @@ final class VisualStyleManager: VisualStyleConfigurable {
...
+        return currentStyle.addressBarTopPaddig(for: type)
+    }
+
+    func addressBarBottomPaddig(for type: AddressBarSizeClass) -> CGFloat {
+        return currentStyle.addressBarBottomPaddig(for: type)
+    }
+
+    // MARK: - Private properties
+
+    private var isEnabled: Bool {
```
ayoy: If I can be honest with you, `isEnabled` sounds more cryptic than `featureFlagger.isFeatureOn(.visualRefresh)` :D because it’s hard to figure out for a reader what actually is enabled. I’d personally remove it and use the feature flagger directly inside `currentStyle`.

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
```diff
@@ -41,18 +111,30 @@ final class VisualStyleManager: VisualStyleConfigurable {
...
+        return currentStyle.addressBarHeight(for: type)
+    }
+
+    func addressBarTopPaddig(for type: AddressBarSizeClass) -> CGFloat {
+        return currentStyle.addressBarTopPaddig(for: type)
+    }
+
+    func addressBarBottomPaddig(for type: AddressBarSizeClass) -> CGFloat {
+        return currentStyle.addressBarBottomPaddig(for: type)
+    }
```
ayoy: I worry that at the end of the project, the `VisualStyleConfigurable` protocol will grow significantly, and there is some code “duplication” (so to speak) here, because you define the same methods in `VisualStyleManager` and in `VisualStyle`. Have you considered conforming `VisualStyle` to `VisualStyleConfigurable` instead, and returning a `VisualStyle` object (or actually `any VisualStyleConfigurable`) from the manager?

Then the call site would look like this:
```swift
manager.currentStyle.addressBarBottomPadding(for: .default)
```

If you think that `currentStyle` in the call site is superfluous and you prefer the way you’re doing it now, that’s completely fine by me – but I just wanted to point out that `VisualStyle` should also conform to `VisualStyleConfigurable`.

ayoy (commented):
@jotaemepereira this works great of course 💪 I’m leaving a number of comments for your consideration – they’re all related to how the code is organized, with some ideas for improving ergonomics. Please have a look and let me know what you think, I’d be happy to have another look at the code. Also happy to jump on a call to discuss it. Thanks a lot!

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
jotaemepereira: Good call. Changed!

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
jotaemepereira: I like legacy and current much more 👍🏼

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
jotaemepereira: Honesty is what I want in these reviews :). Removed the `isEnabled` I will use isFeatureOn instead

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
jotaemepereira: Good idea. Let me create a follow-up task in my project to do this. I want to keep every PR slim for easier reviews

macOS/DuckDuckGo/VisualRefresh/VisualStyleConfigurable.swift
jotaemepereira: I like this idea. I switched, and the manager will be mainly in charge of passing the currently selected style.

jotaemepereira:
Thanks for the feedback @ayoy 🙌🏼 . I believe I addressed all of it.

jotaemepereira:
> Looks great now, thanks @jotaemepereira!
> 
> One thing you can consider for later is to abstract VisualStyleManager interface into a protocol so that classes that use it can take a mock object for tests, if needed.

Thanks for the feedback! I’ve added a protocol for the visual manager :)

PR #440, opened 2025-04-03
macOS - Add version to app menu and copy version item to send feedback

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1209856715496622/f
Tech Design URL:
CC:

### Description
Add version to app level duckduckgo menu and add a copy version menu item to the feedback, both for internal users.

### Testing Steps
1. With internal user state off - check the app level DuckDuckGo menu.  The "About DuckDuckGo" item should not have the version number in it.
2. With internal user state off - check the feedback menu. It should say "Send Feedback" (no version) and there should be no "copy version" submenu item.
1. With internal user state on - check the app level DuckDuckGo menu.  The "About DuckDuckGo" item should include the version number
3. With internal user state on - check the feedback menu. It should say "Send Feedback" and include the version and there should be a sub menu item called "Copy Version" which copies the version in to the pasteboard when selected.

### Impact and Risks
Low: Internal user only

#### What could go wrong?
End users could up seeing the version and that extra item in a non-localised way.

### Quality Considerations
n/a

### Notes to Reviewer
Ensure it's only for internal users

---
###### Internal references:


ayoy (approved):
LGTM 😎

PR #448, opened 2025-04-03
OmniBar visual updates on iOS

dus7:
Task/Issue URL: https://app.asana.com/0/1206226850447395/1209485028986788
Tech Design URL:
CC:

### Description

Introduces a new look and feel for OmniBar. Notable changes:
* OmniBar now has a main ViewController class (`OmniBarViewController`) which is subclassed for each of OmniBar versions (`Default` and `Updated`).
* Keyboard calculation has beed updated and should now work for arbitrary OmniBar height.
* Overlay controller placement in float mode (iPad) has been modified to leverage constraints instead of width. This was needed to place the controller under new OmniBar properly.

### Testing Steps
1. **Disable** experimental appearance.
2. Perform smoke check for OmniBar with both top and bottom position.
3. Perform smoke check for OmniBar on iPad.
4. Do similar checks with experimental updates enabled.

### Impact and Risks

#### What could go wrong?
Current OmniBar and features around it may be broken because of the architectural changes. Keyboard insets / bars animations can be affected.

### Notes to Reviewer

---
###### Internal references:


brindy (approved):
Generally LGTM, good job overall, a few notes though:

1 (optional)/  `m.addressbar.cancel.ntp` seems to fire multiple times when the new UI is enabled (it's fine otherwise).

Note: I a fixed a bug on main where `m.addressbar.clicked.ntp` was firing too much, but your version doesn't, so when resolving the conflict just use your version of the `DefaultOmnibarViewController`.

2 (optional)/ Consider not using `New` in names.  Mainly because something will always be new and eventually we'll have to change this to be the current default.

Maybe we should come up with a codename for this work and use that then we can just delete the old stuff with minimal changes when the day comes.  And you know we'll end up doing something else here in the future so we can do the same thing then - use a codename for the 'new' stuff to minimise the amount of code we have to change when it gets promoted to default.

3 (just a thought)/ I'm not massively keen on having the inheritance, but I understand the advantages and can't really think of an alternative that doesn't leave us open to the two version of the code getting out of sync while also keeping code changes to production minimal.

4 (possibly design feedback)/ bbc.com looks like it loses the rounded corners somehow, but if you scroll the page they are there.  Some weird edge case?

Simulator Screenshot - iPhone 16 Pro - 2025-04-09 at 15 56 14

Simulator Screenshot - iPhone 16 Pro - 2025-04-09 at 15 56 32

5 (possibly out of scope)/ The error page does not appear to be styled correctly. After going to bbc.com then manually navigate to some nonsense domain it looks odd:

Simulator Screenshot - iPhone 16 Pro - 2025-04-09 at 15 55 33

cc @jdjackson, @miasma13 since I'll be out next week so maybe one of them can help with additional reviews if needed

iOS/DuckDuckGo/Refactoring/Common/OmniBarViewController.swift
```diff
@@ -0,0 +1,670 @@
...
+import Core
+
+class OmniBarViewController: UIViewController, OmniBar {
+
+    // MARK: - OmniBar conformance
+
+    // swiftlint:disable:next force_cast
+    var barView: any OmniBarView { view as! OmniBarView }
+
+    var isTextFieldEditing: Bool { false }
```
brindy: Omid noticed that AI Chat isn't submitting text from the address bar and I think it's because this should delegate to the OmniBarView/textField? It doesn't appear to be overridden so always returns false.

PR #451, opened 2025-04-04
Make initialial secure vault failures no longer a crash, and add pixels so we know they happen

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/1201037661562251/1209880191552590/f
Tech Design URL:
CC:

### Description
As title
This consists of two things:
- Going back to making it not a crash
- Adding pixels so it no longer fails silently

This still isn't great as it means in these cases PIR is in an undefined state and just won't work, but at we'll at least have visibility on that and it'll go back to only affecting PIR rather than the rest of the app.

It's not great that this has existed silently forever, so we should endeavor to fix it properly at the earliest opportunity

### Testing Steps

1. Smoke test PIR 
2. In SecureVaultFactory.makeVault, line 63, make it throw an error
3. Run the main app and check the new pixel is fired with the correct name (i.e. the main app one and not the background agent one), and has the usual error parameters
4. Run the background agent and check the same thing

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/macOS/DBPE2ETests/DBPEndToEndTests.swift
+++ b/macOS/DBPE2ETests/DBPEndToEndTests.swift
@@ -47,20 +47,20 @@ final class DBPEndToEndTests: XCTestCase {
         communicationLayer = DBPUICommunicationLayer(webURLSettings:
                                                         DataBrokerProtectionWebUIURLSettings(UserDefaults.standard),
                                                      privacyConfig: PrivacyConfigurationManagingMock())
-        communicationLayer.delegate = pirProtectionManager.dataManager.cache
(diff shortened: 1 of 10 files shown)


THISISDINOSAUR:
Hmm now I've pointed at the hotfix branch the diff is messed up, gimme a sec to fix

tomasstrba (approved):
LGTM! In case storage vault creation fails, there is a pixel sent and no exception.

Screenshot 2025-04-04 at 13 21 07

I would be more comportable if @aataraxiaa could take a look at this as well, as he has more context

aataraxiaa (approved):
Tested and works as expected @THISISDINOSAUR

PR #454, opened 2025-04-04
Fix crash related on onboarding caused by referencing wrong button

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414709148257752/1209892386356515/f


alessandroboron (approved):
🙇

PR #468, opened 2025-04-09
Fix of keyboard shortcuts on various keyboard layouts

tomasstrba:
Task/Issue URL: https://app.asana.com/0/1148564399326804/1208554716251487/f
CC: @mallexxx 

### Description
Aavoid consuming events after they trigger an application menu item, and instead allow them to pass through to the web view.

### Testing Steps
**Test AZERTY keyboard**
1. Add French AZERTY keyboard to your keyboard layouts
2. Open canva.com
3. Create a new presentation
4. Add a text and change it a couple of times
5. Set the keyboard layout to French AZERTY and trigger CMD+Z to undo your actions
6. Make sure the shortcut works

**Test keyboard shortcuts from original fix**
1. See https://github.com/duckduckgo/macos-browser/pull/177 and execute testing steps
### Impact and Risks

---
###### Internal references:


macOS/DuckDuckGo/MainWindow/MainViewController.swift
```diff
@@ -642,14 +642,14 @@ extension MainViewController {
              kVK_ANSI_R where flags == .command:
             guard view.window?.firstResponder is WebView else { return false }
             NSApp.menu?.performKeyEquivalent(with: event)
-            return true
+            return false
```
tomasstrba, line 641: @mallexxx, do you see any problems if events fall through to the web view in these particular cases?

mallexxx (changes requested):
@tomasstrba although the testing steps you described do work, it seems not a correct solution, the logics here was made to handle the main key shortcuts in the browser UI and not letting them get to the web view; 
The breakage here is caused by the actual keyboard layout not matching ANSI character table, so the obvious solution not breaking the logics and the original issue behind it - is to check the `event.keyEquivalent` string to match characters instead of the ANSI map check - that should fix the issue.
That of course needs validation for non-US layouts like japanese/korean etc… I‘ve checked it on Russian and it seems fitting our needs: `event.keyEquivalent` returns "z" when ⌘я is pressed

tomasstrba:
@mallexxx, thank you for the review. I'd still like to revisit the decision that we shouldn't let shortcuts to reach webview, but not as a part of this task. For now, I've refactored the solution based on your suggestion and it's ready for another review

macOS/DuckDuckGo/MainWindow/MainViewController.swift
```diff
@@ -628,35 +630,48 @@ extension MainViewController {
...
+            switch (key, flags, flags.contains(.command)) {
+            case ("n", [.command], _),
+                 ("t", [.command], _), ("t", [.command, .shift], _),
+                 ("w", _, true),
+                 ("q", [.command], _),
+                 ("r", [.command], _):
+                NSApp.menu?.performKeyEquivalent(with: event)
+                return true
+
+            case ("tab", [.control], _),
```
mallexxx: @tomasstrba, this handler isn‘t working, it either needs to be `"\t"` or `kVK_Tab`

macOS/DuckDuckGo/MainWindow/MainViewController.swift
```diff
@@ -628,35 +630,48 @@ extension MainViewController {
...
-             kVK_ANSI_N where flags == .command,
-             kVK_ANSI_W where flags.contains(.command),
-             kVK_ANSI_T where [[.command], [.command, .shift]].contains(flags),
-             kVK_ANSI_Q where flags == .command,
-             kVK_ANSI_R where flags == .command:
-            guard view.window?.firstResponder is WebView else { return false }
-            NSApp.menu?.performKeyEquivalent(with: event)
-            return true
+        // Handle tab switching (CMD+1 through CMD+9)
+        if flags == .command, "123456789".contains(key) {
```
mallexxx: another regression is we‘re loosing the numpad keys (or maybe they never actually worked):
the `key` would match but the flags will be `[.command, .numericPad]` - need to add another check like `if [.command, [.command, .numericPad]].contains(flags)`

mallexxx (approved):
The code looks and works good, I‘ve found 2 issues with an obvious fix and approving the PR to make it go faster

macOS/DuckDuckGo/MainWindow/MainViewController.swift
tomasstrba: Come on, GPT! Good catch 👍

macOS/DuckDuckGo/MainWindow/MainViewController.swift
tomasstrba: Perfect! Thank you, @mallexxx 👍

tomasstrba:
Just a quick update, I'll finalise this tomorrow

macOS/DuckDuckGo/MainWindow/MainViewController.swift
tomasstrba: Resolved ✔️

macOS/DuckDuckGo/MainWindow/MainViewController.swift
tomasstrba: `[.command, [.command, .numericPad]].contains(flags)` added ✔️

tomasstrba:
Thank you @mallexxx, I've fixed the TAB and numpads. Let's merge

PR #493, opened 2025-04-14
improvements in subscription with AuthV2 resiliency

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1209974732411306
CC: @miasma13 

### Description

This fixes a bug I noticed in iOS that causes a couple of issues when the device is connected to a wifi with 100% packet loss, for example, a wifi with a captive portal before authentication.

How to reproduce the bugs:
- Start the app
- Enable AuthV2 feature flag
- Kill the app
- Start the app and subscribe/restore the subscription
- Enable 100% packet loss in the iOS network link conditioner
- Kill the app
- Start the app
- Bug #1: The app needs a long time to start (60s, the API request timeout), this was due to a blocking semaphore in the token provider
- Bug #2: remove the subscription from settings: nothing happens for 60s, the logout API call was blocking the operation.

Fixes:
- iOS token provider is now async; the synchronous code has been moved.
- The logout now runs the logout API call in a Task

### Known bugs

Both with AuthV1 and V2, when the app is opened for the first time without an internet connection, the subscription setting is not displayed.

### Testing Steps

1. Test the iOS app's standard operations while enabling 100% packet loss in the iOS network link conditioner
2. Test the macOS app's standard operations while enabling 100% packet loss in the iOS network link conditioner

---
###### Internal references:


quanganhdo (approved):
Behaved as described

macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionModel.swift
```diff
@@ -911,7 +911,14 @@ hasAnyEntitlement: \(hasAnyEntitlement)
...
         if isUserAuthenticated {
             do {
-                let subscription = try await subscriptionManager.getSubscription(cachePolicy: cachePolicy)
+                var subscription = try? await subscriptionManager.getSubscription(cachePolicy: cachePolicy)
+                if subscription == nil {
+                    subscription = try? await subscriptionManager.getSubscription(cachePolicy: .returnCacheDataDontLoad)
+                }
+                guard let subscription else {
+                    throw SubscriptionEndpointServiceError.noData
+                }
```
miasma13: What is the reason for this logic, there was explicit purpose to pass the required `cachePolicy` and use it. If you check how it is used - on initial load it requests cached data, and then follows with attempt to fetch refreshed data. I am not sure about benefit of this embedded logic that overrides the passed `cachePolicy`.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -482,7 +474,9 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
     public func signOut(notifyUI: Bool) async {
         Logger.subscription.log("SignOut: Removing all traces of the subscription and account")
-        try? await oAuthClient.logout()
+        Task { // Not waiting for an answer
+            try? await oAuthClient.logout()
+        }
```
miasma13: I think this is counterintuitive and we may just be begging for problems. When wrapping it with `Task` you make the logout asynchronous and out of sync with the notification being posted. I imagine somewhere in the code we may receive the `.accountDidSignOut` notification that will execute in a brief state when user will still be signed in.

macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionModel.swift
federicocappelli: I had a second pass at the problem, and I agree. The previous logic was working, but it was not nice.
For now I'm reverting this change and then I'll work on a better solution.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: Agree, I made the `oAuthClient.logout()` sync again and moved the async part inside oAuthClient, now only the API logout call is asynchronous, and that still solves the issue.

SharedPackages/BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -368,8 +368,10 @@ final public actor DefaultOAuthClient: @preconcurrency OAuthClient {
         legacyTokenStorage?.token = nil
 
         if let existingToken {
-            Logger.OAuthClient.log("Logging out")
-            try await authService.logout(accessToken: existingToken)
+            Task { // Not waiting for an answer
+                Logger.OAuthClient.log("Invalidating the V2 token")
+                try? await authService.logout(accessToken: existingToken)
+            }
```
federicocappelli, line 374: this is now async and not stopping the account removal in case of network timeouts

miasma13 (approved):
👍

PR #526, opened 2025-04-17
don't reschedule config update soon on 304

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201037661562251/task/1210005595107104?focus=true

### Description
- Allow status 304 (unchanged) in config manager and don't reschedule to update soon

### Testing Steps

1. Validate config is updated
2. Validate when Config Manager receives status `304` it reschedules config update as it‘ve passed normally

### Impact and Risks

High

#### What could go wrong?

Config update breakage

### Notes to Reviewer

---
###### Internal references:


SabrinaTardio:
@mallexxx 
I am not sure if there is a reason for this but could we instead in `ConfigurationFetcher` do this?

```
public func fetch(_ configuration: Configuration, isDebug: Bool = false) async throws {
        let requirements: APIResponseRequirements = isDebug ? .requireNonEmptyData : .allowHTTPNotModified
        let fetchResult = try await fetch(from: configuration.url, withEtag: etag(for: configuration), requirements: requirements)
        if let data = fetchResult.data {
            try validator.validate(data, for: configuration)
        }
        try store(fetchResult, for: configuration)
    }
```

instead of   let requirements: APIResponseRequirements = isDebug ? .requireNonEmptyData : .default?

I am not sure if there is a reason not to cc @jaceklyp

SabrinaTardio:
> I don't think we should hack this in-place, but rather look at it one level deeper. This is used on iOS, macOS, and in VPN extension - I don't think iOS uses this "tryAgainSoon()" logic but macOS and VPN seem to.
> 
> Now, the `ConfigurationFetcher` has two relevant to us public API methods:
> 
> * `fetch(_:isDebug:)`
> * `fetch(all:)`
> 
> The confusing part, is how the former uses (on prod) the `APIResponseRequirements.default` while the latter uses `APIResponseRequirements.all`
> 
> What I believe should be done (and tested) here, is to unify the usage on production to `.all` which should solve the problem listed here for all customers of this API, not only macOS.
> 
> cc: @diegoreymendez @samsymons

Yeah basically this is what I proposed when I reviewed it (https://github.com/duckduckgo/apple-browsers/pull/526#issuecomment-2813378897 technically I said to change to .allowHTTPNotModified), therefore I agree, but I wasn’t / am not sure if there was a specific reason for it to be .default in this case hence why I tagged Jacek (and then remembered he is on leave.)

bwaresiak:
I think the default is leftover from before the `tryAgainSoon` was implemented. Also just `allowHTTPNotModified ` would not be enough, as I'd at least expect `requireETagHeader` to be included (and jury is still out on `requireUserAgent` :) ).

diegoreymendez:
Agreed on going for `.all`.

Silly questions maybe... but:
1. Why is there a `isDebug` parameter? Why is the distinction necessary at all?
2. Is anything other than `.all` necessary here?

SabrinaTardio:
> Agreed on going for `.all`.
> 
> Silly questions maybe... but:
> 
> 1. Why is there a `isDebug` parameter? Why is the distinction necessary at all?
> 2. Is anything other than `.all` necessary?

I don’t know the context in which the “isDebug” added was added, but, I think having a different behaviour it’s useful for testing, for example trigger the update every time (even if the config was not changed) when reloading the configuration could be useful when making changes to that flow or checking other behaviours (of course we would want to allow this only on the debug build).

PR #568, opened 2025-04-23
macOS / iOS - Change auto-submit behavior on duck.ai

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204167627774280/task/1210025284896479?focus=true

### Description
When the user is on the SERP, we should not auto-submit the queries to duck.ai, instead it should only pre-fill the textbox.

### Testing Steps
Same steps for both iOS and macOS

1. Do a search, while you're on the SERP, click/tap the duck.ai button in the native app. It should open duck.ai with the SERP query pre-filled in the text box but not auto-submitted 
2. Start typing something in the address bar, click/tap the duck.ai button. It should open duck.ai with the query from the address bar automatically submitted.

### Impact and Risks
Low: Improvement to existing features

#### What could go wrong?
Not auto-submitting when the user is typing the query in the address bar

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
@@ -60,7 +60,7 @@ struct AIChatTabOpener: AIChatTabOpening {
     @MainActor
     func openAIChatTab(_ query: String?, target: AIChatTabOpenerTarget) {
         if let query = query {
-            promptHandler.setData(query)
+            promptHandler.setData(.queryPrompt(query, autoSubmit: false))
         }
         WindowControllersManager.shared.openAIChat(aiChatRemoteSettings.aiChatURL, target: target, hasPrompt: query != nil)
     }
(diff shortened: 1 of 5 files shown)


macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
```diff
@@ -54,13 +54,27 @@ struct AIChatTabOpener: AIChatTabOpening {
     @MainActor
     func openAIChatTab(_ value: AddressBarTextField.Value, target: AIChatTabOpenerTarget) {
         let query = addressBarQueryExtractor.queryForValue(value)
-        openAIChatTab(query, target: target)
+
+        // We don't want to auto-submit if the user is opening duck.ai from the SERP
+        // https://app.asana.com/1/137249556945/project/1204167627774280/task/1210024262385459?focus=true
+        let shouldAutoSubmit: Bool
```
jotaemepereira, line 60: @Bunn It would be nice to encapsulate this logic somwhere where it can be unit tested.

jotaemepereira (approved):
Looks good. I’ve added one comment about trying to encapsulate the logic to unit test, or maybe som UI tests can help here too.

macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
Bunn: Agree. I'll create an AOR task for this.

PR #573, opened 2025-04-24
Add Shortcuts to macOS navigation bar icons tooltips

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204167627774280/task/1209902113488172?focus=true

### Description
Add Shortcuts to macOS navigation bar icons tooltips

### Testing Steps
1. Check the navigation bar icons: Back/Forward/Reload/Downloads. They should show a shortcut copy as a suffix in their tooltip

### Impact and Risks
Low: Minor visual changes


jotaemepereira (approved):
It looks good 👍🏼 .

My only concern is that we are not linking these shortcuts with the real shortcut in the Main Menu, which could cause a discrepancy when we change one of them (people we will need to change this new enum as well).

Can you take a look to see how easily would be to connect the Main Menu shortcuts with these ones?

Bunn:
> It looks good 👍🏼 .
> 
> My only concern is that we are not linking these shortcuts with the real shortcut in the Main Menu, which could cause a discrepancy when we change one of them (people we will need to change this new enum as well).
> 
> Can you take a look to see how easily would be to connect the Main Menu shortcuts with these ones?

That’s a good point. It’s a problem we currently have, for example on the menu bar vs context menu.
I took a quick look and the changes would be bigger than I’d want to do it now, this was supposed to be a simple copy change and it’s already getting feature creeped on the SR.

I’ll file another task to unify these properly and take a look in QWD or something. Good comment, thanks!

PR #576, opened 2025-04-24
Authv2 migration pixels improved

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1210072556716592?focus=true
CC: @miasma13 @diegoreymendez 

### Description

Small improvements in the auth v1 migration, and when we send the .migrationStarted pixel
Now we preemptively check if the user is already logged in and in case we stop the migration before sending the pixel

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -27,6 +27,7 @@ public enum OAuthClientError: Error, LocalizedError, Equatable {
     /// When both access token and refresh token are expired
     case refreshTokenExpired
     case invalidTokenRequest
+    case authMigrationNotPerformed
```
miasma13, line 30: I understand the reason for adding this but when reading it this does not sound like a real error, just a state or result that migration was skipped as it was not necessary.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -239,17 +239,20 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
     func migrateAuthV1toAuthV2IfNeeded() async {
 
-        guard v1MigrationNeeded else {
+        guard v1MigrationNeeded, // stops multiple attempts in a session, even in case of unrecoverable failures
+              !isUserAuthenticated else { // warning: using `isUserAuthenticated()` would create a loop
```
miasma13: This line is counter intuitive, since following the logic one would expect to actually do the migration when the user is authenticated. I get the meaning and suggest replacing it with an explicit check if `oAuthClient.currentTokenContainer` is not present

miasma13 (approved):
Two small comments, other than that looks good.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: This makes sense, changed

PR #577, opened 2025-04-24
[macOS] Min window size: Adjust suggestions, Bookmarks, and Settings minimum window size

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1209993929328796?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1210044556037376?focus=true

### Description

This adjusts three parts of the app to handle the smaller minimum window size:

* Suggestions window (when entering text in the address bar)
* Bookmarks manager
* Settings

#### Suggestions window

The “Switch to tab” box is now only shown in the suggestions window if there is space for it (if the suggestions window is greater than a minimum width).

Related to this, the “Switch to tab” behavior in the address bar was updated to match (basing it on the size of the address bar rather than the overall window size).

#### Bookmarks manager

The Bookmarks manager can now be resized to the minimum window size:

* The sidebar can be compressed to a minimum 128 pt.
* When needed, the sidebar contents are truncated to fit and the bookmark count is hidden.
* The detail view (on the right) is updated to match the design spec:
   * The padding around the content is decreased.
   * The minimum width for the search bar was increased.
   * The buttons (new bookmark, new folder, delete, sort) are now compressed one at a time as the window is resized, so some labels don’t need to be truncated (instead of all labels being somewhat truncated).

#### Settings

The Settings (all sections) can now be resized to the minimum window size:

* The sidebar can be compressed to a minimum 128 pt.
* When needed, the sidebar contents are truncated to fit.
* The General section is updated to adjust to the smallest window size (allowing the contents to be resized to fit).

Note: I tested all of the buttons and links in each Settings section. The only view I found that didn’t stay within the minimum window size is Passwords & Autofill > Password Manager > Export Passwords. This displays a wider view (because the text does not wrap) but when it is closed, the window remains at the minimum size.

### Testing Steps

1. Resize the window to the smallest window size.
2. Enter the address of a website open in another tab, and confirm the “Switch to tab” box does not appear in the suggestions window or in the address bar.
3. Open the Bookmarks manager (e.g. go to Bookmarks > Manage Bookmarks).
4. Confirm the window size didn't change, and the sidebar and detail view appear as expected. (Check that nested bookmark folders don’t prevent the window from being the smallest size, even if they aren’t all visible.)
6. Resize the window and confirm the sidebar and then the detail view expands as expected.
7. Resize the window again to the smallest window size.
8. Open Settings (using the menu at the top of the Bookmarks sidebar or e.g. go to DuckDuckGo > Settings in the menu).
9. Confirm the window size didn’t change, and the sidebar and detail view appear as expected.
10. Resize the window and confirm the sidebar and then the settings section expands as expected.
11. Open each settings section and confirm it is displayed as expected, and the settings it contains are visible and can be interacted with without changing the window size.

Confirm it works as expected in the following scenarios:

- Testing across all supported macOS versions.
- Testing with different interface languages (English and at least one verbose, non-English language).
   - Instead of changing the system language, you can edit the product scheme and go to Options > App Language > select Double-length Pseudolanguage. Then, build and run the app as usual to check the UI with very long strings.

### Impact and Risks

Low: Minor visual changes

#### What could go wrong?

Layout in the updated sections could be unexpected, with text unexpectedly truncated/unreadable when the window is resized.

### Notes to Reviewer

Before|After
-|-
Screenshot 2025-04-25 at 10 33 07|Screenshot 2025-04-24 at 17 47 01
Screenshot 2025-04-24 at 17 50 05|Screenshot 2025-04-24 at 17 47 30
Screenshot 2025-04-24 at 17 50 09|Screenshot 2025-04-24 at 17 47 35
Screenshot 2025-04-24 at 17 50 19|Screenshot 2025-04-24 at 18 05 02

—
###### Internal references:


tomasstrba (commented):
This is way better than before 💯 Nice work 👏  Also, your PR description game is really strong 🥇 

I only found three minor issues:
1) "Switch to tab" label in the suggestion row is present if you restore a window and minimise it after. It's not present after in next attempts.
https://github.com/user-attachments/assets/c5ba5726-eb59-485e-ba3b-f709b26cf161

2) The options button for bookmarks folder is present on top or left to the bookmark folder icon in the compressed state
https://github.com/user-attachments/assets/a4032f7c-5cbd-40f1-b83d-fb00d80e17f1

3) Not a blocker:
There is a resizing indicator in the bookmarks view, which in our case isn't correct.
https://github.com/user-attachments/assets/bec11635-d1ab-44dc-869f-0361adf465c9

rachelmcr:
Thanks for catching those issues! This is ready for another look.

> "Switch to tab" label in the suggestion row is present if you restore a window and minimise it after. It's not present in next attempts.

Nice catch! I have a fix for this in 327c296. The check for the frame size was happening too soon (and using the frame’s intrinsic width, which is very large) so it was always showing the “Switch to tab” label in a restored window. I updated it so that any time the view is resized it checks again if the label should be shown:

https://github.com/user-attachments/assets/43652a93-94ca-4c8e-8a9b-ec09653aa96c

> The options button for bookmarks folder is present on top or left to the bookmark folder icon in the compressed state

This one is a little tricky. The same behavior is present in production, but it requires a lot of nested folders to notice it. The fix in 2c21471 feels a little hacky to me, but it does prevent this overlap with the bookmark folder icon. It doesn’t resolve overlap between the menu button and indentation marker (disclosure icon), which is laid out automatically, and it doesn’t resolve other odd layout issues for very nested folders. Those feel out of scope for this project — they are pre-existing layout issues that are more discoverable with the smaller width, but would require more in-depth investigation and layout changes to fully resolve.

https://github.com/user-attachments/assets/b15d356a-61df-4b88-ac62-8a56b8f3c64c

> There is a resizing indicator in the bookmarks view, which in our case isn't correct.

Fixed in 07c9747.

tomasstrba:
@rachelmcr, unfortunately, now I see the “Switch to tab” label in all cases, even when the window is minimised. Can you also reproduce it?
Screenshot 2025-04-30 at 07 26 07

Otherwise, everything else looks good 👍

tomasstrba (approved):
LGMT! 🏅 Everything works as expected ✅  

For future reference, Rachel and I discussed the issue on MM and it turned out I didn't have optional navigation bar buttons enabled.

Screenshot 2025-04-30 at 10 57 00

PR #594, opened 2025-04-25
Return early from navigationDidFail when it contains webkit termination error

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210080097969176?focus=true
CC: @mallexxx 

### Description
Turns out that on slower machines the webkit process termination may call navigationDidFail
callback before it calls webContentProcessDidTerminate. The former is then called with
the WebKit Termination error, and causes the termination error to be handled and webview state
altered to handle navigation error, which in fact isn’t a navigation error. That broke handling
webkit termination by the crash recovery extension.
This change updates navigationDidFail handler to return early in case of webkit termination error,
so that the termination handler can do its job.

### Testing Steps
I’ve tested it on:
* Big Sur, 
* Ventura
* Sequoia.

It was also an issue for Tak, who tested the build on his Ventura machine and found the issue to be fixed.
I consulted the change with Alex as well.

### Impact and Risks

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/Tab/TabExtensions/TabCrashRecoveryExtension.swift
+++ b/macOS/DuckDuckGo/Tab/TabExtensions/TabCrashRecoveryExtension.swift
@@ -130,16 +130,16 @@ extension TabCrashRecoveryExtension: NavigationResponder {
             NSUnderlyingErrorKey: NSError(domain: WKErrorDomain, code: terminationReason)
         ])
 
-        Task {
+        attemptTabCrashRecovery(for: error, in: webView)
+
+        Task.detached(priority: .utility) {
 #if APPSTORE
             let additionalParameters = [String: String]()
 #else
             let additionalParameters = await SystemInfo.pixelParameters()
 #endif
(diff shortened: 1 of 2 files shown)


macOS/DuckDuckGo/Tab/TabExtensions/TabCrashRecoveryExtension.swift
```diff
@@ -130,16 +130,16 @@ extension TabCrashRecoveryExtension: NavigationResponder {
             NSUnderlyingErrorKey: NSError(domain: WKErrorDomain, code: terminationReason)
         ])
 
-        Task {
+        attemptTabCrashRecovery(for: error, in: webView)
+
+        Task.detached(priority: .utility) {
```
ayoy, line 135: Here (and below) I’m just optimizing pixel firing so that it doesn’t happen on main thread (in the scenario below, especially).

Bunn (approved):
I’ve tested [this](https://app.asana.com/1/137249556945/task/1210080097969176/comment/1210082796197464?focus=true) build on Big Sur, and ran it from Xcode on Sequoia. LGTM

PR #605, opened 2025-04-28
Restore basic WebExtension support to our App.

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206580121312550/task/1210045236108793?focus=true

### Description

Make it possible to load and unload web extensions on the fly.

### Testing Steps

1. Download [these extensions](https://dub.duckduckgo.com/duckduckgo/web-extensions-test-suite/tree/main/emoji-substitution).
2. Expand the zip file.
3. Launch the app in debug mode, as an internal user (make sure you're on macOS 15.4).
4. Open [this page](https://en.wikipedia.org/wiki/Main_Page) for testing.
5. Go to Debug > Web Extensions > Install web extension...
6. Install the `emoji-substitution` extension you downloaded
7. Loading should happen on the fly and the page should show emoji substitution.
8. Launching the app when an extension has been loaded should result in the extension being loaded right away.

### Impact and Risks

Low.

#### What could go wrong?

Nothing I can think of.

### Quality Considerations

This is an internal feature.

### Notes to Reviewer

NA

---
###### Internal references:


macOS/Configuration/App/DuckDuckGo.xcconfig
```diff
@@ -39,9 +39,9 @@ GCC_PREPROCESSOR_DEFINITIONS[config=Debug][arch=*][sdk=*] = DEBUG=1 SWIFT_OBJC_I
 GCC_PREPROCESSOR_DEFINITIONS[config=Review][arch=*][sdk=*] = REVIEW=1 SWIFT_OBJC_INTERFACE_HEADER_NAME=$(SWIFT_OBJC_INTERFACE_HEADER_NAME) $(inherited)
 
 SWIFT_ACTIVE_COMPILATION_CONDITIONS[arch=*][sdk=*] = $(FEATURE_FLAGS)
-SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=CI][arch=*][sdk=*] = DEBUG CI $(FEATURE_FLAGS)
-SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=Debug][arch=*][sdk=*] = DEBUG $(FEATURE_FLAGS)
-SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=Review][arch=*][sdk=*] = REVIEW $(FEATURE_FLAGS)
+SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=CI][arch=*][sdk=*] = DEBUG CI $(FEATURE_FLAGS) WEB_EXTENSIONS_ENABLED
+SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=Debug][arch=*][sdk=*] = DEBUG $(FEATURE_FLAGS) WEB_EXTENSIONS_ENABLED
+SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=Review][arch=*][sdk=*] = REVIEW $(FEATURE_FLAGS) WEB_EXTENSIONS_ENABLED
```
diegoreymendez: No need to have this in release builds yet.  Avoiding it helps minimize the risk that these changes might break our public builds.

macOS/DuckDuckGo/WebExtensions/View/WebExtensionsDebugMenu.swift
```diff
@@ -68,9 +68,9 @@ final class WebExtensionsDebugMenu: NSMenu {
         guard case .OK = panel.runModal(),
               let url = panel.url else { return }
 
-        webExtensionManager.addExtension(path: url.absoluteString)
-
-        NSAlert.extensionAlert().runModal()
+        Task {
+            await webExtensionManager.addExtension(path: url.absoluteString)
+        }
```
diegoreymendez, line 73: We no longer need the alert, and the method is now asynchronous.

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
```diff
@@ -18,39 +18,91 @@
...
+
 @available(macOS 15.4, *)
 protocol WebExtensionLoading: AnyObject {
-
-    func loadWebExtensions(from paths: [String]) async -> [WKWebExtension]
-
+    @discardableResult
+    func loadWebExtension(path: String, into controller: WKWebExtensionController) async throws -> WKWebExtensionContext
+    func loadWebExtensions(from paths: [String], into controller: WKWebExtensionController) async -> [Result<WKWebExtensionContext, Error>]
+    func unloadExtension(at path: String, from controller: WKWebExtensionController) throws
```
diegoreymendez, line 28: Several changes here... the loader now takes care of the entirety of loading and unloading onto and from the controller.  This allowed me to also support unloading extensions and simplify the code.

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
```diff
@@ -18,39 +18,91 @@
 
 #if WEB_EXTENSIONS_ENABLED
 
+import CryptoKit
```
diegoreymendez, line 21: Need this to calculate the SHA256 identifier.  We can remove this once we no longer need it.

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
```diff
@@ -18,39 +18,91 @@
...
 @available(macOS 15.4, *)
 final class WebExtensionLoader: WebExtensionLoading {
 
-    private func loadWebExtension(path: String) async -> WKWebExtension? {
+    enum WebExtensionLoaderError: Error {
+        case failedToCreateURLFromPath(path: String)
+        case failedToFindContextForPath(path: String)
+    }
+
+    func loadWebExtension(path: String, into controller: WKWebExtensionController) async throws -> WKWebExtensionContext {
```
diegoreymendez, line 39: We have to think of the context as a loaded extension, which is why it makes sense for the loader to return this instead of just the extension (which isn't very useful on its own).

The extensions can still be referenced from the contexts, and from the controller if necessary.

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
```diff
@@ -18,39 +18,91 @@
...
+        let context = controller.extensionContexts.first {
+            $0.uniqueIdentifier == identifierHash(forPath: path)
+        }
+
+        guard let context else {
+            throw WebExtensionLoaderError.failedToFindContextForPath(path: path)
+        }
+
+        try controller.unload(context)
+    }
```
diegoreymendez, line 75: The loader now also supports unloading extensions.

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
```diff
@@ -18,39 +18,91 @@
...
+
+    private func identifierHash(forPath path: String) -> String {
+        let identifier = Data(path.utf8)
+        let hash = SHA256.hash(data: identifier)
+        let hashString = hash.compactMap { String(format: "%02x", $0) }.joined()
+
+        return hashString
+    }
```
diegoreymendez, line 83: We want each extension to have a separate identifier we can calculate from something known - in this case the path.  This allows us to unload extensions.

macOS/DuckDuckGo/WebExtensions/WebExtensionManager.swift
```diff
@@ -34,14 +34,14 @@ protocol WebExtensionManaging {
...
-    func addExtension(path: String)
+    func addExtension(path: String) async
     func removeExtension(path: String)
 
     // Provides the extension name for the extension resource base path
     func extensionName(from path: String) -> String?
 
     // Controller for tabs
-    var controller: WKWebExtensionController? { get }
+    var controller: WKWebExtensionController { get }
```
diegoreymendez, line 51: There's no need for the controller to be optional since it can be created at `init` time.

macOS/DuckDuckGo/WebExtensions/WebExtensionManager.swift
```diff
@@ -80,14 +83,13 @@ final class WebExtensionManager: NSObject, WebExtensionManaging {
     // Loads web extensions after selection or application start
     var loader: WebExtensionLoading
 
-    // Loaded extensions
-    var extensions: [WKWebExtension] = []
```
diegoreymendez, line 84: We don't need this really.  If we ever need to access the extensions we can access them through `contexts` below, which takes the contexts from the controller, which becomes our single source of truth for loaded extensions.

macOS/DuckDuckGo/WebExtensions/WebExtensionManager.swift
```diff
@@ -103,12 +105,26 @@ final class WebExtensionManager: NSObject, WebExtensionManaging {
...
 
-    func addExtension(path: String) {
+    func addExtension(path: String) async {
         pathsCache.add(path)
+
+        do {
+            try await loader.loadWebExtension(path: path, into: controller)
+        } catch {
+            // This is temporary.  The actual handling of this error should be done outside of this manager.
+            assertionFailure("Failed to unload web extension \(path): \(error)")
```
diegoreymendez, line 140: I'm drawing a line here - we should really just throw and let the parent handle the error, but I need to limit the scope of my changes at some point.

.xcode-version
```diff
@@ -1 +1 @@
-16.2
+16.3
```
diegoreymendez: 🔴🔴🔴 ROLL BACK BEFORE MERGING!

macOS/Configuration/App/DuckDuckGo.xcconfig
diegoreymendez: 🔴🔴🔴 ROLL BACK BEFORE MERGING!

tomasstrba: The webextensions feature is hidden behind the feature flag and open just for internal users. The point of enabling is to explore the API and actually uncover any problems during internal testing

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
tomasstrba: Sounds good 👍 I like it

macOS/DuckDuckGo/WebExtensions/WebExtensionLoader.swift
tomasstrba: Awesome! I tested it and it works great!

tomasstrba (approved):
@diegoreymendez, awesome work! I love the dynamic loading and unloading 🏅 

I've noticed two minor issues, but they aren't blockers so I'm going to approve the PR. If you have some space, it would be great if we can fix them before sending the build.

T1/ Web extensions aren't applied to restored tabs.
Enable tab restoration, open a couple of Wiki sites and relaunch the app. Load an extension, and you can see it wasn't applied to existing tabs even after reload (probably no link to the web extension controller in their web view configuration?)

T2/ I can no longer load Safari web extensions. I remember having same problem before during the implementation and it has something to do with the format of path
Screenshot 2025-04-30 at 13 13 37

macOS/Configuration/App/DuckDuckGo.xcconfig
diegoreymendez: Just adding a note that we discussed this might be tricky to enable until Xcode 16.3 is enabled for our CI, so for now I'll disable before merging and we'll revisit later.

diegoreymendez:
Hey thanks for the review @tomasstrba.  Both T1 and T2 work well for me in debug builds.

The review build is broken though so I'll now look into that.

PR #617, opened 2025-04-29
pixels for feature discovery

brindy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/392891325557410/task/1209194926333594?focus=true
Tech Design URL:
CC:

### Description
Adds a new pixel and parameters to existing pixels in order to measure discovery of features, according to this list:
https://app.asana.com/1/137249556945/project/392891325557410/task/1210059260849216?

### Testing Steps

General
1. Test that pixels are sent according to the list above and that the parameter is sent as indicated.  Then do whatever is needed to do to toggle the parameter and test the pixel is sent with the updated value.

Note: You only need to test the ones that are new or modified - marked in bold.

VPN
1. Enable a subscription on your device, see [this doc](https://app.asana.com/1/137249556945/project/1202500774821704/task/1207149361119036).
4. Visit the VPN settings and check the pixel is sent with the parameter of 0
5. Ideally on a real device, enable the VPN and visit the settings again and check parameter is 1
6. But if using a simulator drop in this code somewhere (e.g. `MainVC.viewDidAppear`) and launch the app.  Wait 5 seconds, then visit the VPN page in settings and check parameter is 1

        DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
            class MockSession: NETunnelProviderSession {
                override var status: NEVPNStatus {
                    return .connected
                }
            }
            let session = MockSession()
            let notification = Notification(name: .NEVPNStatusDidChange, object: session)
            NotificationCenter.default.post(notification)
        }

Duck Player Legacy
1. Visit the duck player settings, pixel should be set parameter set to 0
3. Use duck player using the legacy version
4. Visit the duck player settings, pixel should be set parameter set to 1

Duck Player Native
1. Visit the duck player settings, pixel should be set parameter set to 0
3. Switch to native version (with internal user setting on should be able to switch it on in settings)
4. Use duck player
3. Visit the duck player settings, pixel should be set parameter set to 1

### Impact and Risks
Low: Adding some additional telemetry


iOS/DuckDuckGo/AutofillLoginListViewController.swift
```diff
@@ -217,7 +217,12 @@ final class AutofillLoginListViewController: UIViewController {
                 }
             }
 
-        Pixel.fire(pixel: .autofillManagementOpened, withAdditionalParameters: ["source": source.rawValue])
+        let hasCredentials = ((try? secureVault?.accountsCount()) ?? 0) > 0
```
dus7: ⚠️ As this is inside `init` I'm wondering if this can cause hangs if there are many accounts saved - maybe we can offload this to another queue? It should be safe to do it for GRDB reads.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -1785,6 +1798,14 @@ class MainViewController: UIViewController {
             }
     }
 
+    private func onVPNStatusDidChange(_ notification: Notification) {
+        guard let session = (notification.object as? NETunnelProviderSession),
+           session.status == .connected else {
+            return
+        }
+        self.featureDiscovery.setWasUsedBefore(.vpn)
```
dus7, line 1806: We could unsubscribe from notification once this has been set, but it's not critical.

iOS/DuckDuckGo/AutofillLoginListViewController.swift
```diff
@@ -217,7 +217,12 @@ final class AutofillLoginListViewController: UIViewController {
...
+        Pixel.fire(pixel: .autofillManagementOpened,
```
dus7: Also I'm not necessarily sure the "open" event should be sent on init. I would say `viewDidAppear` is better, but I understand this was here before, so I won't be fighting it much.

dus7 (approved):

iOS/DuckDuckGo/AutofillLoginListViewController.swift
brindy: Yeah, I don't want to make assumptions about why this pixel was fired here or not tbh.

iOS/DuckDuckGo/MainViewController.swift
brindy: I think I'm going to leave it as it's quite simple just now and won't matter on future launches.

PR #627, opened 2025-04-30
Fix SwiftUI element colors used by agents

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207603085593419/task/1210109560256540?focus=true
Tech Design URL:
CC:

### Description

This PR fixes an issue where shared SwiftUI elements used by agents weren't picking up their button colors correctly.

The core issue here was that the colors being referenced by the SwiftUI extensions package existed only in the main app asset catalog - so when the elements were used there, they were fine, but when they were used in an agent then they couldn't be found.

This PR fixes it by moving those colors into the Swift package's asset catalog instead.

### Testing Steps

1. Get your subscription into an expired state - if you want to hardcode this for the VPN, look up `model.shouldShowSubscriptionExpired` and replace any of its uses with `true`
2. Check that the subscription expired button looks good in light and dark mode in both app and menu bar agent
3. Check that there are no other uses of the colors being moved here, so that we aren't accidentally breaking any other SwiftUI elements

### Impact and Risks

Low: Minor visual changes

#### What could go wrong?

This could cause an issue if the color being moved out of the main asset catalog is referenced anywhere else, so that is included in the testing steps.

### Notes to Reviewer

---
###### Internal references:


diegoreymendez (approved):
Looks fantastic in both dark and light modes.

PR #636, opened 2025-04-30
UIComponents skeleton package

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204167627774280/task/1209993325158903?focus=true

### Description
Add UIComponents skeleton package

### Testing Steps
1. on iOS, on any SwiftUI View, import UIComponents and add `SampleiOSView`, `SamplemacOSView` and `SampleSharedView`.
2. The compilation should fail because of `SamplemacOSView`. Remove it and try again
3. Do the same for macOS, but this time you’ll need to remove `SampleiOSView`
4. Check README, see if it makes sense 😅 

### Impact and Risks
None: Internal tooling

#### What could go wrong?
Package is not well configured for specific iOS or macOS project


SharedPackages/UIComponents/Package.resolved
```diff
@@ -0,0 +1,15 @@
+{
+  "originHash" : "9a9153fd0315e8d8297f1d738e08612405ff9980556d89b2e9841c568e9a4c38",
+  "pins" : [
+    {
+      "identity" : "designresourceskit",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/duckduckgo/DesignResourcesKit",
```
samsymons, line 7: Unrelated to this PR, but I'm increasingly convinced that we should move DesignResourcesKit into the monorepo as well, since quite a few packages depend on it and I could see it being a pain to update as a result. I'll get a discussion task up about that.

SharedPackages/UIComponents/Package.swift
```diff
@@ -0,0 +1,52 @@
...
+        .package(url: "https://github.com/duckduckgo/DesignResourcesKit", exact: "4.2.0")
+    ],
+    targets: [
+        .target(
+            name: "UIComponents",
+            dependencies: [
+                "DesignResourcesKit",
+            ],
+            resources: [
+                .process("Resources/Assets.xcassets")
```
samsymons, line 44: The new package technically doesn't have an Asset Catalog, so this shows as a (harmless) warning in Xcode. We can either remove this, or add an empty asset catalog - either way is fine by me. :)

samsymons (approved):
LGTM, thanks!

SharedPackages/UIComponents/Package.swift
Bunn: Interesting. I created the asset catalog, but when it’s empty, it's of a bunch of empty folders that Xcode doesn’t like. I added a SampleColor just so it doesn’t complain. I’ll let the team know that this can be removed, it’s only there to facilitate having an asset catalog ready to go

PR #646, opened 2025-05-02
PIR iOS initial implementation: initial module and iOS manager

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210144315064560?focus=true
Tech Design URL:
CC:

### Description
Adds the initial iOS module and manager class. 
The manager is created in very much the same way as the macOS one.
At the moment it's inert, but if you save something to the DB and call iOSManager.start, it will run jobs. Opt outs will fail though due to some shared setup still being missing. All academic though since it's not being used in this PR by the iOS app at all

Also frustratingly renames the Resources folder due to a build bug (see comment for details)

Also moves some configuration pixels to the macOS specific package since they're only used there, and the dependency was causing problems on iOS (see https://app.asana.com/0/0/1210184938506242)

### Testing Steps

1. Smoke test the iOS app to make sure nothing is broken
2. Also smoke test PIR on macOS

### Notes to Reviewer

---
###### Internal references:


iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/Bundle/DataBrokerProtectionBundleExtension.swift
```diff
@@ -0,0 +1,23 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public extension UserDefaults {
+    static let dbp = UserDefaults(suiteName: "dbp")!
```
THISISDINOSAUR, line 23: This is possibly worth a discussion. I vaguely remember we were trying to avoid the "com.duck...etc" format?

SharedPackages/DataBrokerProtectionCore/Package.swift
```diff
@@ -44,7 +44,7 @@ let package = Package(
                 .product(name: "Configuration", package: "BrowserServicesKit"),
                 .product(name: "Persistence", package: "BrowserServicesKit"),
             ],
-            resources: [.copy("Resources")],
+            resources: [.copy("BundleResources")],
```
THISISDINOSAUR, line 46: A weird build bug makes this necessary. The alternative was using .process instead, but we decided to stick to copy and rename it since the reason it was .copy in the first place was in the past we've relied on the file structure of the resources folder to identify fake brokers

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/Bundle/DataBrokerProtectionBundleExtension.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  DataBrokerProtectionBundleExtension.swift
```
THISISDINOSAUR, line 2: Thanks to these header linting errors, I noticed our format between iOS and macOS is slightly different. Could be a good quick win to fix at some point (obviously we'd have to automate it)

SharedPackages/DataBrokerProtectionCore/Package.swift
samsymons: Yeah shame this is needed, but I think this approach seems the best of the bunch for now.

iOS/LocalPackages/DataBrokerProtection-iOS/Tests/.swiftlint.yml
samsymons, line 1: This has reminded me that I really need to do something about our SwiftLint set up - this approach is correct for now but I'm hoping to consolidate things such that we have a consistent set of rules later.

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/DataBrokerProtectionIOSManager.swift
```diff
@@ -0,0 +1,180 @@
...
+        super.init { event, _, _, _ in
+            switch event {
+            default:
+                print("event happened")
+            }
+        }
+    }
+
+    override init(mapping: @escaping EventMapping<OperationEvent>.Mapping) {
+        fatalError("Use init()")
```
samsymons, line 45: We might also be able to mark this with `@available(*, unavailable)` to prevent it from being used at compile time (hope I'm remembering that correctly!)

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/Bundle/DataBrokerProtectionBundleExtension.swift
samsymons: Yeah I recall this too, can't remember where we got to though - definitely need some kind of expectation around this. Tbh though this works for now, as long as it's easy to tell from the name what it's used for then I'm happy.

samsymons (approved):
Looks great!

PR #647, opened 2025-05-02
White flash while loading websites in dark mode fixed

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1209499005452053?focus=true

### Description
https://github.com/duckduckgo/macos-browser/pull/2116
We are reintroducing delay to webview presentation to fix white background flashing.

Related PRs:
https://github.com/duckduckgo/macos-browser/pull/596
https://github.com/duckduckgo/macos-browser/pull/2116

### Testing Steps

1. Make sure you are internal user in the DEBUG build
2. Execute testing scenarios from related PRs and make sure the white flash is eliminated, and users can use "Sign in with apple" to log in to https://www.ancestry.com/

### Notes to Reviewer

---
###### Internal references:


macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
```diff
@@ -598,26 +598,43 @@ final class BrowserTabViewController: NSViewController {
...
-                        .filter{ $0 >= .started }
-                        .asVoid()
-                )
-                // take the first such event and move forward.
-                .prefix(1)
-                .eraseToAnyPublisher()
+                // If the current content is the native internal site, delay the webview presentation
+                // until a website renders (or edge cases) to avoid white flash
+                // Enabled for internal users for test
+                if [URL.newtab, URL.settings, URL.bookmarks].contains(self?.lastURL) &&
```
tomasstrba, line 609: I was thinking of using the delayed logic only in case the current site is native internal site. Any suggestion or objections to this approach?

ayoy: Yes, this makes perfect sense, however we should be guarded here already by the `tabContent.isUrl` check on line 603. That filters out settings, bookmarks, new tab and a bunch of other internal websites so the check doesn’t need to be repeated here.

macOS/DuckDuckGo/Tab/Model/Tab.swift
```diff
@@ -1355,6 +1366,14 @@ extension Tab/*: NavigationResponder*/ { // to be moved to Tab+Navigation.swift
             webView.setDocumentHtml(html)
         }
     }
+
+    func renderingProgressDidChangeRaw(progressEvents: UInt) {
+        // Emit only after first paint event
```
ayoy: Could you explain the `4` in a comment? My guess, without checking WebKit sources, is that it’s the “first paint” event enum value, but that’s just an assumption that would be great to have explained here.

macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
```diff
@@ -598,26 +598,43 @@ final class BrowserTabViewController: NSViewController {
...
-                    tabViewModel.tab.navigationStatePublisher.compactMap { $0 }
-                        .filter{ $0 >= .started }
-                        .asVoid()
-                )
-                // take the first such event and move forward.
-                .prefix(1)
-                .eraseToAnyPublisher()
+                // If the current content is the native internal site, delay the webview presentation
+                // until a website renders (or edge cases) to avoid white flash
+                // Enabled for internal users for test
```
ayoy: Nitpick about documentation – this line may get outdated quickly and as such, we could remove it.

I think that ultimately we would be interested in rolling out this change behind a Privacy Config flag to be able to disable it remotely if needed. So the flag will stay here long-term, and we’d only change its source in FeatureFlagDescribing protocol.

ayoy (approved):
Great to see this being brought back! 💪 I did a bunch of testing and couldn’t break it with Sign In with Apple, regular websites, non-existent websites and basic-auth protected websites. Thanks a lot @tomasstrba!

Leaving a few comments, mostly related to documentation. I’d also suggest that we connect the new feature flag to Privacy Config already, to avoid having to do another release to make it available to users.

macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
tomasstrba: @ayoy, the `tabContent` and the `lastURL` are two different variables. At the time of the navigation start, the `lastURL` contains the currently visible URL, but `tabContent` contains the content which is about to be presented. This state is present until the navigation finishes and it can be used as an advantage to limit our new algorithm and use more reliable version of publishers for majority of cases. If I understand correctly, we only want to solve the white flash when switching from native sites to a web content.

macOS/DuckDuckGo/Tab/Model/Tab.swift
tomasstrba: Sure, good point 👍

macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
tomasstrba: Agree 👍

tomasstrba:
Thank you, @ayoy! Totally agree on using the remote feature flag — that was my plan as well, but I held off until we confirmed everything works as expected. I’ll go ahead and add the flag, clean up the comments, and get this merged. Appreciate the fast review! 👍

macOS/DuckDuckGo/Tab/View/BrowserTabViewController.swift
ayoy: Ah of course :) Thanks for clarifying that, this of course should stay like that, great catch 💪

PR #671, opened 2025-05-06
PIR iOS initial implementation: background processing

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210171783578081?focus=true
Tech Design URL:
CC:

### Description
Adds rudimentary background processing for PIR
This includes actually setting up the DBPiOSManager, so now the database viewer will work hurray (Although there's still now way to get anything into the database 😅. But for now for debug and alpha builds it hardcodes a profile into the DB )!
For now it only does so for debug and alpha builds since we're a long way from wanting any of this in production

### Testing Steps

1. For now we don't actually want PIR to do anything, so make sure the general app works, and particularly keep an eye on cpu usage etc to check it isn't affected

### Notes to Reviewer

---
###### Internal references:


samsymons (approved):
From the testing instructions, things here seem good - nothing in the code stands out otherwise either.

PR #672, opened 2025-05-06
Prevent multiple sync connection requests simultaneously

graeme:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202926619870900/task/1210171634625852?focus=true

### Description

We discovered that a recent [Rise in macOS login errors after exchange flow rollout](https://app.asana.com/1/137249556945/task/1210047340594631) could well be due to pasting an exchange code multiple times on macOS. Windows doesn't have this issue because they present a "Preparing to Sync" in this case, blocking any further interaction.

We can present the same dialog on macOS too, however this would require a cancel button which is a change that would require a Ship Review (see [Update macOS Preparing to Sync dialog to be cancellable](https://app.asana.com/1/137249556945/task/1210170146955358)).

Instead, let's just block further requests if one is in flight so we don't spam the backend and cause errors.

### Testing Steps
1. Make sure you have sync enabled on one device (let's say iOS for the purposes of this test, can be any though)
2. Build the macOS app from this branch and go to the sync screen (Settings -> Sync and Backup)
3. Make sure sync is *not* enabled
4. Click "Sync with another device" on both platforms
5. From the iOS device, tap Share Text Code and Copy
6. Make sure this code is on the pasteboard of your macOS device
7. Click the "Paste" button **multiple times rapidly.**
8. Confirm there is no error alert shown

### Notes to Reviewer

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/DDGSync/SyncConnectionController.swift
```diff
@@ -86,7 +86,7 @@ public protocol SyncConnectionControlling {
     func loginAndShowDeviceConnected(recoveryKey: SyncCode.RecoveryKey, isRecovery: Bool) async throws
 }
 
-final public class SyncConnectionController: SyncConnectionControlling {
+public actor SyncConnectionController: SyncConnectionControlling {
```
graeme, line 84: Made this an Actor so that the `isCodeHandlingInFlight` check would be safe. This is why there's so many changes. I think this is a reasonable change to make

amddg44 (approved):
LGTM! And nice improvement on the polling cleanup when exiting out of the connection dialog ❤️

PR #700, opened 2025-05-08
PIR iOS initial implementation: Save profile debug UI

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210197361112887?focus=true
Tech Design URL:
CC:

### Description
Adds a debug UI to save a PIR profile.
Once one has done that, everything should magically work. I.e. when you background the app, as long as you're logged in to privacy pro, PIR background tasks will be scheduled.

It also removes the now unnecessary DataManager, for which I think we are all grateful

### Testing Steps

1. Go to the debug settings, and select PIR
2. Select Save profile 
3. Type in whatever data you like and press save
4. An alert should pop up confirming success
5. You can verify it saved correctly in the DB browser. 
6. If you really want you can also test on an actual device and check that background tasks are scheduled when the app is backgrounded (it should log when it happens)

### Notes to Reviewer

---
###### Internal references:


iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
```diff
@@ -71,39 +90,41 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
...
-        case 0:
-            cell.textLabel?.text = "Database browser"
-        default:
-            break
-        }
-    }
-
-    private func didSelectDatabaseBrowser(at indexPath: IndexPath) {
+    private func didSelectDatabase(at indexPath: IndexPath) {
         guard let dbpManager = DataBrokerProtectionIOSManager.shared else {
```
THISISDINOSAUR, line 111: This is something to undo, but I still haven't decided on the best way to do it. I can think of two options:
- Pass it in from the service, to the MainCoordiantor, MainViewController, Settings, etc
- Put it on AppDependencyProvider but init it with the rest of the AppConfiguration things

I don't like either tbh. The constructor for MainViewController is already absurd, and it has no business with this class anyway. The latter I think is acceptable since I think dbpManager should be optional anyway, but it feels a bit like the pixelkit thing where it's reliant on setup happening later, which is a bit of a foot gun

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/DataBrokerProtectionIOSManager.swift
```diff
@@ -157,47 +156,24 @@ public final class DataBrokerProtectionIOSManager {
...
-        let birthYear = year - 58
-        let profile = DataBrokerProtectionProfile(names: [.init(firstName: "Steve", lastName: "Smith")],
-                                                  addresses: [.init(city: "Dallas", state: "TX")],
-                                                  phones: [],
-                                                  birthYear: birthYear)
-        cache.profile = profile
-        Task { @MainActor in
-            _ = try await communicationLayer.saveProfile(params: [], original: WKScriptMessage())
-        }
-#endif
```
samsymons, line 200: 🙌

samsymons (approved):
LGTM overall, thanks!!

> for which I think we are all grateful

yes

PR #712, opened 2025-05-09
[iOS] increase tab switcher close button hit area

brindy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/392891325557410/task/1210169272747567?focus=true
Tech Design URL:
CC:

### Description
Increases the hit area on the close button in tab switcher slightly.

### Testing Steps
1. Open a few tabs
2. Go to the tab switcher
3. Tap the close button
4. Also try tapping around the close button and check that it's not too big
5. Check both styles (grid vs list)

### Impact and Risks


dus7 (approved):
Tested on a device. I don't feel a noticeable difference, but I'm not failing at closing tabs in the App Store version now either, so maybe that's why. LGTM.

PR #726, opened 2025-05-12
Updates should ask the user before restarting the app

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206580121312550/task/1210262964843197?focus=true

### Description

App updates will no longer automatically restart the App.

### Testing Steps

When switching the feature flag ON and OFF please restart the app to make sure you're working on a clean state.

Validate behavior is unchanged with feature-flag off:

1. Automatic updates should stop only at the restart checkpoint.
2. Manual updates should stop only at the download checkpoint.
3. Switching from automatic to manual while a download is ongoing should reset the update and stop at the download checkpoint.
4. Switching from manual to automatic while a download is ongoing should reset the update and start downloading again.
5. When automatic updates are enabled and the update is expired (use Debug Menu > Updates > Restart last update check), clicking "Restart to Update" should download the update again.

Validate new behavior with feature-flag on:

1. Automatic updates should stop only at the restart checkpoint.
2. Manual updates should stop at the download checkpoint and again at the restart checkpoint.
3. Switching from Automatic to Manual while a download is ongoing should reset the update and stop at the download checkpoint
4. Switching from Manual to Automatic while a download is ongoing should reset the update and start downloading again
5. When automatic updates are enabled and the update is expired (use Debug Menu > Updates > Restart last update check), clicking "Restart to Update" should install the update immediately.
6. When automatic updates are enabled and the update is expired (use Debug Menu > Updates > Restart last update check), closing settings should and reopening it should result in a new update being downloaded.  The same goes for the update notification (blue dot).

### Impact and Risks

Meidum: could affect the update feature, but I've done some thorough testing and expect the same will happen during this review.

#### What could go wrong?

We could break the update feature.

### Quality Considerations

The most important bit is to make sure to validate the current experience remains unchanged.

### Notes to Reviewer

NA

---
###### Internal references:


macOS/Configuration/BuildNumber.xcconfig
```diff
@@ -1 +1 @@
-CURRENT_PROJECT_VERSION = 432
+CURRENT_PROJECT_VERSION = 400
```
diegoreymendez: 🔴🔴🔴🔴🔴 ROLL BACK BEFORE MERGING

macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -783,6 +783,7 @@ final class MainMenu: NSMenu {
 
             NSMenuItem(title: "Logging").submenu(setupLoggingMenu())
             NSMenuItem(title: "AI Chat").submenu(AIChatDebugMenu())
+            NSMenuItem(title: "Updates").submenu(UpdatesDebugMenu())
```
diegoreymendez, line 786: Use this menu for testing - no need for multiple builds, if you expire the current build you'll be forcing a new update, which can be useful for simulating new builds becoming available.

macOS/DuckDuckGo/NavigationBar/View/MoreOptionsMenu.swift
```diff
@@ -405,7 +405,12 @@ final class MoreOptionsMenu: NSMenu, NSMenuDelegate {
             return
         }
 
-        addItem(UpdateMenuItemFactory.menuItem(for: update))
+        if featureFlagger.isFeatureOn(.updatesWontAutomaticallyRestartApp) {
+            addItem(UpdateMenuItemFactory.menuItem(for: updateController))
```
diegoreymendez, line 409: I'm passing the update controller so I can have more control in the menu item.

macOS/DuckDuckGo/Preferences/View/PreferencesAboutView.swift
```diff
@@ -48,6 +48,10 @@ extension Preferences {
                     UpdatesSection(areAutomaticUpdatesEnabled: $areAutomaticUpdatesEnabled, model: model)
 #endif
                 }
+            }.task {
+                if model.mustCheckForUpdatesBeforeUserCanTakeAction {
+                    model.checkForUpdate()
+                }
```
diegoreymendez: When the feature flag is ON we may now check for new updates.

macOS/DuckDuckGo/Preferences/View/PreferencesAboutView.swift
```diff
@@ -292,30 +296,38 @@ extension Preferences {
...
                     }
-                    .buttonStyle(UpdateButtonStyle(enabled: false))
-                    .disabled(true)
                 }
+            } else {
+                let configuration = model.updateButtonConfiguration
+
+                Button(configuration.title, action: configuration.action)
+                    .buttonStyle(UpdateButtonStyle(enabled: configuration.enabled))
+                    .disabled(!configuration.enabled)
```
diegoreymendez, line 332: The diff is ugly, but this is what changes... if the feature flag is ON we now use this logic here.

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -186,12 +213,33 @@ final class UpdateController: NSObject, UpdateControllerProtocol {
...
         Logger.updates.log("Checking for updates respecting rollout")
 
+        userDriver?.userCheckedForUpdates()
         updater.checkForUpdatesInBackground()
     }
 
     // Check for updates immediately, bypassing the rollout schedule
     // This is used for user-initiated update checks only
     func checkForUpdateSkippingRollout() {
+        let discardCurrentUpdate = !autoRestartAllowed && shouldForceUpdateCheck
```
diegoreymendez: This is the new expiration logic.  This will be true if the existing update must be discarded.

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -186,12 +213,33 @@ final class UpdateController: NSObject, UpdateControllerProtocol {
 
         Logger.updates.log("Checking for updates respecting rollout")
 
+        userDriver?.userCheckedForUpdates()
```
diegoreymendez: This should be updated when the user checks for updates, immediately.

macOS/DuckDuckGo/Preferences/View/PreferencesAboutView.swift
quanganhdo: Note that `model.checkForUpdate()` calls `updater.checkForUpdateSkippingRollout()`, which bypasses rollout group, so it is a behavior change. 

I.e. Currently we only skips rollout group when the user explicitly clicks on the Check for Updates button, with this change if the user lands on Preferences > About, we immediately pulls the most recent update.

macOS/DuckDuckGo/Updates/UpdatesDebugMenu.swift
```diff
@@ -0,0 +1,48 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import AppKit
+import AIChat
+
+final class UpdatesDebugMenu: NSMenu {
+    private let customURLLabelMenuItem = NSMenuItem(title: "")
+    private let debugStorage = AIChatDebugURLSettings()
```
quanganhdo: I assume this is subject to change?

macOS/DuckDuckGo/Preferences/Model/AboutPreferences.swift
```diff
@@ -28,15 +28,27 @@ final class AboutPreferences: ObservableObject, PreferencesTabOpening {
...
+
+        self.featureFlagger = featureFlagger
         self.internalUserDecider = internalUserDecider
         self.isInternalUser = internalUserDecider.isInternalUser
         self.internalUserCancellable = internalUserDecider.isInternalUserPublisher
             .sink { [weak self] in self?.isInternalUser = $0 }
     }
 
 #if SPARKLE
+    var autoRestartAllowed: Bool {
```
quanganhdo: Maybe something like `allowsRestartWithoutUserConfirmation` would be clearer?

macOS/DuckDuckGo/Updates/UpdateUserDriver.swift
```diff
@@ -261,7 +268,14 @@ final class UpdateUserDriver: NSObject, SPUUserDriver {
...
 
-        if checkpoint == .restart && !ignoresCheckpoint {
+        guard autoRestartAllowed else {
+            onResuming = { reply(.install) }
+            updateProgress = .updateCycleDone(.pausedAtRestartCheckpoint)
+            Logger.updates.log("Updater paused at restart checkpoint")
+            return
+        }
+
+        if areAutomaticUpdatesEnabled || !autoRestartAllowed {
```
quanganhdo: Won’t `!autoRestartAllowed` always be `false` here following the guard statement?

macOS/DuckDuckGo/Preferences/Model/AboutPreferences.swift
quanganhdo: On second thought, having `autoRestartAllowed` as inversion of the feature flag value & using it in `guard` statements make the logic hard(er) to follow (double-negativity). I think this is one of the cases where

```
if isFeatureOn {
  doSomething()
} else {
  doSomethingElse() 
}
```

would be much easier to follow (and later easier to delete when feature flag is permanently ON too)

quanganhdo: Triple-negativity, because the flag itself is negative (`updateWONTAutomaticallyRestartApp`) 😅

macOS/DuckDuckGo/Updates/UpdateController.swift
quanganhdo: Actually this should be inside `checkForUpdateSkippingRollout()` if you want it to happen when the user explicitly/force checks for updates.

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -186,12 +231,44 @@ final class UpdateController: NSObject, UpdateControllerProtocol {
...
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
+            guard let self,
+                  let updater = try? configureUpdater(needsUpdateCheck: true) else {
+                return
+            }
+            self.updater = updater
+
+            userDriver?.userCheckedForUpdates()
+            updater.checkForUpdates()
```
quanganhdo: Similar to a previous comment, this will bypass our rollout group, which likely is not what we want.

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -155,6 +186,20 @@ final class UpdateController: NSObject, UpdateControllerProtocol {
...
+        subscribeToResignKeyNotifications()
+    }
+
+    private var cancellables = Set<AnyCancellable>()
+
+    private func subscribeToResignKeyNotifications() {
+        NotificationCenter.default.publisher(for: NSWindow.didResignKeyNotification)
+            .debounce(for: .seconds(1), scheduler: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.discardCurrentUpdateIfExpiredAndCheckAgain()
```
quanganhdo: If DuckDuckGo.app binary is owned by another user, this leads to unexpected permission prompt and repeated download attempt if the user cancels it (see test case 4 [here](https://app.asana.com/1/137249556945/project/1202500774821704/task/1209475513300639?focus=true)).

Screenshot 2025-05-20 at 1 19 57 AM

macOS/DuckDuckGo/Preferences/Model/AboutPreferences.swift
diegoreymendez: Using guards for the new logic means the old logic won't show up in the diff as changed, which is what I was looking for.

I understand the comment but I don't personally agree that this would be better:

```swift
guard !updatesWontAutomaticallyRestartAppFeatureFlag else {
    // new logic
}

// old logic
```

I've updated the code to use `useLegacyAutoRestartLogic` instead, so it becomes:

```swift
guard useLegacyAutoRestartLogic else {
    // new logic
}

// old logic
```

macOS/DuckDuckGo/Updates/UpdateController.swift
diegoreymendez: Is there documentation backing this up?  I haven't found any so far even though I'm looking for it.

macOS/DuckDuckGo/Preferences/View/PreferencesAboutView.swift
diegoreymendez: I've updated the logic here so it won't skip rollout.

macOS/DuckDuckGo/Updates/UpdateController.swift
diegoreymendez: Updated the name for clarity.

The last update check date should be updated regardless of whether the check was user-initiated or not.  This is especially relevant for automatic updates.

macOS/DuckDuckGo/Updates/UpdatesDebugMenu.swift
diegoreymendez: Good catch.  This was a copy-pasta mistake.

macOS/DuckDuckGo/Updates/UpdateUserDriver.swift
diegoreymendez: Removed.

macOS/DuckDuckGo/Updates/UpdateController.swift
diegoreymendez: Never mind.  I changed the logic so we won't skip rollout now.

macOS/DuckDuckGo/Updates/UpdateController.swift
diegoreymendez: I can't find any difference between the test build and our release build in terms of this.  I don't think this is an issue introduced by my changes.

Could you confirm?

quanganhdo: As discussed on the Zoom call, let’s make a task for this behavior, but by itself it’s not a blocker.

quanganhdo (approved):
Thanks for addressing the feedback! Good to go after reverting the change to the version number.

macOS/Configuration/BuildNumber.xcconfig
diegoreymendez: Removed before merging.

PR #753, opened 2025-05-14
Improve Phishing Datasets Auto-updates

not-a-rootkit:
Task/Issue URL:  https://app.asana.com/1/137249556945/project/1208717418466383/task/1210250090189769?focus=true
Tech Design URL: n/a
CC: @samsymons 

### Description
Now that we have three datasets that are updated separately, it's more often for them to be at different revisions at any given time. As such, this script has been failing so I've:
- Removed the revision check
- Added a new step to ensure the oldest revision number is the one injected into the `MaliciousSiteProtectionManager` class to ensure we don't skip any revisions

Passing run: https://github.com/duckduckgo/apple-browsers/actions/runs/15023347265

### Testing Steps

1. Run `bash scripts/update_phishing_detection_data.sh`
2. Check it updates the datasets
3. Run the GitHub workflow
4. Check it creates a PR with the right data

### Notes to Reviewer

---
###### Internal references:


samsymons:
Overall the changes to the script seem reasonable, but after looking at the embedded file updates I'm concerned about volatility with their file sizes. Here's the before and after:

**Before:**
Malicious Sites Before

**After:**
Malicious Sites After

From this it looks like the scam protection files have dropped in size significantly, but the others have gone up - `phishingFilterSet.json` for example has gone from 23MB to 32.2MB. Since this directly translates into an increased binary size on the App Store, I think we should chat about it quickly with app DRIs - I'll take this to Asana.

samsymons (approved):
Approving after talking through the file size detail in Asana. Thanks!

PR #778, opened 2025-05-16
PIR iOS Alpha: pixels

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210271610309562?focus=true
Tech Design URL:
CC:

### Description
Adds the pixels and parameters for the required alpha build. 
Note that these parameters must never reach production, and are compile flagged and commented appropriately 
Do not merge until privacy triage is approved

Also sneaks unrelated changes in:
- Fixes an issue where we didn't schedule a new background task if the previous one completed without the expiration handler firing
- Changes the events table in the db viewer to show the newest first
- Renames a file to remove the DataBrokerProtection prefix as previously discussed. The iOSManager is now the only one with it. I deliberately kept it, since following the same logic it would just be called "manager" which is definitely confusing. And it is the one object that manages all on DBP on iOS, so I think it's fair for it to still have DBP in the name

### Testing Steps

1. Smoke test macOS with both debug and release builds
2. Test on a real iOS device, both debug and release builds:
- Release build shouldn't do PIR jobs or fire pixels 
- The debug build should fire background task related pixels, and scan and opt out start and end pixels should have device IDs and device names. You check the device ID matches the ID in PIR settings
- You can force a background task to run by backgrounding the app, pausing the debugger, and running the following:
```
e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"com.duckduckgo.app.dbp.backgroundProcessing"]
```

### Notes to Reviewer

---
###### Internal references:


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileScanSubJob.swift
```diff
@@ -276,8 +278,20 @@ struct BrokerProfileScanSubJob {
                     let calculateDurationSinceLastStage = now.timeIntervalSince(attempt.lastStageDate) * 1000
                     let calculateDurationSinceStart = now.timeIntervalSince(attempt.startDate) * 1000
                     pixelHandler.fire(.optOutFinish(dataBroker: attempt.dataBroker, attemptId: attemptUUID, duration: calculateDurationSinceLastStage))
+
+// This should never ever go to production and only exists for internal testing
+#if os(iOS) && (DEBUG || ALPHA)
```
THISISDINOSAUR, line 285: There are many ways to do this with fewer compile flags, but I thought it necessary to do this everywhere a device identifier is even mentioned to: 
1) mitigate the risk of any of this actually going to production (since multiple areas would have to be changed accidentally)
2) mitigate the optics risk given the code is open source. There should be no doubt that anything referencing the device ID isn't in production, even at a glance.

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/DataBrokerProtectionIOSManager.swift
```diff
@@ -226,6 +241,16 @@ public final class DataBrokerProtectionIOSManager {
...
+                let timeTaken = Date.now.timeIntervalSince(startTime)
+                Logger.dataBrokerProtection.log("Background task finshed all operations with time taken: \(timeTaken)")
+// This should never ever go to production due to the deviceID and only exists for internal testing
+#if DEBUG || ALPHA
+                self.iOSPixelsHandler.fire(.backgroundTaskEndedHavingCompletedAllOperations(
+                    duration: timeTaken * 1000.0,
+                    deviceID: DataBrokerProtectionSettings.deviceIdentifier))
+#endif
+
+                self.scheduleBGProcessingTask()
```
THISISDINOSAUR, line 253: I also noticed a we don't schedule a new background task if we complete all jobs before the expiration handler fires

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileScanSubJob.swift
samsymons: Makes sense to me, I'm happy to have this be extremely precise and cumbersome to write.

samsymons (approved):
LGTM from testing locally, confirmed that I had an ID generated and that it was passed along with the pixels defined here. Note that the ID is not included in _all_ pixels, `data_broker_http_error` and `data_broker_action-failed_error` for example, as expected from the Privacy Triage - this does a great job of covering the overall lifecycle.

PR #787, opened 2025-05-19
Adapt Tabs Model to use KeyValueFile store

bwaresiak:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1209702805085421?focus=true
Tech Design URL:
CC:

### Description

Refactor codebase to move persistence location of Tabs Model from User Defaults to local file store.

### Testing Steps

I. Clean install should work and persist tabs between restarts.
II. Migration from older version to this version should keep open tabs.
III. Auto Clear smoke tests.

Additional:
- Thrown an error from TabsModelPersistence init to validate read error on app restart.

### Impact and Risks

### Notes to Reviewer

---
###### Internal references:


iOS/DuckDuckGo/UICoordination/MainCoordinator.swift
```diff
@@ -122,17 +124,16 @@ final class MainCoordinator {
...
     private static func prepareTabsModel(previewsSource: TabPreviewsSource = DefaultTabPreviewsSource(),
-                                         appSettings: AppSettings = AppDependencyProvider.shared.appSettings) -> TabsModel {
+                                         tabsPersistence: TabsModelPersisting,
+                                         appSettings: AppSettings = AppDependencyProvider.shared.appSettings) throws -> TabsModel {
         let isPadDevice = UIDevice.current.userInterfaceIdiom == .pad
         let tabsModel: TabsModel
         if AutoClearSettingsModel(settings: appSettings) != nil {
             tabsModel = TabsModel(desktop: isPadDevice)
-            tabsModel.save()
+            try tabsPersistence.save(model: tabsModel)
```
dus7: This is not throwing so you can remove `try`.

dus7: ⚠️ One downside here is that when this is done before migrating to the new store, we'll set the new store data, but won't clear the legacy tabs. This is not necessarily a bad thing unless we'll run into a case where the new data is empty again and we'll resurrect the legacy tabs from the past. Simple `tabsPersistence.getTabsModel()` call should take care of it though.

bwaresiak: Good point, I've added explicit call to clear() here to clean both stores before saving new state.

PR #790, opened 2025-05-19
Add Protections Report widget to unify Privacy Stats and Recent Activity

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1209353029633684?focus=true

### Description
This change unifies protection stats widgets on New Tab Page, allowing all users to access both Privacy Stats
and Recent Activity. Previously Recent Activity was only available to existing users and new users were seeing Privacy Stats.
The new widget that combines the two is called Protections Report. It provides its own API for visibility settings
and currently active “feed” (Privacy Stats or Recent Activity), plus defines API to request total number of blocked
trackers.
Visibility APIs for Privacy Stats and Recent Activity were removed and totalCount was removed from PrivacyStatsData.

### Testing Steps
1. Run `./clean-app.sh debug` and launch the app. Verify that you’re being presented with Protections Report in expanded state (but no data of course) and the selected feed is “Stats”.
2. Quit the app, run `./clean-app.sh debug` and then also run `defaults write com.duckduckgo.macos.browser.debug new-tab-page.is-new-user -bool false`, and then launch the app. Verify that you’re being presented with Protections Report in expanded state (but no data) and the selected feed is “Details”.
3. Go to Settings -> Appearance and verify that under New Tab Page you see Favorites and Protections Report. Verify that toggling Protections Report from there adjusts the visibility of the protections widget.
4. Do a general smoke-testing of the Privacy Stats and Recent Activity. There are no changes to the widgets themselves and to their logic (besides visibility settings), only to how they’re presented on the NTP.

### Impact and Risks

#### What could go wrong?
The biggest risk is failing to migrate existing users properly to the new widget, in terms of migrating:
* visibility flag,
* collapsed state,
* selected feed.
The migration logic is handled by `NewTabPageProtectionsReportSettingsMigrator` and unit-tested thoroughly. There’s also [an Asana task](https://app.asana.com/1/137249556945/project/72649045549333/task/1210270584197856?focus=true) outlining the logic here (that’s extracted from the unit tests, actually).

---
###### Internal references:


macOS/DuckDuckGo/Preferences/Model/AppearancePreferences.swift
```diff
@@ -102,6 +113,12 @@ struct AppearancePreferencesUserDefaultsPersistor: AppearancePreferencesPersisto
 
     @UserDefaultsWrapper(key: .showTabsAndBookmarksBarOnFullScreen, defaultValue: true)
     var showTabsAndBookmarksBarOnFullScreen: Bool
+
+    init(keyValueStore: KeyValueStoring = UserDefaultsWrapper<Any>.sharedDefaults) {
```
ayoy: Here I’m starting the attempt to remove `@UserDefaultsWrapper` from AppearancePreferences persistor. The newly added `isProtectionsReportVisible` flag uses the `keyValueStore` parameter.

macOS/DuckDuckGo/NewTabPage/NewTabPageCoordinator.swift
```diff
@@ -40,45 +39,42 @@ final class NewTabPageCoordinator {
...
 
         notificationCenter.publisher(for: .newTabPageWebViewDidAppear)
             .prefix(1)
-            .sink { [weak self, weak customizationModel, weak appearancePreferences] _ in
-                guard let self, let customizationModel, let appearancePreferences else {
-                    return
-                }
-                fireNewTabPageShownPixel(appearancePreferences: appearancePreferences, customizationModel: customizationModel)
+            .sink { [weak self] _ in
+                self?.newTabPageShownPixelSender.firePixel()
```
ayoy, line 72: Pixel firing logic is now extracted into a separate struct.

macOS/DuckDuckGo/NewTabPage/NewTabPageCoordinator.swift
```diff
@@ -40,45 +39,42 @@ final class NewTabPageCoordinator {
         notificationCenter: NotificationCenter = .default,
         fireDailyPixel: @escaping (PixelKitEvent) -> Void = { PixelKit.fire($0, frequency: .daily) }
     ) {
+
+        let settingsMigrator = NewTabPageProtectionsReportSettingsMigrator(keyValueStore: keyValueStore)
+        let protectionsReportModel = NewTabPageProtectionsReportModel(
```
ayoy, line 45: NTP Coordinator initializes Protections Report model because it’s needed for the "NTP shown" pixel firing logic. That model is then passed to NewTabPageActionsManager as a parameter when it’s used in its user script client.

macOS/LocalPackages/NewTabPage/Sources/NewTabPage/Configuration/NewTabPageConfigurationClient.swift
```diff
@@ -106,34 +95,20 @@ public final class NewTabPageConfigurationClient: NewTabPageUserScriptClient {
     }
 
     private func fetchWidgets() -> [NewTabPageDataModel.NewTabPageConfiguration.Widget] {
-        var widgets: [NewTabPageDataModel.NewTabPageConfiguration.Widget] = [
+        [
             .init(id: .rmf),
             .init(id: .freemiumPIRBanner),
             .init(id: .nextSteps),
             .init(id: .favorites),
+            .init(id: .protections)
```
ayoy, line 103: Sections availability provider is not needed anymore because Protections Report widget is always available for all users.

macOS/LocalPackages/NewTabPage/Sources/NewTabPage/PrivacyStats/NewTabPagePrivacyStatsClient.swift
```diff
@@ -28,11 +28,8 @@ public final class NewTabPagePrivacyStatsClient: NewTabPageUserScriptClient {
     private var cancellables: Set<AnyCancellable> = []
 
     enum MessageName: String, CaseIterable {
-        case getConfig = "stats_getConfig"
         case getData = "stats_getData"
-        case onConfigUpdate = "stats_onConfigUpdate"
         case onDataUpdate = "stats_onDataUpdate"
-        case setConfig = "stats_setConfig"
```
ayoy, line 35: Config messages for Privacy Stats and Recent Activity are removed because they were only managing visibility, which is now handled in the top-level Protections widget.

SharedPackages/BrowserServicesKit/Sources/PrivacyStats/internal/PrivacyStatsUtils.swift
```diff
@@ -98,6 +98,34 @@ final class PrivacyStatsUtils {
...
+        let sumExpressionDescription = NSExpressionDescription()
+        sumExpressionDescription.name = "totalCount"
+        sumExpressionDescription.expression = sumExpression
+        sumExpressionDescription.expressionResultType = .integer64AttributeType
+
+        request.propertiesToFetch = [sumExpressionDescription]
+        request.resultType = .dictionaryResultType
+
+        let results = try context.fetch(request) as? [[String: Any]]
+        return (results?.first?["totalCount"] as? Int64) ?? 0
```
jotaemepereira, line 126: I have a question here about how we would hit the `?? 0`.

That happens becaue we do not have records? Or does it happen because of an error of the schema and it is something we do not expect? If it is the latter I would be inclined to throw so we can measure that accordingly wiht the `failedToFetchPrivacyStatsTotalCount` error.

macOS/DuckDuckGo/NewTabPage/NewTabPageProtectionsReportSettingsMigrator.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    /**
+     * Returns `activity` if the user wasn't *new*, otherwise returns `privacyStats`
+     * (also when there was no value persisted).
+     */
+    var activeFeed: NewTabPageDataModel.Feed {
+        let isNewUser = keyValueStore.object(forKey: LegacyKey.isNewUser.rawValue) as? Bool
+        return isNewUser == false ? NewTabPageDataModel.Feed.activity : .privacyStats
```
jotaemepereira, line 69: Nit-picky:

```swift
return isNewUser ?  .privacyStats : NewTabPageDataModel.Feed.activity
```

macOS/DuckDuckGo/Tab/Navigation/DuckURLSchemeHandler.swift
```diff
@@ -318,7 +318,7 @@ private extension DuckURLSchemeHandler {
         case "svg": return "image/svg+xml"
         case "ico": return "image/x-icon"
         case "riv": return "application/octet-stream"
-        case "json": return "application/json"
+        case "json", "map": return "application/json"
```
jotaemepereira, line 321: ❓ Out of curiosity, what is this?

jotaemepereira (approved):
This looks great @ayoy, went through the whole scenarios and worked as expected. 

I’ve added some non-blocker comments and questions

SharedPackages/BrowserServicesKit/Sources/PrivacyStats/internal/PrivacyStatsUtils.swift
ayoy: This shouldn’t actually happen:
* When the fetch fails, it would throw an error. When it doesn’t fail, it returns `[[String: Any]]` object unless there’s a bug in Core Data :)
* Then, in line 126 we’re traversing through the known JSON structure and it shouldn’t ever be different than what’s here.

We use the same approach in PrivacyStatsUtils, BookmarkUtils and BookmarkCoreDataImporter so I assumed it’s fine as the likelihood of this becoming broken without a huge bug by Apple is really minimal.

macOS/DuckDuckGo/NewTabPage/NewTabPageProtectionsReportSettingsMigrator.swift
ayoy: Ah actually we need `isNewUser == false` here, because it’s an optional.

Using `isNewUser ?` won’t compile, and using `isNewUser == true` would set privacy stats only when isNewUser is `true`. Here we also want this to happen if `isNewUser` is `nil`. That’s why we compare to `false` specifically, and for other cases we fall back to `.privacyStats`.

macOS/DuckDuckGo/Tab/Navigation/DuckURLSchemeHandler.swift
ayoy: These are [JavaScript source maps](https://dev.to/inspector/what-are-source-maps-and-how-to-properly-use-them-18h4). They are used for debugging and Shane asked me to add support for serving source map files from duck URLs to simplify his workflow.

PR #806, opened 2025-05-20
Use ThrowingKeyValueStoring in AppearancePreferences and remove singleton

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210321544153920?focus=true
CC: @bwaresiak 

### Description
This change updates AppearancePreferencesPersistor to use ThrowingKeyValueStoring instead of UserDefaults.
All new code added to the persistor should use the new store, while the pre-existing properties still
rely on UserDefaults for storage.
To make this happen, AppearancePreferences, StartupPreferences and DataClearingPreferences singletons
were removed and replaced with properties held by AppDelegate.

### Testing Steps
1. Verify that CI is green.
2. Verify that [UI tests workflow](https://github.com/duckduckgo/apple-browsers/actions/runs/15183578651) is green.
3. Smoke test the app to verify that it doesn’t crash at startup and persists appearance settings.

### Impact and Risks

#### What could go wrong?
Removing singletons could cause app crashes, but that would happen on app startup, and my testing went well so it shouldn’t be a problem.

### Notes to the Reviewer
The changes here are mostly about renaming `.shared` to `NSApp.delegateTyped.*Preferences` with small refactoring here and there to inject preferences objects from the app delegate. I deliberately limited the scope of the refactoring for now, because it can be improved gradually once we have this first bit in place.

—
###### Internal references:


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -221,6 +226,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             fileStore = EncryptedFileStore()
         }
 
+        appearancePreferences = AppearancePreferences(keyValueStore: keyValueStore)
```
bwaresiak, line 231: @ayoy  This looks good, there is one interesting addition we should think about: it is possible the first read will fail, so we could (at the place we setup kvs) add dummy read, and crash up on failure - this way consumers will always be able to read data successfully. But I'm leaving that up to you as I'm not sure what is desired startup failure handling on macOS.

ayoy: Thanks @bwaresiak! I’ve added a dummy check at the KVS initialization that will force a crash if throwing (that’s how macOS handles failures in persistent layer setup).

jotaemepereira (approved):
Tested going back from legacy to new, and values are restored as expected 👍🏼

PR #791, opened 2025-05-19
macOS: Provide autoconsent rules from config instead of bundled.

sammacbeth:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1163321984198618/task/1209960043798128?focus=true
Tech Design URL:
CC:

### Description
- Takes CPM rules from config settings and passes to the userscript.
- This allows rules to be configured remotely via privacy-config, rather than needing a bundle update.

### Testing Steps

1. Visit https://privacy-test-pages.site/features/autoconsent/index.html
2. Expected: Cookie popup is not handled.
3. Set the following as the remote-config URL: https://duckduckgo.github.io/privacy-configuration/pr-3150/v4/macos-config.json
4. Reload https://privacy-test-pages.site/features/autoconsent/index.html
5. Expected: Cookie popup is handled.

### Impact and Risks

Medium: Loading the rules incorrectly could break sites.

#### What could go wrong?
 1. Config format is not parsable: parsing error will be caught, and feature will run with no rules.
 2. Remote config has a newer rule format: https://app.asana.com/1/137249556945/project/481882893211075/task/1210142069641181 explictly handles this and will disable the feature.

### Notes to Reviewer

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #807, opened 2025-05-20
iOS: Provide autoconsent rules from config instead of bundling

sammacbeth:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210318211455595
Tech Design URL:
CC:

### Description
Same as #791 but for iOS.

- Takes CPM rules from config settings and passes to the userscript.
- This allows rules to be configured remotely via privacy-config, rather than needing a bundle update.

### Testing Steps

1. Visit https://privacy-test-pages.site/features/autoconsent/index.html
2. Expected: Cookie popup is not handled.
3. Set the following as the remote-config URL: https://ppd.ngrok.io/sam/config
4. Reload https://privacy-test-pages.site/features/autoconsent/index.html
5. Expected: Cookie popup is handled.

### Impact and Risks

Medium: Loading the rules incorrectly could break sites.

#### What could go wrong?
 1. Config format is not parsable: parsing error will be caught, and feature will run with no rules.
 2. Remote config has a newer rule format: https://app.asana.com/1/137249556945/project/481882893211075/task/1210142069641181 explictly handles this and will disable the feature.


aataraxiaa (approved):
Nice one @sammacbeth. Works as expected.

PR #812, opened 2025-05-20
PIR iOS: Pixels take 2

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210271610309562?focus=true
Tech Design URL:
CC:

### Description
Readds what was reverted in https://github.com/duckduckgo/apple-browsers/pull/809

Adds the pixels and parameters for the required alpha build.
Note that these parameters must never reach production, and are compile flagged and commented appropriately
Do not merge until privacy triage is approved

Also sneaks unrelated changes in:

Fixes an issue where we didn't schedule a new background task if the previous one completed without the expiration handler firing
Changes the events table in the db viewer to show the newest first
Renames a file to remove the DataBrokerProtection prefix as previously discussed. The iOSManager is now the only one with it. I deliberately kept it, since following the same logic it would just be called "manager" which is definitely confusing. And it is the one object that manages all on DBP on iOS, so I think it's fair for it to still have DBP in the name

### Testing Steps

1. Smoke test macOS with both debug and release builds
2. Test on a real iOS device, both debug and release builds:
- Release build shouldn't do PIR jobs or fire pixels 
- The debug build should fire background task related pixels, and scan and opt out start and end pixels should have device IDs and device names. You check the device ID matches the ID in PIR settings
- You can force a background task to run by backgrounding the app, pausing the debugger, and running the following:
```
e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"com.duckduckgo.app.dbp.backgroundProcessing"]
```

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileScanSubJob.swift
+++ b/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileScanSubJob.swift
@@ -73,6 +73,10 @@ struct BrokerProfileScanSubJob {
             let event = HistoryEvent(brokerId: brokerId, profileQueryId: profileQueryId, type: .scanStarted)
             try dependencies.database.add(event)
 
+#if os(iOS) && (DEBUG || ALPHA)
+            stageCalculator.fireScanStarted()
+#endif
+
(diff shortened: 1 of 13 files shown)


samsymons (approved):
Thanks for the patience, glad we got there in the end - LGTM based on testing the adhoc build!

PR #819, opened 2025-05-21
Fixes a bug that was showing previous for overflown / hidden tabs

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201048563534612/task/1210330686761245?focus=true

### Description

Fixes a bug that was allowing users to see previews for tabs that were overflowing / hidden.

### Testing Steps

1. Add a ton of tabs until you get overflowing and some tabs are hidden.
2. Make sure a couple of last ones have a site loaded so you can test the issue.
3. Ensure previews aren't shown when the mouse if positioned where those overflown tabs are, if they're hidden.

The same steps can be used to see the original issue in action.

### Impact and Risks

Low.

#### What could go wrong?

These changes could break previews if the logic isn't correct.

### Quality Considerations

Just make sure the logic makes sense, and testing validates it.

### Notes to Reviewer

You might want to ensure previews work well when they should.

---
###### Internal references:


macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
```diff
@@ -681,8 +690,6 @@ final class TabBarViewController: NSViewController, TabBarRemoteMessagePresentin
         // did mouse really exit or is it an event generated by a subview and called via the responder chain?
         guard !isMouseLocationInsideBounds else { return }
 
-        self.hideTabPreview(allowQuickRedisplay: true)
```
diegoreymendez: I removed this because we want to show and hide this based on whether the mouse is inside or outside the collection view visible area, and over a visible part of a tab.

macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
```diff
@@ -666,8 +666,17 @@ final class TabBarViewController: NSViewController, TabBarRemoteMessagePresentin
     override func mouseMoved(with event: NSEvent) {
         super.mouseMoved(with: event)
 
+        guard collectionView.isMouseLocationInsideBounds() else {
+            if tabPreviewWindowController.isPresented {
+                hideTabPreview(allowQuickRedisplay: true)
+            }
+            return
+        }
```
diegoreymendez: If the mouse it outside the visible bounds, we check if there's a preview to hide and bail out.

Otherwise we'll show the preview using the pre-existing code.

mallexxx: this one is not correct, the previews are shown not only for the regular tabs but for pinned tabs as well (see the lines below), so not only the `collectionView` should be inspected for bounds: now the Pinned tab previews got broken.
It would be better to add the 3rd if-else branch into the `mouseMoved` hiding the preview if needed but probably changing the `shouldDisplayTabPreviews` would do the job.

macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
mallexxx: this was performing an extra check on mouse exit the tab bar event, why it needs to be removed?

mallexxx (changes requested):
This change breaks Pinned Tab previews, please see comments inline

macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
diegoreymendez: Thanks, good catch on the pinned tabs.

Changing `shouldDisplayTabPreviews` alone won't resolve the issue @tomasstrba reported: If a tab item is partially hidden then moving the mouse over the hidden area won't ever trigger hiding the preview.

We need a trigger to hide the preview when the mouse is no longer over visible elements that show previews.

mallexxx: I think we still should use shouldDisplayTabPreviews fixing out of sight detection and just add an extra call to it

diegoreymendez: I was trying something similar today, but got stuck with the visibility check not quite working - unclear why.  I'll give it another shot tomorrow morning.

macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
diegoreymendez: It's not necessary anymore as the tab preview is hidden on mouse move.

That said, we can keep it if you prefer, so I restored this line.

macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
diegoreymendez: Ok, I just rolled back my changes and tried again and it's working well now.  The logic in `self.hideTabPreview(allowQuickRedisplay: true)` apparently needs no change.

The logic that is kinda off is the code to get the index for the selected tab since it doesn't properly account for visibility, but this is now fixed because of the added guard checking for `shouldDisplayTabPreviews`.

mallexxx (approved):
LGTM 👍
The issue is not reproduced anymore

PR #822, opened 2025-05-21
Fixes a bug with the bookmarks icon blinking when switching tabs

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201048563534612/task/1210331120635171?focus=true

### Description

Fixes an issue that caused the bookmarks icon to blink when switching tabs.

### Testing Steps

1. Load a site in a tab.
2. Create another tab and switch to it.
3. When going back to the tab with the site loaded you'll see the bookmarks icon blinking.

### Impact and Risks

Low

#### What could go wrong?

I don't think anything should.  But worst thing that could happen is I'm calculating visibility wrong for the icon.

### Quality Considerations

Ensure the icon behaves correctly when moving the mouse into and outside of the address bar area.

### Notes to Reviewer

NA

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/MainWindow/MainView.swift
+++ b/macOS/DuckDuckGo/MainWindow/MainView.swift
@@ -283,12 +283,23 @@ final class MainView: NSView {
     override func updateTrackingAreas() {
         if let mouseAboveWebViewTrackingArea {
             removeTrackingArea(mouseAboveWebViewTrackingArea)
-            isMouseAboveWebView = false
             let trackingArea = makeMouseAboveViewTrackingArea()
             self.mouseAboveWebViewTrackingArea = trackingArea
             addTrackingArea(trackingArea)
+            updateIsMouseAboveWebView(for: trackingArea)
         }
         super.updateTrackingAreas()
+    }
+
+    private func updateIsMouseAboveWebView(for trackingArea: NSTrackingArea) {
+        guard let window else {
+            isMouseAboveWebView = false
+            return
+        }
+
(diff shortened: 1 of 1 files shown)


macOS/DuckDuckGo/MainWindow/MainView.swift
```diff
@@ -283,14 +283,25 @@ final class MainView: NSView {
...
 
+    private func updateIsMouseAboveWebView(for trackingArea: NSTrackingArea) {
+        guard let window else {
+            isMouseAboveWebView = false
+            return
+        }
+
+        let mouseLocation = window.mouseLocationOutsideOfEventStream
+        let viewLocation = convert(mouseLocation, from: nil)
+        isMouseAboveWebView = trackingArea.rect.contains(viewLocation)
```
mallexxx: would it work for Tracking Areas whose `rect` == `.zero` - the ones having `.inVisibleRect` flag?
I suggest reusing `isMouseLocationInsideBounds` from `NSViewExtension`

diegoreymendez: That flag isn't used here so it should be fine.  Regardless I take the point to try and make this more maintainable.

It works well if we just don't try to update the state.  The tracking area should remain the same size when this is called (in fact I'm not even sure recreating it is necessary but that's beyond scope for this PR).

tomasstrba (approved):
LGTM! ✅ 

Hopefully, there aren't any side effects of this fix

PR #838, opened 2025-05-23
iOS PIR: Add run prerequisites UI

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1210338467109188?focus=true
Tech Design URL:
CC:

### Description

This PR adds a new "Health Overview" section to the PIR debug menu to indicate whether PIR is able to run, or if not, then why not. Later this section will also include health statistics from the runs themselves.

### Testing Steps

1. Wipe your local instance of PIR and run the app on a device
2. Sign out of Privacy Pro
3. Open the debug menu and confirm that the Health Overview sections shows :x: for all entries
4. Sign into Privacy Pro and then check that the Account and Entitlement rows show a checkmark
5. Now create a profile, and check that when you return to the debug UI that PIR says it's ready to run when the device is locked and connected to power

### Impact and Risks

None, debug menu only

### Notes to Reviewer

This code isn't particularly high quality, since it's a debug menu feature only, and will evolve later when the health metrics are expanded.

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
+++ b/iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
@@ -172,7 +172,11 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
                 if jobScheduled {
                     cell.textLabel?.text = "✅ PIR will run some time after device is locked and connected to power"
                 } else {
-                    cell.textLabel?.text = "❌ Restart the app to schedule PIR"
+                    if UIApplication.shared.backgroundRefreshStatus == .available {
+                        cell.textLabel?.text = "❌ Restart the app to schedule PIR"
+                    } else {
+                        cell.textLabel?.text = "❌ Enable \"Background App Refresh\" in the app's privacy settings"
+                    }
                 }
             }
 


iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
```diff
@@ -88,6 +145,37 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
...
+                    cell.textLabel?.text = "Profile Saved In DB"
+                    cell.detailTextLabel?.text = hasProfile ? "✅" :"❌"
+                } else {
+                    fatalError("Expected 3 rows for the health overview")
+                }
+            case .runPrerequesitesMet(let jobScheduled):
+                if jobScheduled {
+                    cell.textLabel?.text = "✅ PIR will run some time after device is locked and connected to power"
+                } else {
+                    cell.textLabel?.text = "❌ Restart the app to schedule PIR"
```
samsymons: This should also check whether the app has permission in the iOS Privacy settings. Otherwise, I'm also open to a more appropriate message here - I don't think this will really show up in many cases though, since the job is scheduled immediately.

samsymons: Added a check for the background refresh permission, I'll merge this when CI is green 🎉

PR #854, opened 2025-05-24
Add "delete all data" option to PIR debug menu

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1210338467109183?focus=true
Tech Design URL:
CC:

### Description

This PR adds a "Delete All Data" option to the debug menu.

### Testing Steps

1. Add some data to the database
2. Tap "Delete All Data", then hit cancel
3. Check that the profile in the database still exists
4. Now tap "Delete All Data" again and follow through with it
5. Check that the DB is empty, and that the tester ID has changed
6. Long-press the UUID row and check that a "Copy" option exists and does what it says it will

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UtilsShared/DataBrokerProtectionSettings.swift
+++ b/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UtilsShared/DataBrokerProtectionSettings.swift
@@ -172,6 +172,12 @@ extension DataBrokerProtectionSettings {
         }
     }
 
+    @discardableResult
+    public static func rotateDeviceIdentifier() -> String {
+        defaults.removeObject(forKey: deviceIdentifierKey)
+        return deviceIdentifier
+    }
+
     public static var modelName: String {
         var systemInfo = utsname()
         uname(&systemInfo)
--- a/iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
+++ b/iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
@@ -43,6 +43,7 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
         case databaseBrowser
         case saveProfile
         case deviceIdentifier
+        case deleteAllData
 
         var title: String {
             switch self {
@@ -56,6 +57,8 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
 #else
                 return "No UUID due to wrong build type"
 #endif
+            case .deleteAllData:
+                return "Delete All Data"
             }
         }
     }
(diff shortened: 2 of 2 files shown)


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UtilsShared/DataBrokerProtectionSettings.swift
```diff
@@ -172,6 +172,12 @@ extension DataBrokerProtectionSettings {
         }
     }
 
+    @discardableResult
+    public static func rotateDeviceIdentifier() -> String {
+        defaults.removeObject(forKey: deviceIdentifierKey)
+        return deviceIdentifier
```
THISISDINOSAUR: I think it would be good to explain the side effect of using the getter here.
Although it doesn't really matter since this should definitely be temporary

iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
```diff
@@ -229,11 +250,27 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
...
+
         case .none:
             return
         }
     }
 
+    private func presentDeleteAllDataAlertController() {
+        let alert = UIAlertController(title: "Delete All PIR Data?", message: "This will remove all data and statistics from the PIR database, and reset your Tester ID.", preferredStyle: .alert)
+        alert.addAction(title: "Delete All Data", style: .destructive) { [weak self] in
+            try? self?.manager.database.deleteProfileData()
```
THISISDINOSAUR: This is very tangential and there's nothing actionable here, but I've been wondering about accessing the database through the manager here and just generally thinking what public interface do we want to present from PIR to the rest of the app? I'm keen to avoid the situation on macOS.
I don't have any well developed opinions so far, but my sense is that we probably ought not to be accessing database methods directly

It's much less of a problem now, but we used to have very diffuse decision making on macOS and made some effort to consolidate it

THISISDINOSAUR (approved):
LGTM, although see discussion in asana re UUIDs

iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
samsymons: Yeah, the right move here is likely to expose the debug tools we want on the manager and have it take care of talking to the database, allowing us to control exactly what gets surfaced to the clients rather than giving blanket access to the DB as a whole. I'll update this.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UtilsShared/DataBrokerProtectionSettings.swift
samsymons: Updated this to instead add a count to rotated IDs, so we don't have the same getter shenanigans happening any more.

THISISDINOSAUR (approved):

PR #861, opened 2025-05-26
Refactor remote broker delivery triggers

quanganhdo:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210207632059902?focus=true
Tech Design URL:
CC:

### Description

Currently broker update check is done when operations run. This changes the trigger to be:

1. When the agent finishes launching, we attempt to update the brokers with 5 min timeout before starting the queue manager.
2. Every 1 hour, run a periodic broker update check which uses a new `BrokerProfileJobQueueMode` called `brokerUpdate`. When this is in flight, scheduled operations won’t start. Immediate operations aren’t affected.

### Testing Steps

1. Make sure you’re using Staging for PPro
3. Use `branches/bhall-test-infotracer` as the custom service root
4. InfoTracer will be added to the broker list when broker update check is done
5. PIR should function normally

### Impact and Risks
<!— 

—>

#### What could go wrong?

### Quality Considerations
<!— 
- What edge cases exist?
—>

### Notes to Reviewer

—
###### Internal references:


macOS/LocalPackages/DataBrokerProtection-macOS/Sources/DataBrokerProtection-macOS/BackgroundAgent/DataBrokerProtectionAgentManager.swift
```diff
@@ -298,11 +298,18 @@ extension DataBrokerProtectionAgentManager: DataBrokerProtectionBackgroundActivi
...
         Task {
             do {
                 try await brokerUpdater.checkForUpdates()
+            } catch {
+                Logger.dataBrokerProtection.log("Error checking for broker update: \(error.localizedDescription)")
             }
+            semaphore.signal()
         }
+
+        semaphore.wait()
```
quanganhdo: @THISISDINOSAUR Using a semaphore for this is an ugly hack, but I haven’t found a cleaner alternative. I see 2 other approaches:

1/ Modify the queue manager from using completion blocks to async/await so the async `brokerUpdate.checkForUpdates()` can be used. The issue is that the queue manager relies on passing completion blocks around (via `BrokerProfileJobQueueMode`) so doing this would require more extensive refactoring.

2/ Add a synchronous variant to `RemoteBrokerJSONService.checkForUpdates()`. The blocker is the use of PPro access token to interact with DBP API, which is another async operation (that also uses semaphore underneath).

I went with this as a pragmatic solution. Lmk if you have any suggestions.

THISISDINOSAUR: I'm not sure I understand why this is necessary. What's preventing us from doing the broker update before even calling the queue manager?

quanganhdo: Nothing? Guess I tunnel visioned myself into this; will make an update 🤦‍♂️

THISISDINOSAUR:
1/ What's the reasoning for changing the triggers?

2/ I think making the update process an operation and part of queuemanager is a neat idea, but I think at this point it's worth of a TD

3/ If it is in QueueManager, it should be renamed, since it no longer handlers BrokerProfileJob queues exclusively

4/ Have you considered not making it a new mode? The fact that the initial update doesn't use the queue makes me think we're not getting the maximum value out of this approach, when it could just use the queue and add all the other operations on afterwards

4/ Is it safe to interrupt the broker update process?

PR #885, opened 2025-05-28
Add PIR debug environment overrides

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210338467109189?focus=true
Tech Design URL:
CC:

### Description

This PR adds support for different PIR debug environments. It's fairly crude at the moment, I just wanted to get this running and move onto other debug tools before coming back to polish this where needed.

### Testing Steps

1. Open the PIR debug menu and check that each of the environment options looks correct
2. One note is that we don't have a notice on the Web URL setting that it only works in staging - I'll try to add this before we merge

### Notes to Reviewer

---
###### Internal references:


iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
```diff
@@ -271,6 +336,143 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
...
+
+        let saveAction = UIAlertAction(title: "Save", style: .default) { [weak self, weak alert] _ in
+            guard let textField = alert?.textFields?.first,
+                  let value = textField.text else {
+                return
+            }
+
+            self?.settings.serviceRoot = value
+            try? self?.manager.deleteAllData()
+            // self?.forceBrokerJSONFilesUpdate()
```
samsymons: Need to get this working before merging, then review the macOS implementation one more time to ensure everything's working as expected.

THISISDINOSAUR (commented):
lgtm, besides the outstanding bits

iOS/DuckDuckGo/AppServices/DBPService.swift
```diff
@@ -66,7 +66,7 @@ final class DBPService: NSObject {
     }
 }
 
-private final class DBPFeatureFlagger: RemoteBrokerDeliveryFeatureFlagging {
+final class DBPFeatureFlagger: RemoteBrokerDeliveryFeatureFlagging {
```
samsymons, line 69: This is the one public interface change in this PR – I feel mostly okay about it since the feature flagger class is just an interface to the dependency provider and thus I find it acceptable if various parts of the code are creating their own instance, but I would be happy to have the initialization happen elsewhere and keep this private.

iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
```diff
@@ -271,6 +390,185 @@ final class DataBrokerProtectionDebugViewController: UITableViewController {
...
+            }
+        }
+    }
+
+    private func forceBrokerJSONFilesUpdate() throws {
+        Task {
+            settings.resetBrokerDeliveryData()
+            try await brokerUpdater?.checkForUpdates(skipsLimiter: true)
+        }
+    }
```
samsymons, line 567: I've added support for refreshing the remote broker JSON files. Note that this doesn't work out of the box on iOS since the feature flag is disabled, so I'm going to enable that for internal users so that we can test it during the beta. Hardcoding the flag to `true` and then triggering this does refresh the brokers correctly.

samsymons:
Closing this in favor of https://github.com/duckduckgo/apple-browsers/pull/1341

PR #911, opened 2025-05-30
Makes several improvements to login items

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207603085593419/task/1210422858992201?focus=true

### Description

We want to disable login items if the user has no VPN access

Also improves the code and add tests.

### Testing Steps

1. Connect the VPN so login items are loaded.
2. Disconnect it.
3. Restart the app, ensure the login items are restarted but not disabled.
4. Force [this](https://github.com/duckduckgo/apple-browsers/blob/06a80cdad3e04dc878cc096e4179be5a823a4869/macOS/DuckDuckGo/Waitlist/VPNFeatureGatekeeper.swift#L62) to return `false`.
5. Start the App again, the login items should be disabled.

### Impact and Risks

Medium.

#### What could go wrong?

My changes could have some flaw.

### Quality Considerations

I added testing so that this will be more resilient in the future.

### Notes to Reviewer

Instead of trying to hard-disable the VPN I think it's safer to soft-disable it here.  The login items will be disabled, which is what people find annoying.  As long as the user doesn't enable PPro and start the VPN then they won't see anything related to VPN.

---
###### Internal references:


macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNAppEventsHandler.swift
```diff
@@ -32,69 +32,79 @@ import Subscription
...
-    private let legacyAgentBundleID = "HKE973VLUW.com.duckduckgo.macos.browser.network-protection.system-extension.agent.debug"
-    private let legacySystemExtensionBundleID = "com.duckduckgo.macos.browser.debug.network-protection-extension"
-#elseif REVIEW
-    private let legacyAgentBundleID = "HKE973VLUW.com.duckduckgo.macos.browser.network-protection.system-extension.agent.review"
-    private let legacySystemExtensionBundleID = "com.duckduckgo.macos.browser.review.network-protection-extension"
-#else
-    private let legacyAgentBundleID = "HKE973VLUW.com.duckduckgo.macos.browser.network-protection.system-extension.agent"
-    private let legacySystemExtensionBundleID = "com.duckduckgo.macos.browser.network-protection-extension"
-#endif // DEBUG || REVIEW || RELEASE
-#endif // NETP_SYSTEM_EXTENSION
```
diegoreymendez, line 48: Not used.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNAppEventsHandler.swift
```diff
@@ -32,69 +32,79 @@ import Subscription
...
     private var cancellables = Set<AnyCancellable>()
+    private let defaults: UserDefaults
+    private let featureGatekeeper: VPNFeatureGatekeeper
+    private let loginItemsManager: LoginItemsManaging
+
+    // MARK: - Initializers
 
     init(featureGatekeeper: VPNFeatureGatekeeper,
-         featureFlagOverridesPublishingHandler: FeatureFlagOverridesPublishingHandler<FeatureFlag>,
-         uninstaller: VPNUninstalling = VPNUninstaller(),
```
diegoreymendez, line 59: Unused.

macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionIPCTunnelController.swift
```diff
@@ -76,7 +76,7 @@ final class NetworkProtectionIPCTunnelController {
 
     private func enableLoginItems() async throws {
         do {
-            try loginItemsManager.throwingEnableLoginItems(LoginItemsManager.networkProtectionLoginItems)
+            try loginItemsManager.throwingEnableLoginItems(LoginItemsManager.vpnLoginItems)
```
diegoreymendez, line 79: Just renaming since VPN is so much more readable.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNAppEventsHandler.swift
```diff
@@ -32,69 +32,79 @@ import Subscription
...
-    private let legacySystemExtensionBundleID = "com.duckduckgo.macos.browser.debug.network-protection-extension"
-#elseif REVIEW
-    private let legacyAgentBundleID = "HKE973VLUW.com.duckduckgo.macos.browser.network-protection.system-extension.agent.review"
-    private let legacySystemExtensionBundleID = "com.duckduckgo.macos.browser.review.network-protection-extension"
-#else
-    private let legacyAgentBundleID = "HKE973VLUW.com.duckduckgo.macos.browser.network-protection.system-extension.agent"
-    private let legacySystemExtensionBundleID = "com.duckduckgo.macos.browser.network-protection-extension"
-#endif // DEBUG || REVIEW || RELEASE
-#endif // NETP_SYSTEM_EXTENSION
+    typealias FeatureFlagOverridesPublisher = AnyPublisher<(FeatureFlag, Bool), Never>
```
diegoreymendez, line 35: The diff below is awful and it's easier to read this file in the final code.

The changes:
1. Adjustments to make the code testable.
2. Added handling of `applicationDidBecomeActive` where the login items can become disabled.
3. Added the possibility of disabling the VPN in `applicationDidFinishLaunching`.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/LoginItem+NetworkProtection.swift
```diff
@@ -34,7 +34,7 @@ extension LoginItem {
 }
 
 extension LoginItemsManager {
-    static var networkProtectionLoginItems: Set<LoginItem> {
+    static var vpnLoginItems: Set<LoginItem> {
```
samsymons, line 37: Nice 😄

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNAppEventsHandler.swift
```diff
@@ -32,69 +32,79 @@ import Subscription
...
+    /// Checks whether the VNP login items need to be disabled
+    ///
+    private func loginItemsControlCheckpoint(canRestart: Bool) {
+        Task { @MainActor [loginItemsManager] in
+            guard loginItemsManager.isAnyEnabled(LoginItemsManager.vpnLoginItems) else {
+
+                return
+            }
+
+            switch try? await featureGatekeeper.canStartVPN() {
```
samsymons, line 79: I set up a scenario where the login items are enabled, but I have no Privacy Pro account (which is plausible if something happens to the user's keychain, etc). When this happens, the `canStartVPN()` call throws an error instead of returning false, but because of the `try?` here it's actually the `.none` case that gets executed.

What do you think about running `disableLoginItem` even in the `none` case? What would happen is that the OS would throw `kSMErrorJobNotFound` for anyone who doesn't already have them running, but we aren't pixeling this logic so it wouldn't harm any metrics, and it would guarantee that the login item gets cleaned up in any case where there's no valid entitlement.

diegoreymendez: I pushed a fix for this - it's odd that `noAuthToken` is coming back as an error instead of `false`.

I did a quick fix, but we could follow up with a more proper fix.

samsymons (approved):
LGTM after the fix, my login item was correctly cleaned up. I also ran through a few happy path tests and checked that normal sign-in/sign-out worked as expected across launches/activations.

PR #916, opened 2025-06-02
Branded default user agent for web extensions

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1210054301344157?focus=true

### Description
Setting up a custom user agent for web extensions for them to be able to detect that they are operating in the DuckDuckGo browser environment.

### Testing Steps

1. Run the app and install user-agent-web-extension.zip
2. Open the dashboard by clicking on the icon and verify the user agent contains `Ddg` string
[user-agent-web-extension.zip](https://github.com/user-attachments/files/20548166/user-agent-web-extension.zip)

### Impact and Risks

#### What could go wrong?
Almost nothing, the web extensions feature is now hidden behind the compilation flag

---
###### Internal references:


macOS/DuckDuckGo/WebExtensions/WebExtensionManager.swift
```diff
@@ -106,7 +103,11 @@ final class WebExtensionManager: NSObject, WebExtensionManaging {
     }
 
     // Controller manages a set of loaded extension contexts
-    var controller: WKWebExtensionController
+    lazy var controller: WKWebExtensionController = {
+        let controllerConfiguration = WKWebExtensionController.Configuration.default()
+        controllerConfiguration.webViewConfiguration.applicationNameForUserAgent = UserAgent.brandedDefault
```
tomasstrba: `UserAgent.brandedDefault` needs to read information about the version of WebKit, so we have to delay initialization of the controller to the last possible point, which is `loadInstalledExtensions()` instead of the global init.

macOS/Configuration/App/DuckDuckGo.xcconfig
```diff
@@ -24,7 +24,7 @@ CODE_SIGN_ENTITLEMENTS[config=Debug][arch=*][sdk=*] = DuckDuckGo/DuckDuckGoDebug
 CODE_SIGN_IDENTITY[sdk=macosx*] = Developer ID Application
 CODE_SIGN_IDENTITY[config=Debug][sdk=macosx*] = Apple Development
 
-FEATURE_FLAGS = FEEDBACK SPARKLE STRIPE
+FEATURE_FLAGS = FEEDBACK SPARKLE STRIPE WEB_EXTENSIONS_ENABLED
```
tomasstrba, line 27: This will be rolled back before merge

.xcode-version
```diff
@@ -1 +1 @@
-16.2
+16.3
```
diegoreymendez, line 1: We should roll this change back and let the right version come from main.

macOS/DuckDuckGo/WebExtensions/WebExtensionManager.swift
```diff
@@ -123,7 +124,7 @@ final class WebExtensionManager: NSObject, WebExtensionManaging {
     }
 
     var hasInstalledExtensions: Bool {
-        controller.extensions.count > 0
+        pathsCache.cache.count > 0
```
diegoreymendez, line 127: I don't understand this change.  Are we counting path cache entries or installed extensions?

To me it sounds like the right source of truth should be controller.extensions, which is in essence the installed extensions.

The value might be the same, but semantically speaking this change looks wrong.

diegoreymendez (approved):
Works well, added some comments.  Nothing is blocking but I added a comment.

PR #922, opened 2025-06-02
Enable 60min session for internal users

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204167627774280/task/1210367091666879?focus=true
Tech Design URL:
CC:

### Testing Steps
1. Open the app as a regular user
2. Open duck.ai and send a prompt
3. Close duck.ai
4. Open duck.ai again, check that it’s a new conversation.
5. Enable internal user flag
6. Open duck.ai and send a prompt
7. Close duck.ai
8. Open duck.ai again, check if you can still see the previous conversation.

If you want to test the new feature flag
http://www.jsonblob.com/api/1379160777319178240

### Impact and Risks

#### What could go wrong?
Feature is enabled for users that are not on internal

### Quality Considerations
Make sure the duck assist button is working on both cases.
Search for bread recipe
Click the “Chat” button
it should auto submit the query


SharedPackages/AIChat/Sources/AIChat/iOS/PublicAPI/AIChatSessionTimer.swift
```diff
@@ -0,0 +1,80 @@
...
+    private var startDate: Date?
+
+    public init(durationInSeconds: TimeInterval) {
+        self.durationInSeconds = durationInSeconds
+    }
+
+    public func start(completion: @escaping () -> Void) {
+        cancel()
+        startDate = Date()
+        timer = Timer.scheduledTimer(withTimeInterval: durationInSeconds, repeats: false) { _ in
```
dus7: ⚠️ This is not reliable as Timer will be suspended while the app is in the background. I think best way is to store the start time and recalculate in case going back from background to make sure it's not past the interval.

iOS/DuckDuckGo/AIChat/AIChatViewControllerManager.swift
```diff
@@ -85,19 +111,46 @@ final class AIChatViewControllerManager {
...
         chatViewController = aiChatViewController
+        stopSessionTimer()
     }
 
     // MARK: - Private Helper Methods
 
+    private func startSessionTimer() {
+        guard isKeepSessionEnabled else { return }
+
+        sessionTimer = AIChatSessionTimer(durationInSeconds: TimeInterval(aiChatSettings.sessionTimerInMinutes * 60))
```
dus7: This approach implies the `AIChatViewControllerManager` will be present in the memory at all times, which is fine for now, but I think a better approach is to store the existing session start time and then validate if it's past the deadline before opening the chat. This will make it more resilient to future changes and will solve the issue raised in my previous comment.

iOS/DuckDuckGo/AIChat/AIChatViewControllerManager.swift
```diff
@@ -223,7 +277,11 @@ extension AIChatViewControllerManager: AIChatViewControllerDelegate {
 
 extension AIChatViewControllerManager: RoundedPageSheetContainerViewControllerDelegate {
     func roundedPageSheetContainerViewControllerDidDisappear(_ controller: RoundedPageSheetContainerViewController) {
-        cleanUpUserContent()
+        guard isKeepSessionEnabled == false else { return }
+
+        Task { @MainActor in
```
dus7: I think it's better to mark `cleanUpSession` or callees inside of it as `@MainActor` instead of doing this for the task closure.

dus7 (commented):
Works as expected. Tested the omni bar chat button, remote config and timeout setting. Two comments we need to agree on.

Bunn:
@dus7 I’ve done the changes we talked about yesterday.
Also, please test the https://duckduckgo.github.io/privacy-configuration/pr-3269/v4/ios-config.json config, this one enables the feature just for internal. 

I suggest testing the following:
* Brand new app, open duck.ai, close it, check if the session started. It shouldn’t 
* Add the new config, do the same as the step above, it should have the same behavior
* Change to internal, do the same as before, but this time it should trigger the session

PR #942, opened 2025-06-04
Initial set of cursor rules

afterxleep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204099484721401/task/1210460315061997
Tech Design URL:
CC:

### Description
This is an initial set of cursor rules that I've originally generated using AI, and have been maintaining over the past few months.  These should help the LLM generate better code, based on the existing codebases for macOS and iOS.

### Testing Steps
N/A

### Impact and Risks

#### What could go wrong?
N/A

### Quality Considerations
N/A

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/.cursor/rules/analytics-patterns.mdc
+++ b/.cursor/rules/analytics-patterns.mdc
@@ -1,0 +1,159 @@
+---
+description: Pixel analytics and event tracking patterns for DuckDuckGo browser
+globs:
+  - "**/*.swift"
+alwaysApply: false
+---
+
+# Analytics and Pixel Patterns
+
+## Structured Pixel Events
+Use the existing Pixel.fire pattern with structured event definitions:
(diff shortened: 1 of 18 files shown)


afterxleep:
I did my best to add rules for both.  There are some iOS exclusive and other macOS Exclusive, while the more general ones should be OK for both..  I have not used LLM too much on macOS, but in general I can confirm code is a lot better with these.

aataraxiaa:
Chatted with @afterxleep  on MM to clarity what I meant by "Are some of these rules specific to one platforms, and if so, will they cause issues when for the other platform?”, which was very unclear. As these are just rules for writing code they won’t break any Cursor tooling (e.g macOS builds failing). So, no risk here. Approved.

PR #988, opened 2025-06-09
Improve Feature Flags override menu

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203108348835387/task/1210496567641327?focus=true

### Description

Improve the Feature Flags override menu.

Changes:
- Show different icons for default-off, default-on, user-enabled and user-disabled.  Remove default checkmark.
- Require FeatureFlags to have a category, and populate this menu using those categories.
- Always show default value for a feature flag in a readable manner - probably just (ON) / (OFF).
- The category menu item will show a white flag if any of its feature flag has been overridden, so it's easier to find the ones you'd want to disable without going into all categories.

### Testing Steps

Just test the menu extensively.

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/InternalUserDecider/FeatureFlagOverridesMenu.swift
+++ b/macOS/DuckDuckGo/InternalUserDecider/FeatureFlagOverridesMenu.swift
@@ -34,8 +34,12 @@ final class FeatureFlagOverridesMenu: NSMenu {
             internalUserStateMenuItem()
             NSMenuItem.separator()
 
-            sectionHeader(title: "Feature Flags")
-            featureFlagMenuItems()
+            sectionHeader(title: "Categorized Feature Flags")
+            categotizedFeatureFlagMenuItems()
+            NSMenuItem.separator()
+
+            sectionHeader(title: "Uncategorized Feature Flags")
+            uncategorizedFeatureFlagMenuItems()
             NSMenuItem.separator()
 
             sectionHeader(title: "Experiments")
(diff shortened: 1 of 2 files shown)


diegoreymendez:
> Some categories feel artificial to me and could perhaps be rethinked. What I had in mind is:
> A category for actual features (released to users, or not yet released but ready for testing by internal users)
> A category for experimental and WIP features (not ready for testing yet, but included to help with development)
> A category for flags strictly related to development and debugging (e.g. tab crash debugging, auto update in debug, etc.)

What would this resolve?

The downsides I see in the suggestion:
- It'd force us to align on what those categories are (I don't think they're intuitive enough),
- It'd force us to keep moving flags around, which I'd like to avoid
- It'd limit us to having just 3 categories, and I expect one of those will end up being like our current uncategorized menu.

The categorization is meant to provide some help finding flags and referencing them in written communication (ie: "the flag is in Feature Flags > VPN > someVPNFlag"), which I think is best resolved by a more lose system.

For clarity: I don't think my categories are good, and the system is lose enough that we can adjust as we go.  I'd like to favor simplicity and letting the team grow this organically.

diegoreymendez:
My point is actually rather simple: static categories are enough and don't require any additional process.

> Was your plan to finalize the categories before merging it?

Not at all.  This was very clear in my previous message: I offered a starting point that's not meant to be final.

> Or does it mean you expect people to come back here later and move flags around (which you wrote you would like to avoid).

I've nothing against people moving flags if they want to, that's the whole point.  I'd like to avoid adding process to force people to move flags around - or at least leave it out of scope from this specific step.

> I’m fine with whatever direction we take here, but my feedback that I’d like to see addressed is:
>
> Rename User Interface / User Experience because VPN, Sync and a number of Miscellaneous/Other flags are also related to UI/UX.
> Remove Tab Crash category – tabCrashRecovery is a user-facing feature, and tabCrashDebugging is a developer-only flag.

I'll be frank: I don't think trying to define the categories now has any value, and is blocking a great quality-of-life improvement.  So for the sake of delivering that value I'll move those flags back to "others".

PR #993, opened 2025-06-09
Add a workaround for linking WebKit in the iOS simulator

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1210505420118693?focus=true
Tech Design URL:
CC:

### Description

Apply a workaround for the iOS 18.4 and 18.5 simulator linking issue.

For convenience, I've attached a copy of the script at the bottom of the description.

### Testing Steps

1. Check that the simulator builds and runs for 18.3 (or 18.3.1)
2. Now do a clean build and try the same for 18.4
3. Finally, do another clean build and test 18.5 

#### What could go wrong?

In theory this change is safe since it only affects the simulator, and only two versions of iOS. Worst case the build will fail, but that's already the case. :)

### Notes to Reviewer

---

```
#!/usr/bin/env bash
###############################################################################
#
# Several simulator runtimes (18.4, 18.5, 26.0) ship without the Swift
# overlay libswiftWebKit.dylib. This script copies the missing dylib from the
# runtime into $BUILT_PRODUCTS_DIR so the app can launch under those sims.
# Delete the script once Apple ships corrected runtimes.
###############################################################################
set -euo pipefail

readonly DYLIB_NAME="libswiftWebKit.dylib"

###############################################################################
# 1. Guard – run only for *iphonesimulator* builds
###############################################################################

if [[ "${PLATFORM_NAME:-}" != "iphonesimulator" ]]; then
    echo "ℹ️  Skipping $DYLIB_NAME workaround – PLATFORM_NAME='${PLATFORM_NAME:-unset}'"
    exit 0
fi

###############################################################################
# 2. Validate TARGET_DEVICE_OS_VERSION is available
###############################################################################

if [[ -z "${TARGET_DEVICE_OS_VERSION:-}" ]]; then
    echo "❌ TARGET_DEVICE_OS_VERSION environment variable is not set" >&2
    exit 1
fi

echo "ℹ️  Target device OS version: $TARGET_DEVICE_OS_VERSION"

###############################################################################
# 3. Configuration for supported versions (Bash 3.x compatible)
###############################################################################

get_build_id_for_version() {
    case "$1" in
        18.4*) echo "22E238" ;;
        18.5*) echo "22F77" ;;
        *) return 1 ;;
    esac
}

get_runtime_dylib_path() {
    local version="$1"
    local build_id
    
    build_id="$(get_build_id_for_version "$version")" || return 1
    
    echo "/Library/Developer/CoreSimulator/Volumes/iOS_${build_id}/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS ${version}.simruntime/Contents/Resources/RuntimeRoot/System/Cryptexes/OS/usr/lib/swift/$DYLIB_NAME"
}

###############################################################################
# 4. Check if workaround is needed and find the dylib
###############################################################################

find_matching_runtime() {
    local target_version="$1"
    local runtime_path
    
    runtime_path="$(get_runtime_dylib_path "$target_version")" || {
        return 1
    }
    
    if [[ -f "$runtime_path" ]]; then
        echo "$runtime_path"
        return 0
    else
        echo "⚠️  Runtime dylib not found at: $runtime_path" >&2
        echo "⚠️  Make sure iOS ${target_version%.*} simulator runtime is installed" >&2
        return 1
    fi
}

RUNTIME_DYLIB="$(find_matching_runtime "$TARGET_DEVICE_OS_VERSION")" || {
    echo "✅ Runtime iOS $TARGET_DEVICE_OS_VERSION doesn't require $DYLIB_NAME workaround"
    exit 0
}

###############################################################################
# 5. Validate environment and paths
###############################################################################

if [[ -z "${BUILT_PRODUCTS_DIR:-}" ]]; then
    echo "❌ BUILT_PRODUCTS_DIR environment variable is not set" >&2
    exit 1
fi

if [[ ! -d "$BUILT_PRODUCTS_DIR" ]]; then
    echo "❌ Build products directory does not exist: $BUILT_PRODUCTS_DIR" >&2
    exit 1
fi

###############################################################################
# 6. Copy the overlay into the build products directory
###############################################################################

DESTINATION="$BUILT_PRODUCTS_DIR/$DYLIB_NAME"

# Skip if files are identical
if [[ -f "$DESTINATION" ]] && cmp -s "$RUNTIME_DYLIB" "$DESTINATION"; then
    echo "ℹ️  $DYLIB_NAME already exists and is identical"
    exit 0
fi

echo "🏗️  Copying $RUNTIME_DYLIB → $DESTINATION"

if /bin/cp -f "$RUNTIME_DYLIB" "$DESTINATION" && /bin/chmod 644 "$DESTINATION"; then
    echo "✅ Successfully applied $DYLIB_NAME workaround for iOS $TARGET_DEVICE_OS_VERSION"
else
    echo "❌ Failed to copy $DYLIB_NAME" >&2
    exit 1
fi
```


iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
```diff
@@ -9122,6 +9127,24 @@
...
...######################\n# 5. Validate environment and paths\n###############################################################################\n\nif [[ -z \"${BUILT_PRODUCTS_DIR:-}\" ]]; then\n    echo \"❌ BUILT_PRODUCTS_DIR environment variable is not set\" >&2\n    exit 1\nfi\n\nif [[ ! -d \"$BUILT_PRODUCTS_DIR\" ]]; then\n    echo \"❌ Build products directory does not exist: $BUILT_PRODUCTS_DIR\" >&2\n    exit 1\nfi\n\n###############################################################################\n# 6. Copy the overlay into the build products directory\n###############################################################################\n\nDESTINATION=\"$BUILT_PRODUCTS_DIR/$DYLIB_NAME\"\n\n# Skip if files are identical\nif [[ -f \"$DESTINATION\" ]] && cmp -s \"$RUNTIME_DYLIB\" \"$DESTINATION\"; then\n    echo \"ℹ️  $DYLIB_NAME already exists and is identical\"\n    exit 0\nfi\n\necho \"🏗️  Copying $RUNTIME_DYLIB → $DESTINATION\"\n\nif /bin/cp -f \"$RUNTIME_DYLIB\" \"$DESTINATION\" && /bin/chmod 644 \"$DESTINATION\"; then\n    echo \"✅ Successfully applied $DYLIB_NAME workaround for iOS $TARGET_DEVICE_OS_VERSION\"\nelse\n    echo \"❌ Failed to copy $DYLIB_NAME\" >&2\n    exit 1\nfi\n";
```
ayoy, line 9146: To improve maintainability we can put the script in scripts directory and call it from here. We do that with assert_xcode_version.sh already. Also not sure if we need that script for previews builds.

Screenshot 2025-06-10 at 08 03 00

dus7:
I used this to test running on iOS 26. It does help with the failure, however there's a bunch of warnings about duplicate symbols like so: (also for 18.4)
```
objc[13587]: Class _WKAuthenticationExtensionsClientInputs is implemented in both 
<redacted>/iOS 18.4.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/WebKit.framework/WebKit (0x1f4118778) and 
<redacted>/DerivedData/DuckDuckGo-aqksgbjbkphzrjakkapmazxpmkig/Build/Products/Debug-iphonesimulator/libswiftWebKit.dylib (0x112789448). 
This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
```

samsymons:
@dus7 Thanks, I can reproduce the same issue. Since this is simulator only and they currently don't work at all, I'll merge it as-is and leave the iOS 26 fixing for someone else, since I'm still technically AFK for a while yet. :)

dus7:
I pushed a change allowing to "run" on iOS 26 beta 1. It crashes when trying to create an instance of shared AppDependencyProvider, but that's something we'll need to sort out outside of this.

PR #1027, opened 2025-06-12
iOS VPN: restore shutdown logic for expired subscriptions

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203108348835387/task/1210530444530224?focus=true

### Description

Fixes an issue that was causing the VPN not to stop in iOS for expired subscriptions.

### Testing Steps

1. Start the VPN with a valid subscription
2. Expire the subscription
3. Force a rekey even and make sure the VPN is disabled fully

### Impact and Risks

Medium

#### What could go wrong?

If the logic is off this might turn the VPN off for users with a valid subscription.

### Quality Considerations

NA

### Notes to Reviewer

NA

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/NetworkProtection/Diagnostics/NetworkProtectionError.swift
```diff
@@ -60,7 +60,7 @@ public enum NetworkProtectionError: LocalizedError, CustomNSError {
     case setWireguardConfig(Error)
 
     // Auth errors
-    case noAuthTokenFound
+    case noAuthTokenFound(Error?)
```
diegoreymendez: Adding underlying error information where available, as this is useful to have in pixels.

SharedPackages/BrowserServicesKit/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -102,8 +102,12 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
...
     ///
     public func refreshServerList() async throws -> [NetworkProtectionServer] {
-        guard let token = try? await VPNAuthTokenBuilder.getVPNAuthToken(from: tokenHandler) else {
-            throw NetworkProtectionError.noAuthTokenFound
+        let token: String
+
+        do {
+            token = try await VPNAuthTokenBuilder.getVPNAuthToken(from: tokenHandler)
+        } catch {
+            throw NetworkProtectionError.noAuthTokenFound(error)
```
diegoreymendez, line 110: Does the same as before, with the difference that we now add underlying error information so it shows up in pixels.

SharedPackages/BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -105,7 +105,8 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
         case tokenReset
 
         // Subscription Errors - 100+
-        case vpnAccessRevoked
+        case vpnAccessRevoked(_ underlyingError: Error?)
```
diegoreymendez: Adding underlying error info where available.

SharedPackages/BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1049,11 +1052,14 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                 dnsSettings: dnsSettings,
                 regenerateKey: regenerateKey
             )
-        } catch {
-            if let error = error as? NetworkProtectionError, case .vpnAccessRevoked = error {
-                throw TunnelError.vpnAccessRevoked
+        } catch let error as NetworkProtectionError {
+            switch error {
+            case .vpnAccessRevoked, .noAuthTokenFound:
```
diegoreymendez, line 1057: The diff looks a bit odd, but the actual change is here.  We now consider `noAuthTokenFound` to be indicative of an a expired subscription for the end of disabling the VPN.

Only edge case to evaluate is if this can turn off the VPN for valid users if the keychain is inaccessible.

SharedPackages/BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1366,22 +1372,26 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     @available(iOS 17, *)
     @MainActor
-    public func handleShutDown() async throws {
+    public func handleShutDown(dueTo error: Error) async throws {
+        enum ShutDownError: Error {
+            case noManagerFound
+        }
```
diegoreymendez, line 1378: Not used anywhere else so I declared this within this method.  This is only useful for pixel information at this point and is not intended to be managed.

SharedPackages/BrowserServicesKit/Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: I'm not sure this is correct, the SubscriptionManagerV2 `getTokenContainer(policy: ...)` throws an error even when the token is correctly unavailable.
`throw SubscriptionManagerError.tokenUnavailable(error: OAuthClientError.missingTokenContainer)`
We should distinguish the errors here and disconnect only in case of `SubscriptionManagerError.tokenUnavailable`

federicocappelli: In general, the business logic here doesn't take into account that a "get token" request can now fail for many reasons like network and keychain access

SharedPackages/BrowserServicesKit/Sources/NetworkProtection/Repositories/NetworkProtectionLocationListRepository.swift
```diff
@@ -88,9 +88,14 @@ final public class NetworkProtectionLocationListCompositeRepository: NetworkProt
...
     func fetchLocationListFromRemote() async throws -> [NetworkProtectionLocation] {
         do {
-            guard let authToken = try? await VPNAuthTokenBuilder.getVPNAuthToken(from: tokenHandler) else {
-                throw NetworkProtectionError.noAuthTokenFound
+            let authToken: String
+
+            do {
+                authToken = try await VPNAuthTokenBuilder.getVPNAuthToken(from: tokenHandler)
+            } catch {
+                throw NetworkProtectionError.noAuthTokenFound(error)
```
federicocappelli, line 96: same here, we may need to check the error here and throw `NetworkProtectionError.noAuthTokenFound` only when the token is effectively missing

diegoreymendez: Can you provide me with the precise errors that the subscription lib will throw that should result in  assuming the subscription expired?

federicocappelli: There's no concept of "expired subscription" when retrieving the token but the error you can intercept is `SubscriptionManagerError.tokenUnavailable`, any other error is something like a networking or keychain failure.

federicocappelli: p.s the documentation is available:
```
    /// Get a token container accordingly to the policy
    /// - Parameter policy: The policy that will be used to get the token, it effects the tokens source and validity
    /// - Returns: The TokenContainer
    /// - Throws: A `SubscriptionManagerError`.
    ///     `tokenRefreshFailed` if the token cannot be refreshed, typically due to an expired refresh token.
    ///     `tokenUnavailable` if the token is not available for the reason specified by the underlying error.
    @discardableResult
    func getTokenContainer(policy: AuthTokensCachePolicy) async throws -> TokenContainer
```

The problem is that you are using that function after several abstraction levels and protocols :/

PR #1048, opened 2025-06-13
macOS Duck.ai Sidebar: 4. Support for state restoration, sidebar on pinned tabs and moved between windows, placement param  

miasma13:
Task/Issue URL:
Tech Design URL:
CC:

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


macOS/DuckDuckGo/MainWindow/MainViewController.swift
```diff
@@ -805,7 +807,7 @@ extension MainViewController: BrowserTabViewControllerDelegate {
     )
     bkman.loadBookmarks()
 
-    let vc = MainViewController(bookmarkManager: bkman, autofillPopoverPresenter: DefaultAutofillPopoverPresenter())
+    let vc = MainViewController(bookmarkManager: bkman, autofillPopoverPresenter: DefaultAutofillPopoverPresenter(), aiChatSidebarProvider: AIChatSidebarProvider())
```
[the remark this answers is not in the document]
miasma13: This is only for preview.

macOS/DuckDuckGo/Tab/AIChatSidebar/AIChatSidebarPresenter.swift
```diff
@@ -153,18 +162,22 @@ extension AIChatSidebarPresenter: AIChatSidebarHostingDelegate {
         updateSidebarConstraints(for: tabID, isShowingSidebar: isShowingSidebar, withAnimation: false)
     }
 
-    func sidebarHostDidUpdateTabs(_ currentTabIDs: [TabIdentifier]) {
+    func sidebarHostDidUpdateTabs() {
         guard featureFlagger.isFeatureOn(.aiChatSidebar) else { return }
 
-        sidebarProvider.cleanUp(for: currentTabIDs)
+        Task { @MainActor in
```
ayoy: I assume these `Task`s are here to ensure that AppDelegate (now removed by me and replaced with injected windowControllersManager) was accessed on main actor? And not to achieve async dispatch?

You could mark both `AIChatSidebarHostingDelegate` and `AIChatSidebarViewControllerDelegate` Main Actor protocols, which is a sane requirement, and then you wouldn’t need to use Tasks here (and below).

macOS/DuckDuckGo/Tab/AIChatSidebar/AIChatSidebarProvider.swift
```diff
@@ -51,36 +61,45 @@ final class AIChatSidebarProvider: AIChatSidebarProviding {
...
     }
 
-    private var sidebarsByTabIDs: [TabIdentifier: AIChatSidebar] = [:]
-
     var sidebarWidth: CGFloat { Constants.sidebarWidth }
 
+    private(set) var model: AIChatSidebarProviderModel
+
+    init(currentState: AIChatSidebarProviderModel? = nil) {
+        self.model = currentState ?? [:]
```
ayoy: My personal feeling is that `model` is a broad term and any time we are able to replace it with something more specific, we should. Here it’s a typedef for a dictionary/map. I’d consider renaming it e.g. as `tabSidebars` and `WindowManagerStateRestoration.aiChatSidebars` and keep model for something bigger when we need it.

macOS/DuckDuckGo/Tab/AIChatSidebar/AIChatSidebarViewController.swift
```diff
@@ -231,8 +237,17 @@ final class AIChatSidebarViewController: NSViewController {
         webViewContainer.layer?.mask = shape
     }
 
+    private func subscribeToURLChanges() {
+        aiTab.$content.sink { [weak self] content in
```
ayoy: This will call the closure on subscription, with the current value of `content` – this may be expected, but if you only want to listen to changes (without performing action on the initial value), then use `$content.dropFirst()`.

ayoy (commented):
@miasma13 this is working as expected, but I couldn’t test `chatID`-dependent changes. I assume that part has not yet been released on the FE?

I’m leaving a few comments that would be great to get clarity on before we merge it. Are you going to provide unit tests in a separate PR?

macOS/DuckDuckGo/Tab/AIChatSidebar/AIChatSidebarPresenter.swift
miasma13: Precisely, that was the reason. I will update them with MainActor annotation to remove these Tasks.

PR #1076, opened 2025-06-18
macOS: Convert Freemium PIR from Experiment to Full Feature

aataraxiaa:
Task/Issue URL:
Tech Design URL:
CC:

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


aataraxiaa:
bugbot run

macOS/DuckDuckGo/Freemium/DBP/FreemiumDBPFeature.swift
```diff
@@ -134,55 +146,142 @@ final class DefaultFreemiumDBPFeature: FreemiumDBPFeature {
...
+                .sink { [weak self] canPurchase in
+                    guard let self = self else { return }
+
+                    // Calculate availability manually for App Store to include purchase capability
+                    // This provides more granular control over the availability calculation
+                    // compared to the general isAvailable property
+                    let featureEnabled = isEligibleFeatureFlagEnabled
+                    let notCurrentUser = isEligibleNotCurrentUser
+                    let storeFrontIsUSA = isEligibleAppStorefrontIsUSA
+                    let available = featureEnabled && notCurrentUser && storeFrontIsUSA && canPurchase
```
miasma13: This looks like a duplication of logic in `var isEligible: Bool {`

macOS/DuckDuckGo/Freemium/DBP/FreemiumDBPFeature.swift
```diff
@@ -134,55 +146,142 @@ final class DefaultFreemiumDBPFeature: FreemiumDBPFeature {
...
 
+/// Private extension containing eligibility logic and helper methods.
 private extension DefaultFreemiumDBPFeature {
 
-    /// Returns true IFF:
+    // MARK: - Freemium Eligibility Logic
+
+    /// Determines overall eligibility by combining feature flag, auth status, storefront, and purchase capability.
+    var isEligible: Bool {
+        isEligibleFeatureFlagEnabled && isEligibleNotCurrentUser && isEligibleAppStorefrontIsUSA && isEligibleCanPurchase
```
miasma13: Naming of the main var and the compound ones is not clear to me. I suggest:
- keep the main one as `isEligible` or make if fully descriptive `isEligibleForFreemium...`
- the subsequent condition vars have their name starting with `isEligible...` but to me they do not guarantee eligibility, maybe to be clearer about that we can rename them with a pattern like `isEligibleFeatureFlagEnabled` -> `canBeEligibleDueToFeatureFlagEnabled` or maybe drop the whole `isEligible` prefix and be clear about what condition is checked e.g. `isFeatureFlagEnabled`, WDYT?

miasma13 (approved):
Tested, all worked as expected. Just minor comments, other than that LGTM

macOS/DuckDuckGo/Freemium/DBP/FreemiumDBPFeature.swift
aataraxiaa: Thanks @miasma13  - This is intentional implemented separately to `isEligible`, the reason is `isEligible` relies on the `canPurchase` property. However, we need to rely on the observed property value as at the time of observation notification the `canPurchase` property does not return the correct value. It seems this is due to some quirks of combine, where the observer is notified of the change, and passed the updated value, before the actual property value is updated.

macOS/DuckDuckGo/Freemium/DBP/FreemiumDBPFeature.swift
aataraxiaa: Thanks @miasma13, agree. Naming updated.

PR #1084, opened 2025-06-19
New subscription expired pixel

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200019156869587/task/1210250285874364?focus=true
CC: @diegoreymendez 

### Description

iOS and macOS now detect subscription expirations and send a new pixel.

### Testing Steps

- run the app from xcode
- Subscribe using a sandbox account
- Wait expiration
- check the logs if the pixel was fired

### Impact and Risks

None, just a pixel

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
+++ b/iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
@@ -85,12 +85,15 @@ public class AuthV2PixelHandler: SubscriptionPixelHandler {
             enum State: String {
                 case added
                 case removed
+                case changed
             }
 
-            let state: State // Assumption: The array of entitlements is always complete or empty, we don't have known cases of partial set of entitlements
+            let state: State
             switch (self.previousEntitlements.isEmpty, entitlements.isEmpty) {
                 case (true, false): state = .added
                 case (false, true): state = .removed
+            case (false, false) where self.previousEntitlements.count != entitlements.count:
+                state = .changed
             default:
                 Logger.subscription.fault("Unexpected state")
                 return
--- a/macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
+++ b/macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
@@ -79,12 +79,15 @@ public class AuthV2PixelHandler: SubscriptionPixelHandler {
             enum State: String {
                 case added
                 case removed
+                case changed
             }
 
-            let state: State // Assumption: The array of entitlements is always complete or empty, we don't have known cases of partial set of entitlements
+            let state: State
             switch (self.previousEntitlements.isEmpty, entitlements.isEmpty) {
             case (true, false): state = .added
             case (false, true): state = .removed
+            case (false, false) where self.previousEntitlements.count != entitlements.count:
+                state = .changed
             default:
                 Logger.subscription.fault("Unexpected state")
                 return


SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -316,6 +319,13 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
         if subscription.isActive {
             pixelHandler.handle(pixelType: .subscriptionIsActive)
         }
+
+        if let cachedSubscription,
+           cachedSubscription.isActive && !subscription.isActive {
+            Logger.subscription.log("Subscription expired")
+            pixelHandler.handle(pixelType: .subscriptionExpired)
+        }
+
```
diegoreymendez: The pixel won't fire if the cached subscription is missing or if any error is thrown above this point.  Additionally I'm not sure we can rely on `cachedSubscription` being different here as that's a detail that should be invisible to this method.

Can we instead:
- Keep this pixel but make it daily only (the count volume will be HUGE), and fire it also on `SubscriptionEndpointServiceError.noData`.
- Fire a pixel when entitlements change: probably around [here](https://github.com/duckduckgo/apple-browsers/blob/849b74d09e4b199198cca9f51171241bc5927cf9/SharedPackages/BrowserServicesKit/Sources/Subscription/API/SubscriptionEndpointServiceV2.swift#L179-L184) and [here](https://github.com/duckduckgo/apple-browsers/blob/849b74d09e4b199198cca9f51171241bc5927cf9/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift#L418-L424).
- Fire a pixel when the subscription changes, should be similar to the previous point.
- Consider adding parameters that might help us debug these pixels, if any.  For entitlements I'd include old and new entitlements.

federicocappelli: The only moment we need to fire the pixel is when the subscription changes from subscribed to unsubscribed, so not firing the pixel when the cached subscription is not there is intentional as not firing it when anything goes wrong above it (so we don't have a subscription to compare)

- The pixel is now fired only daily
- There's a misunderstanding here, `SubscriptionEndpointServiceError.noData` doesn't mean that the subscription is expired. This means that the subscription is not there (never been or is expired from more than a month and so removed from the backend)
- The entitlements can change for other reasons, the most reliable way to decide if a subscription is expired is to check `.isActive` and the only way to know if it's expired is to compare to what the app has seen before.

@miasma13 can you double-check my assumptions on detecting a Subscription expiration is correct?

diegoreymendez: > The only moment we need to fire the pixel is when the subscription changes from subscribed to unsubscribed, so not firing the pixel when the cached subscription is not there is intentional as not firing it when anything goes wrong above it (so we don't have a subscription to compare)

I'd like to understand what the subscription library is communicating to the other code that depends on it.

Any response from the library other than "subscription OK and valid" should be captured in this pixel - so any error returned or subscription that's otherwise expired or absent.

> The entitlements can change for other reasons, the most reliable way to decide if a subscription is expired is to check .isActive and the only way to know if it's expired is to compare to what the app has seen before.

These notifications need to be reliable as there's code in `PreferencesSidebarModel`, `NetworkProtectionSubscriptionEventHandler`, `DataBrokerProtectionSubscriptionEventHandler`, `MainViewController.swift` that relies on them.

But I don't think we need to fix anything now - in fact I think we shouldn't fix anything at all.  I'm interested in getting the signal as it is right now.

We just need to fire a pixel for `entitlementsDidChange` and `subscriptionDidChange` so we can understand how badly these are behaving and how urgent it is to fix them.

federicocappelli: > Any response from the library other than "subscription OK and valid" should be captured in this pixel - so any error returned or subscription that's otherwise expired or absent.

Ok, then this is a slightly different requirement, I'll work on this.

> These notifications need to be reliable

let me clarify, the notifications are reliable but `no entitlements` != `subscription expired`. From a feature (VPN, PIR etc) point of view is ok to get disabled when the entitlement disappears, but that doesn't necessarily mean that the subscription has expired.

> We just need to fire a pixel for entitlementsDidChange and subscriptionDidChange so we can understand how badly these are behaving and how urgent it is to fix them.

This is different from "a pixel for tracking subscription expiration", I'll think about how to implement this.

diegoreymendez: It's ok, what I'm trying to get at is that we know from backend how many users have an expired or invalid subscription.  I'm ok here to change the pixel we'll fire from "subscription expired" to something else that works - but what I'd like to capture is the Subs library informing external clients (UI, VPN, DBP, etc) that entitlements have changed and are no longer available, or that the subscription has changed and is no longer available - ie: I'm not interested in coming up with a perfect calculation just for the pixel, but want the pixel to tell us what the subs library is reporting through it's API.

The end goal is being able to look at this and quickly know if we have a bug in our subs library logic right away, and ideally be able to understand where.

miasma13: @federicocappelli I think that the logic you tried to follow here is correct although it will fail in terms of how the subscription cache works - for production the cache validity is set to either subscription's expiry/renewal date (if date is still in the future) or to 20 min if the subscription is expired. When trying to retrieve an expired object from cache we receive nil instead.

Let's imagine this code triggers on first subscription refetch after expiration when the cache still holds the snapshot of subscription when it was active. Attempt to fetch `cachedSubscription` will get you `nil` (since the validity date was set to `expiresOrRenewsAt` which already expired) hence your check for the pixel won't succeed.

iOS/Core/PixelEvent.swift
```diff
@@ -973,6 +973,7 @@ extension Pixel {
         case privacyProAuthV2MigrationFailed
         case privacyProAuthV2MigrationSucceeded
         case privacyProAuthV2GetTokensError
+        case privacyProSubscriptionExpired
```
miasma13, line 976: WDYT about also firing pixel for `.entitlementsDidChange` notification? We could also consider sending one where the entitlements list is in params and what is even more interesting maybe a separate pixel for this notification when entitlements are empty. I would see this as a proxy for the expired subscription scenario but potentially also a signal of something not being right if it differs from the expired subscription pixel.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -63,6 +63,7 @@ public enum SubscriptionPixelType {
     case migrationSucceeded
     case migrationFailed(Error)
     case subscriptionIsActive
+    case subscriptionExpired
```
miasma13: Also this pixel is only added to V2, I think until we are 100% released we should probably add it to both. If we think it would be useful we can optionally add auth version as a param, or maybe as a separate pixel.

miasma13:
Do we have a Privacy Triage task for the new pixel?

iOS/Core/PixelEvent.swift
federicocappelli: agree, I'm working on exactly this

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
diegoreymendez: Honest question: what value do we expect from this specific new logic?

Personally, I'd like to know what the subscription library is _**already**_ returning to clients / notifying to clients, which means looking exclusively at thrown errors, return values and notifications.

My impression is adding new code to do an entirely new disconnected calculation of subscription status will not tell us much that we can use to spot actual issues.  It'll only tell us this newly added logic works or doesn't work.

federicocappelli:
@diegoreymendez  @miasma13 
With the new requirements I've changed approach: I now track the notifications sent by the Subscriptin framework and send pixels if the subscription is expired or the entitlements have changed.

Issue: If the user tries to restore an expired subscription the subscription expired pixel is fired, and I have no way to stop it atm without implementi some more complicated logic.

For now done only on AuthV2 when approved I'll do them for authv1 as well.

iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -37,16 +41,47 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+    private var sourceParam: [String: String] {
+        [Defaults.sourceKey: source.description]
+    }
+    private let notificationCenter: NotificationCenter = NotificationCenter.default
+    private var cancellables = Set<AnyCancellable>()
+
+    init(source: Source) {
+        self.source = source
+
+        notificationCenter.publisher(for: .subscriptionDidChange).sink { param in
```
federicocappelli: main changes here for iOS, the pixel handler is subscribed to the notifications and sends the pixels

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -43,8 +46,34 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
 
     let source: Source
+    private let notificationCenter: NotificationCenter = NotificationCenter.default
+    private var cancellables = Set<AnyCancellable>()
 
-    public func handle(pixelType: Subscription.SubscriptionPixelType) {
+    init(source: Source) {
+        self.source = source
+
+        notificationCenter.publisher(for: .subscriptionDidChange).sink { param in
```
federicocappelli, line 56: main changes here for macOS, the pixel handler is subscribed to the notifications and sends the pixels

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -713,6 +713,7 @@ final class DefaultSubscriptionPagesUseSubscriptionFeatureV2: SubscriptionPagesU
 
         do {
             try await subscriptionManager.adopt(accessToken: subscriptionValues.accessToken, refreshToken: subscriptionValues.refreshToken)
+            try await subscriptionManager.getSubscription(cachePolicy: .remoteFirst)
```
federicocappelli: this now mirrors macOS 
The UI finds now the subscription already cached when appears and the UI feels faster.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -43,8 +46,34 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+
+            guard let userInfo = param.userInfo as? [AnyHashable: PrivacyProSubscription],
+                  let subscription = userInfo[UserDefaultsCacheKey.subscription] else {
+                return
```
diegoreymendez, line 61: What does this case mean to the observer?  Is it an abnormal scenario?  Is the subscription valid or invalid?

iOS/DuckDuckGo/en.lproj/Localizable.strings
```diff
@@ -1,3493 +1 @@
...
扯潫浡牫猠慮搠⽯爠灡獳睯牤猩‪⼊≺楰⹣潮瑥湴⹩浰潲琮瑹灥献瑩瑬攢‽•坨慴⁤漠祯甠睡湴⁴漠業灯牴㼢㬊
```
miasma13: wtf?

federicocappelli: lovely... some localisation script failure maybe? I'll revert it

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
diegoreymendez: Is this a bug fix?  Can you expand on what's being fixed here and whether it's urgent to merge?

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: That means there's no subscription in the userInfo. 
Search `NotificationCenter.default.post(name: .subscriptionDidChange` and you'll see that for AuthV2 the case is not possible.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
federicocappelli: (I posted a comment for this) Just a simple pre-caching; it mirrors macOS behaviour in the same circumstance. With that, the UI seems more reactive and saves some UI delays due to slow connection.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -43,8 +46,34 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+            if !subscription.isActive {
+                PixelKit.fire(PrivacyProPixel.privacyProSubscriptionExpired(source), frequency: .daily)
+            }
+        }.store(in: &cancellables)
+
+        notificationCenter.publisher(for: .entitlementsDidChange).sink { param in
+            let userInfo = param.userInfo as? [AnyHashable: [Entitlement]]
+            let entitlements = userInfo?[UserDefaultsCacheKey.subscriptionEntitlements] ?? []
+            let entitlementsDescriptions = entitlements.map(\.product.rawValue).sorted().joined(separator: ", ")
+            PixelKit.fire(PrivacyProPixel.privacyProEntitlementsDidChange(source, entitlementsDescriptions), frequency: .dailyAndCount)
```
diegoreymendez: The issue with this approach is the pixel won't tell us much on its own.  We'll get a large number of entitlement changes that are hard to get value from.

DISCLAIMER: **The below code is AI generated and untested!!**

The code below helps explain the direction I think would work best for this:
- Compare previous entitlements with current
- Fire two different pixels reporting removed and added entitlements.

Note: Besides removed entitlements, knowing about added entitlements would be valuable as well.

```swift
extension NotificationCenter.Publisher {
    // Just a convenience publisher to scan previous value... this can actually be moved inline if preferred.
    func trackPreviousEntitlements() -> AnyPublisher<([Entitlement], [Entitlement]?), Never> {
        scan((current: [Entitlement](), previous: [Entitlement]?.none)) { state, notification in
            let userInfo = notification.userInfo as? [AnyHashable: [Entitlement]]
            let currentEntitlements = userInfo?[UserDefaultsCacheKey.subscriptionEntitlements] ?? []
            return (current: currentEntitlements, previous: state.current)
        }
        .eraseToAnyPublisher()
    }
}

let cancellable = NotificationCenter.default
    .publisher(for: .entitlementsDidChange)
    .trackPreviousEntitlements()
    .sink { current, previous in
        let currentEntitlements = current
        let previousEntitlements = previous ?? []
        
        // Get the product names, excluding .unknown
        let currentProducts = Set(currentEntitlements.map(\.product).filter { $0 != .unknown })
        let previousProducts = Set(previousEntitlements.map(\.product).filter { $0 != .unknown })
        
        // Find products that were in previous but not in current (removed)
        let removedProducts = previousProducts.subtracting(currentProducts)
        
        // Find products that are in current but were not in previous (added)
        let addedProducts = currentProducts.subtracting(previousProducts)
        
        // Fire pixel for removed products
        if !removedProducts.isEmpty {
            let removedProductsDescription = removedProducts.map(\.rawValue).sorted().joined(separator: ", ")
            let entitlementsDescriptions = currentEntitlements.map(\.product.rawValue).sorted().joined(separator: ", ")
            PixelKit.fire(
                PrivacyProPixel.privacyProEntitlementsRemoved(removedProductsDescription, entitlementsDescriptions),
                frequency: .dailyAndCount
            )
        }
        
        // Fire pixel for added products
        if !addedProducts.isEmpty {
            let addedProductsDescription = addedProducts.map(\.rawValue).sorted().joined(separator: ", ")
            let entitlementsDescriptions = currentEntitlements.map(\.product.rawValue).sorted().joined(separator: ", ")
            PixelKit.fire(
                PrivacyProPixel.privacyProEntitlementsAdded(addedProductsDescription, entitlementsDescriptions),
                frequency: .dailyAndCount
            )
        }
    }
```

iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -37,16 +41,47 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+
+        notificationCenter.publisher(for: .entitlementsDidChange).sink { param in
+            let userInfo = param.userInfo as? [AnyHashable: [Entitlement]]
+            let entitlements = userInfo?[UserDefaultsCacheKey.subscriptionEntitlements] ?? []
+            let entitlementsDescriptions = entitlements.map(\.product.rawValue).sorted().joined(separator: ", ")
+            let params = [Defaults.sourceKey: source.description,
+                          Defaults.entitlementsKey: entitlementsDescriptions] as? [String: String]
+            DailyPixel.fireDailyAndCount(pixel: .privacyProEntitlementsDidChange, withAdditionalParameters: params ?? [:])
+
+        }.store(in: &cancellables)
```
diegoreymendez: Same as in macOS.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
diegoreymendez: While this may be currently true, it may not be in the future if a bug was introduced and the dictionary was missing.

I'm ok with using a different name like `privacyProSubscriptionMissing` if that works best, but we need to cover this exit point too.

diegoreymendez (commented):
Added a few comments.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -43,8 +46,34 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+                  let subscription = userInfo[UserDefaultsCacheKey.subscription] else {
+                return
+            }
+
+            if !subscription.isActive {
+                PixelKit.fire(PrivacyProPixel.privacyProSubscriptionExpired(source), frequency: .daily)
+            }
+        }.store(in: &cancellables)
+
+        notificationCenter.publisher(for: .entitlementsDidChange).sink { param in
```
miasma13: There is a problem with this notification in V2 as it is also posted in `DefaultSubscriptionEndpointServiceV2` where the value of entitlements are subscription features. So this notification can be posted by an expired user (with no entitlements) where the array will contain all values.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: privacyProSubscriptionMissing pixel added

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: Comparing what we had before with what we have now was my original approach.
Sending different pixels for entitlements added/removed is easily achievable, but we need to fix the requirements before proceeding because the requirements seem to be shifting and I'm not sure how these pixels will help us to diagnose issues.
I'm going back to the drawing board.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: I've addressed this point filtering the notification from the source, thanks.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: So... I've addressed this point by comparing the previously known user entitlements with the newly received and adding "added" or "removed" as a parameter in the pixel.
Wdyt? 
I'm still not super sure these kinds of pixels will be very useful, but better than nothing...

PR #1109, opened 2025-06-22
configurations URL overrides refactor

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204186595873227/task/1210111125223821?focus=true

### Description
Refactored how configuration URLs are provided and overridden within the app.  

### Testing Steps
1. As an internal user
2. Set a custom privacy configuration with a change you can track 
3. check the configuration are overridden with the new ones (also you can look at logs in the debugger)
4. Restart the app and check that the configuration are kept
5. Retry as a non internal user and check that the configuration are not overridden

iOS RMF override
1. Override the RMF url with https://www.jsonblob.com/api/1403080308542332928
2. Check you see an RMF
3. Restart the app and check the overrides is still there in the debug menu
4. Try and reset the RMF

General
1. Remove and install the app
2. Start the app
3. Check -> 

### Impact and Risks 
This refactoring affects configuration URL resolution but for production users, behavior should be identical since they don't have access to custom URL functionality.

#### What could go wrong?
1. Configuration Fetching Failures: If ConfigurationURLProvider dependency injection fails during app startup, production users could lose access to critical configurations:
- Privacy Configuration: Could disable tracker blocking and privacy protections
- Tracker Data Set: Could break content blocking rules
- Bloom Filters: Could affect HTTPS upgrade functionality
- Remote Messaging: Could prevent important app notifications
2. App Startup Crashes: Missing ConfigurationURLProvider in the dependency chain could cause crashes on app launch for all production users.
3. Silent Configuration Failures: If the provider returns incorrect URLs, production users might fetch stale or invalid configurations without obvious error symptoms.

### Notes to Reviewer
1. Verify Default URLs: Double-check that AppConfigurationURLProvider returns exactly the same URLs as before for all Configuration cases
2. Test Dependency Injection: Ensure ConfigurationURLProvider is properly created in both:
3. iOS/AppDependencyProvider (line 143)
4. macOS/AppDelegate (line 680)
5. Configuration Fetcher Integration: Verify that ConfigurationFetcher properly receives and uses the ConfigurationURLProvider dependency

—
###### Internal references:

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLStorage.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLStorage.swift
@@ -33,7 +33,7 @@ public final class CustomConfigurationURLStorage: CustomConfigurationURLStoring 
     private let defaults: UserDefaults
     private let keyPrefix: String
 
-    public init(defaults: UserDefaults = .standard,
+    public init(defaults: UserDefaults,
                 keyPrefix: String = "CustomConfigurationURL") {
         self.defaults = defaults
         self.keyPrefix = keyPrefix
--- a/macOS/DuckDuckGo/Menus/MainMenu.swift
+++ b/macOS/DuckDuckGo/Menus/MainMenu.swift
@@ -120,6 +120,7 @@ final class MainMenu: NSMenu {
     private let appearancePreferences: AppearancePreferences
     private let privacyConfigurationManager: PrivacyConfigurationManaging
     private let appVersion: AppVersion
+    private let configurationURLProvider: CustomConfigurationURLProviding
 
     // MARK: - Initialization
 
@@ -134,7 +135,8 @@ final class MainMenu: NSMenu {
          internalUserDecider: InternalUserDecider,
          appearancePreferences: AppearancePreferences,
          privacyConfigurationManager: PrivacyConfigurationManaging,
-         appVersion: AppVersion = .shared) {
+         appVersion: AppVersion = .shared,
+         configurationURLProvider: CustomConfigurationURLProviding) {
 
         self.featureFlagger = featureFlagger
         self.internalUserDecider = internalUserDecider
@@ -145,6 +147,7 @@ final class MainMenu: NSMenu {
         self.defaultBrowserPreferences = defaultBrowserPreferences
         self.aiChatMenuConfig = aiChatMenuConfig
         self.historyMenu = HistoryMenu(historyGroupingDataSource: historyCoordinator, featureFlagger: featureFlagger)
+        self.configurationURLProvider = configurationURLProvider
         super.init(title: UserText.duckDuckGo)
 
         buildItems {
@@ -865,9 +868,7 @@ final class MainMenu: NSMenu {
         }
         configurationDateAndTimeMenuItem.title = dateString
 
-        let urlProvider = Application.appDelegate.configurationURLProvider
-
-        customConfigurationUrlMenuItem.title = "Configuration URL:  \(urlProvider.url(for: .privacyConfiguration).absoluteString)"
+        customConfigurationUrlMenuItem.title = "Configuration URL:  \(configurationURLProvider.url(for: .privacyConfiguration).absoluteString)"
     }
 
     @objc private func toggleAutofillScriptDebugSettingsAction(_ sender: NSMenuItem) {
--- a/iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
+++ b/iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
@@ -140,7 +140,7 @@ final class AppDependencyProvider: DependencyProvider {
             featureFlagger = defaultFeatureFlagger
         }
 
-        configurationURLProvider = ConfigurationURLProvider(defaultProvider: AppConfigurationURLProvider(featureFlagger: featureFlagger), internalUserDecider: internalUserDecider, store: CustomConfigurationURLStorage())
+        configurationURLProvider = ConfigurationURLProvider(defaultProvider: AppConfigurationURLProvider(featureFlagger: featureFlagger), internalUserDecider: internalUserDecider, store: CustomConfigurationURLStorage(defaults: UserDefaults(suiteName: Global.appConfigurationGroupName) ?? UserDefaults()))
         configurationManager = ConfigurationManager(fetcher: ConfigurationFetcher(store: configurationStore, configurationURLProvider: configurationURLProvider, eventMapping: ConfigurationManager.configurationDebugEvents), store: configurationStore)
 
         // Configure Subscription
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -672,7 +672,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         )
         appContentBlocking = contentBlocking
 #endif
-        configurationURLProvider = ConfigurationURLProvider(defaultProvider: AppConfigurationURLProvider(privacyConfigurationManager: privacyConfigurationManager, featureFlagger: featureFlagger), internalUserDecider: internalUserDecider, store: CustomConfigurationURLStorage())
+        configurationURLProvider = ConfigurationURLProvider(defaultProvider: AppConfigurationURLProvider(privacyConfigurationManager: privacyConfigurationManager, featureFlagger: featureFlagger), internalUserDecider: internalUserDecider, store: CustomConfigurationURLStorage(defaults: UserDefaults.appConfiguration))
         configurationManager = ConfigurationManager(
             fetcher: ConfigurationFetcher(store: configurationStore, configurationURLProvider: configurationURLProvider, eventMapping: ConfigurationManager.configurationDebugEvents),
             store: configurationStore,
--- a/macOS/DuckDuckGo/Application/Application.swift
+++ b/macOS/DuckDuckGo/Application/Application.swift
@@ -53,7 +53,8 @@ final class Application: NSApplication {
             aiChatMenuConfig: delegate.aiChatMenuConfiguration,
             internalUserDecider: delegate.internalUserDecider,
             appearancePreferences: delegate.appearancePreferences,
-            privacyConfigurationManager: delegate.privacyFeatures.contentBlocking.privacyConfigurationManager
+            privacyConfigurationManager: delegate.privacyFeatures.contentBlocking.privacyConfigurationManager,
+            configurationURLProvider: delegate.configurationURLProvider
         )
         self.mainMenu = mainMenu
 
--- a/macOS/UnitTests/Menus/MainMenuTests.swift
+++ b/macOS/UnitTests/Menus/MainMenuTests.swift
@@ -20,6 +20,7 @@ import XCTest
 import Combine
 import BrowserServicesKit
 @testable import DuckDuckGo_Privacy_Browser
+@testable import Configuration
 
 class MainMenuTests: XCTestCase {
 
@@ -118,7 +119,8 @@ class MainMenuTests: XCTestCase {
             aiChatMenuConfig: DummyAIChatConfig(),
             internalUserDecider: MockInternalUserDecider(),
             appearancePreferences: appearancePreferences,
-            privacyConfigurationManager: MockPrivacyConfigurationManager()
+            privacyConfigurationManager: MockPrivacyConfigurationManager(),
+            configurationURLProvider: MockCustomURLProvider()
         )
 
         sut.update()
@@ -188,7 +190,8 @@ class MainMenuTests: XCTestCase {
             aiChatMenuConfig: DummyAIChatConfig(),
             internalUserDecider: MockInternalUserDecider(),
             appearancePreferences: appearancePreferences,
-            privacyConfigurationManager: MockPrivacyConfigurationManager()
+            privacyConfigurationManager: MockPrivacyConfigurationManager(),
+            configurationURLProvider: MockCustomURLProvider()
         )
 
         sut.update()
@@ -213,7 +216,8 @@ class MainMenuTests: XCTestCase {
             aiChatMenuConfig: DummyAIChatConfig(),
             internalUserDecider: MockInternalUserDecider(),
             appearancePreferences: appearancePreferences,
-            privacyConfigurationManager: MockPrivacyConfigurationManager()
+            privacyConfigurationManager: MockPrivacyConfigurationManager(),
+            configurationURLProvider: MockCustomURLProvider()
         )
 
         sut.update()
@@ -237,7 +241,8 @@ class MainMenuTests: XCTestCase {
             aiChatMenuConfig: DummyAIChatConfig(),
             internalUserDecider: MockInternalUserDecider(),
             appearancePreferences: appearancePreferences,
-            privacyConfigurationManager: MockPrivacyConfigurationManager()
+            privacyConfigurationManager: MockPrivacyConfigurationManager(),
+            configurationURLProvider: MockCustomURLProvider()
         )
         let bookmarksMenu = try XCTUnwrap(sut.item(withTitle: UserText.bookmarks))
 
@@ -263,7 +268,8 @@ class MainMenuTests: XCTestCase {
             aiChatMenuConfig: aiChatConfig,
             internalUserDecider: MockInternalUserDecider(),
             appearancePreferences: appearancePreferences,
-            privacyConfigurationManager: MockPrivacyConfigurationManager()
+            privacyConfigurationManager: MockPrivacyConfigurationManager(),
+            configurationURLProvider: MockCustomURLProvider()
         )
 
         let fileMenu = try XCTUnwrap(sut.item(withTitle: UserText.mainMenuFile))
@@ -291,7 +297,8 @@ class MainMenuTests: XCTestCase {
             aiChatMenuConfig: aiChatConfig,
             internalUserDecider: MockInternalUserDecider(),
             appearancePreferences: appearancePreferences,
-            privacyConfigurationManager: MockPrivacyConfigurationManager()
+            privacyConfigurationManager: MockPrivacyConfigurationManager(),
+            configurationURLProvider: MockCustomURLProvider()
         )
 
         let fileMenu = try XCTUnwrap(sut.item(withTitle: UserText.mainMenuFile))


SharedPackages/BrowserServicesKit/Sources/BrowserServicesKitTestsUtils/MockInternalUserDecider.swift
```diff
@@ -0,0 +1,33 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import BrowserServicesKit
+import Combine
+
+public class MockInternalUserDecider: InternalUserDecider {
+    public var isInternalUserPublisher: AnyPublisher<Bool, Never> = Just(true).eraseToAnyPublisher()
```
alessandroboron: not sure how we use this for testing but currently isInternalUserPublisher and  isInternalUser return different value. Should we return the same value for consistency?

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
```diff
@@ -40,21 +40,25 @@ public final class ConfigurationFetcher: ConfigurationFetching {
 
     private var store: ConfigurationStoring
     private let validator: ConfigurationValidating
-    private let urlSession: URLSession
+    private let configurationURLProvider: ConfigurationURLProviding
+    private let sessionProvider: () -> URLSession
 
     public convenience init(store: ConfigurationStoring,
-                            urlSession: URLSession = .shared,
+                            urlSession: @autoclosure @escaping () -> URLSession = URLSession.shared,
```
alessandroboron, line 47: out of curiosity, what’s the rationale for using @autoclosure in this scenario? Do we need to lazyily evaluate URLSession or we can just inject the instance like we were doing before?

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
```diff
@@ -125,7 +129,7 @@ public final class ConfigurationFetcher: ConfigurationFetching {
         let configuration = APIRequest.Configuration(url: url,
                                                      headers: APIRequest.Headers(etag: etag),
                                                      cachePolicy: .reloadIgnoringLocalCacheData)
-        let request = APIRequest(configuration: configuration, requirements: requirements, urlSession: urlSession)
+        let request = APIRequest(configuration: configuration, requirements: requirements, urlSession: sessionProvider())
```
alessandroboron, line 132: Should we use NetworkService V2 and pass the APIRequest to the service instead of passing the session to the request?

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationTestSupport/MockCustomURLProvider.swift
```diff
@@ -0,0 +1,54 @@
...
+//  limitations under the License.
+//
+
+#if DEBUG
+import Foundation
+
+final class MockCustomURLProvider: CustomConfigurationURLProviding {
+    var isCustomURLEnabled: Bool = true
+    var url: URL = URL(string: "https://duckduckgo.com")!
+    var capturesURL: URL?
```
alessandroboron:
```suggestion
    var capturedURL: URL?
```

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationTestSupport/MockCustomURLProvider.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    func setCustomURL(_ url: URL?, for configuration: Configuration) {
+        capturedConfiguration = configuration
+        capturedConfigurations.append(configuration)
+        capturesURL = url
+    }
+}
+
+final class MockConfigurationURLProvider: ConfigurationURLProviding {
+    var url: URL?
```
alessandroboron:
```suggestion
    var url: URL = URL(string: "duckduckgo.com")!
```

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationTestSupport/MockCustomURLProvider.swift
```diff
@@ -0,0 +1,54 @@
...
+    var capturedConfiguration: Configuration?
+    var capturedConfigurations: [Configuration] = []
+
+    func url(for configuration: Configuration) -> URL {
+        capturedConfiguration = configuration
+        capturedConfigurations.append(configuration)
+        return url ?? URL(string: "duckduckgo.com")!
```
alessandroboron: We can give a default value to https://github.com/duckduckgo/apple-browsers/pull/1109/files#r2218359401 so it’s clear what we will return and will be consistent with `MockCustomURLProvider`

SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLProvider.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+
+import Foundation
+import BrowserServicesKit
+
+public protocol ConfigurationURLProviding {
+    func url(for configuration: Configuration) -> URL
+}
+
+public protocol CustomConfigurationURLSetting {
```
alessandroboron, line 26: nip:
```suggestion
public protocol CustomConfigurationURLSettingManaging {
```

SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLStorage.swift
```diff
@@ -0,0 +1,80 @@
...
+    var customSurrogatesURL: URL? { get set }
+    var customRemoteMessagingConfigURL: URL? { get set }
+}
+
+public final class CustomConfigurationURLStorage: CustomConfigurationURLStoring {
+
+    private let defaults: UserDefaults
+    private let keyPrefix: String
+
+    public init(defaults: UserDefaults = .standard,
```
alessandroboron: Double checking, should we use the KeyValueFilesStore instead of UserDefaults? There were some recent issues with UserDefaults and there was an [Apple Diary message](https://app.asana.com/0/0/1210382925425419) about using the new storage.

SharedPackages/BrowserServicesKit/Tests/ConfigurationTests/ConfigurationFetcherTests.swift
```diff
@@ -352,4 +361,28 @@ final class ConfigurationFetcherTests: XCTestCase {
...
+        MockURLProtocol.requestHandler = { _ in (HTTPURLResponse.ok, self.privacyConfigurationData) }
+        let expectedURL = URL(string: "https://test.example.com")!
+        configurationURLProvider.url = expectedURL
+
+        let store = MockStore()
+        let fetcher = makeConfigurationFetcher(store: store)
+
+        try? await fetcher.fetch(.privacyConfiguration)
+
+        XCTAssertEqual(MockURLProtocol.lastRequest?.url, expectedURL)
```
alessandroboron, line 386: Should we reset MockURLProtocol.lastRequest and MockURLProtocol.MockURLProtocol.lastRequest in tearDown?

Also probably better to declare the functions throwing instead of using `try?`. If by any chance we have another test that sets MockURLProtocol.lastRequest?.url to the same url of this test and `try? await fetcher.fetch(.privacyConfiguration)` fails, this test still succeeds. I know this is very difficult that will happen but better to avoid these type of situations.

SharedPackages/BrowserServicesKit/Tests/ConfigurationTests/ConfigurationURLProviderTests.swift
```diff
@@ -0,0 +1,298 @@
...
+        mockInternalUserDecider.isInternalUser = false
+
+        // When
+        let result = sut.isCustomURLEnabled
+
+        // Then
+        XCTAssertFalse(result)
+    }
+
+    func testURLForConfiguration_WhenCustomURLsDisabled_AlwaysReturnsDefaultURL() {
```
alessandroboron: Just leaving it here for knowledge sharing: If we use SwiftTesting we can easily build tests where you can pass arguments to the test function. In this case we could pass a collection ofConfiguration so that we could easily cover all the cases when calling `sut.url(for:)`. In this way we could only write a single test instead of repeating the test for each case. Because Configuration has a String rawValue (making it serialisable) we can re-run specific failed test cases without re-running the entire test suite.  e.g.

```
@Test(“Check Default URL Is Returned When No Internal User”, arguments: Configuration.allCases)
func testWhenIsNotInternalUserThenReturnDefaultURL(config: Configuration)
```

SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLStorage.swift
alessandroboron: Actually I noticed that this is just for the override, I think UserDefault is appropriate to use.

macOS/IntegrationTests/Configurations/ConfigurationManagerIntegrationTests.swift
```diff
@@ -70,14 +56,66 @@ final class ConfigurationManagerIntegrationTests: XCTestCase {
...
-        )
+        customURLProvider.setCustomURL(nil, for: .privacyConfiguration)
         await configManager.refreshNow()
         let resetEtag = await Application.appDelegate.privacyFeatures.contentBlocking.trackerDataManager.fetchedData?.etag
         XCTAssertNotEqual(newEtag, resetEtag)
     }
 
 }
+
+class MockInternalUserDecider: InternalUserDecider {
```
alessandroboron: Didn’t we declare the below mocks in the package?

macOS/IntegrationTests/Configurations/ConfigurationManagerIntegrationTests.swift
```diff
@@ -17,35 +17,25 @@
...
     }
 
     override func tearDownWithError() throws {
-        // use default privacyConfiguration link
-        _ = AppConfigurationURLProvider(
-            privacyConfigurationManager: MockPrivacyConfigurationManager(),
-            featureFlagger: MockFeatureFlagger(),
-            customPrivacyConfiguration: AppConfigurationURLProvider.Constants.defaultPrivacyConfigurationURL
-        )
         configManager = nil
```
alessandroboron, line 47: lets nil customURLProvider in tearDown

macOS/IntegrationTests/Configurations/ConfigurationManagerIntegrationTests.swift
```diff
@@ -17,35 +17,25 @@
...
+        let fetcher = ConfigurationFetcher(store: MockConfigurationStoring(), configurationURLProvider: customURLProvider)
         let privacyFeatures = Application.appDelegate.privacyFeatures
-        configManager = ConfigurationManager(
-            trackerDataManager: privacyFeatures.contentBlocking.trackerDataManager,
-            privacyConfigurationManager: privacyFeatures.contentBlocking.privacyConfigurationManager,
-            contentBlockingManager: privacyFeatures.contentBlocking.contentBlockingManager,
-            httpsUpgrade: privacyFeatures.httpsUpgrade
-        )
+        configManager = ConfigurationManager(fetcher: fetcher, store: MockConfigurationStoring(), defaults: MockKeyValueStore(), trackerDataManager: privacyFeatures.contentBlocking.trackerDataManager, privacyConfigurationManager: privacyFeatures.contentBlocking.privacyConfigurationManager, contentBlockingManager: privacyFeatures.contentBlocking.contentBlockingManager, httpsUpgrade: privacyFeatures.httpsUpgrade)
```
alessandroboron: nip: better to use multiline indentation like before given the number of parameters?

iOS/IntegrationTests/Configuration/ConfigurationManagerIntegrationTests.swift
```diff
@@ -19,26 +19,24 @@
...
     override func setUpWithError() throws {
-        // use default privacyConfiguration link
-        customURLProvider = CustomConfigurationURLProvider()
-        customURLProvider.customPrivacyConfigurationURL = URL.privacyConfig
-        Configuration.setURLProvider(customURLProvider)
-        configManager = ConfigurationManager()
+        let internalUserDecider = MockInteranlUserDecider()
+        internalUserDecider.isInternalUser = true
+        customURLProvider = MockCustomURLProvider()
+        let fetcher = ConfigurationFetcher(store: AppDependencyProvider.shared.configurationStore, configurationURLProvider: customURLProvider)
```
alessandroboron: Do we need to use the instance defined in the AppDependencyProvider or we can initialise it here?

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
SabrinaTardio: Great question! I use @autoclosure here to defer the creation of the URLSession instance. Previously, I saw a warning about “Performing I/O on the main thread can cause slow launches” when the session was initialised directly. By using @autoclosure, we ensure that the session is only created when it’s actually needed, which helps avoid main-thread I/O and improves launch performance.

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
SabrinaTardio: I think we should… but I rather not make too many changes at ones and keep the scope of this contained

SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLProvider.swift
SabrinaTardio: I meant it is "allows to set”

SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLStorage.swift
SabrinaTardio: Yeah that was my thinking, also when I started we did not have the new way but then I figured it was not worth changing it

macOS/LocalPackages/DataBrokerProtection-macOS/Sources/DataBrokerProtection-macOS/PrivacyConfiguration/ConfigurationManager.swift
```diff
@@ -35,7 +35,7 @@ public final class ConfigurationManager: DefaultConfigurationManager {
     private let privacyConfigManager: DBPPrivacyConfigurationManager
 
     public init(privacyConfigManager: DBPPrivacyConfigurationManager,
-                fetcher: ConfigurationFetching = ConfigurationFetcher(store: ConfigurationStore(), eventMapping: configurationDebugEvents),
+                fetcher: ConfigurationFetching,
```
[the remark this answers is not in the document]
SabrinaTardio: It is injected :)

SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
```diff
@@ -40,21 +40,25 @@ public final class ConfigurationFetcher: ConfigurationFetching {
...
+    private let sessionProvider: () -> URLSession
 
     public convenience init(store: ConfigurationStoring,
-                            urlSession: URLSession = .shared,
+                            urlSession: @autoclosure @escaping () -> URLSession = URLSession.shared,
+                            configurationURLProvider: ConfigurationURLProviding,
                             eventMapping: EventMapping<ConfigurationDebugEvents>? = nil) {
         let validator = ConfigurationValidator(eventMapping: eventMapping)
-        self.init(store: store, validator: validator)
+        self.init(store: store, validator: validator, sessionProvider: urlSession(), configurationURLProvider: configurationURLProvider)
```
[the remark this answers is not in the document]
alessandroboron: cursor is right. if we evaluate the closure immediately there’s no point in using `@autoclosure`. If we want still take advantage of lazy evaluation we should remove `@autoclosure` from the parameter in the designated initialised and pass the closure with `self.init(store: store, validator: validator, sessionProvider: urlSession, configurationURLProvider: configurationURLProvider)`

SharedPackages/BrowserServicesKit/Tests/ConfigurationTests/ConfigurationURLProviderTests.swift
```diff
@@ -0,0 +1,158 @@
...
+import Foundation
+
+final class ConfigurationURLProviderTests {
+
+    var sut: ConfigurationURLProvider!
+    var mockDefaultProvider: MockConfigurationURLProvider!
+    var mockInternalUserDecider: MockInternalUserDecider!
+    var mockStore: MockCustomConfigurationURLStore!
+
+    func setUp() {
```
alessandroboron: in Swift Testing setup has been replaced by init. The ConfigurationURLProviderTests class will be initialised for each test so you can do your mock setup in init. In that way there’s no need for the tests to call setUp.

Additionally, tearDown has been replaced by deinit (available only for classes) so code to clean up should be added there and it will be executed automatically by the framework when the class is deinitialised. The good news is that there’s no need to implement deinit unless we need to close a file, remove userdefaults suite, reset a global state etc.  The mocks will be nilled after the test executes automatically. You can read more about it here https://app.asana.com/1/137249556945/project/1206329551987282/task/1209479341722484?focus=true

This is a good start for migrating from XCTest to SwiftTesting https://developer.apple.com/documentation/testing/migratingfromxctest

SharedPackages/BrowserServicesKit/Tests/ConfigurationTests/ConfigurationURLProviderTests.swift
```diff
@@ -0,0 +1,158 @@
...
+
+    func tearDown() {
+        sut = nil
+        mockDefaultProvider = nil
+        mockInternalUserDecider = nil
+        mockStore = nil
+    }
+
+    @Test("Check Default URL Is Returned When No Internal User", arguments: Configuration.allCases)
+    func testWhenIsNotInternalUserThenReturnDefaultURL(config: Configuration) {
```
alessandroboron: nip: tests method signature don’t need to start with `test` prefix like in XCTest

SharedPackages/BrowserServicesKit/Tests/ConfigurationTests/ConfigurationURLProviderTests.swift
```diff
@@ -0,0 +1,158 @@
...
+        (.bloomFilterBinary, { $0.customBloomFilterBinaryURL = $1 }, "https://custom-binary.example.com"),
+        (.bloomFilterExcludedDomains, { $0.customBloomFilterExcludedDomainsURL = $1 }, "https://custom-excluded.example.com"),
+        (.privacyConfiguration, { $0.customPrivacyConfigurationURL = $1 }, "https://custom-privacy.example.com"),
+        (.trackerDataSet, { $0.customTrackerDataSetURL = $1 }, "https://custom-tracker.example.com"),
+        (.surrogates, { $0.customSurrogatesURL = $1 }, "https://custom-surrogates.example.com"),
+        (.remoteMessagingConfig, { $0.customRemoteMessagingConfigURL = $1 }, "https://custom-messaging.example.com"),
+    ]
+
+    @Test("Custom URL is returned when set", arguments: customURLCases)
+    func testCustomURLIsReturnedWhenSet(config: Configuration, setCustomURL: (MockCustomConfigurationURLStore, URL?) -> Void, urlString: String) {
```
alessandroboron: nip: can we stick with the // Given // When // Then for consistency?

SharedPackages/BrowserServicesKit/Tests/ConfigurationTests/ConfigurationURLProviderTests.swift
```diff
@@ -0,0 +1,158 @@
...
+
+    @Test("Custom URL is not set when not internal user", arguments: customURLCases)
+    func testSetCustomURL_WhenCustomURLsDisabled_DoesNotUpdateStore(config: Configuration, setCustomURL: (MockCustomConfigurationURLStore, URL?) -> Void, urlString: String) {
+        setUp()
+        defer { tearDown() }
+        mockInternalUserDecider.isInternalUser = false
+        let customURL = URL(string: urlString)!
+        sut.setCustomURL(customURL, for: config)
+        // Should not update the store
+        var value: URL?
```
alessandroboron, line 100: Maybe I’m missing something about this test but it feels we’re testing the mock in the end. Should we not assign a random url to the url store, then call sut.setCustomURL(customURL, for: config) to set the proper url and expect that the url in the store has not changed? e.g.

```
mockInternalUserDecider.isInternalUser = false
<set mockstore with some random values>
let customURL = URL(string: urlString)!
sut.setCustomURL(customURL, for: config)
<expect mockstore URL are still some random values>
```

macOS/IntegrationTests/Configurations/ConfigurationManagerIntegrationTests.swift
```diff
@@ -70,14 +64,22 @@ final class ConfigurationManagerIntegrationTests: XCTestCase {
...
-        )
+        customURLProvider.setCustomURL(nil, for: .privacyConfiguration)
         await configManager.refreshNow()
         let resetEtag = await Application.appDelegate.privacyFeatures.contentBlocking.trackerDataManager.fetchedData?.etag
         XCTAssertNotEqual(newEtag, resetEtag)
     }
 
 }
+
+private class MockKeyValueStore: KeyValueStoring {
```
alessandroboron: in PersistenceTestingUtils we have MockKeyValueStore. Could we re-use that one?

iOS/DuckDuckGo/ConfigurationURLDebugViewController.swift
```diff
@@ -54,25 +54,23 @@ final class ConfigurationURLDebugViewController: UITableViewController {
         return formatter
     }()
 
-    private var customURLProvider = CustomConfigurationURLProvider()
+    weak var viewModel: DebugScreensViewModel?
 
     @UserDefaultsWrapper(key: .lastConfigurationUpdateDate, defaultValue: nil)
     private var lastConfigurationUpdateDate: Date?
 
-    @UserDefaultsWrapper(key: .privacyConfigCustomURL, defaultValue: nil)
```
alessandroboron, line 62: @SabrinaTardio there’s an issue when you try to reset the custom URL after restarting the app. The original code would show an alert with the option to reset the custom URL to the original one but this is not the case anymore.  Screenshot attached

IMG_6891

alessandroboron: Slightly similar issue on macOS. after updating to custom url it works fine but the Configuration URL in the debug menu doesn’t get updated to the real value as per screenshot attached

Screenshot 2025-08-06 at 1 40 37 pm

alessandroboron (commented):
@SabrinaTardio I left few comments, I tested iOS and macOS privacy config and it works as expected (please see comments about debug menu).

I think it would be worth to have someone with RMF knowledge to test the feature as well.

amddg44 (approved):
LGTM from an RMF perspective - thanks @SabrinaTardio!

macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -872,10 +872,8 @@ final class MainMenu: NSMenu {
             dateString = "Last Update Time: -"
         }
         configurationDateAndTimeMenuItem.title = dateString
-        let urlProvider = AppConfigurationURLProvider(
-            privacyConfigurationManager: privacyConfigurationManager,
-            featureFlagger: featureFlagger
-        )
+
+        let urlProvider = Application.appDelegate.configurationURLProvider
```
alessandroboron: should we inject it in Application instead of accessing the app delegate directly?

```
let mainMenu = MainMenu(
    featureFlagger: delegate.featureFlagger,
    bookmarkManager: delegate.bookmarkManager,
    historyCoordinator: delegate.historyCoordinator,
    faviconManager: delegate.faviconManager,
    aiChatMenuConfig: delegate.aiChatMenuConfiguration,
    internalUserDecider: delegate.internalUserDecider,
    appearancePreferences: delegate.appearancePreferences,
    privacyConfigurationManager: delegate.privacyFeatures.contentBlocking.privacyConfigurationManager
)
```

alessandroboron (approved):
Tested Privacy config and work as expected. 👏 

Would recommned to have VPN tested too

SharedPackages/BrowserServicesKit/Sources/Configuration/CustomConfigurationURLStorage.swift
diegoreymendez: This isn't new to your changes, but still relevant.

For macOS it's best if this used `UserDefaults.appConfiguration` which is a shared user default because it means the agents will be able to see the same values / overrides as the main app.

My suggested improvement: don't offer a default value here, and let each app provide its own default storage, using `UserDefaults.appConfiguration` for macOS.

PR #1113, opened 2025-06-23
New subscription expired pixel

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200019156869587/task/1210250285874364?focus=true
CC: @diegoreymendez 

### Description

iOS and macOS now detect subscription expirations and send a new pixel.

### Testing Steps

- Run the app from xcode
- Subscribe using a sandbox account
- Wait expiration
- Check the logs if the pixel was fired

### Impact and Risks

None, just a pixel

---
###### Internal references:


federicocappelli:
@miasma13 @diegoreymendez  I had to open a new PR and point it to main

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -19,8 +19,11 @@
 import Foundation
 import Subscription
 import PixelKit
+import os.log
+import Common
+import Combine
 
-public struct AuthV2PixelHandler: SubscriptionPixelHandler {
+public class AuthV2PixelHandler: SubscriptionPixelHandler {
```
diegoreymendez, line 27: Can we document what this class does?  "Handling" is very broad and a more specific description explaining what's being handled would help the reader here.

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -43,8 +46,53 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
     }
 
     let source: Source
+    private let notificationCenter: NotificationCenter = NotificationCenter.default
```
diegoreymendez: Can we make this an injectable init parameter with a default value of `.default`?

iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -37,16 +41,63 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+
+        // Intercepting and sending a pixel every time a set of entitlements change. We are only interested in changes between full > empty. Any other combination is not possible
+        notificationCenter.publisher(for: .entitlementsDidChange).sink {  [weak self] notification in
+
+            guard let self else { return }
+
+            guard (notification.object as? SubscriptionManagerV2) != nil else {
+                // Sending pixel only for user entitlements, ignoring the Subscription entitlements coming from DefaultSubscriptionEndpointServiceV2
+                return
+            }
```
diegoreymendez, line 85: Why are we excluding some notifications?  Should the rest of the code do the same?  If they should be ignored by all clients: why not avoid firing those notifications or renaming them instead?

This class should observe the exact same notifications that the rest of the code will.

iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
```diff
@@ -37,16 +41,63 @@ public struct AuthV2PixelHandler: SubscriptionPixelHandler {
...
+            let previousEntitlements = userInfo?[UserDefaultsCacheKey.subscriptionPreviousEntitlements] ?? []
+
+            switch (previousEntitlements.isEmpty, entitlements.isEmpty) {
+            case (true, false):
+                DailyPixel.fireDailyAndCount(pixel: .privacyProEntitlementsAdded, withAdditionalParameters: self.sourceParam)
+            case (false, true):
+                DailyPixel.fireDailyAndCount(pixel: .privacyProEntitlementsRemoved, withAdditionalParameters: self.sourceParam)
+            default:
+                Logger.subscription.debug("We shouldn't have received this notification: \(notification.name.rawValue, privacy: .public), ignoring it...")
+            }
```
diegoreymendez: I understand that today single entitlements won't change, but a more robust comparison doesn't need to rely on this assumption, so it won't ever break even if the assumption changes:

(I tested the code and it works well)

```swift
let userInfo = notification.userInfo as? [AnyHashable: [Entitlement]] ?? [:]
let entitlements = Set(userInfo[UserDefaultsCacheKey.subscriptionEntitlements] ?? [])
let previousEntitlements = Set(userInfo[UserDefaultsCacheKey.subscriptionPreviousEntitlements] ?? [])

let addedEntitlements = entitlements.subtracting(previousEntitlements)
let removedEntitlements = previousEntitlements.subtracting(entitlements)

if !addedEntitlements.isEmpty {
    DailyPixel.fireDailyAndCount(pixel: .privacyProEntitlementsAdded, withAdditionalParameters: sourceParam)
}

if !removedEntitlements.isEmpty {
    DailyPixel.fireDailyAndCount(pixel: .privacyProEntitlementsRemoved, withAdditionalParameters: sourceParam)
}

if addedEntitlements.isEmpty && removedEntitlements.isEmpty {
    Logger.subscription.debug("No changes in entitlements for notification: \(notification.name.rawValue, privacy: .public), ignoring...")
}
```

You'll also need:

```swift
extension Entitlement: Hashable {
    public func hash(into hasher: inout Hasher) {
        hasher.combine(product)
    }
}
```

diegoreymendez: Also can we place this specific logic in a method and unit test it?

diegoreymendez:
I misclicked and asked Copilot for re-review, so I just resolved the comments.

iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: In the app there are two kinds of "entitlements"
1. The Subscription entitlements: what the subscription CAN do
2. The user entitlement: what the user is ALLOWED to do

The VPN cares only about the user entitlement (see how the entitlements are checked), and here we want to track only that one. The rest of the app cares about both so we can't stop sending the notification.

diegoreymendez: Some additional observations from testing: I don't see pixels fired when restoring my subscription because this case seems to ignore them.  These notifications should be included.

iOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: done

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: done

macOS/DuckDuckGo/Subscription/AuthV2PixelHandler.swift
federicocappelli: done

diegoreymendez:
As I'm taking this over, I'll close this PR and start over branching off this PR.

PR #1120, opened 2025-06-23
PIR: Allow web views to proceed if they encounter 403 errors and decrease email service retries

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210614061934343?focus=true
Tech Design URL:
CC:

### Description
Allow web views to proceed if they encounter a 403 error
This is because if we get the cloud flare challenge, that's a 403 error, but we want to continue anyway

Also decreases the number of email service retries since if emails are failing then opt outs would stall on this step for a really long time, blocking others from running (so it will only try for five minutes)

### Testing Steps

1. Open up PIR debug mode
2. Run against Wellnut, with information you expect to find a result
3. Run the opt out, and ensure a 403 error isn't thrown, and it's able to proceed to the opt out page
4. Ensure the email step times out after five minutes

### Notes to Reviewer

---
###### Internal references:


brianhall:
@THISISDINOSAUR CloudFlare is specifically returning 403s. I don't know what the impact of not erroring on all 400s would be, could we add a specific exception for 403s to reduce the surface area?

quanganhdo (approved):
Added a note to the linked Asana task, otherwise the changes LGTM.

PR #1147, opened 2025-06-25
VPN Subscription expiration pixel

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210250285874364?focus=true

### IMPORTANT:
After merging we need to cherry-pick this into the iOS release.

### Description

Adds pixels to track subscription status notifications as interpreted in the VPN observer.

### Testing Steps

Please test in both macOS and iOS.

You should see these pixels when the subscription lib fires notifications.  Unfortunately it's hard to know the timing as this PR doesn't seek to fix anything - just record when this happens.

macOS
```
👾[Daily and Count-Fired] m_vpn_subs_notification_signed_in_mac_daily ["isSubscriptionActive": "unknown", "notificationObjectClass": "DefaultAccountManager", "pixelSource": "browser-dmg", "version": "v1", "appVersion": "1.145.0"]
👾[Daily and Count-Fired] m_vpn_subs_notification_signed_in_mac_count ["version": "v1", "appVersion": "1.145.0", "pixelSource": "browser-dmg", "isSubscriptionActive": "unknown", "notificationObjectClass": "DefaultAccountManager"]
👾[Daily and Count-Fired] m_vpn_subs_notification_vpn_enabled_mac_daily ["version": "v1", "isSubscriptionActive": "true", "pixelSource": "browser-dmg", "notificationObjectClass": "DefaultAccountManager", "appVersion": "1.145.0"]
👾[Daily and Count-Fired] m_vpn_subs_notification_vpn_enabled_mac_count ["version": "v1", "isSubscriptionActive": "true", "pixelSource": "browser-dmg", "notificationObjectClass": "DefaultAccountManager", "appVersion": "1.145.0"]
```

iOS
```
👾[Daily and Count-Fired] m_vpn_subs_notification_signed_in_ios_phone_daily ["authVersion": "v2", "pixelSource": "phone", "isSubscriptionActive": "unknown", "notificationObjectClass": "DefaultSubscriptionManagerV2", "appVersion": "7.174.1"]
👾[Daily and Count-Fired] m_vpn_subs_notification_signed_in_ios_phone_count ["authVersion": "v2", "pixelSource": "phone", "isSubscriptionActive": "unknown", "notificationObjectClass": "DefaultSubscriptionManagerV2", "appVersion": "7.174.1"]
👾[Daily and Count-Fired] m_vpn_subs_notification_vpn_enabled_ios_phone_daily ["notificationObjectClass": "SubscriptionEndpointServiceV2", "pixelSource": "phone", "isSubscriptionActive": "true", "appVersion": "7.174.1", "authVersion": "v2"]
👾[Daily and Count-Fired] m_vpn_subs_notification_vpn_enabled_ios_phone_count ["notificationObjectClass": "SubscriptionEndpointServiceV2", "pixelSource": "phone", "isSubscriptionActive": "true", "appVersion": "7.174.1", "authVersion": "v2"]
```

### Impact and Risks

None

#### What could go wrong?

Nothing as there are no changes to existing logic.

### Quality Considerations

NA

### Notes to Reviewer

Please keep in mind this is merely recording something that already exists, and doesn't seek to fix anything.

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -445,6 +445,11 @@ public final class PixelKit {
         }
 
     private func prefixedAndSuffixedName(for event: Event, namePrefix: String?) -> String {
+
+        if let v3Event = event as? PixelKitEventV3 {
+            return v3Event.namePrefix + v3Event.name + platformSuffix
+        }
```
diegoreymendez: Support for firing pixels that define their own prefix, and use a standard platform suffix.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -472,7 +477,20 @@ public final class PixelKit {
...
+        case Source.iOS.rawValue:
+            return "_ios_phone"
+        case Source.iPadOS.rawValue:
+            return "_ios_tablet"
+        case Source.macStore.rawValue, Source.macDMG.rawValue:
+            return "_mac"
+        default:
+            return ""
+        }
+    }
```
diegoreymendez: The only change here is that V3 pixel events will have a `_mac` suffix.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKitEventV3.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import Foundation
+
+/// New version of this protocol that allows us to maintain backwards-compatibility with PixelKitEvent
+///
+/// This allows us to introduce support for new features without having to immediately migrate every other pixel to it.
+///
+public protocol PixelKitEventV3: PixelKitEventV2 {
+    var namePrefix: String { get }
+}
```
diegoreymendez: A protocol to be able to improve PixelKitEvent without compromising any other existing pixel.

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
```diff
@@ -0,0 +1,89 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import PixelKit
+import Subscription
+
+public enum VPNSubscriptionNotificationPixel: PixelKitEventV3 {
+    case vpnEnabled(isSubscriptionActive: Bool?,
```
diegoreymendez: I was going to use a string here, but having a parameter named `isSubscriptionActive` with a string value looked extremely weird.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -1872,8 +1873,20 @@ class MainViewController: UIViewController {
...
+            let isSubscriptionActive = try? await subscriptionManager.getSubscription(cachePolicy: .cacheOnly).isActive
+
+            PixelKit.fire(
+                VPNSubscriptionNotificationPixel.signedIn(
+                    isSubscriptionActive: isSubscriptionActive,
+                    isAuthV2Enabled: isAuthV2Enabled,
+                    notificationSourceObject: notification.object),
+                frequency: .dailyAndCount)
+            tunnelDefaults.resetEntitlementMessaging()
+            Logger.networkProtection.info("[NetP Subscription] Reset expired entitlement messaging")
```
diegoreymendez, line 1888: These lines are the original code, and they should be 100% unchanged.

macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionSubscriptionEventHandler.swift
```diff
@@ -48,7 +49,7 @@ final class NetworkProtectionSubscriptionEventHandler {
         Task {
             let hasEntitlement = await subscriptionManager.isFeatureEnabledForUser(feature: .networkProtection)
             Task {
-                await handleEntitlementsChange(hasEntitlements: hasEntitlement)
+                await handleEntitlementsChange(hasEntitlements: hasEntitlement, sourceObject: subscriptionManager)
```
diegoreymendez: This call is not triggered from a notification.

I would use this very class, but it's not accessible from the location where the pixel tests it.  `subscriptionManager` is the bridge class so it's quite unique anyways.

We could also pass `nil` if that works best.

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
```diff
@@ -0,0 +1,102 @@
...
+            return "DefaultAccountManager"
+        case is SubscriptionEndpointServiceV2:
+            return "SubscriptionEndpointServiceV2"
+        case is DefaultSubscriptionManagerV2:
+            return "DefaultSubscriptionManagerV2"
+        case is SubscriptionAuthV1toV2Bridge:
+            return "SubscriptionAuthV1toV2Bridge"
+        default:
+            return "Unknown - add class!"
+        }
```
miasma13: To avoid that you can just:
`String(describing: sourceObject).trimmingPrefix("Subscription.")`

Otherwise if you prefer to keep explicit list of classes we miss `DefaultSubscriptionManager` on the list.

macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionSubscriptionEventHandler.swift
miasma13: I think keeping `sourceObject` makes sense. What I am thinking about is maybe we should add another parameter to discern the call's origin - during initialisation (like this one) or as a result of notification WDYT?

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
```diff
@@ -0,0 +1,102 @@
...
+//  limitations under the License.
+//
+
+import PixelKit
+import Subscription
+
+public enum VPNSubscriptionNotificationPixel: PixelKitEventV3 {
+    case vpnEnabled(isSubscriptionActive: Bool?,
+                    isAuthV2Enabled: Bool,
+                    sourceObject: Any?)
```
miasma13: Maybe we can avoid passing whole object here and can already pass stringified class name as a parameter? (applies to all events)

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
```diff
@@ -0,0 +1,102 @@
...
+    public var parameters: [String: String]? {
+        switch self {
+        case .signedIn(let isSubscriptionActive, let isAuthV2, let sourceObject),
+                .signedOut(let isSubscriptionActive, let isAuthV2, let sourceObject),
+                .vpnEnabled(let isSubscriptionActive, let isAuthV2, let sourceObject),
+                .vpnDisabled(let isSubscriptionActive, let isAuthV2, let sourceObject):
+
+            let isSubscriptionActiveString = {
+                guard let isSubscriptionActive else {
+                    return "unknown"
```
miasma13: Let's change the value to "no subscription"

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
miasma13: Can we please add test for that new v3 logic?

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
```diff
@@ -0,0 +1,102 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import PixelKit
+import Subscription
+
+public enum VPNSubscriptionNotificationPixel: PixelKitEventV3 {
```
miasma13: I suggest the following naming changes, mainly for the enabled/disabled as it easily can be read in a wrong context:
- `vpnEnabled` -> `vpnEntitlementPresent`
- `vpnDisabled` -> `vpnEntitlementMissing`
- `signedIn` -> `didSignIn`
- `signedOut` -> `didSignOut`

miasma13 (approved):
Overall looks good, suggested some naming changes and proposed additional parameter to the pixel. If anything is unclear feel free to reach out to me.

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
diegoreymendez: What's the rationale / win?

We'd still need to calculate the stringified class name, ideally in a shared location.  That shared location might as well be within the pixel?

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
diegoreymendez: I think "no subscription" would be misread as claiming there's no subscription available - this is truly an error condition, and either "unknown" or "error" are more fitting IMO.

I'd consider sticking with "unknown" as it really conveys what we know (ie: we don't know).

macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionSubscriptionEventHandler.swift
diegoreymendez: I can adjust this parameters to be an enum explaining source.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
diegoreymendez: I've just added tests for checking the naming works well.

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
diegoreymendez: Actually I'm not sure this is correct: this is a pixel that's has a clear semantic meaning: the subscription notification has told the VPN it needs to be disabled.

This is a contract that's clear and semantically correct.  That said if notifications are firing at the wrong times then that's a separate issue - but the fact remains that the VPN has determined it needs to be disabled.

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
miasma13: No strong opinion here, the fact is that `nil` value means that there is no subscription present due to an error or because the code runs when no subscription is on the device.

SharedPackages/VPN/Sources/VPN/Subscription/VPNSubscriptionNotificationPixel.swift
miasma13: I don't agree. The notification is about change of entitlements. I find having pixel events named `vpnEnabled` and `vpnDisabled` confusing, as `vpnEnabled` does not mean that VPN was running at that time or was turned on as a result of the event. I would not focus on what the logic of the class does (e.g. disabling the VPN if it was running or uninstalling) but on a raw meaning of the event that e.g. "code was notified that entitlements for the VPN are present or missing"

PR #1178, opened 2025-06-27
Fix DuckPlayer test timing issues

afterxleep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204099484721401/task/1210657037372047
Tech Design URL: N/A
CC: N/A

### Description
Fixed timeout failures in DuckPlayerNativeUIPresenterTests that were causing CI failures. Tests were not accounting for asynchronous operations in the presenter's dismissPublisher logic, which includes a 0.3-second delay before presenting re-entry pills.

### Testing Steps
- CI should be 🥬 

### Impact and Risks
**Low**: Minor timing adjustments to test infra

#### What could go wrong?
- Tests might run slightly longer due to increased timeouts, but this is necessary to avoid flaky failures

### Quality Considerations
- Maintained existing test coverage and assertions


afterxleep:
> I don't quite like this async after approach to tests, as event based ones would be deterministic and this is prone for timing issues.

Understand.  DuckPlayer has inherent presentation delays and async behavior so this the most pragmatic approach and  reflects real-world behavior - The UI presentation itself has these delays, so tests accommodate the actual timing characteristics

PR #1181, opened 2025-06-27
Fix threading issue in subscription test code

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205237866452338/task/1210655849349291?focus=true
Tech Design URL:
CC:

### Description
Fix threading issue in subscription test code

### Testing Steps
1. Run repeatedly (100/1000/10000) affected test cases:
- SubscriptionPagesUseSubscriptionFeatureTests 
- SubscriptionPagesUseSubscriptionFeatureForStripeTests

### Impact and Risks

#### What could go wrong?
None, fixing the tests.

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGoTests/Subscription/SubscriptionPagesUseSubscriptionFeatureTests.swift
+++ b/iOS/DuckDuckGoTests/Subscription/SubscriptionPagesUseSubscriptionFeatureTests.swift
@@ -114,7 +114,9 @@ final class SubscriptionPagesUseSubscriptionFeatureTests: XCTestCase {
                 let pixelName = path.dropping(prefix: "/t/")
                     .dropping(suffix: "_ios_phone")
                     .dropping(suffix: "_ios_tablet")
-                self.pixelsFired.append(pixelName)
+                Task { @MainActor in
+                    self.pixelsFired.append(pixelName)
+                }
             }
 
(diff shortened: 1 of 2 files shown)


iOS/DuckDuckGoTests/Subscription/SubscriptionPagesUseSubscriptionFeatureTests.swift
```diff
@@ -114,7 +114,9 @@ final class SubscriptionPagesUseSubscriptionFeatureTests: XCTestCase {
                 let pixelName = path.dropping(prefix: "/t/")
                     .dropping(suffix: "_ios_phone")
                     .dropping(suffix: "_ios_tablet")
-                self.pixelsFired.append(pixelName)
+                Task { @MainActor in
```
diegoreymendez, line 117: So.  I've run 10k tests twice without failures with this fix, so it's clearly working well.

That said, I think the AI is right and if this task took long enough to start, the tests will definitely fail too since they don't wait to check expectations - the check executes immediately.

One way to resolve this is to wait for expectations in each test, and fulfill expectations within the fire callback.

That said, given tests are succeeding I don't think this is a high priority change, so I'm approving as-is.

We should consider scoping a follow-up task though.

PR #1194, opened 2025-06-30
Add background task to Statistics Loader

bwaresiak:
Task/Issue URL: https://app.asana.com/1/137249556945/project/856498667320406/task/1210672892676451?focus=true
Tech Design URL:
CC:

### Description

Statistic Loader can trigger two async operations (fetches) that could take some time to complete. When done, notification is posted, that can invoke `AutofillPixelReporter` and cause SecureVault fetches. If that coincide with transition from Background to Suspended, app is getting killed as it is not allowed to hold an active lock on a DB file.

### Testing Steps

This is actually impossible to trigger with debugger attached, and nearly impossible to trigger in general unless you prolong the operations - so just sanity check for the logic is needed here.

Current setup does not cancel the in flight requests in case of the timeout, assuming it will be extremely rare situation for that handler to be invoked.

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/iOS/Core/QRunInBackgroundAssertion.swift
+++ b/iOS/Core/QRunInBackgroundAssertion.swift
@@ -1,0 +1,123 @@
+//
+//  QRunInBackgroundAssertion.swift
+//  DuckDuckGo
+//
+//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


iOS/Core/QRunInBackgroundAssertion.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    deinit {
+        // We don’t apply this assert because it’s hard to force the last object
+        // reference to be released on the main thread.  However, it should be
+        // safe to call through to `consumeValidTaskID(_:)` because no other
+        // thread can be running inside this object (because that would have its
+        // own retain on us).
+        //
+        // dispatchPrecondition(condition: .onQueue(.main))
```
brindy: In order to reduce confusion, imo it's better to remove the code and say "we don't check the main thread condition because..."

iOS/Core/QRunInBackgroundAssertion.swift
```diff
@@ -0,0 +1,123 @@
...
+        self.consumeValidTaskID { }
+    }
+
+    private func consumeValidTaskID(_ body: () -> Void) {
+        // Move this check to all clients except the deinitialiser.
+        //
+        // dispatchPrecondition(condition: .onQueue(.main))
```
brindy: As above, better off phrasing it in the comment as it looks like forgotten code at a glance.

PR #1207, opened 2025-07-01
Add missing vpn subscription status pixel

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1210682416409282?focus=true

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -387,6 +388,7 @@ class MainViewController: UIViewController {
         subscribeToEmailProtectionStatusNotifications()
         subscribeToURLInterceptorNotifications()
         subscribeToSettingsDeeplinkNotifications()
+        checkSubscriptionEntitlements()
```
diegoreymendez: Extremely nitpicky, so feel free to ignore.  I'd consider moving this call out before or after the subscribe* calls to make it stand out more.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -1902,10 +1904,34 @@ class MainViewController: UIViewController {
...
+                    frequency: .dailyAndCount)
+                return
+            }
+
+            PixelKit.fire(
+                VPNSubscriptionStatusPixel.vpnFeatureDisabled(
+                    isSubscriptionActive: isSubscriptionActive,
+                    isAuthV2Enabled: isAuthV2Enabled,
+                    source: .clientCheck(sourceObject: self)),
+                frequency: .dailyAndCount)
```
diegoreymendez: This specific pixel will run for all users and give us a huge amount of failures - which will mostly end up being noise in our end.  Additionally it won't only detect changes, but fixed states as well.

I think it would be ideal to remember the last known value like in macOS's `lastKnownEntitlementsExpired` so we can fire this only for actual changes and eliminate all the noise.

I know we don't want additional sources of truth and I'm happy to add any documentation that's necessary to clarify that.  This value would be useful for the pixels I want to add on "wake", so we can only fire changes there as well.

iOS/DuckDuckGo/MainViewController.swift
federicocappelli: done

iOS/DuckDuckGo/MainViewController.swift
federicocappelli: ok, I've changed the iOS approach to copy macOS, I'm still testing though, please have a look in the meantime

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -118,6 +118,9 @@ class MainViewController: UIViewController {
     @UserDefaultsWrapper(key: .syncDidShowSyncPausedByFeatureFlagAlert, defaultValue: false)
     private var syncDidShowSyncPausedByFeatureFlagAlert: Bool
 
+    @UserDefaultsWrapper(key: .networkProtectionEntitlementsExpired, defaultValue: false)
```
diegoreymendez: This is interesting because I don't think we do this in macOS either (we probably should): the default should likely be `true` here as users don't have entitlements on first launch.

Having this as false by default will fire expiration pixels on first launch.

federicocappelli:
@diegoreymendez note [Cursor comment](https://github.com/duckduckgo/apple-browsers/pull/1207#pullrequestreview-2978156413), it is correct, this will be fixed in my bugfix PR where the entitlement check can't throw an error anymore

diegoreymendez (commented):
Adding one comment before testing, so you can see it.

iOS/DuckDuckGo/MainViewController.swift
federicocappelli: - making @UserDefaultsWrapper more generic and cross-platform is on my todo list
- It doesn't seem that `lastKnownEntitlementsExpired` is set to true by default in mac either, is that intentional or did I miss it somewhere?

PR #1250, opened 2025-07-04
Auth v2 Improved migration

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1210718749358230
CC: @miasma13 @diegoreymendez 

### Description

The migration is now performed in 2 steps
1. While using V1 we migrate to V2 and store the new token, but we continue to use V2
2. When V2 feature flag is enabled, we start using V2 which is already migrated

### Testing Steps

Cases:
1. AuthV1, start without subscription, should use V1
2. AuthV2, start without subscription, should use V2
3. AuthV1, subscribed, should use V1
4. Enable AuthV2, subscribed, not migrated, should use V1
5. Restart AuthV2, subscribed, migrated, should use V2
6. (Roll-back) disable V2, restart, should use V1
---
- Install the previous app version and use AuthV2
- subscribe
- Install the new version on top without changing FF
- Everything should work as expected
---
- Install the previous app version and use AuthV1
- subscribe
- Install the new version on top without changing FF
- Everything should work as expected
---
- AuthV1, subscribe
- Enable AuthV2
- Restart the app
- See that in console authv1 token is migrated to V2
- Restart the app
- See that, in console, authv2 token is used
---
- AuthV1, subscribe
- Enable AuthV2
- Enable Network link conditioner: 100% loss
- Restart the app
- See that in console authv1 token is not migrated
- Kill the app
- Disable Network link conditioner
- Restart the app
- See that, in console, subscription **v1**  is still used

### Impact and Risks

High: 
- PPro features could not be enabled or disabled in time after sign in or out if the notifications are not sent
- The user could not be migrated to AuthV2

---
###### Internal references:


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -430,29 +431,88 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             return Application.appDelegate.windowControllersManager
         })
 
-        self.isAuthV2Enabled = featureFlagger.isFeatureOn(.privacyProAuthV2)
```
federicocappelli, line 435: From here to line 515 is just the usual subscription init swapped around V1<>V2 and the new `AuthMigrator `

iOS/PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -550,11 +557,14 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
...
                    defaults: .networkProtectionGroupDefaults,
                    entitlementCheck: entitlementsCheck)
-
         accountManager?.delegate = self
         startMonitoringMemoryPressureEvents()
         observeServerChanges()
         APIRequest.Headers.setUserAgent(DefaultUserAgentManager.duckDuckGoUserAgent)
+
+        Task {
+            await self.subscriptionAuthMigrator.migrateAuthV1toAuthV2IfNeeded()
```
federicocappelli: Attempting migration

iOS/PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -444,61 +445,40 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
 
         var tokenHandler: any SubscriptionTokenHandling
         var entitlementsCheck: (() async -> Result<Bool, Error>)
-        Self.isAuthV2Enabled = settings.isAuthV2Enabled
```
federicocappelli, line 447: From here to line 534 is just the usual subscription init swapped around V1<>V2 and the new `AuthMigrator `

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
```diff
@@ -154,75 +156,30 @@ final class AppDependencyProvider: DependencyProvider {
                                          pixelNameSuffixes: DailyPixel.Constant.legacyDailyPixelSuffixes,
                                          withAdditionalParameters: parameters)
         }
-        self.isAuthV2Enabled = featureFlagger.isFeatureOn(.privacyProAuthV2)
```
federicocappelli, line 157: From here to line 282 is just the usual subscription init swapped around V1<>V2 and the new `AuthMigrator `

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -163,13 +124,13 @@ extension DefaultSubscriptionManagerV2: SubscriptionAuthV1toV2Bridge {
 
     public func isFeatureEnabledForUser(feature: Entitlement.ProductName) async -> Bool {
         do {
-            guard let tokenContainer = try self.oAuthClient.currentTokenContainer() else {
-                return false
-            }
-            return tokenContainer.decodedAccessToken.subscriptionEntitlements.contains(where: { $0.product == feature })
+            guard isUserAuthenticated else { return false }
+            let tokenContainer = try await getTokenContainer(policy: .localValid)
```
federicocappelli, line 128: This guarantees that the token is still valid

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -257,28 +278,6 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
     // MARK: - Subscription
 
-    func migrateAuthV1toAuthV2IfNeeded() async {
```
federicocappelli, line 260: Migration moved to `AuthMigrator`

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -162,7 +177,13 @@ public protocol SubscriptionManagerV2: SubscriptionTokenProvider, SubscriptionAu
 /// Single entry point for everything related to Subscription. This manager is disposable, every time something related to the environment changes this need to be recreated.
 public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
-    var oAuthClient: any OAuthClient
+    @globalActor
+    public struct CacheActor {
```
federicocappelli: New CacheActor made for serialising any access to the entitlements and isUserAuthenticated cache

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
```diff
@@ -154,75 +156,30 @@ final class AppDependencyProvider: DependencyProvider {
...
+        let authClient = DefaultOAuthClient(tokensStorage: tokenStorageV2,
+                                            legacyTokenStorage: legacyAccountStorage,
+                                            authService: authService)
+        let pixelHandler = AuthV2PixelHandler(source: .mainApp)
+        let isAuthV2Enabled = featureFlagger.isFeatureOn(.privacyProAuthV2)
+        subscriptionAuthMigrator = AuthMigrator(oAuthClient: authClient,
+                                                    pixelHandler: pixelHandler,
+                                                    isAuthV2Enabled: isAuthV2Enabled)
+
+        shouldUseAuthV2 = isAuthV2Enabled && subscriptionAuthMigrator.isMigrated
```
diegoreymendez: Two comments here:
- Semantically this sounds like users that go straight into v2 would never get this to succeed, because they never migrated.  Of course you can go into the implementation and check - but it'd be nice for that to be more transparent to the reader.
- We're checking `isAuthV2Enabled` both here and inside `isMigrated`.
- The would "should" implies there's an option to not use auth v2... we're either using auth v2 or not, so a stronger wording would help clarify.

There are two ways in which we could address this.

```swift
isUsingAuthV2 = subscriptionAuthMigrator.isReadyToUseAuthV2
```

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AuthMigrator.swift
```diff
@@ -0,0 +1,67 @@
...
+        if isAuthV2Enabled {
+            if oAuthClient.isMigrationPossible {
+                return oAuthClient.isUserAuthenticated
+            } else {
+                return true
+            }
+        } else {
+            return oAuthClient.isUserAuthenticated
+        }
+    }
```
diegoreymendez: This method is confusing to me for a couple of reasons:
- This method isn't really checking if there was an auth migration (semantics).  In fact it can succeed even if `isMigrationPossible` returns `false`.
- This method could method really be simplified to just `return oAuthClient.isUserAuthenticated` right now.

I think we can make this code much easier to read and maintain.

```swift
public var isReadyToUseAuthV2: Bool {
    guard isAuthV2Enabled else {
        return false
    }

    return oAuthClient.isUserAuthenticated
}
```

I'm assuming you want to check for authV2 here, since it's passed to this class' initializer, but the authV2 check can happen either outside or inside this method.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -521,8 +509,8 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
     public func exchange(tokenV1: String) async throws -> TokenContainer {
         let tokenContainer = try await oAuthClient.exchange(accessTokenV1: tokenV1)
-        cachedIsUserAuthenticated = true
-        cachedUserEntitlements = tokenContainer.decodedAccessToken.subscriptionEntitlements
+            cachedIsUserAuthenticated = true
+            cachedUserEntitlements = tokenContainer.decodedAccessToken.subscriptionEntitlements
```
diegoreymendez, line 513: Indentation seems off here and in a few lines below too.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AuthMigrator.swift
federicocappelli: I'm ok with the name change.
Unfortunately, the code change doesn't take into account if the user has AuthV2 FF enabled but is not subscribed.
This is why I introduce the oAuthClient.isMigrationPossible check

federicocappelli: p.s I'm trying to adapt it

diegoreymendez: So the idea is that a user with FF-enabled, but not yet a subscriber would be "using" auth V2 classes, is my understanding correct?

federicocappelli: Correct, it will be using all AuthV2 classes

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: fixed

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AuthMigrator.swift
federicocappelli: I refactored the method and changed some function names, have a look if this looks clearer

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
federicocappelli: external isAuthV2Enabled check removed

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AuthMigrator.swift
```diff
@@ -0,0 +1,71 @@
...
+            // AuthV2 enabled but the user is not logged in, this could be:
+            if oAuthClient.isV1TokenPresent {
+                return false // Migration not performed yet
+            } else {
+               return true // User without a subscription
+            }
+        case (false, _):
+            return false // AuthV2 FF disabled
+        }
+    }
```
diegoreymendez, line 47: For clarity: the reason I'm so focused on this is because I believe this is going to be the break-it or make-it snippet of code.

So I asked AI to help a bit reduce this logic a bit more, to help me make it more readable.  Can we consider using this?  Again this shouldn't change the logic you have - just focus on make the flow slightly clearer.

```swift
public var isReadyToUseAuthV2: Bool {
    guard isAuthV2Enabled else {
        return false // No auth V2 FF means no auth v2 at all
    }
    
    if oAuthClient.isUserAuthenticated {
        return true // Auth V2 FF enabled + user authenticated means clearly the user is on authV2
    }

    let isUserLoggedIntoAuthV1 = oAuthClient.isV1TokenPresent
    return !isUserLoggedIntoAuthV1 // Users not logged into auth V1 go straight to auth V2
}
```

**Question:** I believe this works well as long as we clear out the v1 token when logging out - otherwise a V2 user with a V1 token will never be able to log out.  Is that the case?

federicocappelli: - I agree, this simple piece of logic is fundamental.
- IMO the AI code is absolutely correct, but just a reshuffle of my code. I still find the version with the switch easier to read and more like a table with all the permutations, but... 🤷 
- Correct, the V1 token is also removed when v2 logs out and the v2 token is removed as soon v1 is used (in case of rollback)

diegoreymendez: Ok thanks, I'll just double check this while testing.

diegoreymendez (changes requested):
This is mostly working fine for me, expect when rolling back to v1 the extension logs are still reporting they're on V2.

SharedPackages/VPN/Sources/VPN/PacketTunnelProvider.swift
```diff
@@ -273,12 +273,12 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     private let tokenHandlerProvider: () -> any SubscriptionTokenHandling
     @objc
-    public static var isAuthV2Enabled: Bool {
+    public static var shouldUseAuthV2: Bool {
```
diegoreymendez: Could we rename to `isUsingAuthV2`?  The word "should" makes it sound optional.

iOS/PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -444,61 +445,40 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
...
+        }
+        let legacyAccountStorage = SubscriptionTokenKeychainStorage(keychainType: .dataProtection(.named(subscriptionAppGroup)))
+        let authClient = DefaultOAuthClient(tokensStorage: tokenStorage,
+                                            legacyTokenStorage: legacyAccountStorage,
+                                            authService: authService)
+        let pixelHandler = AuthV2PixelHandler(source: .systemExtension)
+        subscriptionAuthMigrator = AuthMigrator(oAuthClient: authClient,
+                                                    pixelHandler: pixelHandler,
+                                                    isAuthV2Enabled: settings.isAuthV2Enabled)
+        Self.shouldUseAuthV2 = subscriptionAuthMigrator.isReadyToUseAuthV2
```
diegoreymendez: IMHO  the issue is here - the extension shouldn't migrate to authV2 on its own, but rely on the parent app to decide all this for it.  The extension shouldn't do a migration at all, ever.

diegoreymendez (commented):
Added a comment that might explain the issue.

SharedPackages/VPN/Sources/VPN/PacketTunnelProvider.swift
federicocappelli: changed

macOS/DuckDuckGo/NetworkProtection/NetworkExtensionTargets/NetworkExtensionTargets/MacPacketTunnelProvider.swift
```diff
@@ -593,12 +594,6 @@ final class MacPacketTunnelProvider: PacketTunnelProvider {
     override func startTunnel(options: [String: NSObject]? = nil) async throws {
 
         try await super.startTunnel(options: options)
-
-        if !Self.isAuthV2Enabled {
-            // Auth V2 cleanup in case of rollback
-            Logger.subscription.debug("Cleaning up Auth V2 token")
```
miasma13, line 599: Is it no longer needed?

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
miasma13: Maybe not within the scope of this PR but an idea for the future, wouldn't it be enough to just rely on `cachedUserEntitlements` here?

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -163,13 +124,13 @@ extension DefaultSubscriptionManagerV2: SubscriptionAuthV1toV2Bridge {
...
+            return tokenContainer.decodedAccessToken.subscriptionEntitlements.contains(feature.subscriptionEntitlement)
         } catch {
             // Fallback to the cached user entitlements in case of keychain reading error
-            return self.cachedUserEntitlements.contains(where: { $0.product == feature })
+            Logger.subscription.debug("Failed to read user entitlements from keychain: \(error, privacy: .public)")
+            return await self.cachedUserEntitlements.contains(feature.subscriptionEntitlement)
```
miasma13: WDYT about adding pixel for this case where we failed to use the designed path to read the entitlements?

iOS/PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
miasma13: Should we even attempt migration from non-app context, I think in the standard `SubscriptionManager` init the `legacyTokenStorage` is set to nil by default.

macOS/DuckDuckGo/NetworkProtection/NetworkExtensionTargets/NetworkExtensionTargets/MacPacketTunnelProvider.swift
federicocappelli: Well spotted, thanks, reverted

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
federicocappelli: do you mean in case `getTokenContainer` fails? we have that and we send the error as param: https://grafana.duckduckgo.com/goto/fpnuAysNg?orgId=1 I'm sure if we digg in the errors we can see if we have keychain errors

diegoreymendez:
Ok, it seems it bas a small one liner regression that's now resolved.  Back to testing, will update with more feedback soon.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AuthMigrator.swift
diegoreymendez: Marking this as resolved to clean up a bit.

diegoreymendez (approved):
Works well for me.  Let's move forward and merge.

PR #1264, opened 2025-07-07
Dynamic new tab page omnibar availability

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1210694811141075?focus=true

### Description
 Making omnibar available on new tab page dynamicly based on the feature flag

### Testing Steps

1. _TBD_
2.

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/Common/Localizables/UserText.swift
+++ b/macOS/DuckDuckGo/Common/Localizables/UserText.swift
@@ -1276,6 +1276,7 @@ struct UserText {
 
     // Favorites
     static let newTabFavoriteSectionTitle = NSLocalizedString("newTab.favorites.section.title", value: "Favorites", comment: "Title of the Favorites section in the home page")
+    static let newTabOmnibarSectionTitle = NSLocalizedString("newTab.favorites.section.omnibar", value: "Search & Duck.ai", comment: "Title of the Search & Duck.ai section in the home page")
 
     // Set Up
     static let newTabSetUpDefaultBrowserCardTitle = NSLocalizedString("newTab.setup.default.browser.title", value: "Default to Privacy", comment: "Title of the Default Browser card of the Set Up section in the home page")
--- a/macOS/DuckDuckGo/NewTabPage/NewTabPageSectionsAvailabilityProvider.swift
+++ b/macOS/DuckDuckGo/NewTabPage/NewTabPageSectionsAvailabilityProvider.swift
@@ -1,0 +1,59 @@
+//
+//  NewTabPageSectionsAvailabilityProvider.swift
+//
+//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import NewTabPage
+import BrowserServicesKit
+import Combine
+import FeatureFlags
+
+extension Notification.Name {
+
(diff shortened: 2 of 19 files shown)


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -346,10 +346,41 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         )
 #endif
 
+        let featureFlagger: FeatureFlagger
```
tomasstrba, line 349: This part is just moved from below to be able to pass `FeatureFlagger` to `AppearancePreferences`

macOS/DuckDuckGo/Common/Localizables/UserText.swift
```diff
@@ -1290,6 +1290,7 @@ struct UserText {
 
     // Favorites
     static let newTabFavoriteSectionTitle = NSLocalizedString("newTab.favorites.section.title", value: "Favorites", comment: "Title of the Favorites section in the home page")
+    static let newTabOmnibarSectionTitle = NSLocalizedString("newTab.favorites.section.omnibar", value: "Search", comment: "Title of the Search section in the home page")
```
tomasstrba, line 1293: In C-S-S 10.6.0, it’s called `Search` instead of `Search & Duck.ai` so I’ve aligned it with the newer name.

macOS/DuckDuckGo/NewTabPage/NewTabPageSectionsAvailabilityProvider.swift
```diff
@@ -0,0 +1,59 @@
...
+            return
+        }
+
+        overridesHandler.flagDidChangePublisher
+            .filter { $0.0 == .newTabPageOmnibar }
+            .sink { _ in
+                NotificationCenter.default.post(name: .newTabPageSectionsAvailabilityDidChange, object: nil)
+            }
+            .store(in: &cancellables)
+    }
```
ayoy, line 53: This would only work for feature flag overrides (not for privacy config updates), hence it would only help internal users. I think we should be fine skipping the subscription and the notification (also in `AppearancePreferences`) – for production release, it will be completely OK to only apply the new flag state on new window creation (that’s how the HTML NTP feature flag was rolled out, for instance). So we can have simpler code and just ask internal users to reopen a browser window or restart the app.

ayoy (commented):
@tomasstrba the code is looking pretty great! I’m gonna test it out later, but dropping a comment for now to indicate that we don’t need to listen to flag overrides and can rely on users opening a new app window. This will simplify the implementation by quite a bit and we’ve followed that path already in the past. Let me know what you think. Thanks!

macOS/LocalPackages/NewTabPage/Tests/NewTabPageTests/NewTabPageConfigurationClientTests.swift
```diff
@@ -35,6 +36,7 @@ final class NewTabPageConfigurationClientTests: XCTestCase {
         contextMenuPresenter = CapturingNewTabPageContextMenuPresenter()
         eventMapper = CapturingNewTabPageConfigurationEventHandler()
         client = NewTabPageConfigurationClient(
+            sectionsAvailabilityProvider: MockNewTabPageSectionsAvailabilityProvider(),
```
ayoy: We could add tests to cover omnibar not being available, by adding a variant of `testThatInitialSetupReturnsConfiguration`.

ayoy (approved):
Works great! 💪  I’ve added a comment to add a test case to configuration client tests to cover the case when omnibar isn’t available. Feel free to merge when it’s ready.

One final thing – I’d move this PR to the release branch so that we have it in the latest release. I know there’s not much to test, but if it goes to main, you’ll have to keep adding code to main for the remainder of the week, even if you end up having something shippable internally.

macOS/DuckDuckGo/NewTabPage/NewTabPageSectionsAvailabilityProvider.swift
tomasstrba: _Cross-posting my answer from Asana for future reference:_
If there is a percentage of users, who have the window open for ages, they won’t get the search feature until they restart the window, which is not ideal. Let’s keep the code for internal users there for now, we can remove it once the feature flag becomes global. If there is enough time, perhaps we could think about how to notify UI layer when the remote feature flag is changed

PR #1289, opened 2025-07-08
Change modes via swipe

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210641337067869?focus=true

### Description
Allow users to change chat/search mode by swiping the content area

### Testing Steps
1. Swipe left and right and check if the state change is being respected
2. Change the state using the toggle at the top to confirm it is still working
3. Test with and without favorites to see if we're displaying them when we should (ie. show favorites when no text is in the suggestion tray)

### Impact and Risks

#### What could go wrong?
Broken animation when swiping 
Not setting correct state when swiping

### DIFF
--- a/iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/DaxLogoManager.swift
+++ b/iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/DaxLogoManager.swift
@@ -1,0 +1,89 @@
+//
+//  DaxLogoManager.swift
+//  DuckDuckGo
+//
+//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import UIKit
+import SwiftUI
+
+/// Manages the Dax logo view display and positioning
+final class DaxLogoManager {
+    
+    // MARK: - Types
+    
+    struct Configuration {
+        let logoOffset: CGFloat
+        
+        static let `default` = Configuration(logoOffset: 18)
+    }
+    
+    // MARK: - Properties
+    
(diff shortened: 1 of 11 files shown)


iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/DaxLogoManager.swift
```diff
@@ -0,0 +1,89 @@
...
+        
+        NSLayoutConstraint.activate([
+            hostingController.view.centerXAnchor.constraint(equalTo: viewController.view.centerXAnchor),
+            logoCenterYConstraint!
+        ])
+        
+        hostingController.didMove(toParent: viewController)
+        
+        // Send to back so it appears behind other views
+        viewController.view.sendSubviewToBack(hostingController.view)
```
dus7: Technically this should be before `hostingController.didMove(toParent: viewController)`.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/FlowManagers/KeyboardAdjustmentManager.swift
```diff
@@ -0,0 +1,128 @@
...
+        
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(keyboardWillHide(_:)),
+            name: UIResponder.keyboardWillHideNotification,
+            object: nil
+        )
+    }
+    
+    @objc private func keyboardWillShow(_ notification: Notification) {
```
dus7, line 83: I'm wondering why we're not using `UIKeyboardLayoutGuide` instead. It probably would be much simpler.

dus7 (approved):
Tested. Works as advertised, however I have noticed a few minor things that could be improved, but can be done at the later phase:
1. It looks somewhat weird that the Dax logo from Duck.ai page is static while the search content is swiping over it.
2. When I'm switching via the switcher it's not scrolling - I'd expect those two methods to be equal transition-wise.
3. The webRag button is just appearing, would be nice for it to fade at least.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/DaxLogoManager.swift
Bunn: fixed

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/FlowManagers/KeyboardAdjustmentManager.swift
Bunn: No particular reason, this was extracted from other places in the app

PR #936, opened 2025-06-03
AuthV2 - isUserAuthenticated and entitlements check improvements

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1210449733678950
CC: @miasma13 

### Description

- `isUserAuthenticated` now relies on cached (UserDefaults) values and can't fail
- `func isFeatureEnabledForUser(feature: Entitlement.ProductName)` The entitlements check used by the VPN, is now relying on cached user entitlements and can't fail

### Testing Steps

Subscription + VPN functioning as expected

### Impact and Risks

High, the subscription and VPN could be impacted

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #1299, opened 2025-07-08
Fix v1 entitlements regression

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1210746879873137?focus=true

### Description

Fixes a regression with entitlement checks on AuthV1.

The issue was introduced in [this line](https://github.com/duckduckgo/apple-browsers/pull/936/files#diff-ef31b1cf1db6fe730b6cadd9926e1466fd19b34e0fd567a3144ee6ecee597cf3R90) and [this line](https://github.com/duckduckgo/apple-browsers/pull/936/files#diff-ef31b1cf1db6fe730b6cadd9926e1466fd19b34e0fd567a3144ee6ecee597cf3R95).

Local cache expires after 20 minutes, so the first line would often return an error about no cache being available.  The second line would then proceed to inform there are no entitlements to the app.

### Testing Steps

1. Throw an error here.
2. Start the app.
3. See that the VPN and DBP are disabled.

### Impact and Risks

Medium

#### What could go wrong?

This is making changes to AuthV1 logic to fix a regression.

### Quality Considerations

NA

### Notes to Reviewer

NA

---
###### Internal references:

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/API/SubscriptionEndpointServiceV2.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/API/SubscriptionEndpointServiceV2.swift
@@ -58,17 +58,12 @@ public enum SubscriptionCachePolicy {
     /// Returns the cached subscription if it exists; otherwise, attempts to fetch from the remote source.
     case cacheFirst
 
-    /// Returns only the cached subscription. No remote fetch is attempted.
-    case cacheOnly
-
     public var apiCachePolicy: APICachePolicy {
         switch self {
         case .remoteFirst:
             return .reloadIgnoringLocalCacheData
         case .cacheFirst:
             return .returnCacheDataElseLoad
-        case .cacheOnly:
-            return .returnCacheDataDontLoad
         }
     }
 }
@@ -211,7 +206,11 @@ New: \(subscription.debugDescription, privacy: .public)
             } catch SubscriptionEndpointServiceError.noData {
                 throw SubscriptionEndpointServiceError.noData
             } catch {
-                return try await getSubscription(accessToken: accessToken, cachePolicy: .cacheOnly)
+                if let cachedSubscription = getCachedSubscription() {
+                    return cachedSubscription
+                } else {
+                    throw SubscriptionEndpointServiceError.noData
+                }
             }
 
         case .cacheFirst:
@@ -219,13 +218,6 @@ New: \(subscription.debugDescription, privacy: .public)
                 return cachedSubscription
             } else {
                 return try await getRemoteSubscription(accessToken: accessToken)
-            }
-
-        case .cacheOnly:
-            if let cachedSubscription = getCachedSubscription() {
-                return cachedSubscription
-            } else {
-                throw SubscriptionEndpointServiceError.noData
             }
         }
     }
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AccountManager.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AccountManager.swift
@@ -256,13 +256,6 @@ public final class DefaultAccountManager: AccountManager {
             } else {
                 return await fetchRemoteEntitlements()
             }
-
-        case .returnCacheDataDontLoad:
-            if let cachedEntitlements: [Entitlement] = entitlementsCache.get() {
-                return .success(cachedEntitlements)
-            } else {
-                return .failure(EntitlementsError.noCachedData)
-            }
         }
 
     }
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AccountManager.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AccountManager.swift
@@ -209,7 +209,7 @@ public final class DefaultAccountManager: AccountManager {
     private func fetchRemoteEntitlements() async -> Result<[Entitlement], Error> {
         guard let accessToken else {
             entitlementsCache.reset()
-            return .failure(EntitlementsError.noAccessToken)
+            return .success([])
         }
 
         switch await authEndpointService.validateToken(accessToken: accessToken) {
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
@@ -315,13 +315,10 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
                 throw SubscriptionEndpointServiceError.noData
             } catch {
                 // Failed to get a valid token, fall back on cache
-                subscription = try await subscriptionEndpointService.getSubscription(accessToken: nil, cachePolicy: .cacheOnly)
+                subscription = try await subscriptionEndpointService.getSubscription(accessToken: nil, cachePolicy: .cacheFirst)
                 break
             }
             subscription = try await subscriptionEndpointService.getSubscription(accessToken: tokenContainer.accessToken, cachePolicy: cachePolicy)
-
-        case .cacheOnly:
-            subscription = try await subscriptionEndpointService.getSubscription(accessToken: nil, cachePolicy: cachePolicy)
         }
 
         if subscription.isActive {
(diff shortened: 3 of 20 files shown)


SharedPackages/BrowserServicesKit/Sources/Subscription/API/SubscriptionEndpointServiceV2.swift
```diff
@@ -220,13 +219,6 @@ New: \(subscription.debugDescription, privacy: .public)
             } else {
                 return try await getRemoteSubscription(accessToken: accessToken)
             }
-
-        case .cacheOnly:
-            if let cachedSubscription = getCachedSubscription() {
-                return cachedSubscription
-            } else {
-                throw SubscriptionEndpointServiceError.noData
-            }
```
diegoreymendez, line 229: This goes away.  Caching expires after 20 minutes so we can't rely on that as a source of truth.

Caching serves the purpose of avoiding unnecessary network calls, so `cacheFirst` is appropriate and enough (and safer).

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/AccountManager.swift
```diff
@@ -209,7 +209,7 @@ public final class DefaultAccountManager: AccountManager {
     private func fetchRemoteEntitlements() async -> Result<[Entitlement], Error> {
         guard let accessToken else {
             entitlementsCache.reset()
-            return .failure(EntitlementsError.noAccessToken)
+            return .success([])
```
samsymons, line 212: Nice, good catch - we should definitely treat this case as not having an entitlement, since we want to ignore most errors as something that can happen transiently.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -30,7 +30,7 @@ public protocol SubscriptionAuthV1toV2Bridge: SubscriptionTokenProvider, Subscri
     func isFeatureAvailableAndEnabled(feature: Entitlement.ProductName, cachePolicy: APICachePolicy) async throws -> Bool
     /// If the user is allowed to use the feature, base on the TokenContainer entitlements
     /// This is used by VPN and PIR
-    func isFeatureEnabledForUser(feature: Entitlement.ProductName) async -> Bool
+    func isFeatureEnabledForUser(feature: Entitlement.ProductName) async throws -> Bool
```
diegoreymendez, line 33: `true` means there are entitlements, `false` means no entitlements, and error is for recoverable errors such as network failures.  We should **NEVER** conflate recoverable errors with regular results.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -62,13 +62,18 @@ extension SubscriptionAuthV1toV2Bridge {
...
         case .success(let hasEntitlements):
             return hasEntitlements
-        case .failure:
-            return false
+        case .failure(let error):
+            switch error {
+            case APIServiceError.invalidToken:
+                return false
+            default:
+                throw error
```
samsymons, line 75: 🎉

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -62,13 +62,18 @@ extension SubscriptionAuthV1toV2Bridge {
...
+        let result = await accountManager.hasEntitlement(forProductName: feature, cachePolicy: .returnCacheDataElseLoad)
         switch result {
         case .success(let hasEntitlements):
             return hasEntitlements
-        case .failure:
-            return false
+        case .failure(let error):
+            switch error {
+            case APIServiceError.invalidToken:
+                return false
```
diegoreymendez, line 73: This is to avoid having to do a larger refactor in the API service... but in truth this would be best handled closer to the API call.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
diegoreymendez: Yeah, this is one of the two lines that was causing the regression, the other one is above.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -62,13 +62,18 @@ extension SubscriptionAuthV1toV2Bridge {
 
 extension DefaultSubscriptionManager: SubscriptionAuthV1toV2Bridge {
 
-    public func isFeatureEnabledForUser(feature: Entitlement.ProductName) async -> Bool {
-        let result = await accountManager.hasEntitlement(forProductName: feature, cachePolicy: .returnCacheDataDontLoad)
+    public func isFeatureEnabledForUser(feature: Entitlement.ProductName) async throws -> Bool {
+        let result = await accountManager.hasEntitlement(forProductName: feature, cachePolicy: .returnCacheDataElseLoad)
```
diegoreymendez, line 66: This is part of the original regression.

samsymons (approved):
Fix looks good to me once CI is green - thanks!

SharedPackages/BrowserServicesKit/Tests/SubscriptionTests/API/SubscriptionEndpointServiceV2Tests.swift
```diff
@@ -117,17 +117,6 @@ final class SubscriptionEndpointServiceV2Tests: XCTestCase {
...
-        do {
-            _ = try await endpointService.getSubscription(accessToken: "token", cachePolicy: .cacheOnly)
-            XCTFail("Expected noData error")
-        } catch SubscriptionEndpointServiceError.noData {
-            // Success
-        } catch {
-            XCTFail("Unexpected error: \(error)")
-        }
-    }
-
```
samsymons, line 130: This is no longer accurate with the caching change, so I removed the test

SharedPackages/BrowserServicesKit/Sources/Subscription/API/SubscriptionEndpointServiceV2.swift
federicocappelli: @diegoreymendez The cache doesn't expires in 20 minutes, but it expires when the subscription expires, see `SubscriptionEndpointServiceV2.swift:179`

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -315,13 +315,10 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
                 throw SubscriptionEndpointServiceError.noData
             } catch {
                 // Failed to get a valid token, fall back on cache
-                subscription = try await subscriptionEndpointService.getSubscription(accessToken: nil, cachePolicy: .cacheOnly)
+                subscription = try await subscriptionEndpointService.getSubscription(accessToken: nil, cachePolicy: .cacheFirst)
```
federicocappelli, line 318: Here we don't pass the token, the end result is identical to the now defunct `.cacheOnly`, no changes. See `SubscriptionEndpointServiceV2.swift:198`

SharedPackages/BrowserServicesKit/Sources/Subscription/API/SubscriptionEndpointServiceV2.swift
diegoreymendez: The regression is in V1 not V2, which is why this is affecting so many users.  Try lowering the expiration to 10 seconds in V1 and you'll be able to reproduce this issue faithfully.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
diegoreymendez: Can you more clearly detail what you're saying here?  My reading is we're fixing a V1 issue and this is likely the same for V2 - so the change is good.

But I imagine that's not what you meant?

federicocappelli: I'm saying that this change has no effect.
Passing `.cacheOnly` without a token or `.cacheFirst` without a token has the same effect.
I'm not saying the entire change is not valid, this is more a note to myself for a cleanup.

federicocappelli:
Do you have the steps for reproducing the bug?

PR #1301, opened 2025-07-09
Malicious site protection datasets serialization fix

alessandroboron:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206329551987282/task/1209638992004956?focus=true
Tech Design URL:
CC:

### Description

I found an issue where background tasks could run in parallel and fetch data while another update was in progress. This pull request ensures that if there are ongoing updates, we skip the background update.

This is an attempt to fix the [Malicious Site Protection crash](https://app.asana.com/1/137249556945/project/1206329551987282/task/1209638992004956?focus=true) we have been monitoring.

Regardless, I believe it would make sense to build a serializer for tasks so that no matter how many tasks are initiated, we can enqueue them and execute them sequentially.

I’m currently targeting the release branch so the fix can reach the users. Let me know if we want to spend more time testing the fix internally first. 

### Testing Steps
Ensure tests pass.

### Impact and Risks
Risk is low given this crashes in some scenarios.

#### What could go wrong?
The fix does not fix the crash

### Quality Considerations
I wrote a test to ensure that if there’s an inflight task and a background task fires it should not fetch the dataset again.

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
+++ b/iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
@@ -93,12 +93,13 @@ extension MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDataset
     func startFetching() -> Task<Void, Never> {
         guard
             canFetchDatasets,
-            !isDatasetsFetchInProgress
+            !isDatasetsFetchInProgress,
+            shouldUpdateHashPrefixSets || shouldUpdateFilterSets
         else {
             return Task {}
         }
 
-        isDatasetsFetchInProgress = shouldUpdateHashPrefixSets || shouldUpdateFilterSets
+        isDatasetsFetchInProgress = true
         Logger.MaliciousSiteProtection.datasetsFetcher.debug("Start Updating Datasets...")
 
         return Task {


alessandroboron:
> ### Bug: Background Task Handler Updates Incorrectly
> The `backgroundRefreshTaskHandler` now calls `startFetching()`, which updates both malicious site protection datasets (hashPrefixSet and filterSet), regardless of the specific `datasetType` that triggered the background task. This leads to unnecessary updates and inconsistent background task rescheduling, as only the original `datasetType`'s task is rescheduled. Furthermore, `scheduleBackgroundRefreshTask(datasetType:)` is called from a `@MainActor` context without being marked `@MainActor`, potentially causing actor isolation issues.
> 
> `iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift#L220-L231`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmVmYWEwOTZjLWQyZjktNGJmZS05M2IzLTMxN2I2OTFhMGI4MSIsImVuY3J5cHRpb25LZXkiOiJYWXlhZjVEeHcybUctSWFqckg0ellyM0h2Z283bFUtb0ZDSHY1SjNWdjBrIiwiYnJhbmNoIjoiYWxlc3NhbmRyby9tYWxpY2lvdXMtc2l0ZS1wcm90ZWN0aW9uLWRhdGFzZXRzLXNlcmlhbGl6YXRpb24taW9zIn0sImlhdCI6MTc1MjA0MjQ2MCwiZXhwIjoxNzUyNjQ3MjYwfQ.IU5RmWs7UJZwdsqvrBRQq2Mynksd5UGOywCN9D50dlM13o49vZNU9UXJwN3oNsg-R9NgqvkfvdBXocBBBIu0ob1vcE45TjWiX85P1vOthVdln5jPgng4HPa8nb0qqGQCsv89wlgtkyrdyqpt-iZfEOp5Bm_OVdkgs4mysL0ewYqLLHrOtJddz45SchewbE_OK9EcyIBKXANQH05ToF73eL2x5gnzc0quR5sEAmE1nRaY4hapscdKwiMsBtIPEcaF886or18VR_K5-OH_Y1ZFM9l0nKA6AFD64lj-EKGsMFMNAXQ3w6h6294I0c0VssbeITdlPUDTuqENhb5TyG2zRw) • [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmVmYWEwOTZjLWQyZjktNGJmZS05M2IzLTMxN2I2OTFhMGI4MSIsImVuY3J5cHRpb25LZXkiOiJYWXlhZjVEeHcybUctSWFqckg0ellyM0h2Z283bFUtb0ZDSHY1SjNWdjBrIiwiYnJhbmNoIjoiYWxlc3NhbmRyby9tYWxpY2lvdXMtc2l0ZS1wcm90ZWN0aW9uLWRhdGFzZXRzLXNlcmlhbGl6YXRpb24taW9zIiwicmVwb093bmVyIjoiZHVja2R1Y2tnbyIsInJlcG9OYW1lIjoiYXBwbGUtYnJvd3NlcnMiLCJwck51bWJlciI6MTMwMSwiY29tbWl0U2hhIjoiZDg2NjAzYmM3ZTdmNzNlNjNjN2UwZjUzYTZiYWQxZTBkOWE0MDNlNyJ9LCJpYXQiOjE3NTIwNDI0NjAsImV4cCI6MTc1MjY0NzI2MH0.kyc_Z6j89ABgyIIBLzBN50a80c3lFMquq27D5Rdind2qHZIMnKYThnAXsuPY1ykBPUf_hCEig8O2bYhMWgoh7oWsjcfxS6v-1kELqn2HTiSZI-etUj3FDZZKK_64On03hLTy1nskxF_pMc3hGYwJ7fGPbpr78B9djQymuoadHORW-PL49-rF62EBMCYNjzPTs1nslOTtlTxYOjNkel3rxHAPDwjXRjk088AA-ULtWyiDgxmGL5W6myhMDjtIsXknGdE0RBHDbQ1mzCdn2l1gTszIW_N6WGTTvwuqmkMbEcgT0MagM_oQHw7MEtZqQHpxeybPyb63XxlV4myEZmVlnQ)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_



> ### Bug: Background Task Handler Fails and Misbehaves
> The `backgroundRefreshTaskHandler` contains multiple issues: The `defer` block, which marks the background task as complete, is placed after `guard let self else { return }`, meaning the task will remain incomplete if `self` is `nil`. Additionally, the `Task` calls the `@MainActor` method `startFetching()` without being marked as `@MainActor`, causing an actor isolation compilation error. Finally, the handler now calls `startFetching()` instead of updating the specific `datasetType` it was scheduled for, broadening the update scope to potentially both `hashPrefixSet` and `filterSet` and removing granular control.
> 
> `iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift#L219-L231`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmU5ODc3N2ViLWUyYTAtNDUwZS1iNWJjLTliZTc1NTc0ZDJlOCIsImVuY3J5cHRpb25LZXkiOiJmTlhuYlljQWx2b3BsM0JzdFNmZWNpbXhqQ1NJenh3OFJyQi1UOU5IcnU0IiwiYnJhbmNoIjoiYWxlc3NhbmRyby9tYWxpY2lvdXMtc2l0ZS1wcm90ZWN0aW9uLWRhdGFzZXRzLXNlcmlhbGl6YXRpb24taW9zIn0sImlhdCI6MTc1MjA0NDEyNiwiZXhwIjoxNzUyNjQ4OTI2fQ.a7XLmFWlzE_aCtqDdL-bMRXN9-wg2l_tStPYBGFwNeYKRKdmyXOZaXOmGbkpwttev8qzjxHCWJ40OTxH-GAcOP-gMTcUypZ4XpYyEVsri9gBQVZgCi8z-Cw2Om9w_r4p-kM04Ey2mo4-GUkYVFVdbdLeekOoubOnzCLxUGV610GlIN6SkYNeD-LOQ8TYOlzgqWB3P7tyc39qhYHRvCDvbb_BFBiTn6aKQaMbsRxZaaN0lKEIu9LkXsSGLH6ZCijZon3NapvjGpYrDRInLZQSnsKTJa869wj9VfCa-_aV7OdzonnocF3Lfv5zIib1LnaR4redbsJYw52e_okEdJPSVQ) • [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmU5ODc3N2ViLWUyYTAtNDUwZS1iNWJjLTliZTc1NTc0ZDJlOCIsImVuY3J5cHRpb25LZXkiOiJmTlhuYlljQWx2b3BsM0JzdFNmZWNpbXhqQ1NJenh3OFJyQi1UOU5IcnU0IiwiYnJhbmNoIjoiYWxlc3NhbmRyby9tYWxpY2lvdXMtc2l0ZS1wcm90ZWN0aW9uLWRhdGFzZXRzLXNlcmlhbGl6YXRpb24taW9zIiwicmVwb093bmVyIjoiZHVja2R1Y2tnbyIsInJlcG9OYW1lIjoiYXBwbGUtYnJvd3NlcnMiLCJwck51bWJlciI6MTMwMSwiY29tbWl0U2hhIjoiYmY2ZmQxMTZjY2YyZDI1YzU3ZmQ4MTk1ZmJiNzU2ZjM5ZDgyM2E3OCJ9LCJpYXQiOjE3NTIwNDQxMjYsImV4cCI6MTc1MjY0ODkyNn0.FnGEOKsRUMiGqN-jBJ0ZT16YQbhLFlSa7fnklOBpnNXbhcMw2OK_8j69yaINrNa07iwdAkDTXS9dUd2Re_qLIJeFItnTzEb-pb4zGdyhZNYEuXtQVlhAlv4OpuDn5URiHLMlfTOxZ8IyNZYImsg-_MnXxQhVpQmxk3joYapgKrYvmp9HrfKxVAuFt9gWTPY1xHAvRahv7gg9dGWgsA4HVkpWTMJLyLqi0OoOgiCGicQngHA_KSYguyFGjzjljfq1BYfFiDGJ9Ud8Fw9iTUhFIj8zGb_pfW_pyJNfEInI8YdhICSyhzK54z_dYT3BroAiQ3p-Bt6mITmmg2BwpPu_qg)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

About calling `startFetching()`, which updates both malicious site protection datasets (hashPrefixSet and filterSet), regardless of the specific `datasetType`. In that method we check if we need to update the datasets. So if there’s no reason we won’t update a dataset that does not need to be updated.

SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/UpdateManager.swift
```diff
@@ -146,7 +146,7 @@ public struct UpdateManager: InternalUpdateManaging {
 
     #if os(iOS)
     public func updateData(datasetType: DataManager.StoredDataType.Kind) -> Task<Void, Never> {
-        Task.detached {
+        Task {
```
bwaresiak, line 149: Can you tell me rationale behind this change?

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
```diff
@@ -90,7 +90,7 @@ extension MaliciousSiteProtectionDatasetsFetcher: MaliciousSiteProtectionDataset
 
     @MainActor
     @discardableResult
-    func startFetching() -> Task<Void, Error> {
+    func startFetching() -> Task<Void, Never> {
```
bwaresiak, line 93: I think there is super slight chance of race condition below - we most likely should check `shouldUpdateHashPrefixSets || shouldUpdateFilterSets` in the guard

alessandroboron: I think if we do that it would be more problematic. Lets consider this scenario:

1. startFetching() is called.
2. The datasets are in the middle of being downloaded but the new update date is not set yet.
3. The background task fires and call startFetching()
4. Because the new update date is not set yet it would pass the guard.

Unless I’m missing something?

bwaresiak: From what I see we have here something similar to double-check locking problem:
- Each call to this function done while there is no fetch pending, will schedule a Task on line 104, even if no work is needed. In such case `isDatasetsFetchInProgress` will be false.
- It means if the status of "work needed" changes between when Task is created and executed (as it s time-based), we will get into state where: 1/ Task is being executed 2/ `isDatasetsFetchInProgress` is false 3/ We in fact start doing updates, while we are not guarded against parallel execution.

This is a timing issue with timing conditions that not likely to happen, but not impossible either. 

The fix I suggested may not be the perfect one, but my intention is to check if we both need to do any work (if not skip) and if any work is being done right now (if yes - also skip). Feel free to suggest different one based on your knowledge :)

SharedPackages/BrowserServicesKit/Sources/MaliciousSiteProtection/Services/UpdateManager.swift
alessandroboron: I think it does not make much difference in this case as It’s a top level task. I thought there’s no actual need to be detached in this case.

iOS/DuckDuckGo/MaliciousSiteProtection/Services/MaliciousSiteProtectionDatasetsFetcher.swift
alessandroboron: Thanks for the explanation, I now understand what you meant. Really good catch!

PR #1305, opened 2025-07-09
PIR: fix debug commands

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1210752661340615?focus=true
Tech Design URL:
CC:

### Description
1/ Debug commands (E.g. execute scans, execute opt outs) currently don't interrupt the queue manager or set up the queue manager state, so frequently don't work. This changes it so they do

Also two different params were no longer being passed to subjobs correctly:
2/ They also no longer correctly show the web view if requested
3/ All scans also no longer correctly pass the isManual flag which allows us to differentiate the first set of scans from others. This affects pixel params.

### Testing Steps

1. Smoke test PIR on macOS
2. Test the PIR debug commands, particularly the show web view versions

### Notes to Reviewer

---
###### Internal references:


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/QueueManaging/BrokerProfileJobQueueManager.swift
```diff
@@ -177,6 +177,8 @@ public final class BrokerProfileJobQueueManager: BrokerProfileJobQueueManaging {
                                           let errorHandler,
                                           let completion) = command else { return }
 
+        cancelCurrentModeAndResetIfNeeded()
+        mode = .immediate(errorHandler: nil, completion: nil)
```
THISISDINOSAUR, line 181: There's an argument for defining a new debug mode here IMO. For now I've gone with immediate since I think that is an accurate description of how we want debug commands to behave anyway

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileOptOutSubJob.swift
```diff
@@ -40,7 +40,7 @@ struct BrokerProfileOptOutSubJob {
 
     public func runOptOut(for extractedProfile: ExtractedProfile,
                           brokerProfileQueryData: BrokerProfileQueryData,
-                          showWebView: Bool = false,
```
THISISDINOSAUR, line 43: I think in general we should be much more careful with default values (and avoid them much more often) to avoid bugs like this

PR #1319, opened 2025-07-10
Fcappelli/invalid token request iinvestigation

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1210747547494908?focus=true
CC: @miasma13 

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -230,13 +226,6 @@ final public actor DefaultOAuthClient: @preconcurrency OAuthClient {
     public func getTokens(policy: AuthTokensCachePolicy) async throws -> TokenContainer {
         let localTokenContainer = try tokenStorage.getTokenContainer()
 
-        if policy != .local,
-           let localTokenContainer,
-            localTokenContainer.decodedRefreshToken.isExpired() {
-            // The refresh token is expired, the token is un-refreshable
-            throw OAuthClientError.refreshTokenExpired
```
federicocappelli, line 237: This error was artificial; we want a failure from the API instead.

SharedPackages/BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -260,9 +249,6 @@ final public actor DefaultOAuthClient: @preconcurrency OAuthClient {
             if localTokenContainer.decodedAccessToken.isExpired() || expiresSoon {
                 Logger.OAuthClient.log("Refreshing local already expired token")
                 return try await getTokens(policy: .localForceRefresh)
-            } else if expiresSoon {
```
federicocappelli, line 263: unreachable if branch

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -22,10 +22,12 @@ import os.log
 
 public actor DeadTokenRecoverer {
 
-    private static var recoveryAttemptCount: Int = 0
+    private var recoveryAttemptCount: Int = 0
```
federicocappelli: all var and func are not instance and not static, so they are isolated in the actor

iOS/PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -475,9 +475,6 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
                                                                    subscriptionEndpointService: subscriptionEndpointService,
                                                                    subscriptionEnvironment: subscriptionEnvironment,
                                                                    pixelHandler: pixelHandler,
-                                                                   tokenRecoveryHandler: {
-                Logger.networkProtection.error("Expired refresh token detected")
```
federicocappelli, line 479: this is the main change, it guarantees that the recovery fails in the AppExt

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -42,24 +44,11 @@ public actor DeadTokenRecoverer {
             }
         case .stripe:
             Logger.subscription.debug("Subscription purchased via Stripe can't be restored automatically, notifying the user...")
-            NotificationCenter.default.post(name: .expiredRefreshTokenDetected, object: self, userInfo: nil)
             throw SubscriptionManagerError.noTokenAvailable
         }
     }
 
-    public static func reportDeadRefreshToken() async throws {
```
federicocappelli, line 50: This was used if the purchase platform is Stripe, and was doing nothing except logging and sending an unused notification.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -495,16 +498,22 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
     }
 
     func attemptTokenRecovery() async throws -> TokenContainer {
+
+        Logger.subscription.log("Attempting token recovery...")
+
         guard let tokenRecoveryHandler else {
+            Logger.subscription.log("Recovery not possible, no handler configured.")
```
federicocappelli, line 502: This is a valid case for all extensions

SharedPackages/BrowserServicesKit/Sources/Subscription/NSNotificationName+Subscription.swift
```diff
@@ -32,7 +32,6 @@ public extension NSNotification.Name {
     static let entitlementsDidChange = Notification.Name("com.duckduckgo.subscription.EntitlementsDidChange")
     static let subscriptionDidChange = Notification.Name("com.duckduckgo.subscription.SubscriptionDidChange")
     static let availableAppStoreProductsDidChange = Notification.Name("com.duckduckgo.subscription.AvailableAppStoreProductsDidChange")
-    static let expiredRefreshTokenDetected = Notification.Name("com.duckduckgo.subscription.ExpiredRefreshTokenDetected")
```
federicocappelli, line 35: unused

macOS/DuckDuckGo/NetworkProtection/NetworkExtensionTargets/NetworkExtensionTargets/MacPacketTunnelProvider.swift
```diff
@@ -471,7 +471,6 @@ final class MacPacketTunnelProvider: PacketTunnelProvider {
                                                                subscriptionEndpointService: subscriptionEndpointServiceV2,
                                                                subscriptionEnvironment: subscriptionEnvironment,
                                                                pixelHandler: pixelHandler,
-                                                               tokenRecoveryHandler: nil,
```
federicocappelli, line 474: redundant, the default is already nil

SharedPackages/BrowserServicesKit/Sources/Networking/Auth/OAuthClient.swift
diegoreymendez: If this was preventing calls to the backend, am I right that we should now expect a rise in backend errors?

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -22,45 +22,29 @@ import os.log
...
+    public func attemptRecoveryFromPastPurchase(purchasePlatform: SubscriptionEnvironment.PurchasePlatform,
+                                                restoreFlow: any AppStoreRestoreFlowV2) async throws {
+        defer {
+            recovering = false
+        }
+
+        guard !recovering else {
             Logger.subscription.debug("Recovery attempt already in progress, skipping...")
-            try reportFailure()
+            return
```
diegoreymendez: This exit point is undefined behavior / non-deterministic: is the token recovered?  is it not recovered?

This will likely result in the caller not really knowing how to continue, probably assuming the token was recovered and trying to move forward.

Maybe we can have a task instead of `recovering` and await its result here?

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -495,16 +498,22 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
...
             throw SubscriptionManagerError.errorRetrievingTokenContainer(error: nil)
```
diegoreymendez: This error seems off to me - I know it's ignored in the parent call, but that's part of what IMHO is confusing about our current error handling.  Could be standardize on `noTokenAvailable` here, and just pass on the error in the parent call?

federicocappelli:
@diegoreymendez 
> If this was preventing calls to the backend, am I right to see we should now expect a rise in backend errors?
Yes, in theory, this will help us size any problem correctly. I would prefer to see the real number of expired refresh token in the backend instead of failing silently.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: Yes, I'm ok with throwing just `SubscriptionManagerError.noTokenAvailable`
Changed

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: I just serialised DeadTokenRecoverer, now every call waits for the previous one to complete, and only one, overall subscription recovery is performed.

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -22,45 +22,56 @@ import os.log
 
 public actor DeadTokenRecoverer {
 
-    private static var recoveryAttemptCount: Int = 0
+    var hasRecovered = false
+    var isRecovering = false
+    private var recoveryWaiters: [CheckedContinuation<Void, Error>] = []
```
diegoreymendez: We're in a much better place now in terms of thread safety.  That said I'm not sure we want more than one recovery attempt to run.

Especially since a large number of calls here can queue up rather quickly.

This recovery mechanism is called from `DefaultSubscriptionManagerV2` when necessary by `getTokenContainer`.  When the recovery fails, the user will be signed out - further recovery attempts aren't necessary and probably aren't desirable.

Just asked Cursor to provide a simpler alternative that will not attempt to run the recovery more than once, and it gave me this - which I think might be good.  Let's sync over Zoom though.

```swift
public actor DeadTokenRecoverer {
    private var recoveryTask: Task<Void, Error>?
    
    public func attemptRecoveryFromPastPurchase(
        purchasePlatform: SubscriptionEnvironment.PurchasePlatform,
        restoreFlow: any AppStoreRestoreFlowV2
    ) async throws {
        // If recovery is in progress or completed, wait for it
        if let existingTask = recoveryTask {
            try await existingTask.value
            return
        }
        
        // Start new recovery
        recoveryTask = Task {
            try await performRecovery(purchasePlatform: purchasePlatform, restoreFlow: restoreFlow)
        }
        
        defer { recoveryTask = nil }
        try await recoveryTask!.value
    }
    
    private func performRecovery(
        purchasePlatform: SubscriptionEnvironment.PurchasePlatform,
        restoreFlow: any AppStoreRestoreFlowV2
    ) async throws {
        switch purchasePlatform {
        case .appStore:
            try await restoreFlow.restoreSubscriptionAfterExpiredRefreshToken()
        case .stripe:
            throw SubscriptionManagerError.noTokenAvailable
        }
    }
}
```

federicocappelli: Much cleaner yes, I agree, let me try to use it in the tests.

federicocappelli: Applied, and tests updated

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -21,46 +21,36 @@ import Networking
...
-        recoveryAttemptCount = 0
-        throw SubscriptionManagerError.errorRetrievingTokenContainer(error: error)
+    @available(macOS 12.0, *)
+    private func performRecovery(purchasePlatform: SubscriptionEnvironment.PurchasePlatform,
+                                 restoreFlow: any AppStoreRestoreFlowV2) async throws {
+        switch purchasePlatform {
+        case .appStore:
+            try await restoreFlow.restoreSubscriptionAfterExpiredRefreshToken()
+        case .stripe:
+            throw SubscriptionManagerError.noTokenAvailable
```
diegoreymendez: It looks to me like we shouldn't even instantiate `DeadTokenRecoverer` if we're on stripe, or at the very least consider bailing out much earlier.

That is unless we're planning to add support for this case.

diegoreymendez (approved):
Approving since this worked well for me.

Two notes:
- I added a comment, can be ignored though.
- Please make sure you target this fix to the correct target branch.  It may be worth pushing to internal this week, IMO - since otherwise we'll get this live in the last week of the rollout.

SharedPackages/BrowserServicesKit/Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: Fair, I simplified it

PR #1323, opened 2025-07-10
Simplifies subscription feature-availability/enabled checks

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210756866798692?focus=true

### Description

Simplifies subscription feature-availability/enabled checks

The goal here is:
- Simplify the logic so it's clearer and easier to maintain.
- Standardize v1 and v2 a bit more in terms of a feature being available or enabled.
- Surface some issues where the subscription is returning no features available.  This will be tackled in [a follow-up task](https://app.asana.com/1/137249556945/task/1210799126744217).

### Test Greyed our features

Remove a user entitlement manually and make sure it's greyed out correctly in the UI.

### Test missing features

Remove a feature manually and make sure the UI doesn't show the feature.

### Impact and Risks

Medium: risk should be medium here, so if you spot anything that looks dangerous let me know and we'll revise.

#### What could go wrong?

This could affect PPro feature availability / enabled state, possibly preventing users from accessing their subscriptions.

### Quality Considerations

NA

### Notes to Reviewer

NA

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -45,7 +45,7 @@ public protocol SubscriptionManager: SubscriptionTokenProvider, SubscriptionAuth
     /// Purchase page URL when launched as a result of intercepted `/pro` navigation.
     /// It is created based on current `SubscriptionURL.purchase` and inherits designated URL components from the source page that triggered redirect.
     func urlForPurchaseFromRedirect(redirectURLComponents: URLComponents, tld: TLD) -> URL
-    func currentSubscriptionFeatures() async -> [Entitlement.ProductName]
+    func currentSubscriptionFeatures() async throws -> [Entitlement.ProductName]
```
diegoreymendez, line 48: Transient errors need to be handled by UI - we can't assume features are either present or absent if the API calls can fail.

If the UI needs to definitely show something then we need to re-evaluate how to remember the last values without an expiration date.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -145,7 +145,7 @@ public protocol SubscriptionManagerV2: SubscriptionTokenProvider, SubscriptionAu
     /// Get the current subscription features
     /// A feature is based on an entitlement and can be enabled or disabled
     /// A user cant have an entitlement without the feature, if a user is missing an entitlement the feature is disabled
-    func currentSubscriptionFeatures(forceRefresh: Bool) async throws -> [SubscriptionFeatureV2]
+    func currentSubscriptionFeatures(forceRefresh: Bool) async throws -> [SubscriptionEntitlement]
```
diegoreymendez, line 148: Separating feature-availability from user-entitlements as in V1.  It doesn't make sense to try and group them both here for two basic reasons:
1. This is not how V1 worked and in a migration we should try to respect the previous API contract as much as possible.
2. We either want to check if a feature is part of the subscription, or if the user can use a feature.  These are separate things that we'll currently check separately.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -564,43 +564,24 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
...
     }
 
     public func isSubscriptionFeatureEnabled(_ entitlement: SubscriptionEntitlement) async throws -> Bool {
-        guard isUserAuthenticated else { return false }
-
-        let currentFeatures = try await currentSubscriptionFeatures(forceRefresh: false)
-        return currentFeatures.contains { feature in
-            feature.entitlement == entitlement && feature.isAvailableForUser
-        }
+        try await currentSubscriptionFeatures(forceRefresh: false).contains(entitlement)
```
diegoreymendez: This method only needs to check against `currentSubscriptionFeatures`.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -25,14 +25,21 @@ import os.log
...
-    func isFeatureEnabledForUser(feature: Entitlement.ProductName) async throws -> Bool
-
-    func currentSubscriptionFeatures() async -> [Entitlement.ProductName]
+    /// Whether a feature is included in the Subscription.
+    ///
+    /// This is mostly useful pre-purchase so that the user will know which features the Subscription includes for them.
+    /// This should really not be used for much else, as the client UI should check ``isFeatureAvailable``
+    /// to know if a feature can be used by the user.
+    ///
+    func isFeatureIncludedInSubscription(_ feature: Entitlement.ProductName, cachePolicy: APICachePolicy) async throws -> Bool
```
diegoreymendez: I think this method should possibly not throw and return an expired cache if necessary.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -25,14 +25,21 @@ import os.log
...
+
+    /// Whether the feature is enabled for use.
+    ///
+    /// This is mostly useful post-purchases.
+    ///
+    func isFeatureEnabled(_ feature: Entitlement.ProductName) async throws -> Bool
```
diegoreymendez, line 38: Same as above, I think this should return an expired cache if necessary, instead of throwing.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -54,15 +61,19 @@ public protocol SubscriptionAuthV1toV2Bridge: SubscriptionTokenProvider, Subscri
 }
 
 extension SubscriptionAuthV1toV2Bridge {
-
-    public func isEnabled(feature: Entitlement.ProductName) async throws -> Bool {
-        try await isFeatureAvailableAndEnabled(feature: feature, cachePolicy: .returnCacheDataElseLoad)
```
diegoreymendez, line 59: A third way to check shouldn't be necessary necessary.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -54,15 +61,19 @@ public protocol SubscriptionAuthV1toV2Bridge: SubscriptionTokenProvider, Subscri
...
 
 extension DefaultSubscriptionManager: SubscriptionAuthV1toV2Bridge {
 
-    public func isFeatureEnabledForUser(feature: Entitlement.ProductName) async throws -> Bool {
+    public func isFeatureEnabled(_ feature: Entitlement.ProductName) async throws -> Bool {
+
+        // Note by Diego: we should probably check if the feature is included in the subscription
+        //   here, but I'm trying to minimize the amount of changes, and up until today I see the
+        //   code is only checking entitlements.  I'll keep that logic for now, but we should
+        //   try to understand if this is important to change.
```
diegoreymendez: As discussed with @federicocappelli , I'm not checking if the feature is included in the subscription because that could introduce troublesome regressions.

SharedPackages/BrowserServicesKit/Sources/Subscription/SubscriptionFeatureV2.swift
```diff
@@ -1,31 +0,0 @@
...
-/// A `SubscriptionFeature` is **available** if the specific feature is `on` for the specific subscription. Feature availability if decided based on the country and the local and remote feature flags.
-/// A `SubscriptionFeature` is **availableForUser** if the logged in user has the required entitlements.
-public struct SubscriptionFeatureV2: Equatable, CustomDebugStringConvertible {
-    public var entitlement: SubscriptionEntitlement
-    public var isAvailableForUser: Bool
-
-    public var debugDescription: String {
-        "\(entitlement.rawValue) is \(isAvailableForUser ? "available" : "unavailable")"
-    }
-}
```
diegoreymendez, line 31: This is part of the simplification - it's not clear we need subscription features and entitlements mixed in a new structure at this time.  We currently don't have a need to check both lists of entitlements together.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1048,14 +1048,22 @@ extension SettingsViewModel {
...
-            self.state.subscription.subscriptionFeatures = await subscriptionAuthV1toV2Bridge.currentSubscriptionFeatures()
+
+            // Note by Diego: using try? and then defaulting to [] below is not good!!!
+            //   I'm doing this because the original code did the same - it just did it inside
+            //   the call to currentSubscriptionFeatures so it wasn't as noticeable.  I'd like
+            //   to change this to have proper error handling but I see no quick solution
+            //   right now with the current code implementation.
+            //   In short: this is not new, I just surfaced it here instead of having it hidden
+            //   inside the call.
+            self.state.subscription.subscriptionFeatures = (try? await subscriptionAuthV1toV2Bridge.currentSubscriptionFeatures()) ?? []
```
diegoreymendez, line 1061: Should we consider allowing the cache to return the last known value even if expired - so we can avoid error handling in the UI level, which is tricky?  That'd default to showing the last known good state.

federicocappelli: One less bad option is defaulting to all features. This can be incorrect but less bad better than removing them.

federicocappelli (commented):
1. The modified methods use a mix of
- V1: Entitlement.ProductName
- V2: SubscriptionEntitlement
as params, should we move everything to the V2 model and use the extensions in EntitlementsBridging for bridging v1 and v2?

2. I would remove the `Note by Diego:` comments, I don't think that's the best way to explain code choices and planning future improvements. The codebase being open source I would avoid leaking internal discussions, maybe you can add a link to a to an Asana task/comment instead?

iOS/DuckDuckGo/Subscription/Feedback/UnifiedFeedbackFormViewModel.swift
```diff
@@ -164,7 +164,14 @@ final class UnifiedFeedbackFormViewModel: ObservableObject {
...
 
         Task {
-            let features = await subscriptionManager.currentSubscriptionFeatures()
+            // Note by Diego: using try? and then defaulting to [] below is not good!!!
+            //   I'm doing this because the original code did the same - it just did it inside
+            //   the call to currentSubscriptionFeatures so it wasn't as noticeable.  I'd like
+            //   to change this to have proper error handling but I see no quick solution
+            //   right now with the current code implementation.
+            //   In short: this is not new, I just surfaced it here instead of having it hidden
+            //   inside the call.
```
diegoreymendez: 🔴🔴🔴🔴🔴🔴 - Resolve before merging, or replace this comment with a follow-up task.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1048,14 +1048,22 @@ extension SettingsViewModel {
...
             self.state.subscription.entitlements = currentEntitlements
-            self.state.subscription.subscriptionFeatures = await subscriptionAuthV1toV2Bridge.currentSubscriptionFeatures()
+
+            // Note by Diego: using try? and then defaulting to [] below is not good!!!
+            //   I'm doing this because the original code did the same - it just did it inside
+            //   the call to currentSubscriptionFeatures so it wasn't as noticeable.  I'd like
+            //   to change this to have proper error handling but I see no quick solution
+            //   right now with the current code implementation.
+            //   In short: this is not new, I just surfaced it here instead of having it hidden
+            //   inside the call.
```
diegoreymendez: 🔴🔴🔴🔴🔴🔴 - Resolve before merging, or replace this comment with a follow-up task.

iOS/DuckDuckGo/SettingsViewModel.swift
diegoreymendez: I think it was @miasma13  that mentioned we might want to allow expired cache to be returned as the least bad of all.  The only important thing in that case would be properly clearing caches on sign out.  IMHO that'd be ideal.

Defaulting to `true` can have negative effects if this code ever runs for non-subscribers.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
federicocappelli: @diegoreymendez Should I implement the cache while you are looking at the rest??

federicocappelli (commented):
The approach seems good, let me know if you want me to help in any part (currentSubscriptionEntitlements cache?)

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
diegoreymendez: @federicocappelli - Not sure tbh... I think we need to first decide what we want from our cache.  Let's chat.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -579,7 +579,7 @@ class MainViewController: UIViewController {
     func presentNetworkProtectionStatusSettingsModal() {
         Task {
             let subscriptionManager = AppDependencyProvider.shared.subscriptionAuthV1toV2Bridge
-            if let hasEntitlement = try? await subscriptionManager.isEnabled(feature: .networkProtection), hasEntitlement {
+            if let hasEntitlement = try? await subscriptionManager.isFeatureEnabled(.networkProtection), hasEntitlement {
```
diegoreymendez: Let's roll this back.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1048,14 +1048,22 @@ extension SettingsViewModel {
             let entitlementsToCheck: [Entitlement.ProductName] = [.networkProtection, .dataBrokerProtection, .identityTheftRestoration, .identityTheftRestorationGlobal, .paidAIChat]
 
             for entitlement in entitlementsToCheck {
-                if let hasEntitlement = try? await subscriptionAuthV1toV2Bridge.isEnabled(feature: entitlement),
+                if let hasEntitlement = try? await subscriptionAuthV1toV2Bridge.isFeatureEnabled(entitlement),
```
diegoreymendez, line 1051: Let's roll this back.

iOS/DuckDuckGo/Subscription/Feedback/VPNMetadataCollector.swift
```diff
@@ -242,7 +242,7 @@ final class DefaultVPNMetadataCollector: VPNMetadataCollector {
     }
 
     func collectPrivacyProInfo() async -> VPNMetadata.PrivacyProInfo {
-        let hasVPNEntitlement = (try? await subscriptionManager.isEnabled(feature: .networkProtection)) ?? false
+        let hasVPNEntitlement = (try? await subscriptionManager.isFeatureEnabled(.networkProtection)) ?? false
```
diegoreymendez: We should collect both here, subscription entitlements and user entitlements.

iOS/DuckDuckGo/UICoordination/MainCoordinator.swift
```diff
@@ -170,7 +170,7 @@ final class MainCoordinator {
 
     func presentNetworkProtectionStatusSettingsModal() {
         Task {
-            if let hasEntitlement = try? await subscriptionManager.isEnabled(feature: .networkProtection),
+            if let hasEntitlement = try? await subscriptionManager.isFeatureEnabled(.networkProtection),
```
diegoreymendez: Let's roll this back.

macOS/DuckDuckGo/PrivacyDashboard/View/PrivacyDashboardViewController.swift
```diff
@@ -324,7 +324,7 @@ extension PrivacyDashboardViewController {
     }
 
     private func isPirEnabledAndUserHasProfile() async -> Bool {
-        let isPIRFeatureEnabled = try? await Application.appDelegate.subscriptionAuthV1toV2Bridge.isEnabled(feature: .dataBrokerProtection)
+        let isPIRFeatureEnabled = try? await Application.appDelegate.subscriptionAuthV1toV2Bridge.isFeatureEnabled(.dataBrokerProtection)
```
diegoreymendez: Let's roll this back.

macOS/DuckDuckGo/UnifiedFeedbackForm/MetadataCollectors/VPNMetadataCollector.swift
```diff
@@ -317,7 +317,7 @@ final class DefaultVPNMetadataCollector: VPNMetadataCollector {
     }
 
     func collectPrivacyProInfo() async -> VPNMetadata.PrivacyProInfo {
-        let hasVPNEntitlement = (try? await subscriptionManager.isEnabled(feature: .networkProtection)) ?? false
+        let hasVPNEntitlement = (try? await subscriptionManager.isFeatureEnabled(.networkProtection)) ?? false
```
diegoreymendez: Again we should collect both subscription entitlements and user entitlements.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/V1toV2Bridging/SubscriptionAuthV1toV2Bridge.swift
```diff
@@ -54,15 +61,19 @@ public protocol SubscriptionAuthV1toV2Bridge: SubscriptionTokenProvider, Subscri
 }
 
 extension SubscriptionAuthV1toV2Bridge {
-
-    public func isEnabled(feature: Entitlement.ProductName) async throws -> Bool {
-        try await isFeatureAvailableAndEnabled(feature: feature, cachePolicy: .returnCacheDataElseLoad)
+    public func isFeatureIncludedInSubscription(_ feature: Entitlement.ProductName) async throws -> Bool {
+        try await currentSubscriptionFeatures().contains(feature)
```
diegoreymendez, line 64: I think this should be the correct way to to check a feature is available within the subscription.  Can you verify it checks out?

iOS/DuckDuckGo/SettingsViewModel.swift
diegoreymendez: I've updated the comments to reference a follow-up work task.

iOS/DuckDuckGo/Subscription/Feedback/UnifiedFeedbackFormViewModel.swift
diegoreymendez: I replaced this with a comment referencing follow-up work.

iOS/DuckDuckGo/SettingsViewModel.swift
diegoreymendez: I replaced this with a comment referencing follow-up work.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionNavBarButtonModel.swift
```diff
@@ -133,16 +133,22 @@ final class NetworkProtectionNavBarButtonModel: NSObject, ObservableObject {
...
+                pinNetworkProtectionToNavBarIfNeverPinnedBefore()
+            } else {
+                pinningManager.unpin(.networkProtection)
+                showVPNButton = false
+                return
+            }
 
-        showVPNButton = false
+            showVPNButton = isPinned || popoverManager.isShown || isHavingConnectivityIssues
+        }
```
diegoreymendez, line 151: This logic was just off.  When user entitlements are disabled the shortcut should be removed (unpinned).  When the feature is available for the first time, the shortcut should be pinned automatically.

This updates the logic to behave properly.

iOS/DuckDuckGo/UICoordination/MainCoordinator.swift
```diff
@@ -170,7 +170,7 @@ final class MainCoordinator {
 
     func presentNetworkProtectionStatusSettingsModal() {
         Task {
-            if let hasEntitlement = try? await subscriptionManager.isEnabled(feature: .networkProtection),
+            if let hasEntitlement = try? await subscriptionManager.isFeatureIncludedInSubscription(.networkProtection),
```
federicocappelli: I think here we have a change in behaviour.
Previously, `isEnabled` was calling `isFeatureAvailableAndEnabled(feature: feature, cachePolicy: .returnCacheDataElseLoad)` that was a merge of Subscription and user entitlements. Now `isFeatureIncludedInSubscription` is only checking the Subscription entitlements.

This is valid for all places where we used `isEnabled(feature`

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionNavBarButtonModel.swift
```diff
@@ -133,16 +133,22 @@ final class NetworkProtectionNavBarButtonModel: NSObject, ObservableObject {
 
     @MainActor
     func updateVisibility() {
-        guard !isPinned,
-              !popoverManager.isShown,
-              !isHavingConnectivityIssues else {
+        Task { @MainActor in
+            guard let canStartVPN = try? await vpnGatekeeper.canStartVPN() else {
+                // If there's an error, don't make any changes
```
diegoreymendez, line 138: Just remain in the previous state in this scenario.  Shortcut pins are persisted so no need to change anything here.

iOS/DuckDuckGo/UICoordination/MainCoordinator.swift
diegoreymendez: For clarity: this is intentional.  I think the previous behavior was actually broken.

A greyed out feature requires `featureInSubscription==true && featureEnabled==false` which was impossible with the previous logic.

So far in testing I'm finding feature availability is looking good.

macOS/DuckDuckGo/Preferences/Model/PreferencesSidebarModel.swift
```diff
@@ -309,26 +309,19 @@ final class PreferencesSidebarModel: ObservableObject {
...
+            for entitlement in entitlements {
+                if let hasEntitlement = try? await subscriptionManager.isFeatureEnabled(entitlement.product), hasEntitlement == true {
                     currentUserEntitlements.append(entitlement)
-
-                }
-            } else {
-                for entitlement in entitlements {
-                    if let hasEntitlement = try? await subscriptionManager.isEnabled(feature: entitlement.product), hasEntitlement == true {
-                        currentUserEntitlements.append(entitlement)
-                    }
```
diegoreymendez, line 331: Diff looks awful here, I'm just standardizing the way to check if the feature should be greyed out between v1 and v2 because on v2 this was NOT working fine.

See screenshot:
Screenshot 2025-07-15 at 14 37 06

iOS/DuckDuckGo/UICoordination/MainCoordinator.swift
miasma13: I agree with @federicocappelli here. The new condition actually does not make sense here as the logic is if user has access to VPN (meaning is subscribed and has present entitlement to VPN) then we should segue to VPN settings, otherwise we should segue to Privacy Pro purchase.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1048,14 +1048,17 @@ extension SettingsViewModel {
             let entitlementsToCheck: [Entitlement.ProductName] = [.networkProtection, .dataBrokerProtection, .identityTheftRestoration, .identityTheftRestorationGlobal, .paidAIChat]
 
             for entitlement in entitlementsToCheck {
-                if let hasEntitlement = try? await subscriptionAuthV1toV2Bridge.isEnabled(feature: entitlement),
+                if let hasEntitlement = try? await subscriptionAuthV1toV2Bridge.isFeatureIncludedInSubscription(entitlement),
```
miasma13: This is not equivalent to the previous check as `hasEntitlement` should be calculated using new function `isFeatureEnabled`

iOS/DuckDuckGo/AppServices/VPNService.swift
```diff
@@ -114,8 +114,7 @@ final class VPNService: NSObject {
     @MainActor
     private func refreshVPNShortcuts() async {
         guard await vpnFeatureVisibility.shouldShowVPNShortcut(),
-              let hasEntitlement = try? await subscriptionManager.isFeatureAvailableAndEnabled(feature: .networkProtection,
-                                                                            cachePolicy: .returnCacheDataElseLoad),
+              let hasEntitlement = try? await subscriptionManager.isFeatureIncludedInSubscription(.networkProtection),
```
miasma13: I suggest changing the variable name from `hasEntitlement` to `isFeatureIncludedInSubscription` (or similar) to better reflect fixed logic.

SharedPackages/BrowserServicesKit/Sources/SubscriptionTestingUtilities/Managers/SubscriptionManagerMock.swift
```diff
@@ -121,14 +121,14 @@ public final class SubscriptionManagerMock: SubscriptionManager {
...
-        guard let hasEntitlement = try? await isFeatureAvailableAndEnabled(feature: feature, cachePolicy: .returnCacheDataElseLoad) else {
+    public func isFeatureEnabled(_ feature: Entitlement.ProductName) async -> Bool {
+        guard let hasEntitlement = try? await isFeatureIncludedInSubscription(feature: feature, cachePolicy: .returnCacheDataElseLoad) else {
             return false
         }
         return hasEntitlement
     }
 
-    public func isFeatureAvailableAndEnabled(feature: Entitlement.ProductName, cachePolicy: APICachePolicy) async throws -> Bool {
+    public func isFeatureIncludedInSubscription(feature: Entitlement.ProductName, cachePolicy: APICachePolicy) async throws -> Bool {
```
miasma13: I know it is mock only but the logic in these 2 functions seems to be inverted:
- `func isFeatureEnabled` should call `let result = await accountManager.hasEntitlement`
- `func isFeatureIncludedInSubscription` should call `try? await isFeatureIncludedInSubscription`

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -579,7 +579,7 @@ class MainViewController: UIViewController {
     func presentNetworkProtectionStatusSettingsModal() {
         Task {
             let subscriptionManager = AppDependencyProvider.shared.subscriptionAuthV1toV2Bridge
-            if let hasEntitlement = try? await subscriptionManager.isEnabled(feature: .networkProtection), hasEntitlement {
+            if let hasEntitlement = try? await subscriptionManager.isFeatureIncludedInSubscription(.networkProtection), hasEntitlement {
```
miasma13: I suggest changing the variable name from `hasEntitlement` to `isFeatureIncludedInSubscription` (or similar) to better reflect fixed logic.

iOS/DuckDuckGo/SettingsViewModel.swift
miasma13: Reviewing this comment after our zoom call, I still find my previous comment was pointing out the right issue and we should be using `isFeatureEnabled` instead.

miasma13 (changes requested):

iOS/DuckDuckGo/AppServices/VPNService.swift
diegoreymendez: Changed.

iOS/DuckDuckGo/MainViewController.swift
diegoreymendez: Changed.

iOS/DuckDuckGo/SettingsViewModel.swift
diegoreymendez: You might be right because of the assignment I see below.  I'll test it to see if there's any issue.

diegoreymendez: You were right, I changed this.

SharedPackages/BrowserServicesKit/Sources/SubscriptionTestingUtilities/Managers/SubscriptionManagerMock.swift
diegoreymendez: Good catch, reverted and fixed tests.

federicocappelli (approved):
LGTM

miasma13 (approved):
Checked after updates and all looks good 👍

PR #1326, opened 2025-07-10
Separate crash reports between targets

bwaresiak:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1210727563819390?focus=true
Tech Design URL:
CC: @diegoreymendez @samsymons @THISISDINOSAUR @quanganhdo @amddg44 

### Description

Update the crash pixel to differentiate between DDG apps.

### Testing Steps
Test on device, just break in the code that has been modified.

iOS
 - Open app without debugger attached.
 - Crash the app from debug menu.
 - Restart the app with debugger attached - regular crash pixel should be sent.

macOS - MetricKit path - App Store target
 - Open app without debugger attached.
 - Crash the app from debug menu.
 - Restart the app with debugger attached - regular crash pixel should be sent.

macOS - Crash log path - Regular target
 - Remove `debug` checks from `CrashReporter.swift`
 - Open app without debugger attached.
 - Crash the app from debug menu.
 - Restart the app with debugger attached - regular crash should be found.

VPN:
We don't have a way to trigger crash, so breaking the extension is needed in order to test this properly. 

### Notes to Reviewer

---
###### Internal references:


macOS/DuckDuckGo/CrashReports/Model/CrashPixelAppIdentifier.swift
```diff
@@ -0,0 +1,40 @@
...
+    init?(_ bundleID: String?) {
+        guard let bundleID, let mainBundleID = Bundle.main.bundleIdentifier, bundleID != mainBundleID else {
+            return nil
+        }
+        if bundleID.hasSuffix("vpn") {
+            self = .vpnAgent
+        } else if bundleID.hasSuffix("vpn.network-extension") {
+            self = .vpnExtension
+        } else if bundleID.hasSuffix("DBP.backgroundAgent") {
+            self = .dbp
```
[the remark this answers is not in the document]
diegoreymendez: Copilot has a point here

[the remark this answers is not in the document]
ayoy: Yes, I’m fixing it.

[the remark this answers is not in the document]
ayoy: Ok, I’ll at least add tests :)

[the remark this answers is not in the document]
ayoy: @diegoreymendez this should now be correct. Please give it a try. Thanks!

macOS/DuckDuckGo/CrashReports/Model/CrashPixelAppIdentifier.swift
```diff
@@ -0,0 +1,66 @@
...
+        switch self {
+        case .dbp:
+            return []
+        case .vpnAgent:
+            return ["com.duckduckgo.macos.vpn"]
+        case .vpnExtension:
+            return ["com.duckduckgo.macos.vpn.network-extension"]
+        }
+    }
+}
```
diegoreymendez: This didn't work very well for me in testing.  I think this logic might work well on release but likely not in debug and other build configurations unless we remove the build configuration identifiers.

I would suggest making sure the logic works the same in all build configurations, even if it's not meant to run in `DEBUG`.

A simple alternative that worked well for me is this:

```swift
enum CrashPixelAppIdentifier: String, CaseIterable {
    case dbp
    case vpnAgent = "vpnagent"
    case vpnExtension = "vpnextension"

    /// Initializes the identifier by checking for specific substrings in the bundle ID, excluding the main bundle ID.
    init?(_ bundleID: String?, mainBundleID: String? = Bundle.main.bundleIdentifier) {
        guard let bundleID, let mainBundleID, bundleID != mainBundleID else { return nil }

        if Self.isVPNAgentBundleID(bundleID) {
            self = .vpnAgent
        } else if Self.isVPNExtensionBundleID(bundleID) {
            self = .vpnExtension
        } else if Self.isDBPBundleID(bundleID) {
            self = .dbp
        } else {
            return nil
        }
    }

    private static func isVPNAgentBundleID(_ bundleID: String) -> Bool {
        bundleID.contains("vpn") && !isVPNExtensionBundleID(bundleID)
    }

    private static func isVPNExtensionBundleID(_ bundleID: String) -> Bool {
        bundleID.contains("vpn") && (bundleID.contains("network-extension") || bundleID.contains("network-protection-extension") || bundleID.contains("proxy"))
    }

    private static func isDBPBundleID(_ bundleID: String) -> Bool {
        bundleID.contains("DBP.backgroundAgent")
    }
}
```

macOS/DuckDuckGo/CrashReports/Model/CrashPixelAppIdentifier.swift
```diff
@@ -0,0 +1,66 @@
...
+    case dbp, vpnAgent = "vpnagent", vpnExtension = "vpnextension"
+
+    init?(_ bundleID: String?, mainBundleID: String? = Bundle.main.bundleIdentifier) {
+        guard let bundleID, let mainBundleID, bundleID != mainBundleID else {
+            return nil
+        }
+
+        if let matchingBundleID = Self.allCases.first(where: { $0.bundleIDs.contains(bundleID) }) {
+            self = matchingBundleID
+        } else if let matchingSuffix = Self.allCases.first(where: { $0.bundleSuffixes.contains(where: { bundleID.hasSuffix($0) }) }) {
```
diegoreymendez, line 34: `com.duckduckgo.macos.vpn.network-extension.debug` is not matching here.  There's an alternative approach suggested below but if you want to fix with minimal changes I'd recommend replacing `hasSuffix` with `contains` or `hasPrefix`.

macOS/DuckDuckGo/CrashReports/Model/CrashPixelAppIdentifier.swift
diegoreymendez: I'm re-reviewing this because I don't understand why it's not matching.

macOS/DuckDuckGo/CrashReports/Model/CrashPixelAppIdentifier.swift
```diff
@@ -0,0 +1,66 @@
...
+/// For crashes happening in main bundle it should remain `nil`, otherwise it can take one of the predefined values for known bundles.
+enum CrashPixelAppIdentifier: String, CaseIterable {
+    case dbp, vpnAgent = "vpnagent", vpnExtension = "vpnextension"
+
+    init?(_ bundleID: String?, mainBundleID: String? = Bundle.main.bundleIdentifier) {
+        guard let bundleID, let mainBundleID, bundleID != mainBundleID else {
+            return nil
+        }
+
+        if let matchingBundleID = Self.allCases.first(where: { $0.bundleIDs.contains(bundleID) }) {
```
diegoreymendez, line 32: Ok this is the one that's not matching properly for me.  In debug, this is checking that one of the bundle IDs below contains `com.duckduckgo.macos.vpn.network-extension.debug` which will never happen.

We should revert the check and see if the provided bundleID contains any of the bundle IDs below.

macOS/DuckDuckGo/CrashReports/Model/CrashPixelAppIdentifier.swift
diegoreymendez: Ok now yes, replaced with the suggestion above.

PR #1341, opened 2025-07-12
PIR iOS debug menu improvements

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210338467109187
Tech Design URL:
CC:

### Description

This PR adds new options for debugging PIR operations on iOS. This includes:
* Environment toggles
* A dedicated Debug Mode screen which lets you scan a profile on whatever brokers you choose and view the results
* The ability to run all queued scan/optouts/both, letting you view the webview that's running in the background if you choose

### Testing Steps

1. Open the PIR debug menu
2. Tap "Run PIR debug mode"
3. Enter some information and pick a broker
4. Run the operation, and wait for results to appear (or not, depending on the info you entered)
5. If some results appear, trigger an opt out and check that console logs appear throughout the process
6. Now go back to the base debug mode screen and save a profile if you haven't already, which will cause a bunch of scan operations to get queued up
7. Tap "Run Pending Scans" and watch as the scan count goes down, and the opt out count goes up (assuming you entered a profile that will get some results)
8. Wait for the scans to complete and then trigger the opt outs
9. Check that the "Show WebView" button tends to work - note that it won't always work, since there is a brief period between jobs where the old webview is torn down and the new one is created, but it should work overall

### Notes to Reviewer

---
###### Internal references:


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/CCF/WebViewHandler.swift
```diff
@@ -88,10 +88,12 @@ final class DataBrokerProtectionWebViewHandler: NSObject, WebViewHandler {
...
             let viewController = UIViewController.init()
             viewController.view = webView
-            window?.rootViewController = viewController
+            let navigationController = UINavigationController(rootViewController: viewController)
+            navigationController.title = "PIR Debug Mode"
 
+            window?.rootViewController = navigationController
             window?.windowLevel = UIWindow.Level.alert
-            window?.makeKeyAndVisible()
+            window?.tag = 42
```
samsymons: Need to replace this with a more rigorous approach

iOS/DuckDuckGo/DataBrokerProtectionDebugViewController.swift
```diff
@@ -1,276 +0,0 @@
-//
-//  DataBrokerProtectionDebugViewController.swift
```
samsymons, line 2: I just moved this file into a `DBP` directory, git didn't handle it that well apparently

iOS/DuckDuckGo/DBP/RunDBPDebugModeViewController.swift
```diff
@@ -0,0 +1,737 @@
...
+    }
+}
+
+// MARK: - WebView Window Manager
+
+class WebViewWindowManager {
+    private weak var webViewHandler: WebViewHandler?
+
+    private var isWebViewVisible: Bool {
+        if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene {
```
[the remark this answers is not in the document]
samsymons: I normally completely agree, but I'm not too worried about this given it's a singleton provided by the system and only used for a debug menu.

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
```diff
@@ -0,0 +1,962 @@
...
+                return "Custom Web URL"
+            }
+        }
+    }
+
+    private var manager: DataBrokerProtectionIOSManager
+    private let settings = DataBrokerProtectionSettings(defaults: .dbp)
+    private let webUISettings = DataBrokerProtectionWebUIURLSettings(.dbp)
+
+    private lazy var brokerUpdater: BrokerJSONServiceProvider? = {
```
THISISDINOSAUR: I know it's just debug stuff, but I'm worried that's most of our PIR surface area in the iOS app at the moment and potentially setting precedents. I think it would be better if this was obtained through the DBPIOSManager so that things in the iOS app don't have to know about/depend on the details of the dependancies.
I know it's just debug stuff, but I'm worried that's most of our PIR surface area in the iOS app at the moment and potentially setting precedents. I think it would be better if this was obtained through the DBPIOSManager so that things in the iOS app don't have to know about/depend on the details of the dependancies.

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
```diff
@@ -0,0 +1,962 @@
...
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        stopJobCountRefreshTimer()
+    }
+
+    private func loadHealthOverview() {
+        Task {
+            if await manager.validateRunPrerequisites() {
+                let allScheduledTasks = await BGTaskScheduler.shared.pendingTaskRequests()
```
THISISDINOSAUR: As above, I know it's debug code, but given how important debug mode is to us I think we should avoid this kind of thing.As above, I know it's debug code, but given how important debug mode is to us I think we should avoid this kind of thing.

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
```diff
@@ -0,0 +1,962 @@
...
+        }
+    }
+    
+    @objc private func showWebViewTapped() {
+        webViewWindowHelper.showWebView(title: "PIR Debug Mode")
+    }
+    
+    private func calculatePendingJobCounts() async -> (pendingScans: Int, pendingOptOuts: Int) {
+        guard let allData = try? manager.database.fetchAllBrokerProfileQueryData() else {
+            return (0, 0)
```
THISISDINOSAUR, line 303: Might be worth having an assertion failure here?

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
samsymons: Yeah glad to hear this because I was thinking the same way after I did it, and hoped to come back and correct it before you'd had a look. 😅 Lemme clean this up and have it such that the debug menu mostly just needs to talk to the manager, so that the manager's in control of how these dependencies are created.

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
samsymons: Yep definitely

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
samsymons: The `BrokerJSONServiceProvider` has been moved to the iOS manager as a private property, and the iOS manager has a new `refreshRemoteBrokerJSON` method for use by the debug menu - there's no way for the app to reach into the broker JSON provider any more.

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
samsymons: I've moved the background task logic into the iOS manager, so now the debug menu just asks the manager for `hasScheduledBackgroundJob`. This also let me make the background job identifier private, and further keep that logic encapsulated within the manager.

iOS/DuckDuckGo/DBP/DataBrokerProtectionDebugViewController.swift
```diff
@@ -0,0 +1,897 @@
...
+            showWebViewButton()
+        }
+    }
+    
+    @objc private func showWebViewTapped() {
+        webViewWindowHelper.showWebView(title: "PIR Debug Mode")
+    }
+    
+    private func calculatePendingJobCounts() async -> (pendingScans: Int, pendingOptOuts: Int) {
+        guard let allData = try? manager.database.fetchAllBrokerProfileQueryData() else {
```
THISISDINOSAUR, line 301: One option as opposed to refactoring this all into DBPManager would be just to remove the database call (I'd love to make database private)

THISISDINOSAUR (commented):
Mostly LGTM, a couple of things:
- Would be good to have some feedback when pressing opt out
- I found the UI to be quite unresponsive when running pending scans

THISISDINOSAUR (approved):
LGTM!

PR #1361, opened 2025-07-15
Voice search button visibility

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210777323867681?focus=true

### Description
Change voice-search visibility based on user interaction + text availability 

### Testing Steps
The Voice Input button should be hidden when the user types in the Input Field.  This means:
User first enters Input Screen → Voice Input button is visible (because user has not input text themselves)
User types text in Input Field → Voice Input button is hidden
User clears all text in Input field (with ⓧ button or by hand) → Voice Input button is visible

### Impact and Risks

#### What could go wrong?
Display the voice search button when it shouldn't or vice-versa


iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/NavigationActionBarViewModel.swift
```diff
@@ -25,85 +25,104 @@ import SwiftUI
...
                 self?.isSearchMode = toggleState == .search
             }
             .store(in: &cancellables)
-        
+
         switchBarHandler.currentTextPublisher
             .receive(on: DispatchQueue.main)
             .sink { [weak self] (text: String) in
-                self?.hasText = !text.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).isEmpty
+                let hasText = !text.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).isEmpty
```
dus7, line 73: That's different vs current production state. Currently any text (including whitespace) that's input into the search is causing the mic button to be invisible.

Bunn: Yeah, I believe this is an improvement, the main goal is to avoid text being replaced by the audio search, and empty spaces are not valid prompts or queries so it's fine to replace them with audio. Easily revertible in case stakeholders disagree. Thanks!

PR #1362, opened 2025-07-15
App startup improvements

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210437247680862?focus=true
Tech Design URL: 
https://app.asana.com/1/137249556945/project/481882893211075/task/1210741508940252
https://app.asana.com/1/137249556945/project/481882893211075/task/1210787475496146

### Description

1. Introduce LaunchTaskManager as a wrapper around a non-critical task queue executed after app startup.
2. Migrate interaction state clearing from the main startup flow into this queue.
3. Add default user agent retrieval into this queue.
4. Introduce caching for default user agent.

### Testing Steps

1. Run the tests.
2. Validate if the interaction state clearing is called.
3. Validate if the UA is cached after the first startup.
4. Validate if the cached UA is used after on the consecutive launch.
5. Change the logic        
from: guard osVersionProvider.osVersion != cached.osVersion else { return }
to: guard osVersionProvider.osVersion == cached.osVersion else { return }
and see if the LaunchTask to update user agent is spawned and the user agent is refetched.

### Impact and Risks


iOS/DuckDuckGo/AppLifecycle/AppStates/Launching.swift
```diff
@@ -159,12 +158,17 @@ struct Launching: LaunchingHandling {
                                defaultBrowserPromptService: defaultBrowserPromptService
         )
 
+        // MARK: - Register background tasks that run after app is ready
+        // Use this to avoid delaying app startup
+        launchTaskManager.register(task: ClearInteractionStateTask(autoClearService: autoClearService,
+                                                                   mainViewController: mainCoordinator.controller))
```
bwaresiak: Can we directly pass `TabInteractionStateSource` here instead of MainVC?

bwaresiak: I see, there is a call to `excluding: model.tabs` so apparently not.

PR #1369, opened 2025-07-15
Update PIR Status Indicator on Settings Screen to consider stored profiles

edulpn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1210733520275433?focus=true
Tech Design URL: 
CC:

### Description

This PR extracts the `DataBrokerProtectionIOSManager` dependency from `SettingsSubscriptionView` to `SettingsViewModel`. Then `SettingsSubscriptionView` compounds the status indicator logic for the PIR settings item to take the existence of a valid stored profile into consideration.

### Testing Steps

1. You’ll need to be authenticated with your Privacy Pro account. Launch the debug menu > `Subscription` > `I Have a Subscription` then follow the flow to authenticate.
2. You should see Privacy Pro settings items, but your PIR item should be `off` (you’ll have the entitlement, but no stored profile).
3. Tap the PIR item and follow the flow until you see your scanning results.
4. Dismiss the screen, now your PIR setting item status should be `on` (now you have both entitlement and a stored profile).

### Impact and Risks
Low

#### What could go wrong?

### Quality Considerations
<!— 
- What edge cases exist?
—>

### Notes to Reviewer

—
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/SettingsSubscriptionView.swift
+++ b/iOS/DuckDuckGo/SettingsSubscriptionView.swift
@@ -266,13 +266,14 @@ struct SettingsSubscriptionView: View {
             let hasDBPEntitlement = userEntitlements.contains(.dataBrokerProtection)
             let hasValidStoredProfile = settingsViewModel.dataBrokerProtectionIOSManager
                 .flatMap { try? $0.meetsProfileRunPrequisite } ?? false
-            let statusIndicator: StatusIndicator = hasDBPEntitlement && hasValidStoredProfile ? .on : .off
+            var statusIndicator: StatusIndicator = hasDBPEntitlement && hasValidStoredProfile ? .on : .off
 
             let destination: LazyView<AnyView> = {
                 if let dbpManager = settingsViewModel.dataBrokerProtectionIOSManager,
                    DataBrokerProtectionIOSManager.isDBPStaticallyEnabled {
                     return LazyView(AnyView(DataBrokerProtectionViewControllerRepresentation(dbpViewControllerProvider: dbpManager)))
                 } else {
+                    statusIndicator = .on
                     return LazyView(AnyView(SubscriptionPIRMoveToDesktopView()))
                 }
             }()
--- a/iOS/DuckDuckGo/SettingsSubscriptionView.swift
+++ b/iOS/DuckDuckGo/SettingsSubscriptionView.swift
@@ -261,13 +261,13 @@ struct SettingsSubscriptionView: View {
             }
             .disabled(!hasVPNEntitlement)
         }
-        
+
         if subscriptionFeatures.contains(.dataBrokerProtection) {
             let hasDBPEntitlement = userEntitlements.contains(.dataBrokerProtection)
             let hasValidStoredProfile = settingsViewModel.dataBrokerProtectionIOSManager
                 .flatMap { try? $0.meetsProfileRunPrequisite } ?? false
             let statusIndicator: StatusIndicator = hasDBPEntitlement && hasValidStoredProfile ? .on : .off
-            
+
             let destination: LazyView<AnyView> = {
                 if let dbpManager = settingsViewModel.dataBrokerProtectionIOSManager,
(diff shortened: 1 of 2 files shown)


iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -515,7 +517,8 @@ final class SettingsViewModel: ObservableObject {
          featureDiscovery: FeatureDiscovery = DefaultFeatureDiscovery(),
          subscriptionFreeTrialsHelper: SubscriptionFreeTrialsHelping = SubscriptionFreeTrialsHelper(),
          urlOpener: URLOpener = UIApplication.shared,
-         keyValueStore: ThrowingKeyValueStoring
+         keyValueStore: ThrowingKeyValueStoring,
+         dataBrokerProtectionIOSManager: DataBrokerProtectionIOSManager = .shared!
```
samsymons: I forgot to mention one important detail when we chatted earlier: this is disabled in production, so the force unwrap here will cause a crash in that case. Check out `isDBPStaticallyEnabled` if you want to see where this is defined.

For now we should make this property optional instead, and unwrap it when attempting to push the PIR flow.

iOS/DuckDuckGo/SettingsSubscriptionView.swift
```diff
@@ -264,28 +264,28 @@ struct SettingsSubscriptionView: View {
...
-                        statusIndicator: StatusIndicatorView(status: hasDBPEntitlement ? .on : .off),
-                        isGreyedOut: !hasDBPEntitlement
-                    )
+            let hasValidStoredProfile = settingsViewModel.dataBrokerProtectionIOSManager
+                .flatMap { try? $0.meetsProfileRunPrequisite } ?? false
+            let statusIndicator: StatusIndicator = hasDBPEntitlement && hasValidStoredProfile ? .on : .off
+
+            let destination: LazyView<AnyView> = {
+                if let dbpManager = settingsViewModel.dataBrokerProtectionIOSManager,
+                   DataBrokerProtectionIOSManager.isDBPStaticallyEnabled {
```
edulpn, line 273: Theoretically, successfully unwrapping `settingsViewModel.dataBrokerProtectionIOSManager ` should be enough, no need to check if it’s statically enabled. But I thought it would be nice to keep that check here.

THISISDINOSAUR: Cool, I'd definitely prefer we have the check here as well since I'd rather we didn't hardcode any assumptions about if dbpIOSManager exists or not. And the settings page shouldn't change if isDBPStaticallyEnabled is false

iOS/DuckDuckGo/SettingsSubscriptionView.swift
```diff
@@ -264,28 +264,28 @@ struct SettingsSubscriptionView: View {
...
-                NavigationLink(destination: LazyView(SubscriptionPIRMoveToDesktopView()), isActive: $isShowingDBP) {
-                    SettingsCellView(
-                        label: UserText.settingsPProDBPTitle,
-                        image: Image(uiImage: DesignSystemImages.Color.Size24.identity),
-                        statusIndicator: StatusIndicatorView(status: hasDBPEntitlement ? .on : .off),
-                        isGreyedOut: !hasDBPEntitlement
-                    )
+            let hasValidStoredProfile = settingsViewModel.dataBrokerProtectionIOSManager
+                .flatMap { try? $0.meetsProfileRunPrequisite } ?? false
+            let statusIndicator: StatusIndicator = hasDBPEntitlement && hasValidStoredProfile ? .on : .off
```
samsymons: One last tweak to request then I think we're good to go here: today in production, the PIR badge is hardcoded to `.on`, but after this change it will appear as `.off`. Would we be able to fall back to `.on` as a default in the case where we don't pass the `isDBPStaticallyEnabled` check? Thanks!

samsymons (changes requested):

samsymons (approved):
LGTM - thanks! I've tested both production and debug cases, with and without PIR data, the badge is correct in all situations.

PR #1400, opened 2025-07-17
Add navigation pixel

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201621708115095/task/1210812990564441?focus=true

### Description
This change adds navigation pixel that is fired on every navigation, regular and same-page, excluding replaceState().

### Testing Steps
1. Launch the app and filter logs by `m_mac_navigation [`
2. Verify that `m_mac_navigation` pixel is fired on every navigation and URL change triggered by the user, and that it’s not fired twice or multiple times for some navigations.
3. Check the linked Asana task for the full Specification and try to break it.

### Impact and Risks

—
###### Internal references:


macOS/DuckDuckGo/Tab/Model/Tab.swift
```diff
@@ -1326,16 +1326,30 @@ extension Tab/*: NavigationResponder*/ { // to be moved to Tab+Navigation.swift
         permissions.tabDidStartNavigation()
         userInteractionDialog = nil
 
+        let shouldFireNavigationPixel: Bool = {
```
mallexxx: since the code grows larger maybe it worths bringing it out to a small NavigationResponder struct and add it to Tab+Navigation (no protocols and dedicated vars, but just nav responder)?

ayoy: Sure, let me do it 👍

mallexxx (approved):
LGTM 👍
As expected, reloading a page from the error page fires no pixel, but it's probably fine as technically it‘s the same navigation that was causing the error page to appear

macOS/DuckDuckGo/Tab/Model/Tab.swift
ayoy: @mallexxx could you have just a quick look at the tab extension code, please?

PR #1407, opened 2025-07-17
Update alpha build version text

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1210825527588180?focus=true
Tech Design URL:
CC:

### Description

This PR makes two changes:
1. The settings screen's version string has been updated to show the git revision after the version number
2. The about screen has been updated to clarify that this build is an alpha, and also show the revision

This change lets employees know for sure that they're using an alpha, and they can tell exactly which revision it was built from. In debug, nothing changes since the SHA isn't set outside of GitHub Actions.

### Testing Steps

1. Grab the ad-hoc alpha build from [this workflow](https://github.com/duckduckgo/apple-browsers/actions/runs/16356419098) and check that it shows the git SHA and "Alpha" text in the about screen

### Impact and Risks

None, this is an internal build change

### Notes to Reviewer

---
###### Internal references:


samsymons:
One catch with this change is that anyone running the alpha in debug mode will have their Info.plist modified, which runs the risk of checking it in. Before merging I will likely remove the "alpha debug" support so that the revision is only written on actual alpha builds, avoiding this problem.

iOS/DuckDuckGo/AboutView.swift
```diff
@@ -84,8 +84,19 @@ struct AboutViewVersion: View {
...
+            let version = "\(viewModel.state.version) (Alpha)"
+#else
+            let version = viewModel.state.version
+#endif
+
+            SettingsCellView(label: UserText.settingsVersion, accessory: .rightDetail(version))
+
+#if ALPHA
+            if let buildVersion = Bundle.main.ShortGitRevision {
+                SettingsCellView(label: "Build", accessory: .rightDetail(buildVersion))
```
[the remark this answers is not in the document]
samsymons: This is a non-localized string, so it being hardcoded is deliberate - I don't want it translated.

samsymons:
There's a helper coming later which can be reused in this PR, I'll mark this Do Not Merge and update this PR once that's available: https://github.com/duckduckgo/apple-browsers/compare/dominik/macos-alpha-version?expand=1

iOS/DuckDuckGo/AboutView.swift
```diff
@@ -84,8 +85,15 @@ struct AboutViewVersion: View {
 
     var body: some View {
         Section(header: Text("DuckDuckGo for iOS"), footer: Text(UserText.settingsSendCrashReportsDescription)) {
-            SettingsCellView(label: UserText.settingsVersion,
-                             accessory: .rightDetail(viewModel.state.version))
+#if (ALPHA && !DEBUG)
```
samsymons, line 88: The SHA isn't set when running the alpha in debug mode, so I'm suppressing it entirely to avoid others on the dev team getting alarmed by the blank "Build" row

iOS/DuckDuckGo/SettingsOthersView.swift
```diff
@@ -31,9 +32,16 @@ struct SettingsOthersView: View {
         Section {
             // About
             NavigationLink(destination: AboutView().environmentObject(viewModel)) {
+#if (ALPHA && !DEBUG)
+                // The commit SHA is only set for release alpha builds, so debug alpha builds won't show it
+                let version = "v\(viewModel.state.version) (\(AppVersion.shared.commitSHAShort))"
```
samsymons, line 37: We could probably put "alpha" in here too, it's pretty cramped on small screens but should  be fine

samsymons:
The AI comments are fine, this logic only ever applies to alpha builds and we set the SHA in GitHub Actions. It would be blank if someone were to make an alpha build for distribution locally, but IMO including the SHA there is less interesting since it could be made with unpushed or even uncommitted changes anyway. Personally I think only setting it from GitHub Actions makes sense since we know the commit is available to everyone.

ayoy (approved):
Looks great, thanks @samsymons!

PR #1433, opened 2025-07-21
Fix flaky address bar test

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205237866452338/task/1210837193154122?focus=true
Tech Design URL: 
CC:

### Description

This PR fixes a flaky address bar test. The previous implementation was sleeping for a short time before asserting its conditions, but I was able to reproduce this failing consistently using the Run Repeatedly feature.

Increasing the sleep duration helped, suggesting that this was simply not waiting long enough, so the implementation has been replaced with XCTWaiter with a generous timeout instead.

After this change, I wasn't able to reproduce the failure even after running the test for 1000 iterations - previously I would always hit a failure within the first 100.

### Testing Steps

1. Run this test repeatedly locally and verify that it doesn't fail

### Notes to Reviewer

---
###### Internal references:


ayoy (approved):
Looks great, thanks a lot @samsymons!

PR #1434, opened 2025-07-21
Fix assertion failure during autoconsent tests.

sammacbeth:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210836027639389?focus=true
Tech Design URL:
CC:

### Description
Prevent decode error due to a file:// URL being used by the test runner.

### Testing Steps
1. Run AutoconsentBackgroundTests
2. Should no longer trigger an assertionFailure.


samsymons (approved):
Looks great - thanks for the quick fix!

PR #1437, opened 2025-07-21
Fix loc_export.sh scripts to properly isolate platform exports

afterxleep:
Task/Issue URL: N/A
Tech Design URL: N/A
CC: N/A

### Description
Fixed both iOS and macOS `loc_export.sh` scripts to ensure they only export strings from their respective platforms when run from the repository root directory.

The issue was that running these scripts from the root directory (e.g., `macOS/scripts/loc_export.sh`) could result in xcodebuild detecting and using the workspace file, which would export strings from both iOS and macOS platforms.

### Testing Steps
1. From the repository root, run `macOS/scripts/loc_export.sh`
2. Verify that only macOS strings are exported (no iOS phantom strings)
3. From the repository root, run `iOS/scripts/loc_export.sh`
4. Verify that only iOS strings are exported
5. Run the scripts from within their respective directories and verify they still work correctly

### Impact and Risks
**Low**: This is a build script improvement that ensures proper isolation of platform-specific localization exports.

#### What could go wrong?
- The scripts could fail if the directory structure changes, but the changes include checks to ensure we don't unnecessarily change directories if already in the correct location

### Quality Considerations
- Added directory change logic that checks current location before changing
- Both scripts now explicitly specify the project file to prevent workspace detection
- Scripts return to the original directory after completion to maintain expected behavior

### Notes to Reviewer
The key changes are:
1. Both scripts now change to their respective project directories before running xcodebuild
2. The macOS script now explicitly specifies `-project "DuckDuckGo-macOS.xcodeproj"`
3. Directory changes are conditional to avoid unnecessary operations

### DIFF
--- a/iOS/scripts/loc_export.sh
+++ b/iOS/scripts/loc_export.sh
@@ -2,7 +2,7 @@
 
 # Get the directory where the script is stored
 script_dir=$(dirname "$(readlink -f "$0")")
-base_dir="${script_dir}/.."
+ios_dir="${script_dir}/.."
 
 echo "Updating..."
 "${script_dir}/loc_update.sh"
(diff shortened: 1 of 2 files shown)


rachelmcr (commented):
I ran both the iOS and macOS export scripts from the root directory and their project directory, and confirmed I got the expected exported strings in the xliff files they generated.

I see that `shellcheck` is failing for iOS on this PR. We could use the `-x` flag to allow external sources (e.g. `-x -P iOS/scripts`), which should silence this error. Here’s how it’s used in the macOS check:

https://github.com/duckduckgo/apple-browsers/blob/db373a3d1d153c98c658c0fe313ca1f7bd4bf862/.github/workflows/macos_pr_checks.yml#L73

iOS/scripts/loc_export.sh
```diff
@@ -2,13 +2,19 @@
 
 # Get the directory where the script is stored
 script_dir=$(dirname "$(readlink -f "$0")")
-base_dir="${script_dir}/.."
+ios_dir="${script_dir}/.."
```
rachelmcr, line 5: As cursor suggested, this will fix the equality check for the current directory in `run_in_directory()`:

```suggestion
ios_dir=$(dirname "$script_dir")
```

rachelmcr (approved):
Works great for me! 👍

PR #1467, opened 2025-07-23
wkwebview abstractions to improve testing

brindy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205237866452338/task/1210740006914450?focus=true
Tech Design URL:
CC:

### Description
Adds WKAbstractions package and updates iOS to use it.

Removes unused subscriptions cookie code.

Updates web cache manager and fire button reference tests to use abstractions. 

### Testing Steps
1. Smoke test subscriptions on iOS and macOS
2. Smoke test data clearing on iOS

### Impact and Risks
Medium 

#### What could go wrong?
Could break subscriptions and data clearing

### Notes to Reviewer

---
###### Internal references:


iOS/Core/Fireproofing.swift
```diff
@@ -57,7 +57,6 @@ public class UserDefaultsFireproofing: Fireproofing {
     private var allowedDomainsIncludingDuckDuckGo: [String] {
         allowedDomains + [
             URL.ddg.host ?? "",
-            SubscriptionCookieManager.cookieDomain
```
bwaresiak, line 60: @miasma13 I was under impression that we still set something for subscription, but my knowledge is out of date - so just pointing this place to you to validate. :)

SharedPackages/BrowserServicesKit/Sources/WKAbstractions/WKAbstractions.swift
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public func dataRecords(ofTypes types: Set<String>) async -> [any DDGWebsiteDataRecord] {
+        await wrapped.dataRecords(ofTypes: types).map { record in
+            WebsiteDataRecordWrapper(wrapped: record)
+        }
+    }
+
+    public func removeData(ofTypes types: Set<String>, for records: [any DDGWebsiteDataRecord]) async {
+        let unwrappedRecords = records.compactMap { ($0 as? WebsiteDataRecordWrapper)?.wrapped }
```
bwaresiak: Since we are casting here anyway, maybe we could get rid of `WebsiteDataRecordWrapper` and just cast to `WKWebsiteDataRecord` ? I think this is possible as `WebsiteDataStoreWrapper` is not used in tests (or tested), what do you think?

iOS/Core/Fireproofing.swift
brindy: yeah cookies are no longer used for subscriptions - it's now an explicit communication between the app and FE

SharedPackages/BrowserServicesKit/Sources/WKAbstractions/WKAbstractions.swift
bwaresiak: Also on top of this: I wonder if we could get rid of casting completely, if we were to add `associatedtype Record: DDGWebsiteDataRecord` to `DDGWebsiteDataStore` protocol, and have app use `WKWebsiteDataRecord` types and tests use Mocked ones.

brindy: You mean change the signature so that it takes `WKWebsiteDataRecord` ? the problem with that is there's no way to create and configure a WKWebsiteDataRecord for the purpose of tests, so these have to be abstractions as well. Therefore the only way to pass them around is by wrapping them.  Unless I'm missing something?

brindy: As discussed on MM and with the diff you supplied, using an associated type in the data store and then adding an extension of WKWebsiteDataStore comply with the type, works!  Thanks.

iOS/Core/Fireproofing.swift
miasma13: @bwaresiak I was in touch with @brindy when he was planing out this work and I reconfirm that cookie route for sharing subscription route is not used (was briefly introduced but due to issues was rolled back and never introduced). The initial cookie based approach was replaced with more robust one using JS messaging based one.

bwaresiak: That's good, what I had more or less in mind, was to make sure serp (or rather subscriptions.duckduckgo.com) is not setting up a cookie that is supposed to be fireproofed. If that's not a case, then we're good :)

miasma13: I don't think it is a case, the whole `SubscriptionCookieManager` logic was dormant as the privacy config flags after the rollback were disabled.

bwaresiak (approved):
Great job, much needed changes. The if for DEBUG is not nice, but from what I understand you want to move this, right? 

I've tested this, fireproofing etc and saw no regressions.

miasma13 (approved):
I have reviewed the removal of legacy and obsolete subscription cookie code, all looks good to me 👍

PR #1522, opened 2025-07-29
Attempt fixes for occassional blank webView issue

graeme:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1210155968610460?focus=true
Tech Design URL:
CC:

### Description
Fixed an issue where users would see blank tabs after switching between tabs, requiring them to switch away and back to restore visibility. Investigation revealed the webView was getting stuck in an inconsistent state where it was hidden but no error message was being displayed. Added two defensive fixes: prevent hiding the webView when there's no meaningful error message, and automatically restore webView visibility when the inconsistent state is detected.

### Testing Steps
- Open multiple tabs with different websites loaded
- Switch between tabs rapidly, including tabs that may have loading errors or interrupted navigation
- Verify that tabs never appear completely blank (should show either web content or a proper error message)
- Test error scenarios by navigating to invalid URLs to ensure error pages still display correctly (https://privacy-test-pages.site/network-error/)

### Impact and Risks
Low - This is a defensive bug fix that should improve the reliability of tab switching and error handling.

### What could go wrong?

The guard clause in showError could potentially prevent legitimate error messages from displaying if they're unexpectedly empty, but this would preserve existing web content visibility rather than showing a blank screen. The automatic state correction in fireWebViewDebugPixels calls the existing hideErrorMessage() method, so it follows the same code path as normal error recovery.

### Quality Considerations

The existing debug pixel (`debugWebViewInVisibleTabHidden`) will continue to fire when the inconsistent state is detected, providing monitoring of how often this issue occurs and whether the fix is effective. Edge case consideration: error messages that are whitespace-only are now treated the same as empty messages. Performance impact is minimal as these are defensive checks in existing code paths.

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/TabViewController.swift
+++ b/iOS/DuckDuckGo/TabViewController.swift
@@ -643,17 +643,12 @@ class TabViewController: UIViewController {
     }
 
     private func fireWebViewDebugPixels() {
-        if !webView.isDescendant(of: view) {
-            DailyPixel.fireDailyAndCount(pixel: .debugWebViewNotInVisibleTabHierarchy)
-        }
-        if webView.window == nil {
-            DailyPixel.fireDailyAndCount(pixel: .debugWebViewNotAttachedToWindow)
-        }
         if webView.isHidden && (errorMessage.text.isNilOrEmpty || error.isHidden) {
             DailyPixel.fireDailyAndCount(pixel: .debugWebViewInVisibleTabHidden)
-        }
-        if webView.frame == .zero {
-            DailyPixel.fireDailyAndCount(pixel: .debugWebViewHasZeroFrameSize)
+            
+            // Fix inconsistent state - if webView is hidden but no error shown, show webView
+            // https://app.asana.com/1/137249556945/project/414709148257752/task/1210155968610460?focus=true
+            hideErrorMessage()
         }
     }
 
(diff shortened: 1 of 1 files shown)


iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -642,18 +642,13 @@ class TabViewController: UIViewController {
...
 
-    private func fireWebViewDebugPixels() {
-        if !webView.isDescendant(of: view) {
-            DailyPixel.fireDailyAndCount(pixel: .debugWebViewNotInVisibleTabHierarchy)
-        }
-        if webView.window == nil {
-            DailyPixel.fireDailyAndCount(pixel: .debugWebViewNotAttachedToWindow)
-        }
+    private func checkWebViewVisibilityConsistency() {
         if webView.isHidden && (errorMessage.text.isNilOrEmpty || error.isHidden) {
```
alessandroboron: @graeme Perhaps we should only check for `webView.isHidden && error.isHidden` to determine an inconsistent state. If the error view is visible but the message is nil or empty, it should still be valuable to display it to users, as the error image and header indicate that there was an issue loading the page (See picture attached). 

Screenshot 2025-07-30 at 2 28 21 pm

iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -1051,6 +1046,11 @@ class TabViewController: UIViewController {
     }
     
     private func showError(message: String) {
+        guard !message.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
```
alessandroboron: See comment https://github.com/duckduckgo/apple-browsers/pull/1522/files#r2241511329.

alessandroboron (commented):
Thanks for the thourough PR description @graeme. 

I tested and could not reproduce the issue. I left a comment about not showing the error view  if the error message is empty or nil.

iOS/DuckDuckGo/TabViewController.swift
graeme: Hmm yeah, good point. I think I should just remove this part then.

PR #1525, opened 2025-07-29
VPN Upsell: Allow manual triggering for internal users

jozsef-vesza:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210869582295671?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1210649122070619
CC:

### Description

This PR has 2 main updates:
1. Adds debug menu options for VPN upsell, located under Debug > VPN > Upsell
2. Updates the VPN icon to support showing a blue dot (required for upsell)

### Testing Steps
Make sure you’re not a subscriber. 

To test showing/hiding the upsell button:
1. Show the button via `Debug > VPN > Upsell > Show Upsell Button`-> button should become visible (but non functional yet)
2. This will directly set the state on `VPNUpsellVisibilityManager`, bypassing prerequisite checks
3. You can then hide it via `Debug > VPN > Upsell > Hide Upsell Button`

To test the flow of showing the VPN button automatically (without bypassing prerequisites):
1. Enable the `vpnToolbarUpsell` flag via `Debug > Feature Flag Overrides > VPN`
4. Reset upsell state via `Debug > VPN > Reset Upsell State`
5. Relaunch the app
6. Skip onboarding via `Debug > Skip Onboarding`
7. Set as default browser -> VPN button should appear after 10 seconds

### Impact and Risks
Low: I’ve updated the VPN button to support notification dots, and that affects the regular flow too.

#### What could go wrong?
We might show an unwanted notification dot on the VPN button.

### Quality Considerations
~I’ve added tests to make sure that we only set the notification dot when upsell should be visible.~
👆this will come in the form of a UI test in a later PR

### Notes to Reviewer
N/A

—
###### Internal references:


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -206,15 +206,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     lazy var vpnUpsellVisibilityManager: VPNUpsellVisibilityManager = {
         return VPNUpsellVisibilityManager(
-            isFirstLaunch: AppDelegate.isFirstLaunch,
+            isFirstLaunch: LocalStatisticsStore().atb == nil,
```
jozsef-vesza: I removed `AppDelegate.isFirstLaunch`, because this lazy property got evaluated earlier than we would set it, which caused it to always be false when `VPNUpsellVisibilityManager` evaluated it.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNUpsellVisibilityManager.swift
```diff
@@ -218,3 +218,16 @@ final class VPNUpsellVisibilityManager: ObservableObject {
         timer?.invalidate()
     }
 }
+
+// MARK: - Debug Menu
+/// These methods are triggered from the VPN Debug Menu (Debug > VPN > Upsell)
+/// They explicitly set the state, bypassing all eligibility checks.
+extension VPNUpsellVisibilityManager {
```
jozsef-vesza, line 297: @aataraxiaa is there a better way to limit access to these methods? They are meant to work only for testing the feature.

macOS/DuckDuckGo/Onboarding/ContextualOnboarding/ContextualDialogsManager.swift
```diff
@@ -101,7 +101,7 @@ public class ContextualDialogsManager: ObservableObject, ContextualOnboardingDia
     private weak var lastTab: Tab?
 
     // Publisher for contextual onboarding completion
-    @Published private(set) var isContextualOnboardingCompleted: Bool = true
+    @Published private(set) var isContextualOnboardingCompleted: Bool
```
jozsef-vesza, line 104: Changed to set this value up in the init below, otherwise we might kick off the timer early by publishing a `true` value.

macOS/DuckDuckGo/Application/AppDelegate.swift
jozsef-vesza: @aataraxiaa just realized that this can cause a crash in tests, probably due to a circular dependency between `AppDelegate` and `LocalStatisticsStore`. Looking into it

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -231,19 +237,19 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     var dockCustomization: DockCustomization?
 #endif
 
-    @UserDefaultsWrapper(key: .firstLaunchDate, defaultValue: Date.monthAgo)
-    static var firstLaunchDate: Date
+    @UserDefaultsWrapper(key: .firstLaunchDate)
+    static var firstLaunchDate: Date?
```
jozsef-vesza: @aataraxiaa I removed the default value here, and made the date optional. I think it makes sense (new users shouldn’t have a first launch date). Cursor Bot thinks otherwise, so let me know what you think, I might be missing something.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/VPNUpsellVisibilityManager.swift
aataraxiaa: I don’t think so. I thought about `#if DEBUG` compilation directives, but we want these to be available for any internal user, even on a release build. So, I think we can leave it as it is. Clear documentation is enough.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -206,15 +206,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
             contextualOnboardingPublisher: onboardingContextualDialogsManager.isContextualOnboardingCompletedPublisher.eraseToAnyPublisher(),
-            featureFlagger: featureFlagger
+            featureFlagger: featureFlagger,
+            persistor: vpnUpsellUserDefaultsPersistor,
+            timerDuration: vpnUpsellUserDefaultsPersistor.expectedUpsellTimeInterval
         )
     }()
 
+    private lazy var vpnUpsellUserDefaultsPersistor: VPNUpsellUserDefaultsPersistor = {
+        return VPNUpsellUserDefaultsPersistor(keyValueStore: UserDefaults.standard)
```
aataraxiaa: We could use [this](https://app.asana.com/1/137249556945/project/1200194497630846/task/1210288620537332?focus=true) here instead of `UserDefaults`. Based on the documentation it’s the recommended best practice going forward.

macOS/DuckDuckGo/Application/AppDelegate.swift
aataraxiaa: I found a related problem and change here - https://github.com/duckduckgo/apple-browsers/pull/144#issue-2895233738. However, I don’t think we can apply the same approach as our type is automatically created early via storyboard init etc.

aataraxiaa: Without knowing why it was at one point set to `.monthAgo` I think there is risk in changing it. So, as discussed on MM it’s worth reaching out to the wider team to see if anyone has context. Also, could we try some alternate approaches like:

  Initialize with default state, then update:

```
  lazy var vpnUpsellVisibilityManager: VPNUpsellVisibilityManager = {
      return VPNUpsellVisibilityManager(
          isFirstLaunch: false, // Default to false initially
          // ... other params
      )
  }()

  // Later in app lifecycle, after ATB is set:
  func updateFirstLaunchState() {
      let isFirstLaunch = LocalStatisticsStore().atb == nil
      vpnUpsellVisibilityManager.updateFirstLaunchState(isFirstLaunch)
  }
```

jozsef-vesza: Good point on the 2-step init, I think that’s the only real alternative here.

aataraxiaa (commented):
Code changes LGTM. The following test still does not consistently pass for me:

* Reset state, relaunch, skip onboarding, set as default, wait 10 seconds

The toolbar icon was added 1/3 times I ran the test. Maybe something related to window focus, as at least on one of the failing attempts the window was unfocused.

aataraxiaa (approved):
Tested the latest changes and LGTM @jozsef-vesza ! Nice job solving the issue! 🚀

PR #1536, opened 2025-07-30
Fix rules compilation crash by recreating tmp directory right after it is being removed

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210925187026095?focus=true

### Description
See details in the task.

### Testing Steps

1. Run the app for the first time and notice the rules are compiled just fine.
2. Remove the app and add a following piece of code to Launching.init() -> _ = WKWebView()
3. Run the app and notice the rules are compiled just fine.

### Impact and Risks

#### What could go wrong?
If, for any reason, the folder recreation were to fail (which is highly unlikely imho), the app would crash when attempting to compile the rules.


brindy (approved):
This seems to work.

Please change the base to the current release branch before merging. Then it'll end up on main for free and in the alpha quickly. We def want to run this in the alpha for a few days before releasing. 

Good job!

PR #1561, opened 2025-08-01
Show correct URL on SERP after opening URL in external app

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1209365034718375?focus=true
CC:

### Description

This change addresses two distinct scenarios where users interact with advertisements within search engine results pages (SERPs), which currently lack optimal handling:

- **Product Advertisement Click:** Upon clicking on a product advertisement, the Bing.com URL is opened in a new, empty tab.
- **SERP Advertisement Click:** In the case of a search result for Amazon, clicking on the first advertisement for amazon.co.uk opens the Bing.com URL in the same tab while simultaneously removing the content of the SERP.

In both scenarios, the intention is to avoid opening a new tab or displaying an incorrect URL. Instead, the desired outcome is to maintain the current state of the browser before clicking on the advertisement.

#### Preparation

1. Install the Amazon iOS app

#### Test 1

1. Search "Bluetooth speaker"
Screenshot 2025-08-01 at 17 10 02

2. Click on an Amazon ad
3. The native Amazon app should be opened
4. Go back to DDG
5. The only tab open should be SERP and the address bar should show the search terms or the DDG URL

#### Test 2

1. Search "Amazon"
Screenshot 2025-08-01 at 17 10 24

2. Click on the first, sponsored result for Amazon 
3. The native Amazon app should be opened
4. Go back to DDG
5. The only tab open should be SERP, and the address bar should show the search terms or the DDG URL

#### Try to break the mitigation

- If you test on a device while using a keyboard, try to run one of the test scenarios, and as soon as the DDG app goes in the background CMD+Tab back to it.
- Try to open links in tabs manually and immediately send the app to the background. Also, without a connection to simulate a load error.

### Impact and Risks

If the mitigation is executed in the wrong case, we could close a tab that shouldn't be closed.

#### What could go wrong?

The tab is closed in the wrong circumstances because the user's action manages to fool the state machine.

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/AdAttribution/AdClickExternalOpenDetector.swift
+++ b/iOS/DuckDuckGo/AdAttribution/AdClickExternalOpenDetector.swift
@@ -27,15 +27,14 @@ extension Logger {
 }
 
 /**
- * This class detects and mitigates a specific DDG SERP issue with ad clicks that attempt to redirect users to external applications.
- * It implements a state machine that monitors navigation events and app background transitions to identify when an ad click sequence indicates malicious behaviour.
+ * This class detects and mitigates a specific navigation issue with redirects to external applications.
+ * It implements a state machine that monitors navigation events and app background transitions to identify when we leave the app due to navigation and need to emulate Safari's behaviour: tab cleanup or URL updates.
+ * This is to ensure correct URL is being presented to the user, and that no "empty new tabs" are being left open in the app post-navigation.
  *
- * We handle 2 different scenarios:
- * 1. Click on a product AD (Opens the bing.com url in a new tab)
- * 2. Click on a SERP result AD, e.g. searching amazon and clicking on the first AD result for amazon.co.uk (Opens the bing.com url in the same tab)
+ * We handle two scenarios: for same-tab navigations and for navigation that are being done in a new tab.
  *
  * The detector follows a specific event sequence to trigger mitigation:
- * 1. Navigation starts (user clicks an ad)
+ * 1. Navigation to external domain starts.
  * 2. Navigation fails with a specific error (redirect doesn't complete normally)
  * 3. App enters background (indicating external app launch attempt)
  *


iOS/DuckDuckGo/AdAttribution/AdClickExternalOpenDetector.swift
```diff
@@ -0,0 +1,196 @@
...
+    private func isTimeoutExpired() -> Bool {
+        guard let navigationFailedDate = navigationFailedDate else { return false }
+        return Date().timeIntervalSince(navigationFailedDate) > operationTimeout
+    }
+
+    // MARK: - Private Helpers
+    
+    private func isRelevantWebKitError(_ error: Error) -> Bool {
+        let nsError = error as NSError
+        return nsError.domain == "WebKitErrorDomain" && nsError.code == 102
```
afterxleep: Extract this error code to constant?

iOS/DuckDuckGoTests/AdAttribution/AdClickExternalOpenDetectorTests.swift
```diff
@@ -0,0 +1,302 @@
...
+        detector.mitigationHandler = {
+            XCTFail("Mitigation should not be triggered")
+        }
+
+        detector.startNavigation()
+        detector.finishNavigation()
+        detector.appDidEnterBackground()
+
+        // Small delay to allow any async events to fire
+        RunLoop.current.run(until: Date().addingTimeInterval(0.5))
```
afterxleep: Any particular reason not to use XCTest expectation here?

iOS/DuckDuckGo/AdAttribution/AdClickExternalOpenDetector.swift
federicocappelli: Modified

iOS/DuckDuckGoTests/AdAttribution/AdClickExternalOpenDetectorTests.swift
federicocappelli: Good point, modified

PR #1581, opened 2025-08-05
Subscription keychain error 25291 mitigation

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1210830784500963?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/task/1210830784500963/comment/1210963320974998?focus=true
CC: @diegoreymendez @miasma13 

### Description

This PR implements a mitigation for keychain error 25291 (-25291, which is errSecNotAvailable) by introducing a new KeychainManager that provides automatic retry capabilities for subscription keychain operations. When the keychain is unavailable, operations are queued in a writing backlog and automatically retried when the keychain becomes available again.

Introduction of a new KeychainManager class with thread-safe keychain operations and automatic retry mechanism
Refactoring of SubscriptionTokenKeychainStorageV2 to use the new KeychainManager instead of direct keychain operations
Addition of new pixel events with improved error tracking using PixelKit's error field instead of string descriptions

KeychainManager is fully tested

### Testing Steps

- Subscription should work as usual
- Probably the best way of testing this is to hardcode an `errSecNotAvailable` error in the KeychainManager `store(data...)`
- in console filter for "KeychainManager" categoty for seeing the keychain use across main app and system extension

#### What could go wrong?

- The recovery could not work

---
###### Internal references:


macOS/PixelDefinitions/pixels/subscription.json5
```diff
@@ -0,0 +1,123 @@
+// Privacy Pro subscription-related pixel definitions for macOS
+{
+    'm_mac_direct_privacy-pro_keychainmanager_data_added_to_backlog': {
+        description: 'Fired when Privacy Pro keychain manager adds data to the backlog because keychain is unavailable (direct distribution)',
+        owners: ['federicocappelli'],
+        triggers: ['other'],
+        suffixes: ['first_daily_count'],
+        parameters: [
+            'appVersion',
+            'pixelSource',
```
miasma13, line 10: You don't need to define `pixelSource` every time as it is already present in the shared params definition.
Please use `parameters: ['appVersion', 'pixelSource'],`

federicocappelli: fixed

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
```diff
@@ -0,0 +1,351 @@
...
+import Foundation
+import os.log
+import Common
+import Combine
+
+#if canImport(UIKit)
+import UIKit
+#elseif canImport(AppKit)
+import AppKit
+#endif
```
miasma13, line 28: Probably not needed anymore.

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
```diff
@@ -0,0 +1,351 @@
...
+    // MARK: - Constants
+
+    private enum Constants {
+        static let accessQueueLabel = "keychain.subscription.access"
+        static let keychainAccessibilityLevel = kSecAttrAccessibleAfterFirstUnlock
+    }
+
+    // MARK: - Properties
+
+    private let keychainOperations: KeychainOperationsProtocol
```
miasma13: i suggest renaming `KeychainOperationsProtocol` to something without the `Protocol` suffix (e.g. `KeychainOperationsProviding`, `KeychainAccessProviding`...) In general in most cases we follow the pattern to use bare name for the protocol and `Default` suffix for implementation.

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
```diff
@@ -0,0 +1,351 @@
...
+    public func store(data: Data, forKey key: String) throws {
+        let writingResultStatus = try accessQueue.sync {
+            try internalStore(data: data, forKey: key)
+        }
+
+        if writingResultStatus == errSecSuccess {
+            // Trigger backlog processing for that contexts that can't rely on notifications for detecting keychain availability (SysExt)
+            accessQueue.async {
+                self.processWritingBacklog()
+            }
```
miasma13: Isn't the order of things risky here first we store a value that came as a parameter to this call then we attempt to process and store backlog. What if the same key is already in the backlog and I think it is safe to assume that the backlog is definitely older. Due to the order of operation the backlog value may overwrite one from the input argument.

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
```diff
@@ -0,0 +1,351 @@
...
+
+        if !writingBacklog.isEmpty {
+            self.pixelHandler.handle(pixel: .deallocatedWithBacklog)
+            Logger.keychainManager.warning("Deallocating with \(self.writingBacklog.count) unprocessed backlog items")
+        }
+    }
+
+    // MARK: - Public API
+
+    public func retrieveData(forKey key: String) throws -> Data? {
```
miasma13, line 133: WDYT about triggering backlog processing in the read operation? I guess stores may not be that frequent compared to reads, and if read succeeds it means that the keychain is available. No strong opinions here we can probably follow up when we get initial signal in pixels.

iOS/PixelDefinitions/pixels/subscription.json5
```diff
@@ -0,0 +1,52 @@
+// Privacy Pro subscription-related pixel definitions
+{
+    'm_privacy-pro_keychainmanager_data_added_to_backlog': {
+        description: 'Fired when Privacy Pro keychain manager adds data to the backlog because keychain is unavailable',
+        owners: ['federicocappelli'],
+        triggers: ['other'],
+        suffixes: ['first_daily_count'],
+        parameters: ['appVersion', 'pixelSource'],
```
miasma13: `source` param and its definition is missing for the new pixels

macOS/DuckDuckGo/Statistics/PrivacyProPixel.swift
```diff
@@ -138,6 +143,11 @@ enum PrivacyProPixel: PixelKitEventV2 {
         case .privacyProAuthV2MigrationFailed: return "m_mac_\(appDistribution)_privacy-pro_auth_v2_migration_failure"
         case .privacyProAuthV2MigrationSucceeded: return "m_mac_\(appDistribution)_privacy-pro_auth_v2_migration_success"
         case .privacyProAuthV2GetTokensError: return "m_mac_\(appDistribution)_privacy-pro_auth_v2_get_tokens_error"
+            // KeychainManager
+        case .privacyProKeychainManagerDataAddedToTheBacklog: return "m_mac_privacy-pro_keychainmanager_data_added_to_backlog"
```
miasma13: A nit pick can we update the new pixels value in all relevant places `_keychainmanager...` -> `..._keychain_manager... `

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
federicocappelli: It's still needed by the notification subscriber that use UIApplication and NSApplication depending on the platform.
I could use if os(xxx) but that's more precise.

Edit: with this comment, you made me think that my use of `#if os(xxx)` in the class is dangerous, I reverted to `#if canImport(UIKit)` so we are sure that the use of UIApplication and NSApplicationhappens in a context where the classes are imported correctly

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
federicocappelli: renamed `KeychainOperationsProviding`

iOS/PixelDefinitions/pixels/subscription.json5
federicocappelli: done!

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
federicocappelli: So, that was my first approach, but the problem is that, at `KeychainManager:136`, if data is in the backlog, we never really query the keychain and therefore never reach the switch errSecSuccess :/
I'm still thinking about this and how to remove the backlog check from store...

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
federicocappelli: I changed approach, now the backlog writing happens on read and doesn't rely on a successfull read, but inteas checks the keychain availability before

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/V2/KeychainManager.swift
federicocappelli: I changed approach, now the backlog writing happens on read and doesn't rely on a successfull read, but inteas checks the keychain availability before

macOS/DuckDuckGo/InfoPlist.xcstrings
```diff
@@ -61,18 +61,6 @@
         }
       }
     },
-    "NSAppDataUsageDescription" : {
```
federicocappelli: this damn thing changes every time, I'll revert the change

iOS/PixelDefinitions/pixels/subscription.json5
```diff
@@ -0,0 +1,78 @@
...
+        description: 'Fired when Privacy Pro keychain manager adds data to the backlog because keychain is unavailable',
+        owners: ['federicocappelli'],
+        triggers: ['other'],
+        suffixes: ['first_daily_count'],
+        parameters: [
+            'appVersion',
+            'pixelSource',
+            {
+                key: 'source',
+                description: 'The app component that generated teh pixel, liek MainApp, SysExt, VPNApp or DBP',
```
miasma13: Typos: `teh` and `liek`

miasma13 (approved):

iOS/PixelDefinitions/pixels/subscription.json5
federicocappelli: ffs XD fixed

PR #1586, opened 2025-08-05
Fix the iOS debug menu's Configuration Info ETag viewer

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1210976624719060?focus=true
Tech Design URL:
CC:

### Description

This PR fixes the debug menu in the iOS app to show the stored ETag version correctly. ETag storage was changed to use an app group but the debug menu wasn't updated too, so this PR fixes that.

It also fixes an issue where non-ETag rows were able to show the ETag, due to a cell reuse issue. Now the detail label value is set to nil right after fetching a cell, wiping out any reused state.

### Testing Steps

1. Open the debug menu's Configuration Info screen
2. Check that ETags are visible

### Notes to Reviewer

---
###### Internal references:


jaceklyp (approved):
Nice work, thanks for looking into this :)

PR #1619, opened 2025-08-07
macOS: Fixes for Xcode 26 builds on macOS 26

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1210989790892557?focus=true
Tech Design URL:
CC:

### Description

This fixes breakage with Xcode 26 builds on macOS 26. 

Notes: 
* Some fixes require Xcode 26; we should only merge this when we are ready to transition to building with Xcode 26.
* Currently the macOS TabBarTests fail because of a known issue on macOS 26.

### Testing Steps

1. Set up VM with latest macOS 26 beta.
2. Install Xcode 26 build of the app.
3. Open the app and confirm the tab bar looks good (e.g. no white space/lines at the top, no significant overlap with the navigation/address bar).
4. In the main menu, go to Debug > UI Triggers > Append tabs > 50 tabs. Confirm you can comfortably scroll the tabs in both directions using the left/right arrows next to the tabs.
5. Open the Passwords & Autofill popover and confirm the "Add New" button fills the available width and works as expected.

Also tested: 
* macOS versions 12-15 to confirm no breaking changes.
* smoke tests on macOS 26 to track other known issues not addressed with these changes.

### Impact and Risks
Medium: Fixes here are relatively low risk bug fixes, but missing critical breakage could disrupt user flows on macOS 26.

—
###### Internal references:


rachelmcr:
I took a quick look at the iOS failures, but the issue is with missing destinations although they haven’t been changed in this branch. I believe the change to Xcode 26 is making xcodebuild look for iOS 26 by default, which [should be available](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md), so we may want to double check this behavior closer to release. (Currently Github is still using Xcode beta 1, which could be a factor here.)

.github/workflows/macos_pir_end_to_end_tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       matrix:
         runner: [macos-15-xlarge]
         include:
-          - xcode-version: "16.4"
+          - xcode-version: "26.0"
```
samsymons: I learned the other day that this test suite is disabled at the moment anyway, with `throw XCTSkip("Flaky")` - going to look into how we can restore it on Monday.

macOS/DuckDuckGo/SecureVault/View/PasswordManagementItemList.swift
```diff
@@ -535,3 +536,16 @@ private struct HideMenuIndicatorModifier: ViewModifier {
...
+
+private struct FlexibleButtonSizingModifier: ViewModifier {
+
+    func body(content: Content) -> some View {
+        if #available(macOS 26, *) {
+            content
+                .buttonSizing(.flexible)
+        } else {
+            content
+        }
```
samsymons: Like you mentioned in your PR description, we'll want to find a way to keep these changes but disable them until we're ready to move to Xcode 26 on CI.

We could probably just comment out Xcode 26 blocks with a `TODO` comment before we merge this, and we create a follow-up task to uncomment them as part of the Xcode 26 upgrade later - what do you think?

rachelmcr: > We could probably just comment out Xcode 26 blocks with a TODO comment before we merge this, and we create a follow-up task to uncomment them as part of the Xcode 26 upgrade later - what do you think?

Ah yeah, that seems like a reasonable compromise to allow us to merge the PR now. I’ll remove the Xcode version changes and comment out anything requiring it for now, with the necessary followup task to keep track.

samsymons (approved):
Thanks a lot, the new changes to make this compatible with the current version of Xcode look great! I ran through a bunch of tests on new and current macOS versions, couldn't see any issues besides the remainders we know about, and running this branch on macOS 15 looks good.

PR #1652, opened 2025-08-12
migrate AI Chat settings to shared user defaults - app-configuration

brindy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/392891325557410/task/1210873370382228?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/392891325557410/task/1211000221536588?focus=true
CC: @samsymons @bwaresiak @jaceklyp 

### Description
Migrates AIChatSettings to the shared user defaults `app-configuration` and some minor tidy up.

### Testing Steps
In addition to ensuring the unit test has passed you can validate that settings are migrated and then read/written to the shared user defaults by performing these steps:

1. Use a simulator where Widgets work (for me that's 18.2)
1. Run the app from `main`
2. Add the favorites widget to the home screen
3. Turn off AI Settings explicitly (turn it on first if needed)
4. Check out this branch
5. Apply this [pr.patch](https://github.com/user-attachments/files/21738212/pr.patch) which adds some hacky logic to the widget to hide/show the icon based on the setting
6. Run the app 
7. Migration should run
8. Check the widget has been updated to hide the duck ai button (it will look rubbish)
9. Turn AI back on
10. Restart the app (the widget re-loading isn't part of the change settings call yet)
11. After the app has launched check the widget is now showing the icon

### Impact and Risks

#### What could go wrong?
Anything that users the app-configuration group could get messed up:
* ContentBlocker storage
* Etag storage
* Pixel / Experiment data
* Network protection

### Notes to Reviewer

---
###### Internal references:


iOS/DuckDuckGo/AIChat/AIChatSettingsMigration.swift
```diff
@@ -0,0 +1,58 @@
...
+        guard settings.count > 0 else { return }
+
+        if settings.first(where: { store.object(forKey: $0.key) != nil }) != nil {
+            assertionFailure("Secondary migration of AIChatSettings has ocurred")
+        }
+
+        // Write to the new store
+        settings.forEach {
+            store.set($0.value, forKey: $0.key)
+        }
```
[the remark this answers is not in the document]
brindy: Yes, that shouldn't happen but there's no atomic way to ensure that. Which I thought I had made clear in the code.  The assertion failure will at least let us catch it in development.

iOS/DuckDuckGo/AppConfiguration/AppConfiguration.swift
```diff
@@ -53,10 +53,25 @@ struct AppConfiguration {
...
+    private func migrateAIChatSettings() throws {
+        enum Error: Swift.Error {
+            case unableToCreatedSharedUserDefaults
+        }
+        guard let sharedUserDefaults = UserDefaults(suiteName: Global.appConfigurationGroupName) else {
+            throw Error.unableToCreatedSharedUserDefaults
+        }
+        AIChatSettingsMigration.migrate(from: UserDefaults.standard, to: sharedUserDefaults)
+    }
+
```
brindy: This might be too harsh. @jaceklyp do you have any guidance here?

jaceklyp: hey, question is if we want to crash? If we can't create shared UD I guess it makes sense. If you throw an error on Launching (and this is inside Launching) the app will transition to Terminating state. Please, see the implementation of Terminating.swift. You need to throw TerminationError if you want to properly handle it (send pixel etc.) otherwise it will just execute fatalError()

brindy: Thanks. Yeah, I am just wondering if this important enough to crash the app or if `UserDefaults(suiteName: "")` is flakey enough that it might sometimes just not work.

brindy: I'm gonna change it to the same pattern used elsewhere which is just to default to en empty UserDefaults()

jaceklyp: Sure, I would be tempted to send a pixel if this happens, just so we know for the future :)

iOS/DuckDuckGo/AppConfiguration/AppConfiguration.swift
```diff
@@ -53,10 +53,20 @@
...
         PrivacyFeatures.httpsUpgrade.loadDataAsync()
     }
 
+    /// Perform AI Chat settings migration, and needs to happen before AIChatSettings is created
+    ///  and the widgets needs to be reloaded after.
+    /// Moves settings from `UserDefaults.standard` to the shared container.
+    private func migrateAIChatSettings() throws {        
+        let sharedUserDefaults = UserDefaults(suiteName: Global.appConfigurationGroupName) ?? UserDefaults()
+        AIChatSettingsMigration.migrate(from: UserDefaults.standard, to: sharedUserDefaults)
+    }
```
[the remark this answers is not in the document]
brindy: It's either this or crash the app because we can't create the app-configuration UserDefaults.  This pattern is already used throughout.

iOS/DuckDuckGo/AIChat/AIChatSettings.swift
```diff
@@ -48,17 +49,17 @@ struct AIChatSettings: AIChatSettingsProvider {
...
         privacyConfigurationManager.privacyConfig.settings(for: .aiChat)
     }
-    private let userDefaults: UserDefaults
+    private let keyValueStore: KeyValueStoring
     private let notificationCenter: NotificationCenter
     private let featureFlagger: FeatureFlagger
     init(privacyConfigurationManager: PrivacyConfigurationManaging = ContentBlocking.shared.privacyConfigurationManager,
          debugSettings: AIChatDebugSettingsHandling = AIChatDebugSettings(),
-         userDefaults: UserDefaults = .standard,
+         keyValueStore: KeyValueStoring = UserDefaults(suiteName: Global.appConfigurationGroupName) ?? UserDefaults(),
```
dus7, line 57: Is the fallback happening here something what we are ok with? Upside is that migration should take care of moving this data in case this happens. Alternative could be to create `FailingKeyValueStore` and throw an exception or send a pixel so that we are aware of this happening.

I don't have strong opinion on this, just thinking out loud.

dus7: Now I read [this response](https://github.com/duckduckgo/apple-browsers/pull/1652/files#r2270104594) to one of Cursors comments and I'm leaning towards having `FailingKVS` created with a suiteName parameter to be able to notify ourselves about this, WDYT?

EDIT: Also reading the thread under https://github.com/duckduckgo/apple-browsers/pull/1652#discussion_r2269865523 I think it's a good idea to have something ready to just use whenever we need shared defaults.

brindy: kvfsInitError seems specific to AppKeyValueFileStoreService so I'll just fall back to default but send a pixel when it happens.  I also don't want to attempt to even create the store unless it's needed so I'm going to inject a store factory.

brindy: Ah, I misread and thought you were talking about the migration.  I'd say this is OK since we do it in loads of other places in the app.  

I don't want to make this more complicated because AIChatSettings is instantiated directly in a few places rather than injected so they may end up getting complicated as well:
* DefaultOmniBarViewController
* VoiceSearchViewController
* VoiceSearchFeedbackView_Previews

And fixing all those seems out of the scope of what I was trying to do here.

PR #1666, opened 2025-08-13
[Quick Win Days] iOS: Dax Easter Egg Logos

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1210989706758207?focus=true
CC:

### Description
This Quick Win/Hack Days PR adds Dax Easter Egg logos to the Omnibar. 

Note 1: This feature is behind a feature flag, which is currently local only (i.e failsafe feature flag) and is set to disabled. It can be enabled via the debug FF menu.

Note 2: There are some issues still to be address, which I think can be done after this is merged, as it’s behind a FF. Some issues:
1. Logo positioning looks a bit off compared to the default (I think the default PDF is actually not centered, so maybe design input here is needed)
2. Navigating directly from a search with dynamic icon to a specific URL, and then back, results in dynamic logo not being displayed.

### Testing Prequisites
- Enable feature flag via Debug menu -> Feature Flags -> daxEasterEggLogos -> Force quit app and relaunch

### Testing Steps
1. Search for any of the terms from [here](https://makdulac.com/daxduckdex.html)
2. Observe the dynamic logo displayed in the omnibar in place of the default logo
3. Tap it to enlarge it
4. Tap again to minimize it
5. Navigate back/forth to other searches/pages

### Impact and Risks
Impacts core usage scenario (search and navigation UX)

#### What could go wrong?
* Impacts existing UI

### Quality Considerations
* How does the image look?

—
###### Internal references:


iOS/DuckDuckGo/DaxEasterEggLogos/DaxEasterEggHandler.swift
```diff
@@ -0,0 +1,170 @@
...
+            return
+        }
+        
+        Logger.daxEasterEgg.debug("extractLogosForCurrentPage - executing JavaScript directly on URL: \(webView.url?.absoluteString ?? "no-url")")
+        
+        // Execute logo extraction JavaScript directly without UserScript dependency
+        executeLogoExtraction(webView: webView)
+    }
+    
+    private func executeLogoExtraction(webView: WKWebView) {
```
aataraxiaa: @brindy after our previous chat about the need for a UserScript, and after trying to resolve some timing issues with script injection, I moved to a simple direct execution method with the JS. Advantages:
1. Less code
2. Script is not injected for every page
3. Timing issues resolved.

iOS/DuckDuckGo/DaxEasterEggLogos/DaxEasterEggHandler.swift
```diff
@@ -0,0 +1,170 @@
...
+                
+                return findLogo();
+            } catch (error) {
+                console.error('DaxEasterEgg: Error in logo extraction:', error);
+                return null;
+            }
+        })();
+        """
+        
+        webView.evaluateJavaScript(extractionJS) { [weak self] result, error in
```
brindy: Why this rather than async/await?

brindy (changes requested):
Code looks good but:

1/ Sent you a video on MM - it's a little jarring when it changes from dax to the cosplay as the images don't appear to be the same size.

2/ one question about use of completion handler vs async/await - I don't feel strongly, but have preference for async/await. Not sure your view or if you were aware there was an async/await version. 

3/ Maybe more for the ship review:
a) I think it'd be good to something to the favicon when you click on it, maybe even blur the entire background

b) If we could cross-fade the cosplay icon into the favicon that would be nice

iOS/DuckDuckGo/DaxEasterEggLogos/DaxEasterEggHandler.swift
aataraxiaa: Thanks @brindy. No reason, just habit. Converted to async/await now.

brindy (approved):
Looks much better now the image sizes match.  Thanks for the other changes.

PR #1671, opened 2025-08-13
Allow dynamically looking up the iOS simulator version

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1211034038467055?focus=true
Tech Design URL:
CC:

### Description

This PR adds a new action for checking simulator versions in order of preference, in cases where the GitHub Actions runner is in the process of rolling out.

### Testing Steps

1. Check that CI is green

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/.github/actions/select-ios-simulator/action.yml
+++ b/.github/actions/select-ios-simulator/action.yml
@@ -1,0 +1,61 @@
+name: 'Select iOS Simulator Version'
+description: 'Selects the first available iOS simulator version from a list of preferred versions'
+inputs:
+  preferred-versions:
+    description: 'Comma-separated list of preferred iOS versions (e.g., "18.6,18.2,18.4")'
+    required: true
+  device-name:
+    description: 'iOS simulator device name'
+    required: false
+    default: 'iPhone 16'
+outputs:
+  destination:
+    description: 'Complete xcodebuild destination string'
+    value: ${{ steps.select-version.outputs.destination }}
+  ios-version:
+    description: 'Selected iOS version'
+    value: ${{ steps.select-version.outputs.ios-version }}
+runs:
+  using: 'composite'
+  steps:
+    - name: Select iOS simulator version
+      id: select-version
+      shell: bash
+      run: |
+        # Parse preferred versions from input
+        IFS=',' read -ra PREFERRED_VERSIONS <<< "${{ inputs.preferred-versions }}"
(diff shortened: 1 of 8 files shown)


.github/actions/select-ios-simulator/action.yml
```diff
@@ -0,0 +1,57 @@
...
+        # Get list of available iOS runtimes
+        available_runtimes=$(xcrun simctl list runtimes | grep "iOS")
+        echo "Available iOS runtimes:"
+        echo "$available_runtimes"
+        
+        for version in "${PREFERRED_VERSIONS[@]}"; do
+          version=$(echo "$version" | xargs)
+          echo "Checking for iOS $version..."
+          
+          if echo "$available_runtimes" | grep -q "iOS $version"; then
```
noisysocks: Is it a problem that this matches partial strings? e.g. `iOS 18.1` would match `iOS 18.10`

.github/actions/select-ios-simulator/action.yml
```diff
@@ -0,0 +1,57 @@
...
+            exact_version=$(echo "$available_runtimes" | grep "iOS $version" | head -1 | sed -n 's/.*(\([0-9.]*\) - .*/\1/p')
```
noisysocks: Is it likely that formatting of the output will change, breaking the regex? Could use `xcrun simctl list --json` and `jq` as an alternative.

noisysocks (approved):
Looks good to me. Can see it working well here: https://github.com/duckduckgo/apple-browsers/actions/runs/16953269018/job/48050306608?pr=1671

I left notes for the two questions that I had while reading the code but feel free to ignore them.

.github/actions/select-ios-simulator/action.yml
samsymons: Yeah I took a stab at the JSON approach but decided to just keep it as-is – good idea though, the JSON approach is more reliable long-term.

.github/actions/select-ios-simulator/action.yml
samsymons: Good catch, yeah that should be addressed before merging - lemme fix it now.

samsymons:
Duck Player UI test failures are known and have been fixed on main, unrelated to the changes here.

PR #1676, opened 2025-08-14
Fix DuckPlayer compatibility on macOS 11 (Big Sur)

afterxleep:
Task URL: https://app.asana.com/1/137249556945/project/1204099484721401/task/1210968460248838?focus=true

## Description
PR #1331 introduced a breaking change for macOS 11 (Big Sur) users causing blank screens in DuckPlayer. The async delay workaround uses `loadSimulatedRequest` which is only available on macOS 12+.

This fix applies the workaround only for macOS 12+ while keeping the original fallback method for macOS 11.

## How to Test
- [ ] Smoke Test DuckPlayer functionality on macOS 11 (Big Sur) - Should load content without blank screen
- [ ] Smoke Test DuckPlayer functionality on macOS 12+ - Should use the async delayed loading method

MacOS 11 Screenshot:
DuckPlayer


afterxleep:
@ayoy Please merge it once approved - Thank you!

ayoy (approved):
LGTM

PR #1686, opened 2025-08-17
Sanitize VPN errors before emitting them

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1211073812765067?focus=true
Tech Design URL:
CC:

### Description

This PR updates the VPN error handling chain to avoid emitting those with issues encoding for XPC.

### Testing Steps

1. Check that the method of XPC encoding is good enough
2. Check that the tests are looking for the right conditions

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
+++ b/SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
@@ -1,6 +1,5 @@
 //
 //  XPCErrorSanitizer.swift
-//  VPN
 //
 //  Copyright © 2024 DuckDuckGo. All rights reserved.
 //
--- a/SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
+++ b/SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
@@ -1,0 +1,60 @@
+//
+//  XPCErrorSanitizer.swift
+//  VPN
+//
+//  Copyright © 2024 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public struct XPCErrorSanitizer {
+    
+    public static func sanitize(_ error: Error) -> Error {
+        let nsError = error as NSError
+        return isXPCSafe(nsError) ? error : PacketTunnelProvider.TunnelError.xpcIncompatibleError(underlyingError: error)
+    }
+
+    static func isXPCSafe(_ error: NSError) -> Bool {
+        do {
+            _ = try NSKeyedArchiver.archivedData(withRootObject: error, requiringSecureCoding: true)
+            return true
+        } catch {
+            return false
+        }
+    }
+
+}
+
+extension Error {
+    func sanitizedForXPC() -> Error {
+        return XPCErrorSanitizer.sanitize(self)
+    }
+}
+
+extension PacketTunnelProvider.TunnelError {
+    var sanitizedForXPC: PacketTunnelProvider.TunnelError {
+        switch self {
+        case .couldNotGenerateTunnelConfiguration(let internalError):
+            return .couldNotGenerateTunnelConfiguration(internalError: internalError.sanitizedForXPC())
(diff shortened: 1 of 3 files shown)


SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
```diff
@@ -0,0 +1,59 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
```
diegoreymendez, line 20: I'd suggest moving the error here, so this code stays more generic and better modularized.  Also, this error might apply to underlying errors, and not just the top-level tunnel errors.

See my other suggestions below for more context.

```suggestion
struct SanitizedError: CustomNSError {
    let wrappedError: Error
    let underlyingError: Error?

    init(wrappedError: Error, underlyingError: Error?) {
        self.wrappedError = wrappedError
        self.underlyingError = underlyingError
    }

    public var errorUserInfo: [String: Any] {
        let ns = wrappedError as NSError
        var userDictionary: [String: Any] = [
            "OriginalErrorDomain": ns.domain,
            "OriginalErrorCode": ns.code,
            "OriginalErrorDescription": ns.localizedDescription,
        ]

        if let underlyingError = underlyingError as NSError? {
            userDictionary["UnderlyingError"] = underlyingError
        }

        return userDictionary
    }
}
```

SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
```diff
@@ -0,0 +1,59 @@
...
+public struct XPCErrorSanitizer {
+
+    public static func sanitize(_ error: Error) -> Error {
+        let nsError = error as NSError
+        return isXPCSafe(nsError) ? error : PacketTunnelProvider.TunnelError.xpcIncompatibleError(underlyingError: error)
+    }
```
diegoreymendez: The second suggestion is to keep underlying error info.  The underlying error info will be sanitized too.  This way we can figure out which error is introducing this issue if it's deeper down the chain.

```suggestion
    public static func sanitize(_ error: Error) -> Error {
        let nsError = error as NSError

        guard !XPCErrorSanitizer.isXPCSafe(nsError) else {
            return error
        }

        let underlyingError = nsError.userInfo[NSUnderlyingErrorKey] as? NSError
        return SanitizedError(wrappedError: error, underlyingError: underlyingError?.sanitizedForXPC())
    }
```

diegoreymendez (commented):
Hey @samsymons !  I think the approach is really good, but I think we could make it even better by trying to keep underlying error information past level 2 in the case there's an XPC-incompatible error.

I tried to provide an approach that's simple and that hopefully doesn't require a ton of extra work - I think the tests should mostly keep working with those changes.

But please let me know if I'm just making things more difficult - I'm open to dropping my suggestions if they introduce problems.

SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
```diff
@@ -0,0 +1,59 @@
...
+        case .vpnAccessRevoked(let underlyingError):
+            return .vpnAccessRevoked(underlyingError.sanitizedForXPC())
+        case .startingTunnelWithoutAuthToken(let internalError):
+            return .startingTunnelWithoutAuthToken(internalError: internalError?.sanitizedForXPC())
+        default:
+            let nsError = self as NSError
+            return XPCErrorSanitizer.isXPCSafe(nsError) ? self : .xpcIncompatibleError(underlyingError: self)
+        }
+    }
+}
```
diegoreymendez: If you accept the changes above, I'd remove this entirely, and I'd remove xpcIncompatibleError. The code above should deal with all of what's being done here.

```suggestion
```

SharedPackages/VPN/Sources/VPN/XPCErrorSanitizer.swift
```diff
@@ -0,0 +1,75 @@
...
+        let ns = wrappedError as NSError
+
+        var userDictionary: [String: Any] = [
+            "OriginalErrorDomain": ns.domain,
+            "OriginalErrorCode": ns.code,
+            "OriginalErrorDescription": ns.localizedDescription,
+        ]
+
+        if let underlyingError = underlyingError as NSError? {
+            userDictionary["UnderlyingError"] = underlyingError
```
diegoreymendez: Nit: You can use `NSUnderlyingError`

samsymons: Whoops, yep let me fix that - thanks!

samsymons: Done, this now uses `NSUnderlyingErrorKey` instead.

diegoreymendez (approved):

PR #1713, opened 2025-08-19
macOS: Open Fire Window by default

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1209904613939706?focus=true
Tech Design URL:
CC:

### Description
We are introducing a new feature that allows our users to set the Fire Window as their default. What do these mean?
- When opening a new window (tapping CMD+N), it will open a Fire Window instead of a non-Fire one.
- Shortcuts will be swapped (CMD+N will open a Fire Window instead of a non-Fire)
- Menu items will be swapped in order (if the setting is ON, Open Fire Window will be first)
- Users can still open non-Fire windows, and the restore session works with this setting enabled.

The change is behind a feature flag, which is currently internal for users.

#### Setting image 
Screenshot 2025-08-19 at 6 17 13 PM

#### Demo
https://github.com/user-attachments/assets/57396f09-5a59-4e1f-af09-316fdeb729cf

### Testing Steps
1. Open the application, and make sure you are an internal user
2. Go to Settings -> Data Clearing, check the ‘Open Fire Window by default’
3. Open the File menu, check that `New Fire Window` comes before `New Window`, and CMD+N is set to open a Fire Window. Use CMD+N to verify that it opens a Fire window; ensure that CMD+SHIFT+N opens a non-Fire window.
4. The above should also be checked for the `MoreOptionsMenu`. Also the right tapping the application if in the dock, should show the items in the new order (check the demo video if in doubt)
5. Check that General -> On Startup -> `Reopen all windows from last session` is off.
6. Restart the application
7. Make sure that a Fire Window is opened by default
8. Open a couple of non-Fire windows and turn on General -> On Startup -> `Reopen all windows from last session`.
9. Restore the application; the non-Fire windows should be restored.
10. Go to Settings -> Data Clearing, uncheck the ‘Open Fire Window by default’
11. Check that shortcuts are in the menu items and positions are back to normal

### Impact and Risks

#### What could go wrong?
Edge cases that weren’t addressed.

### Quality Considerations
I do not know of any edge cases that might exist.

### Notes to Reviewer
Given your experience, Tom, make sure I’m not missing anything.

—
###### Internal references:


macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -693,7 +697,7 @@ final class MainMenu: NSMenu {
                 NSMenuItem(title: "Reset MakeDuckDuckYours user settings", action: #selector(AppDelegate.resetMakeDuckDuckGoYoursUserSettings))
                 NSMenuItem(title: "Experiment Install Date more than 5 days ago", action: #selector(AppDelegate.changePixelExperimentInstalledDateToLessMoreThan5DayAgo(_:)))
                 NSMenuItem(title: "Change Activation Date") {
-                    NSMenuItem(title: "Today", action: #selector(AppDelegate.changeInstallDateToToday), keyEquivalent: "N")
+                    NSMenuItem(title: "Today", action: #selector(AppDelegate.changeInstallDateToToday))
```
jotaemepereira, line 705: ℹ️ I removed this because this was wrong and conflicting with other shorcuts (it is a debug menu item). It seems it was just a slip on a copy/paste situation, given that all the other items do not have a shortcut.

jotaemepereira:
> @jotaemepereira, I really like the simplicity of this feature 👍
> 
> 1. I suppose the testing step 5 should be "Check that General -> On Startup -> Reopen all windows from last session is **on.**", correct?
> 2. I've noticed opening AI Chat -> Option + Command + N opens a regular window even when the setting is enabled. Perhaps this should be open in Fire Window too?

@tomasstrba Let me add some replies:

1. For step 5, it is correct; it should be **off**. The reason is that if it is on and you have a non-Fire window, it will be restored. I wanted to verify the scenario where the Fire Window is opened by default.

2. Great catch. I’ve pushed a fix for it.

PR #1721, opened 2025-08-19
Internal subscription alert

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1211088979568167?focus=true

### Description

This PR implements a user-friendly alert message for DDG internal subscriptions when users attempt to manage their subscription. Previously, when a subscription had no platform associated with it (DDG internal subscriptions), the "update plan or cancel" button would do nothing.

- Adds standardized alert messaging for both iOS and macOS platforms
- Replaces assertion failure with proper user feedback for internal subscriptions
- Implements consistent UI flow across both platforms

### Testing Steps

1. Restore your internal privacy pro subscription (the one on your ddg email)
2. Click/tap on Subscription > Subscription Settings > Update plan or cancel
3. You should see the new alert
4. Remove subscription
5. Subscribe with Apple sandbox account
6. Click/tap on Subscription > Subscription Settings > Update plan or cancel
7. You should be redirected to App Store

### Impact and Risks

Low, the new alert is visible only when a subscription has no platform associated with it, which happens only for DDG internal subscriptions.

---
###### Internal references:

### DIFF
--- a/macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionSettingsModelV2.swift
+++ b/macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionSettingsModelV2.swift
@@ -155,7 +155,12 @@ hasActiveTrialOffer: \(hasTrialOffer, privacy: .public)
     func changePlanOrBillingAction() async -> ChangePlanOrBillingAction {
         userEventHandler(.didClickChangePlanOrBilling)
 
-        switch subscriptionPlatform {
+        guard let platform = subscriptionPlatform else {
+            assertionFailure("Missing or unknown subscriptionPlatform")
+            return .navigateToManageSubscription { }
+        }
+
+        switch platform {
         case .apple:
             return .navigateToManageSubscription { [weak self] in
                 self?.changePlanOrBilling(for: .appStore)
@@ -166,7 +171,7 @@ hasActiveTrialOffer: \(hasTrialOffer, privacy: .public)
             return .navigateToManageSubscription { [weak self] in
                 self?.changePlanOrBilling(for: .stripe)
             }
-        default:
+        case .unknown:
             return .showInternalSubscriptionAlert
         }
     }


iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -178,6 +179,7 @@ final class SubscriptionSettingsViewModel: ObservableObject {
         case .stripe:
             Task { await manageStripeSubscription() }
         default:
+            manageInternalSubscription()
```
rachelmcr, line 188: To confirm, is it only possible to encounter the default case as an internal user?

It might be worth explicitly handling each case so we’re sure we only calling this for internal users. (For example, we don’t want to accidentally call it if a new `Platform` case is introduced.)

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModelV2.swift
```diff
@@ -184,8 +185,7 @@ final class SubscriptionSettingsViewModelV2: ObservableObject {
         case .stripe:
             Task { await manageStripeSubscription() }
         default:
-            assertionFailure("Invalid subscription platform")
-            return
+            manageInternalSubscription()
```
rachelmcr, line 194: Same question as above: is it only possible to encounter the default case as an internal user, and are we sure that will be true in the future?

macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionSettingsModelV2.swift
```diff
@@ -166,8 +167,7 @@ hasActiveTrialOffer: \(hasTrialOffer, privacy: .public)
                 self?.changePlanOrBilling(for: .stripe)
             }
         default:
-            assertionFailure("Missing or unknown subscriptionPlatform")
-            return .navigateToManageSubscription { }
+            return .showInternalSubscriptionAlert
```
rachelmcr, line 175: Same question as the iOS implementation: Are we confident that the default case will only be triggered by internal users?

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: Fair observation, I've removed the previously present default case and used the `.unknown` case that was not handled.

I can confirm that the only case where the platform is `.unknown` is when it's a DDG internal subscription.

macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionSettingsModelV2.swift
federicocappelli: fixed

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModelV2.swift
federicocappelli: modified as above

macOS/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/Preferences/PreferencesSubscriptionSettingsModelV2.swift
federicocappelli: modified as above

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
rachelmcr: Looks great! 👍

PR #1739, opened 2025-08-20
Fix - NTP address bar click pixel is not sent when using experimental search bar

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/715106103902962/task/1211080567413562?focus=true
Tech Design URL:
CC:

### Description
Fix - NTP address bar click pixel is not sent when using experimental search bar

### Testing Steps
1. Enable the experimental address bar under AI Features
2. Check if the correct pixels are sent on these actions on NTP, SERP and website:
3. Select the address bar
4. Dismiss the address bar
5. Clear the text in the address bar

### Impact and Risks

#### What could go wrong?
Send the wrong pixels

### DIFF
--- a/iOS/DuckDuckGo/DefaultOmniBarViewController.swift
+++ b/iOS/DuckDuckGo/DefaultOmniBarViewController.swift
@@ -60,7 +60,7 @@ final class DefaultOmniBarViewController: OmniBarViewController {
 
     override func textFieldShouldBeginEditing(_ textField: UITextField) -> Bool {
         if aiChatSettings.isAIChatSearchInputUserSettingsEnabled {
-            omniDelegate?.onExperimentalAddressBarTappedForPixelsOnly()
+            omniDelegate?.onExperimentalAddressBarTapped()
             presentExperimentalEditingState(for: textField)
             return false
         }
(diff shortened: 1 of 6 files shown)


iOS/DuckDuckGo/DefaultOmniBarViewController.swift
```diff
@@ -58,6 +60,7 @@ final class DefaultOmniBarViewController: OmniBarViewController {
 
     override func textFieldShouldBeginEditing(_ textField: UITextField) -> Bool {
         if aiChatSettings.isAIChatSearchInputUserSettingsEnabled {
+            omniDelegate?.onExperimentalAddressBarTapped()
```
dus7: This will be fired on every first responder attempt (even programmatic - e.g. opening a new tab) so you need to check `OmniBarViewController.textFieldTapped` property. See `MainViewController.onTextFieldWillBeginEditing`.

dus7 (changes requested):

iOS/DuckDuckGo/DefaultOmniBarViewController.swift
Bunn: Fixed, could you take another look?

dus7 (approved):

PR #1767, opened 2025-08-25
Refactor sync for presentation from new entry points

graeme:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1142021229838617/task/1210785630825151?focus=true
Tech Design: https://app.asana.com/1/137249556945/project/481882893211075/task/1211137446160146?focus=true

### Description

The macOS browser's sync functionality was previously tightly coupled to the sync settings screen through a monolithic SyncPreferences class. This architecture made it difficult to launch sync device connection flows from other parts of the application preventing sync onboarding in contexts where it would be most valuable.

The refactoring addresses the need to display sync device connection flows from multiple entry points throughout the app, not just the dedicated sync settings screen. This enables features like onboarding new users to sync during data import flows, bookmark management, and other natural sync adoption moments.

### Testing Steps
1. Smoke test sync (particularly device connection flows).

### Impact and Risks

#### What could go wrong?
Regressions in Sync device connection flow

### Quality Considerations
I'm shipping this behind a feature flag to de-risk.

### Notes to Reviewer

---
###### Internal references:


macOS/DuckDuckGo/Sync/DeviceSyncCoordinator.swift
```diff
@@ -0,0 +1,164 @@
...
+        dialogController.coordinationDelegate = self
+    }
+
+    private let managementDialogModel: ManagementDialogModel
+    private let dialogController: SyncDialogController
+    private var syncWindowController: NSWindowController?
+
+    @MainActor
+    private func presentDialog(completion: (() -> Void)? = nil) {
+        guard !(syncWindowController?.window?.isVisible ?? false) else {
```
graeme: The contents of this used to be in `SyncPreferences.presentDialog`

macOS/DuckDuckGo/Preferences/Model/SyncPreferences.swift
```diff
@@ -189,25 +176,31 @@ final class SyncPreferences: ObservableObject, SyncUI_macOS.ManagementViewModel
...
-        self.userAuthenticator = userAuthenticator
         self.syncPausedStateManager = syncPausedStateManager
-        self.connectionControllerFactory = connectionControllerFactory
-        self.featureFlagger = featureFlagger
 
         self.isFaviconsFetchingEnabled = syncBookmarksAdapter.isFaviconsFetchingEnabled
         self.isUnifiedFavoritesEnabled = appearancePreferences.favoritesDisplayMode.isDisplayUnified
 
-        self.managementDialogModel = managementDialogModel
+        let syncSettingsHandler = SyncDialogController(
```
ayoy: Maybe call it `dialogController`?

macOS/DuckDuckGo/Preferences/Model/SyncPreferences.swift
```diff
@@ -189,25 +176,31 @@ final class SyncPreferences: ObservableObject, SyncUI_macOS.ManagementViewModel
...
+            syncService: syncService,
+            managementDialogModel: managementDialogModel,
+            userAuthenticator: userAuthenticator,
+            syncPausedStateManager: syncPausedStateManager,
+            connectionControllerFactory: connectionControllerFactory,
+            featureFlagger: featureFlagger
+        )
+
+        self.syncSettingsHandler = DeviceSyncCoordinator(managementDialogModel: managementDialogModel, dialogController: syncSettingsHandler) ?? syncSettingsHandler
```
ayoy: As I understand, the fallback here shouldn’t be happening in production and it’s only to satisfy the compiler because technically DeviceSyncCoordinator may be nil (because it accesses optional AppDelegate properties). Could you perhaps add a comment explaining what’s happening here?

macOS/DuckDuckGo/Sync/DeviceSyncCoordinator.swift
ayoy: How about changing this guard to this, which I believe is simpler?
```suggestion
        guard syncWindowController?.window?.isVisible != true else {
```

macOS/DuckDuckGo/Sync/DeviceSyncCoordinator.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        parentWindowController.window?.beginSheet(syncWindow) { _ in
+            completion?()
+        }
+    }
+
+    @MainActor
+    private static func createDialogController(managementDialogModel: ManagementDialogModel) -> SyncDialogController? {
+        guard let syncService = NSApp.delegateTyped.syncService, let errorHandler = NSApp.delegateTyped.syncDataProviders?.syncErrorHandler else {
+            assertionFailure("Sync: Core dependencies not available")
```
ayoy: Just a reminder that if you think this may be happening in production, we could be sending a pixel here.

ayoy (approved):
Overall this looks great and works well for me, great job @graeme!

I’m leaving a few minor comments, and I’d love to see more documentation, at least for the new protocols' APIs: SyncDeviceFlowLaunching, SyncSettingsViewHandling and DeviceSyncCoordinationDelegate.

Thanks!

macOS/DuckDuckGo/Preferences/Model/SyncPreferences.swift
graeme: I've actually cleared this up a bit in the following PR so this doesn't have any fallback and instead calls a non-optional initialiser

macOS/DuckDuckGo/Sync/DeviceSyncCoordinator.swift
graeme: Same as above: This isn't happening any more after a fix in the following PR. Instead, the only places we're handling the nil case are ViewModels which will just hide the Sync button in that case. Will probably never happen in production but it's a slightly more elegant fallback in the slim chance that sync dependencies were not ready yet.

PR #1783, opened 2025-08-26
Add actionlint

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1211153127406668?focus=true
Tech Design URL:
CC:

### Description

This PR adds a dedicated actionlint workflow.

As part of this PR I've also removed some old workflows that relied on old runners. They've been disabled for a while, but we may we well remove them now since macOS 11 and 12 runners are never coming back. 💔

### Testing Steps

1. See failing checks on this PR (such as [this](https://github.com/duckduckgo/apple-browsers/actions/runs/17246479671/job/48937373692)) to confirm that it is identifying syntax issues correctly
2. Check that CI is now green after fixing the issues
3. Check that the fixes made to satisfy `actionlint` are legit

### Notes to Reviewer

---
###### Internal references:


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,17 @@
...
+  actionlint:
+    name: Actionlint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check workflow files
+        uses: docker://rhysd/actionlint:latest
+        with:
+          args: -color -shellcheck=
```
samsymons, line 17: We can remove the `-shellcheck=` argument to have it also run `shellcheck` on scripts within the actions. I've disabled it for now just to assess whether we want to proceed with integrating `actionlint` to begin with, and would be happy to come back with fixes for the `shellcheck` issues if so. (Maybe in a separate PR, so that we can get the core syntax checking available ASAP.)

.github/workflows/ios_hotfix.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     outputs:
       release_branch_name: ${{ steps.make_release_branch.outputs.release_branch_name }}
-      asana_task_url: ${{ steps.create_release_task.outputs.asana_task_url }}
+      asana_task_url: ${{ steps.make_release_branch.outputs.asana_task_url }}
```
ayoy: Fortunately neither of the outputs here nor in the macOS hotfix workflow is in use, so we could remove all outputs from these two.

ayoy (approved):
❤️

.github/workflows/ios_hotfix.yml
samsymons: Done, these have been removed and I checked the Fastlane plugin to confirm that we set them through `GitHubActionsHelper.set_output`.

PR #1731, opened 2025-08-20
[iOS] Add Push Notification Pixel Definition

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211003501974958?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1211032403913073?focus=true

### Description

Added a new pixel definition: `m_push-notification_local-provisional_inactive-user-tap`
- Fires when a user taps a local provisional push notification scheduled due to inactivity to measure engagement with provisional push notifications on iOS
- Captures: appVersion, pixelSource, daysInactive, copyNumber, and form_factor.

### Testing Steps

1. Build and run the app.
2. Confirm parameters are correctly defined.

### Impact and Risks

No risks — pixel definition only

#### What could go wrong?

Likely none. Incorrect parameter type definitions could lead to dropped events.

### Quality Considerations

Definition follows schema conventions

### Notes to Reviewer
- Added param `daysInactive` due to https://app.asana.com/1/137249556945/project/481882893211075/task/1211032403913073?focus=true showing a possibility of changing this value on the fly
- Added param `copyNumber` due to product asking about multiple copies https://app.asana.com/1/137249556945/task/1211041375427085/comment/1211100532889037?focus=true

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #1794, opened 2025-08-27
[iOS] Schedule Inactivity Push Notification + Pixel Firing

hanyutang-sandra:
@alessandroboron Sorry for the HUGE pr. Feel free to look at the last 2 commits to see new changes after your last review.

Screenshot 2025-08-28 at 16 29 33

Task/Issue URL:  https://app.asana.com/1/137249556945/task/1211003501974964?focus=true + https://app.asana.com/1/137249556945/task/1211003501974968?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1211032403913073?focus=true

### Description
**Update 8/28 (Thursday)**
- Responded to comments
- Added `UNUserNotificationCenterRepresentable` and `MockUNUserNotificationCenter` for injecting `UNUserNotificationCenter` related attributes for testing
- Added more test cases in unite test file `InactivityNotificationSchedulerServiceTests`, including stress test, error path tests and unit test for `resume`, `schedule` and `requestAuthorization`
- Added integration test in `InactivityNotificationSchedulerServiceIntegrationTests`
- Tested locally via local override
-------------
This PR creates a `InactivityNotificationSchedulerService` with it’s  helpers: 
**InactivityNotificationSchedulerService:** Cancels any existing request and reschedules a local notification when authorization status is .provisional.
**NotificationServiceManager:** Small refactor to centralize `UNUserNotificationCenterDelegate`
**Pixel firing:** Added pixel new cases aligned with https://github.com/duckduckgo/apple-browsers/pull/1731
**Unit tests:** Added unit tests for helper methods in `InactivityNotificationSchedulerService`

### Testing Steps

Hard to test at the moment. Will introduce debug menu item for manual testing in https://app.asana.com/1/137249556945/task/1211035568337399?focus=true

Tested via local override:
IMG_8973

### Impact and Risks

Medium: Affects VPN notifications and tap flows due to changes in `UNUserNotificationCenterDelegate` in `NotificationServiceManager`. The new feature is guarded behind feature toggle.

#### What could go wrong?

- Misconfiguration of daysInactive (non-numeric or unexpected shape)
   Mitigation: Parser is defensive; on error we log and fall back to default (7 days) so we don’t spam or under-notify.
- Duplicate or repeated notifications after rapid app relaunches
   Mitigation: We cancel pending requests for our identifier before scheduling, ensuring at most one pending inactivity notification.
- User annoyance from unexpected notifications
   Mitigation: Remote kill-switch; conservative default; staged rollout; easy rollback if feedback or metrics indicate issues.

### Quality Considerations

- Performance: The scheduler always cancels by identifier, then reschedules one request on resume(). In rare situations (e.g., rapid foregrounding), this can result in multiple remove/re-add cycles within a session. Work is O(1) and limited to a single identifier, so impact is limited; logic is idempotent to prevent growth of pending requests.
- Edge cases: Missing/empty config; malformed JSON; non-numeric daysInactive. Parser is defensive and falls back to default (7 days).
- Trigger validity: UNTimeIntervalNotificationTrigger requires timeInterval >= 1. Our default ensures this; malformed config logs an error and falls back, avoiding invalid triggers.
- Observability: Errors on parsing/scheduling are logged; tap engagement is measured via the pixel. Remote feature control provides an immediate kill-switch if metrics or feedback regress.

### Notes to Reviewer

Sorry for the _HUGE_ pr. Most are tests.

---
###### Internal references:


iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -0,0 +1,148 @@
...
+         userNotificationCenter: UNUserNotificationCenter = .current(),
+         privacyConfigurationManager: PrivacyConfigurationManaging = ContentBlocking.shared.privacyConfigurationManager,
+         notificationServiceManager: NotificationServiceManaging? = nil
+    ) {
+        self.featureFlagger = featureFlagger
+        self.userNotificationCenter = userNotificationCenter
+        self.privacyConfigurationManager = privacyConfigurationManager
+        self.notificationServiceManager = notificationServiceManager
+        
+        if let notificationServiceManager {
```
hanyutang-sandra: This is to avoid global side effects in unit tests. Making the delegate parameter optional (default nil) lets tests construct the scheduler without mutating the global notification center.

iOS/Core/PixelEvent.swift
```diff
@@ -1341,6 +1341,9 @@ extension Pixel {
         case systemSettingsPiPTutorialFailedToLoadVideo
 
         case appDidTerminateWithUnhandledError
+        
+        // MARK: - Push Notifications
+        case provisionalPushNotificationTapped
```
alessandroboron: Perhaps we should give this pixel a more specific name to distinguish it from future cases involving other notification tap events. How about naming it `inactiveUserProvisionalNotificationTapped`?

iOS/DuckDuckGo/AppServices/VPNService.swift
```diff
@@ -30,21 +30,28 @@ final class VPNService: NSObject {
...
 
     private let mainCoordinator: MainCoordinator
     private let subscriptionManager: any SubscriptionAuthV1toV2Bridge
     private let application: UIApplication
     init(mainCoordinator: MainCoordinator,
          subscriptionManager: any SubscriptionAuthV1toV2Bridge = AppDependencyProvider.shared.subscriptionAuthV1toV2Bridge,
          application: UIApplication = UIApplication.shared,
-         notificationCenter: UNUserNotificationCenter = .current()) {
+         notificationCenter: UNUserNotificationCenter = .current(),
+         notificationServiceManager: NotificationServiceManaging? = nil,
```
alessandroboron: I notice that VPNService is only initialised from Launching, and it doesn’t appear that there are any cases where notificationServiceManager should be nil. I believe we should enforce setting a delegate by removing the optional.

iOS/DuckDuckGo/UserText.swift
```diff
@@ -817,6 +817,9 @@ public struct UserText {
     // MARK: Notifications
 
     public static let macWaitlistAvailableNotificationTitle = NSLocalizedString("mac-waitlist.available.notification.title", value: "DuckDuckGo for Mac is ready!", comment: "Title for the macOS waitlist notification")
+    /// https://app.asana.com/1/137249556945/project/38424471409662/task/1211092472261869
+    public static let inactivityNotificationTitle = NSLocalizedString("inactivity.notification.title", value: "Trackers are lurking!", comment: "Title for inactivity notification")
```
alessandroboron: Are we still runinng this only for en_US users? If so we want to exclude the strings from translations. We can do that by using `NotLocalizedString` instead of `NSLocalizedString`.

iOS/DuckDuckGo/en.lproj/Localizable.strings
```diff
@@ -1858,6 +1858,12 @@
 /* Title for the Import Passwords Promotion banner */
 "import.passwords.promo.title" = "Import your passwords to DuckDuckGo";
 
+/* Body for inactivaity notification */
```
alessandroboron: See my above comment https://github.com/duckduckgo/apple-browsers/pull/1794/files#r2305744096

iOS/DuckDuckGo/NotificationServiceManager.swift
```diff
@@ -0,0 +1,69 @@
...
+                                willPresent notification: UNNotification,
+                                withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+        completionHandler(.banner)
+    }
+
+    func userNotificationCenter(_ center: UNUserNotificationCenter,
+                                didReceive response: UNNotificationResponse,
+                                withCompletionHandler completionHandler: @escaping () -> Void) {
+        
+        guard response.actionIdentifier == UNNotificationDefaultActionIdentifier else { return }
```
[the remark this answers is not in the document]
alessandroboron: As per Apple documentation `completionHandler:  You must execute this block at some point after processing the user’s response to let the system know that you are done`

Looking at the below code

```
case let raw where NetworkProtectionNotificationIdentifier(rawValue: raw) != nil:
            Task { @MainActor [weak self] in
                self?.mainCoordinator.presentNetworkProtectionStatusSettingsModal()
                completionHandler()
            }
```
            
Since we’re using closure and tasks, why don’t we use the async counterpart of the method? We would make the code nicer and don’t have to worry about calling the closure from every path. We could also mark the class @MainActor to be safe. Or we could also split the code and mark only the functions that need to run on main thread.

```
func userNotificationCenter(
    _ center: [UNUserNotificationCenter]
) async {
  // Check notification id and call to the right method
}

private func handleInactiveNotification() {
}

@MainActor
private func handleVPNNotification() {
}
```
   
Let me know your thoughts :)

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -0,0 +1,148 @@
...
+    // MARK: - Dependencies
+    
+    private let featureFlagger: FeatureFlagger
+    private let userNotificationCenter: UNUserNotificationCenter
+    private let privacyConfigurationManager: PrivacyConfigurationManaging
+    private let notificationServiceManager: NotificationServiceManaging?
+    
+    // MARK: - Constants
+    
+    static let notificationIdentifier = "com.duckduckgo.inactivity.notification"
```
alessandroboron: Can we organise these constants in an enum at the top? We do for other classes :)

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -0,0 +1,148 @@
...
+    static let defaultDaysInactive: Double = 7.0 // default to 7 days
+    static let daysInactiveSettingKey: String = "daysInactive"
+    private static let subfeature: any PrivacySubfeature = iOSBrowserConfigSubfeature.inactivityNotification
+    
+    init(featureFlagger: FeatureFlagger = AppDependencyProvider.shared.featureFlagger,
```
alessandroboron: We do have featureFlagger and privacyConfigurationManager already initialised in Launching.swift, check line 44 and 47. Lets just pass them to be explicit on what they need from Launching.

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -0,0 +1,148 @@
...
+                Logger.pushNotification.error("Inactivity notification authorization request failed with \(error.localizedDescription, privacy: .public)")
+            }
+        default:
+            break
+        }
+    }
+    
+    private func buildUNNotificationRequest() -> UNNotificationRequest {
+        let daysInactive = makeDaysInactive()
+        let trigger = UNTimeIntervalNotificationTrigger(timeInterval: daysInactive.toSeconds(), repeats: false)
```
alessandroboron: We already have an extension that does this so you can use `UNTimeIntervalNotificationTrigger(timeInterval: .days(daysInactive), repeats: false)` so we can remove 

```
private extension Double {
    func toSeconds() -> Double {
        return self * 60 * 60 * 24
    }
}
```

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -0,0 +1,148 @@
...
+import UserNotifications
+import Core
+import BrowserServicesKit
+
+final class InactivityNotificationSchedulerService {
+    
+    // MARK: - Dependencies
+    
+    private let featureFlagger: FeatureFlagger
+    private let userNotificationCenter: UNUserNotificationCenter
```
alessandroboron: I saw your comment on the PR review regarding re: UNUserNotificationCenter, but I believe we should unit test the behaviour of this class. Additionally, we could write integration tests to check how the class behaves under stress, such as scheduling and cancelling notifications multiple times in a row.

alessandroboron (commented):
Great job @hanyutang-sandra! Thanks for the torough PR description. I left some comments, happy to sync if anything is not clear.

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
hanyutang-sandra: Ok makes sense. I will follow up with a commit to test it

iOS/DuckDuckGo/AppServices/VPNService.swift
hanyutang-sandra: It was set as optional because I’m avoiding unit test side effects. If we will go ahead with protocoling `UNUserNotificationCenter` for test, this will be removed as well

iOS/Core/PixelEvent.swift
hanyutang-sandra: Using `inactiveUserProvisionalPushNotificationTapped` to align with pixel name `m_push-notification_local-provisional_inactive-user-tap`

iOS/DuckDuckGo/NotificationServiceManager.swift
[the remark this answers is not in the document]
hanyutang-sandra: Good callout. Yes I agree. I’m also changing `willPresent`

iOS/DuckDuckGo/UserText.swift
```diff
@@ -817,6 +817,9 @@ public struct UserText {
     // MARK: Notifications
 
     public static let macWaitlistAvailableNotificationTitle = NSLocalizedString("mac-waitlist.available.notification.title", value: "DuckDuckGo for Mac is ready!", comment: "Title for the macOS waitlist notification")
+    /// https://app.asana.com/1/137249556945/project/38424471409662/task/1211092472261869
+    public static let inactivityNotificationTitle = NotLocalizedString("inactivity.notification.title", value: "Trackers are lurking!", comment: "Title for inactivity notification. Used on for en_US locale only.")
+    public static let inactivityNotificationBody = NotLocalizedString("inactivity.notification.body", value: "Over 85% of popular sites contain trackers. Search and browse with DuckDuckGo to help avoid them.", comment: "Body for inactivaity notification. Used on for en_US locale only.")
```
alessandroboron: typo: `inactivaity` -> `inactivity`. Do we need to add "Used on for en_US locale only”? We would need to remember to remove it if we localise the strings, so we don’t confuse the translator.

iOS/DuckDuckGo/UNUserNotificationCenterRepresentable.swift
```diff
@@ -0,0 +1,46 @@
...
+/// Protocol abstraction of `UNUserNotificationCenter`.
+///
+/// This exists to increase testability by allowing dependency injection.
+/// Instead of referencing `UNUserNotificationCenter` directly, depend on this
+/// protocol so that a mock or stub implementation can be provided in unit tests.
+///
+/// For example, production code can use the system `UNUserNotificationCenter.current()`,
+/// while test code can inject a lightweight mock implementation.
+
+protocol UNUserNotificationCenterRepresentable: AnyObject {
```
alessandroboron, line 31: nip: I don’t have a strong opinion on this but how about `UserNotificationCenter`?

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -0,0 +1,145 @@
...
+        guard isFeatureEnabled() else {
+            cancelPendingNotifications()
+            return Task {} // noop
+        }
+        return Task {
+            await schedule()
+        }
+    }
+    
+    private func isFeatureEnabled() -> Bool {
```
alessandroboron: nip: Can we move these private functions together with the other private functions?

iOS/DuckDuckGoTests/AppServices/InactivityNotificationSchedulerServiceTests.swift
```diff
@@ -0,0 +1,326 @@
...
+        
+        // When
+        await service.resume().value
+        
+        // Then
+        XCTAssertTrue(mockUserNotificationCenter.removedIdentifiers.contains { $0.contains(InactivityNotificationSchedulerService.Constants.notificationIdentifier) })
+        XCTAssertEqual(mockUserNotificationCenter.addedRequests.count, 1)
+        XCTAssertEqual(mockUserNotificationCenter.addedRequests.first?.identifier, InactivityNotificationSchedulerService.Constants.notificationIdentifier)
+        
+        let notificationRequest = mockUserNotificationCenter.addedRequests.first!
```
alessandroboron: Let’s not use force unwrapping in this case so if for any reason this is not working as expected we do not crash the test suite. Lets mark the function `throws` and use XCTUnwrap. `let notificationRequest = try XCTUnwrap(mockUserNotificationCenter.addedRequests.first)`

iOS/DuckDuckGo/UserText.swift
hanyutang-sandra: I added `Used for en_US locale only` in the comment. What is the comment used for in this case if we don’t need to have them translated

iOS/DuckDuckGoTests/AppServices/InactivityNotificationSchedulerServiceTests.swift
hanyutang-sandra: Ok that makes sense

iOS/DuckDuckGo/UNUserNotificationCenterRepresentable.swift
hanyutang-sandra: I want to make sure folks are clear that this is **representing** the original one. `UserNotificationCenter` sounds really similar to the original `UNUserNotificationCenter` so it can be confusing

PR #1807, opened 2025-08-29
Improve Navigation and Configuration tests resources usage

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205237866452338/task/1211168459107850?focus=true

### Description
- Fix DefaultConfigurationManager leak in ConfigurationTests by adding the `tearDown` method unregistering it as a file presenter (I‘ve noticed there‘s a lot of file presenter event logs during the Navigation tests running, so it possibly can affect performance)
- Disable Apple SafeBrowsing checks in test WebView configuration (I‘ve noticed the logs contain Safe Browsing checks on each request as well)
- Fixed leaking Navigation objects in the Responder mock and leaking TestSchemeHandler initialized for every test
- Increase standard navigation expectation wait time

### Testing Steps
1. Ensure CI is green

### Impact and Risks

No

#### What could go wrong?

Only tests affected 

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/SharedPackages/BrowserServicesKit/Tests/NavigationTests/Helpers/TestNavigationSchemeHandler.swift
+++ b/SharedPackages/BrowserServicesKit/Tests/NavigationTests/Helpers/TestNavigationSchemeHandler.swift
@@ -28,7 +28,7 @@ final class TestNavigationSchemeHandler: NSObject, WKURLSchemeHandler {
 
     public var onRequest: ((WKURLSchemeTask) -> Void)?
 
-    init(requestHandlers: [URL : RequestResponse] = [URL: RequestResponse](), onRequest: ((WKURLSchemeTask) -> Void)? = nil) {
+    init(requestHandlers: [URL: RequestResponse] = [URL: RequestResponse](), onRequest: ((WKURLSchemeTask) -> Void)? = nil) {
         self.requestHandlers = requestHandlers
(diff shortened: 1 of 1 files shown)


SharedPackages/BrowserServicesKit/Sources/Configuration/DefaultConfigurationManager.swift
```diff
@@ -70,7 +70,7 @@ open class DefaultConfigurationManager: NSObject {
         NSFileCoordinator.addFilePresenter(self)
     }
 
-    deinit {
+    public func tearDown() {
```
ayoy, line 73: Shouldn’t we still call removeFilePresenter in deinit? tearDown is otherwise not called by the app code.

ayoy (approved):
LGTM except that one comment about DefaultConfigurationManager

SharedPackages/BrowserServicesKit/Sources/Configuration/DefaultConfigurationManager.swift
mallexxx: The issue is File Coordinator holding the ref strongly and `deinit` would never happen.
I‘m assuming the DefaultConfigurationManager is created once per app run so tearDown is only used in tests

PR #1814, opened 2025-08-29
Make keychain item for sync available after first unlock

bwaresiak:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1211172720872304?focus=true
Tech Design URL:
CC:

### Description

Update keychain item access to avoid logging people out when app update is automatically applied while device screen is locked.

### Testing Steps

This will be tricky to test locally with a version built from Xcode.

Regular testing steps:
Get some older version (1-2 versions ago). 
Setup sync.
Set it to download and apply browser update.
Lock the screen.
Wait 10min - unlock screen, see that new browser version is there, sync should be still enabled.

### Notes to Reviewer

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
```diff
@@ -323,6 +324,15 @@ public class DDGSync: DDGSyncing {
...
         try syncQueue.prepareDataModelsForSync(needsRemoteDataFetch: account.state == .addingNewDevice)
 
+        do {
+            if try dependencies.keyValueStore.object(forKey: Constants.keychainAttrMigratedKey) == nil {
+                //Remove account from keychain to re-apply attribute changes
+                try? dependencies.secureStore.removeAccount()
+
+                try dependencies.keyValueStore.set(true, forKey: Constants.keychainAttrMigratedKey)
+            }
+        } catch {}
```
bwaresiak, line 334: @graeme We have a balance here: either fix it by migrating people (removing - saving) risking that in some bizarre scenario we could just lose the account, or ignore it and hope that the problem disappears with time (as people reconfigure sync).

I'm rather leaning towards migrating now, what's your take?

graeme: I think we migrate, but do it in `initializeIfNeeded` directly rather than in `updateAccount` as there's only one entry point (`AppDelegate`).

bwaresiak: I was thinking to use updateAccount to cover the account creation/login as well - so newly created accounts would be already marked as "migrated". We could do it explicitly if you want.

graeme: Okay yeah, I see what you mean. Yeah, I think this is fine then.

PR #1820, opened 2025-08-30
Stop the VPN expiry notification showing when signing out of the subscription

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1211187909384612?focus=true
Tech Design URL:
CC:

### Description

This PR fixes a bug where the VPN is reported as expired when you sign out of the subscription.

### Testing Steps

1. Sign into the subscription
2. Activate the VPN
3. Sign out of the subscription
4. Check that the VPN is deactivated, and that you don't get an alert saying that your account has expired

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
@@ -413,6 +413,8 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
         return (try? oAuthClient.currentTokenContainer())?.decodedAccessToken.email
     }
 
+    var userInitiatedSignOut: Bool = false
+
     var cachedUserEntitlements: [SubscriptionEntitlement] {
         get {
             userDefaults.userEntitlements
@@ -425,7 +427,10 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
             if !SubscriptionEntitlement.areEntitlementsEqual(currentCachedUserEntitlements, newValue) {
                 Logger.subscription.debug("Entitlements changed - New \(String(describing: newValue)) Old \(String(describing: currentCachedUserEntitlements))")
                 let payload = EntitlementsDidChangePayload(entitlements: newValue)
-                NotificationCenter.default.post(name: .entitlementsDidChange, object: self, userInfo: payload.notificationUserInfo)
+
+                var userInfo = payload.notificationUserInfo
+                userInfo[EntitlementsDidChangePayload.userInitiatedEntitlementChangeKey] = userInitiatedSignOut
+                NotificationCenter.default.post(name: .entitlementsDidChange, object: self, userInfo: userInfo)
             }
         }
     }
@@ -551,6 +556,12 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
     public func signOut(notifyUI: Bool) async {
         Logger.subscription.log("SignOut: Removing all traces of the subscription and account. Notify UI: \(notifyUI ? "true" : "false")")
+        userInitiatedSignOut = true
+
+        defer {
+            userInitiatedSignOut = false
+        }
+
         try? await oAuthClient.logout()
         clearSubscriptionCache()
         if notifyUI {
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -2027,6 +2027,8 @@ class MainViewController: UIViewController {
                 Logger.subscription.fault("Missing entitlements payload")
                 return
             }
+
+            let userInitiatedSignOut = (userInfo[EntitlementsDidChangePayload.userInitiatedEntitlementChangeKey] as? Bool) ?? false
             let hasVPNEntitlements = payload.entitlements.contains(.networkProtection)
             let isAuthV2Enabled = AppDependencyProvider.shared.isUsingAuthV2
             let isSubscriptionActive = try? await subscriptionManager.getSubscription(cachePolicy: .cacheFirst).isActive
@@ -2046,7 +2048,7 @@ class MainViewController: UIViewController {
                         sourceObject: notification.object),
                     frequency: .dailyAndCount)
 
-                if await networkProtectionTunnelController.isInstalled {
+                if await networkProtectionTunnelController.isInstalled && !userInitiatedSignOut {
                     tunnelDefaults.enableEntitlementMessaging()
                 }
 
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/NSNotificationName+Subscription.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/NSNotificationName+Subscription.swift
@@ -37,6 +37,8 @@ public extension NSNotification.Name {
 public struct EntitlementsDidChangePayload {
 
     private let entitlementsKey = "entitlements"
+    public static let userInitiatedEntitlementChangeKey = "userInitiatedEntitlementChange"
+
     public let entitlements: [SubscriptionEntitlement]
     public var notificationUserInfo: [AnyHashable: Any] {
         return [entitlementsKey: entitlements]


SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -551,6 +556,10 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
 
     public func signOut(notifyUI: Bool) async {
         Logger.subscription.log("SignOut: Removing all traces of the subscription and account. Notify UI: \(notifyUI ? "true" : "false")")
+
+        if notifyUI { userInitiatedSignOut = true }
```
samsymons: I'm relying on `notifyUI` to determine whether we want to show the prompt. It's not only set when the user initiated the sign out, but the other cases are less likely to occur when the VPN is installed.

samsymons: Actually, I think this approach won't be good enough as it'll get false positives - I'll improve it before Monday.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -226,7 +226,7 @@ final class SubscriptionSettingsViewModel: ObservableObject {
     }
     
     func removeSubscription() {
-        subscriptionManager.accountManager.signOut()
+        subscriptionManager.accountManager.signOut(skipNotification: false, userInitiated: true)
```
samsymons, line 229: This is the key change - this gets propagated through to the entitlement change notification that is fired, and used to determine whether this was a user-initiated sign out or due to the subscription entitlements being removed.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -2046,12 +2048,17 @@ class MainViewController: UIViewController {
...
                 }
 
                 await networkProtectionTunnelController.stop()
-                await networkProtectionTunnelController.removeVPN(reason: .entitlementCheck)
+
+                if userInitiatedSignOut {
+                    await networkProtectionTunnelController.removeVPN(reason: .signedOut)
+                } else {
+                    await networkProtectionTunnelController.removeVPN(reason: .entitlementCheck)
+                }
```
samsymons, line 2098: This change allows us to fix a pixel bug too, where we were reporting VPN removals as being due to an entitlement check, when really it could have been due to the user signing out.

SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlowV2.swift
```diff
@@ -141,7 +141,7 @@ public final class DefaultAppStorePurchaseFlowV2: AppStorePurchaseFlowV2 {
         case .failure(let error):
             Logger.subscriptionAppStorePurchaseFlow.error("purchaseSubscription error: \(error.localizedDescription, privacy: .public)")
 
-            await subscriptionManager.signOut(notifyUI: false) // TBD see if true is needed
+            await subscriptionManager.signOut(notifyUI: false, userInitiated: false) // TBD see if true is needed
```
diegoreymendez: I'd consider either removing this comment, or clarifying which parameter it refers to.

But in truth I'd just remove it since it's lacking useful details or follow-up tasks.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -413,6 +413,8 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
         return (try? oAuthClient.currentTokenContainer())?.decodedAccessToken.email
     }
 
+    private var userInitiatedSignOut: Bool = false
+
     var cachedUserEntitlements: [SubscriptionEntitlement] {
```
diegoreymendez, line 437: Could we convert this to a method, so we can also pass the `userInitiated` parameter more deterministically?

From what I could see, this is not called from a lot of places so it shouldn't be a huge change.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -549,8 +554,17 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
...
+    public func signOut(notifyUI: Bool, userInitiated: Bool) async {
+        Logger.subscription.log("SignOut: Removing all traces of the subscription and account. Notify UI: \(notifyUI ? "true" : "false"), User Initiated: \(userInitiated ? "true" : "false")")
+
+        if userInitiated {
+            userInitiatedSignOut = true
+        }
+
+        defer {
+            userInitiatedSignOut = false
+        }
```
diegoreymendez: Related to my comment above, this is the only bit concerning me basically because of what cursor-bot commented above.  I realize this shouldn't have a huge impact - but if we can pass these as parameters to a method I think we'd end up with much cleaner code.

federicocappelli:
@samsymons  @diegoreymendez Hold on, there seems to be a lot of changes for something that should be much more straightforward.
1. In the rest of the code shouldn't (!!!) matter which notification arrives first; it matters only in this case, so we should explore the idea of changing the Subscription framework behaviour.
2. Logically, the "sign out" notification should arrive before the entitlements change notification, so I'm not against changing the behaviour.
3. TBH, the OS doesn't guarantee the notifications' delivery order, but that's an optimisation we can approach later.

I can have a look and play with the SubscriptionManagerV2 `public func signOut(notifyUI: Bool) async {`implementation, we may be able to fix this with a 1 line PR.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
samsymons: So the reason for this is that the notification gets fired from a setter, rather than in a function. We could replace the getter-based approach with a function entirely which would make this possible - but I'm curious to see @federicocappelli's approach first, I'll wait for that before doing anything further.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
samsymons: @diegoreymendez Done, the getters have been replaced with computed properties, and the setters have been replaced with functions that now take `userInitiated` as a parameter - the `userInitiatedSignOut` property has been removed.

samsymons:
@federicocappelli I looked at the notification flow again and confirmed what I mentioned in our last call, where `accountDidSignOut` gets called first - this triggers the VPN uninstallation which _should_ stop the entitlement notification handler from showing the expiry logic, but the calls happen too close together and the uninstallation hasn't completed:

```
DEBUG: Received accountDidSignOut
DEBUG: Got onEntitlementsChange
DEBUG: Removing VPN with reason signedOut
```

Given the above I'd like to proceed with this approach unless you feel strongly, since it lets the subscription library tell us exactly what has changed and why.

diegoreymendez (approved):
This is good.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -136,7 +136,7 @@ public protocol SubscriptionManagerV2: SubscriptionTokenProvider, SubscriptionAu
     var userEmail: String? { get }
 
     /// Sign out the user, clear and invalidate the access token and clear the subscription cache
-    func signOut(notifyUI: Bool) async
+    func signOut(notifyUI: Bool, userInitiated: Bool) async
```
federicocappelli, line 154: What about adding an extension to this protocol that implements `func signOut(notifyUI: Bool, userInitiated: Bool = false) async`? so we don't need to change it everywhere?

PR #1830, opened 2025-09-01
Only lock passwords view when non-empty DB

graeme:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210881113393590?focus=true

### Description

iOS and Android only require authentication once you have an item stored in the password manager. Both macOS and Windows require authentication even when empty. Seems we should remove this requirement and match the mobile apps.

### Testing Steps
1. If necessary, clear your secure vault data by going to Debug -> Reset Data -> Reset Autofill Data
2. Open More menu -> Passwords and Autofill
3. **You should immediately see the empty state with no unlock prompt**
4. Add an item either by saving a login or directly in the Passwords and Autofill screen
5. (if necessary) Close the Passwords and Autofill popover and reopen (step 2 again)
6. **You should be prompted to unlock**

Also @amddg44 feel free to do some other smoke tests as you know this stuff better than me! 😅 

### Impact and Risks

#### What could go wrong?
If the logic is wrong, the screen could be unlocked when there are in fact passwords. However, I've made the logic such that the screen is completely inaccessible until the actual data in the database has been fetched. So I think this is pretty unlikely 

---
###### Internal references:


amddg44 (approved):
I’ve tested every scenario I could think of and this is looking solid! Some minor suggestions:
- All items are being fetched twice from the vault (before and after authentication check), perhaps some optimisation can be made here?
- When the vault contains a lot of items, the unlock screen flashes each time you go into the management screen (screen recording below). Perhaps a side effect of the double vault fetch?
- Not sure if this is an issue or not :) When the vault is empty, and I create a new login in the management screen, I’m not prompted to authenticate when I exit and come straight back in. I think this is acceptable, especially as you are required to authenticate once the autolock timer expires; but wanted to flag just in case. 

https://github.com/user-attachments/assets/78d70628-c9ac-4394-bdc8-6fa9876331ed

PR #1831, opened 2025-09-01
[iOS] Add Debug Support

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211035568337399?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1211032403913073?focus=true

### Description
- Copied the debug playground setup from https://github.com/duckduckgo/apple-browsers/tree/alessandro/hack-phase-push-provisional-authorization
- Added supporting utils to connect the debug playground with the notification scheduling logic.
- This allows easier local testing of InactivityNotificationScheduler.

### Testing Steps

1. Launch the app in Debug build and navigate to the new debug playground view.
2. Toggle on the scheduler option and verify that notifications can be scheduled/canceled as expected.

Simulator Screenshot - iPhone 16 - 2025-09-03 at 11 10 00

### Impact and Risks

**Impact:** Low. Debug menu option is only available for debug views
**Risk Level:** Low. Debug menu option is only available for debug views

#### What could go wrong?

- Playground view is not set up correctly. Scheduler not correctly wired in the debug view.
    - Safe to fail, since it only impacts developer testing.

### Quality Considerations

- Verified that debug playground works as expected.
- Slight impact on runtime memory
- No changes to user-facing privacy, security, or data handling.
- Documentation: inline comments included for playground wiring.

### Notes to Reviewer

Added inline

—
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
+++ b/iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
@@ -98,7 +98,6 @@ final class InactivityNotificationSchedulerService {
     
     func makeUNNotificationContent(with daysInactive: Int = Constants.defaultDaysInactive) -> UNNotificationContent {
         let content = UNMutableNotificationContent()
-        
         content.title = UserText.inactivityNotificationTitle
         content.body = UserText.inactivityNotificationBody
         content.userInfo = [Constants.daysInactiveSettingKey: daysInactive]


iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
```diff
@@ -53,6 +62,16 @@ final class InactivityNotificationSchedulerService {
         self.userNotificationCenter.delegate = notificationServiceManager
     }
     
+    /// Playground Initializer
+    convenience init() {
```
hanyutang-sandra: I wanted to wrap this (and other debug utils) inside `#if_DEBUG` to avoid them taking up prod memory, but I got errors from `makingReleaseBuild` job which seems to compile playground in release build and cannot get things wrapped inside `#if_DEBUG`.

alessandroboron: I can see now we have two instances of `InactivityNotificationSchedulerService`, one initialised in Launching.swift and one in `InactivityNotificationSchedulerService`. 

I believe if we want to use this service in the settings we should pass it around, in the same way we do for other services. For example:

```
// DebugScreensViewModel.swift

.view(title: "Default Browser Prompt", { d in // ‘d’ represents the DebugScreen.Dependencies.
    DefaultBrowserPromptDebugView(model: DefaultBrowserPromptDebugViewModel(keyValueFilesStore: d.keyValueStore))
    })
```

Having said that I think we can simplify it. We don’t really need the `InactivityNotificationSchedulerService` to test a local notification. We can schedule a local notification by directly using UNUserNotificationCenter.

For the ship review, what we can do is to make an ad-hoc build from a special branch where we hardcode a short time trigger. When the ship review is approved we just discard the branch without merging it.

I think in this way we keep the InactivityNotificationScheduler simpler and unaware of playground mode. Let me know your thoughts. I’m happy to sync if something is not clear.

iOS/DuckDuckGo/LocalNotificationsPlaygroundView.swift
```diff
@@ -0,0 +1,301 @@
...
+    var body: some View {
+        List {
+            Section {
+                Toggle(isOn: $model.useScheduler) {
+                    Text("Use Scheduler")
+                }
+            } header: {
+                Text(verbatim: "Mode:")
+            } footer: {
+                (Text("Test through ") + Text(model.useScheduler ? "InactivityNotificationScheduler" : "UNUserNotificationCenter").bold())
```
alessandroboron: When we work add copy to Debug Menu we don’t want to make the strings exportable for translations. In SwiftUI we do that by using .init(verbatim: “”) as explained in [macOS: Localization / translation guidelines](https://app.asana.com/1/137249556945/project/904401899170/task/1206727265537758?focus=true) :)

iOS/DuckDuckGo/NotificationServiceManager.swift
```diff
@@ -27,9 +27,9 @@ protocol NotificationServiceManaging: UNUserNotificationCenterDelegate {}
 
 final class NotificationServiceManager: NSObject, NotificationServiceManaging {
     
-    private let mainCoordinator: MainCoordinator
```
alessandroboron: Please see my comment in https://github.com/duckduckgo/apple-browsers/pull/1831/files#r2317852424. If we do that we don’t need to make this change either

iOS/DuckDuckGo/AppServices/InactivityNotificationSchedulerService.swift
hanyutang-sandra: My intention was to provide an easy way to test out this scheduler service since it’s hard to test. Since we can make a special build with a short time trigger for ship review test, yeah I agree there is no need to put it inside debug playground

alessandroboron (approved):
LGTM!

PR #1844, opened 2025-09-03
DDGError pixelkit adoption

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1211174278010319

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -187,11 +186,207 @@ public final class PixelKit {
         logger.debug("👾 PixelKit initialised: dryRun: \(self.dryRun, privacy: .public) appVersion: \(self.appVersion, privacy: .public) source: \(self.source ?? "-", privacy: .public) defaultHeaders: \(self.defaultHeaders, privacy: .public) pixelCalendar: \(self.pixelCalendar, privacy: .public)")
     }
 
+    // MARK: - Public Fire
+
+    public func fire(_ event: Event,
```
federicocappelli, line 191: I'm sorry, here I moved all "fire" functions around because they were all over the place, now they are divided in public and private but, unfortunately, the diff is a mess.

.claude/settings.json
```diff
@@ -22,9 +22,9 @@
         "Bash(git fetch:*)",
         "Bash(domain:github.com)",
         "WebFetch(domain:raw.githubusercontent.com)",
-        "Bash(set -e -o pipefail && xcodebuild:*)",
-        "Bash(xcodebuild:*)",
-        "Bash(xcrun simctl:*)"
+        "set -e -o pipefail && xcodebuild **",
+        "xcodebuild **",
+        "xcrun simctl*"
```
samsymons: Interesting, is that change intentional? I made that `Bash` change to account for a recent Claude Code update, and it's been working well for me locally, but keen to hear if that wasn't the case for you. 😄

SharedPackages/BrowserServicesKit/Sources/PixelKit/WidePixel/WidePixelError.swift
```diff
@@ -43,6 +44,38 @@ public enum WidePixelError: Error, LocalizedError {
...
+            return 5
+        }
+    }
+
+    public var underlyingError: Error? {
+        switch self {
+        case .flowNotFound, .typeMismatch, .invalidFlowState, .invalidParameters:
+            return nil
+        case .serializationFailed(let error), .storageError(let error):
+            return error
```
samsymons, line 75: Thanks for adding this! 👏

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -187,11 +186,207 @@ public final class PixelKit {
...
+                return
+            }
+        }
+
+        let newError: (any DDGError)?
+
+        /*
+         We have 2 options here:
+         - The event is a PixelKitEvent: We handle the error passed in this call
+         - The event is a PixelKitEventV2: We only consider the error in the returned by the PixelKitEventV2 `var error` and assert if an error is also passed to this function
```
samsymons, line 239: Since this is a new API, what do you think about requiring that V2 is used? (FWIW I think the approach here is likely the right way to help it become adopted, but throwing the question out there anyway)

samsymons (approved):
Left a question about constraining the usage of the new API, but approving this since the new subscription parameter handling would be great to include in next week's internal, the improvements to that here are nice. Thanks!

.claude/settings.json
federicocappelli: Claude did it on itself, reverted

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
federicocappelli: I tried to do it, but it is too disruptive. I would need to make another version of the fire function, so we'll end up with 4 different ones :(
The reality is that PixelKitEventV2 is useless, the only difference from PixelKitEvent is the error and we should simply pass errors as event enum parameters... I may do that change, actually, but in a different PR.

PR #1849, opened 2025-09-03
Send debug pixel when BrowserServicesKit fails to load JS file

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1211054087327709
Tech Design URL:
CC: @ayoy @bwaresiak 

### Description

This updates `UserScript.loadJS(_:from:withReplacements:)` (in `BrowserServicesKit`) to throw an error when it fails to load the JS from the provided file/bundle. The thrown error bubbles up to the iOS and macOS clients (and their local DBP packages), which then fire a corresponding pixel and crash the app.

This pixel contains the JS file name, path, and underlying error to help us investigate the underlying cause of the crash that sometimes occurs when loading JS files.

Notable changes:
* Introduces `UserScriptError` to throw the relevant information (JS file name, path, and underlying error) so they can be included in the pixel.
* Introduces an extension to `UserScriptError` in each client to fire the pixel and wait a second, so the pixel has time to be fired before the app crashes.

### Testing Steps

1. Confirm iOS and macOS apps build and run as expected.
2. Force an error to be thrown in each app (see notes to reviewer) and confirm the expected pixel is fired and the app crashes with the expected `fatalError`.

### Impact and Risks

None: Should not affect users (app already crashes at this point)

#### What could go wrong?

There is potential for introducing different kinds of crashes (instead of the expected `fatalError`) if the app ends up in a bad state because of the thrown error, e.g. if an init method throws before its properties are initialized.
I have manually tested the points where a `fatalError` is called to confirm when the expected error is thrown it fires the pixel and results in the expected crash. I also added unit tests to confirm the expected error is thrown and the helper methods fire the expected pixel for each client.

### Quality Considerations

* Pixels added (needs privacy review)

### Notes to Reviewer

You can force a failure in testing by e.g. adding a deliberate `throw` to `UserScript.loadJS(_:from:withReplacements:)` and triggering it from one of the points in the app that calls this method:

```swift
    public static func loadJS(_ jsFile: String, from bundle: Bundle, withReplacements replacements: [String: String] = [:], forceThrow: Bool) throws -> String {

        let path = bundle.path(forResource: jsFile, ofType: "js")!

        if forceThrow {
            let error = NSError(domain: "Test", code: 1)
            throw UserScriptError.failedToLoadJS(jsFile: jsFile, path: path, error: error)
        }
```

Here are sample pixels from each client (directly from the client and from the local DBP package):

**iOS (direct)**

```
Pixel fired m.ios.debug.user.script.load.js.failed.count ["path": "/var/containers/Bundle/Application/4636DC77-CE08-4C18-872B-6FEFF174FE34/DuckDuckGo.app/BrowserServicesKit_BrowserServicesKit.bundle/contentblockerrules.js", "d": "Test", "e": "1", "jsFile": "contentblockerrules"]
```

**macOS (direct)**

```
👾[Daily and Count-Fired] m_mac_debug_user_script_load_js_failed_count ["appVersion": "1.155.0", "d": "Test", "e": "1", "jsFile": "contentblockerrules", "path": "/Applications/DEBUG/DuckDuckGo.app/Contents/Resources/BrowserServicesKit_BrowserServicesKit.bundle/Contents/Resources/contentblockerrules.js", "pixelSource": "browser-dmg"]
```

**DBP iOS**

```
👾[Daily and Count-Fired] m_ios_debug_user_script_load_js_failed_count ["appVersion": "7.185.0", "d": "Test", "e": "1", "jsFile": "contentScope", "path": "/var/containers/Bundle/Application/11EED80E-2891-413E-A470-F5FAB4AA7AE6/DuckDuckGo.app/BrowserServicesKit_ContentScopeScripts.bundle/contentScope.js", "pixelSource": "phone"]
```

**DBP macOS**

```
👾[Daily and Count-Fired] m_mac_debug_user_script_load_js_failed_count ["appVersion": "1.155.0", "d": "Test", "e": "1", "jsFile": "contentScope", "path": "/Applications/DEBUG/DuckDuckGo.app/Contents/Resources/BrowserServicesKit_ContentScopeScripts.bundle/Contents/Resources/contentScope.js", "pixelSource": "browser-dmg"]
```

---
###### Internal references:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211313424738527


SharedPackages/BrowserServicesKit/Sources/UserScript/UserScript.swift
```diff
@@ -61,19 +61,21 @@ extension UserScript {
...
-        }
+            for (key, value) in replacements {
+                js = js.replacingOccurrences(of: key, with: value, options: .literal)
+            }
 
-        return js
+            return js
+        } catch {
+            throw UserScriptError.failedToLoadJS(jsFile: jsFile, path: path, error: error)
+        }
```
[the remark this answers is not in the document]
rachelmcr: Note: I decided not to change the force unwrapping we’re doing for the file path, since this isn’t the source of the crashes we’re investigating and doesn’t seem to be causing any issues at the moment.

iOS/Core/PixelEvent.swift
```diff
@@ -2633,6 +2642,9 @@ extension Pixel.Event {
         case .systemSettingsPiPTutorialFailedToLoadVideo: return "m_picture-in-picture-tutorial_failed-to-load-video"
 
         case .appDidTerminateWithUnhandledError: return "m_app-did-terminate-with-unhandled-error"
+
+        // MARK: UserScript
+        case .userScriptLoadJSFailed: return "m_ios_debug_user_script_load_js_failed"
```
bwaresiak: On iOS convention is to append `ios_<form-factor>` and it happens automatically, just use `m_debug_..` as prefix

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Parameters.swift
```diff
@@ -81,6 +81,10 @@ public extension PixelKit {
 
         // Data import
         public static let importedFavorites = "saved_favorites"
+
+        // UserScript
+        public static let jsFile = "jsFile"
+        public static let path = "path"
```
bwaresiak: I wonder - do we need path (are file names not unique enough?), also full path contains file name - I'm asking from the perspective of reducing the number of things we send in the pixel parameters

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UIWeb/DBPUIWebConfiguration.swift
```diff
@@ -23,21 +23,30 @@ import UserScript
...
+                                                   prefs: prefs,
+                                                   delegate: delegate,
+                                                   webUISettings: webUISettings,
+                                                   vpnBypassService: vpnBypassService)
+        } catch {
+            // Finish initialization before throwing
+            self.dbpUIUserScripts = nil
+            super.init()
+            throw error
+        }
```
bwaresiak: It would be good to have someone from DBP as a stakeholder here to give 👍 just to sanity-check the overall changes in that area

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/SubJobWebRunner/BrokerProfileScanSubJobWebRunner.swift
```diff
@@ -106,7 +106,12 @@ public final class BrokerProfileScanSubJobWebRunner: SubJobWebRunning, BrokerPro
                 }
 
                 task = Task {
-                    await initialize(handler: webViewHandler, isFakeBroker: context.dataBroker.isFakeBroker, showWebView: showWebView)
+                    do {
+                        try await initialize(handler: webViewHandler, isFakeBroker: context.dataBroker.isFakeBroker, showWebView: showWebView)
+                    } catch {
+                        failed(with: error)
```
bwaresiak, line 112: Do I get it right - in the past it used to crash, but now it can complete with error? Where is that error handled, will it be seen anywhere?

The only place I've found was `print("Error scanning \(broker.name): \(error)")` in `RunDBPDebugModeViewModel` - I want to make sure we are not silencing anything here.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UIWeb/DBPUIWebConfiguration.swift
bwaresiak: Also,  do we need to finish the initialization? I'd expect throw to completely invalidate object that is being created, so no optionality would be needed on `dbpUIUserScripts` property as it would never become a proper member of a living object.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Parameters.swift
rachelmcr: Good point, file names are unique enough so we can just send that parameter.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UIWeb/DBPUIWebConfiguration.swift
rachelmcr: We do need to finish the initialization. If we don’t, the app will crash with `EXC_BAD_ACCESS` because the object isn’t fully initialized when it throws.

I was also surprised that throwing didn’t just invalidate the object, so perhaps I’m missing something in how to handle it, but the only approach that worked for me was to ensure `super.init()` was called to finish initialization before throwing.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/SubJobWebRunner/BrokerProfileScanSubJobWebRunner.swift
rachelmcr: This initialize method ultimately calls `loadJS` and was previously crashing here. Now, when `loadJS` throws its error that this `catch` handles it. `failed(with: error)` calls the `continuation` block throwing the error, which is [caught and crashes the app (and fires the corresponding pixel) in `DataBrokerRunCustomJSONViewModel.runJSON(jsonString:)`](https://github.com/duckduckgo/apple-browsers/pull/1849/files#diff-91b1cf98b39edbe9f860dc832f3fba6031ea2111e1c25d58791a721b6cb292e8R431-R441).

rachelmcr: I missed that `RunDBPDebugModeViewModel` also calls this method, so I’ll update that to make sure it’s handled correctly there as well.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/UIWeb/DBPUIWebConfiguration.swift
bwaresiak: > We do need to finish the initialization. If we don’t, the app will crash with EXC_BAD_ACCESS because the object isn’t fully initialized when it throws.

Interesting, this seems `WKUserContentController` we inherit from explodes the moment we throw in derived init.

Maybe we can change it so `DBPUIUserContentController` takes `DBPUIUserScript` as parameter so we can throw outside of the init, directly from `applyDBPUIConfiguration`?

rachelmcr: That’s a great suggestion, thanks @bwaresiak! It makes the throw clearer and more readable; I made that change in 5b4f01c.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/SubJobWebRunner/BrokerProfileScanSubJobWebRunner.swift
THISISDINOSAUR: I think there should be a return after this. Obviously at the moment it makes no difference if it crashes the app, but if we ever change that it will

PR #1864, opened 2025-09-04
PIR: New email confirmation handling 

quanganhdo:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211009642249869
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1211046211583710
CC:

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/OperationPreferredDateCalculator.swift
```diff
@@ -83,14 +83,16 @@ struct OperationPreferredDateCalculator {
             return date.now.addingTimeInterval(calculateNextRunDateOnError(schedulingConfig: schedulingConfig, historyEvents: historyEvents))
         case .optOutStarted, .scanStarted, .noMatchFound:
             return currentPreferredRunDate
-        case .optOutConfirmed, .matchRemovedByUser, .optOutSubmittedAndAwaitingEmailConfirmation:
```
quanganhdo, line 86: This was done down stack as a placeholder.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/EmailConfirmationJob.swift
```diff
@@ -0,0 +1,327 @@
...
+    public init(jobData: OptOutEmailConfirmationJobData,
+                showWebView: Bool,
+                errorDelegate: EmailConfirmationErrorDelegate?,
+                jobDependencies: EmailConfirmationJobDependencyProviding) {
+        self.jobData = jobData
+        self.showWebView = showWebView
+        self.errorDelegate = errorDelegate
+        self.jobDependencies = jobDependencies
+        super.init()
+    }
```
quanganhdo, line 58: Followed-up PR (https://github.com/duckduckgo/apple-browsers/pull/1877) includes the additions of `webRunnerForTesting`/`webViewHandlerForTesting` to enable mocking.

This PR is already big so I left it off.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/EmailConfirmationJob.swift
```diff
@@ -0,0 +1,327 @@
...
+    ) async throws {
+        guard let optOutStep = broker.steps.first(where: { $0.type == .optOut }) else {
+            throw DataBrokerProtectionError.noOptOutStep
+        }
+
+        guard let profileQuery = try? jobDependencies.database.fetchProfileQuery(with: jobData.profileQueryId) else {
+            throw DataBrokerProtectionError.dataNotInDatabase
+        }
+
+        let actionsHandler = ActionsHandler.forEmailConfirmationContinuation(optOutStep, confirmationURL: linkURL)
```
quanganhdo, line 160: Followed-up PR (https://github.com/duckduckgo/apple-browsers/pull/1877) uses `Variant` enum instead of setting a custom `actionsHandler` here.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/OptOutServices/EmailConfirmationDataService.swift
```diff
@@ -19,12 +19,28 @@
...
 import Common
 import Algorithms
+import os.log
+
+public protocol EmailConfirmationDataChecking {
+    var emailConfirmationDataService: EmailConfirmationDataServiceProvider? { get }
+    func checkForEmailConfirmationData() async
+}
+
+extension EmailConfirmationDataChecking {
```
THISISDINOSAUR: I think I need selling on this, what's the purpose of squashing the error like this?

iOS/DuckDuckGo/AppServices/DBPService.swift
```diff
@@ -25,10 +25,13 @@ import BrowserServicesKit
 import PixelKit
 import Networking
 
-final class DBPService: NSObject {
-
+final class DBPService: NSObject, EmailConfirmationDataChecking {
     private let dbpIOSManager: DataBrokerProtectionIOSManager?
 
+    public var emailConfirmationDataService: EmailConfirmationDataServiceProvider? {
```
THISISDINOSAUR: What's this for?

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileOptOutSubJob.swift
```diff
@@ -115,26 +115,40 @@ struct BrokerProfileOptOutSubJob {
...
-                brokerId: brokerId,
-                profileQueryId: profileQueryId,
-                extractedProfileId: extractedProfileId
-            )
+            if dependencies.featureFlagger.isEmailConfirmationDecouplingFeatureOn,
+               brokerProfileQueryData.dataBroker.requiresEmailConfirmationDuringOptOut() {
+                // Halt the opt-out process
+                // The EmailConfirmationJob will handle obtaining and clicking the confirmation link,
+                // then resume the opt-out from this point
+                try dependencies.database.add(.init(
```
THISISDINOSAUR, line 123: Am I right in thinking the fact that this only saves a history event at this point and the email confirmation object is saved earlier means we'll try and fetch links for opt outs that didn't progress this far?

Also want to double check we won't try and do opt outs straight away again after this?

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/EmailConfirmationJob.swift
```diff
@@ -0,0 +1,327 @@
...
+
+import Foundation
+import Common
+import os.log
+
+public protocol EmailConfirmationErrorDelegate: AnyObject {
+    func emailConfirmationOperationDidError(_ error: Error, withBrokerName brokerName: String?, version: String?)
+}
+
+public class EmailConfirmationJob: Operation, @unchecked Sendable {
```
THISISDINOSAUR, line 27: This is really nicely laid out

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/EmailConfirmationJob.swift
```diff
@@ -0,0 +1,327 @@
...
+
+    private func runJob() async {
+        guard let emailConfirmationLink = jobData.emailConfirmationLink,
+              let linkURL = URL(string: emailConfirmationLink) else {
+            await handleError(EmailError.invalidEmailLink)
+            return
+        }
+
+        // Fetch the broker data
+        guard let broker = try? jobDependencies.database.fetchBroker(with: jobData.brokerId) else {
```
THISISDINOSAUR, line 100: Nice, much better than what the other subjobs are doing

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/EmailConfirmationJob.swift
```diff
@@ -0,0 +1,327 @@
...
+
+        let stageDurationCalculator = DataBrokerProtectionStageDurationCalculator(
+            dataBroker: broker.url,
+            dataBrokerVersion: broker.version,
+            handler: jobDependencies.pixelHandler,
+            vpnConnectionState: jobDependencies.vpnBypassService?.connectionStatus ?? "unknown",
+            vpnBypassStatus: jobDependencies.vpnBypassService?.bypassStatus.rawValue ?? "unknown"
+        )
+
+        while attemptCount < Self.maxRetries {
```
THISISDINOSAUR, line 123: I was imagining retries would work differently. If we retry them straight away, if they failed the first time it seems likely the outcome will be the same this time.

I think it would be better if a job consisted of a single attempt, and aren't retried until the queue is populated again

quanganhdo: That retry approach sounds better. I’ll reimplement that on top of this PR and the one up stack (https://github.com/duckduckgo/apple-browsers/pull/1877) ([Asana task](https://app.asana.com/1/137249556945/project/72649045549333/task/1211009642249877?focus=true))

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileOptOutSubJob.swift
quanganhdo: > Am I right in thinking the fact that this only saves a history event at this point and the email confirmation object is saved earlier means we'll try and fetch links for opt outs that didn't progress this far?

We’ll try to fetch links for any opt outs that have requested a generated email from BE & have no confirmation link yet. So yeah even for opt outs that didn’t progress to the step right ahead of `emailConfirmation`. 

We should add an additional check for the history event then?

> Also want to double check we won't try and do opt outs straight away again after this?

Correct. We won’t try and do opts out until the next time the scheduler triggers.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/OptOutServices/EmailConfirmationDataService.swift
quanganhdo: Placeholder, I’ll add relevant pixels too.

iOS/DuckDuckGo/AppServices/DBPService.swift
quanganhdo: Nvm, this has been removed as `EmailConfirmationDataChecking` has been removed in favor of `EmailConfirmationDataDelegate`

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileOptOutSubJob.swift
THISISDINOSAUR: I'm trying to figure out if it could come back to bite us at all. I can't think of any, although I would be slightly more comfortable if we don't leave this gap.

I'm also wondering if we should do something in case of an error, e.g. delete the local email db object, and even delete the email on the backend. I guess the latter is fine since if the opt out doesn't go through there's no privacy risk?

"Correct. We won’t try and do opts out until the next time the scheduler triggers."
So what I'm wondering is could the following flow happen:
- Opt out runs successfully 
- Scheduler fires again
- We try to fetch the email link and it doesn't exist yet (or it does and we try and run the confirm step and it fails for whatever reason)
- We run the opt out again and it overwrites that email object in the db 

Basically what I'm asking is do we have logic to stop opt outs from running again until their expected retry cadence? (i.e. some number of days)

quanganhdo: @THISISDINOSAUR 

> I can't think of any, although I would be slightly more comfortable if we don't leave this gap.

Yeah we can have an extra query to ensure only opt-outs with `optOutSubmittedAndAwaitingEmailConfirmation` have their email data requested -> I’ll add this to the PR.

> I'm also wondering if we should do something in case of an error, e.g. delete the local email db object, and even delete the email on the backend. I guess the latter is fine since if the opt out doesn't go through there's no privacy risk?

Email addresses are stored for 7 days on the BE, but it’s not associated with the user in any way so Thomas suggested us not to worry about it -> Nothing to be done here. 

> Basically what I'm asking is do we have logic to stop opt outs from running again until their expected retry cadence? (i.e. some number of days)

We have 5 days (`hoursUntilNextAttemptForOptOutWithEmailConfirmation`, currently set to the same value as `maintenanceScan`) from `optOutSubmittedAndAwaitingEmailConfirmation` being set to when the opt-out is retried again (and having the email x attemptId combo being overwritten).

> We try to fetch the email link and it doesn't exist yet (or it does and we try and run the confirm step and it fails for whatever reason)

We should have enough scheduler triggers during that 5 days for email confirmation to come and 3 retry attempts to go through.

Earlier we discussed during a TD that stopping the opt-out from running again at all while email confirmation is going on would have the risk of having stuck opt-outs, so I propose we do nothing extra here and rely on just the `hoursUntilNextAttemptForOptOutWithEmailConfirmation` period.

THISISDINOSAUR: I just mean I want to check we're filtering the opt outs currently waiting on an email step when adding opt outs to the queue? I don't remember if I've seen that logic in a PR or not

quanganhdo: Understood, good catch. Based on existing logic:

- When adding jobs to the queue in `scheduled` mode, we have an attached `priorityDate` set to `Date()` and it’ll be used to filter out jobs with `preferredRunDate` in the future -> we are good here
- When adding jobs in `immediate` mode (`profileSave()`), we don’t have a `priorityDate`, so all opt-out jobs not being manually removed (“This isn’t me”) will be included 🤦‍♂️ 

Adding an extra check here for `latest history event != optOutSubmittedAndAwaitingEmailConfirmation`?

```
        if let priorityDate = priorityDate {
            filteredAndSortedJobData = jobsData
                .filteredByNilOrEarlierPreferredRunDateThan(date: priorityDate)
                .sorted(by: areInIncreasingOrder)
        } else {
            filteredAndSortedJobData = jobsData
                .excludingUserRemoved()
        }
```

THISISDINOSAUR: Re the first case, where/when do we update the preferredRunDate of opt outs waiting on an email confirmation?

quanganhdo: [This](https://github.com/duckduckgo/apple-browsers/blob/anh/shared/pir/email-decoupling/email-confirmation-job/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/SubJob/BrokerProfileOptOutSubJob.swift#L89-L99)

```
        // 7. Set up a defer block to report opt-out job completion regardless of its success:
        defer {
            reportOptOutJobCompletion(
                brokerProfileQueryData: brokerProfileQueryData,
                extractedProfileId: extractedProfileId,
                brokerId: brokerId,
                profileQueryId: profileQueryId,
                database: dependencies.database,
                notificationCenter: dependencies.notificationCenter
            )
        }
```

updates the `preferredRunDate` based on scheduling config.

PR #1867, opened 2025-09-05
Add more metrics and diagnostics to better understand bookmarks lost structure issue

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1211050177427683?focus=true
CC: @brindy 

### Description
1. Added 5 diagnostic parameters to identify root causes (sync status, recent errors, file marker presence)
2. Replaced 1 generic database error pixel with 10 specific error pixels for precise failure identification
3. Added pixel to track sync attempts to delete root folders
4. Added pixel to correlate structure loss with app crashes
5. Refactored legacy fatalError calls to structured error throwing with detailed tracking
6. Added file marker system to possibly distinguish backup restores from actual db corruption

### Testing Steps
1. Since there is no straightforward way to confirm that all errors are handled, the recommended approach is to review each error type originating from the Terminating class related to bookmarks. Attempt to reproduce conditions that trigger these errors and verify that they are properly propagated back to the Terminating class.
2. After triggering an error, recompile the app and simulate the structure_lost case. Verify that the recent-bookmark-error-* parameters are populated (non-nil).
3. Enable sync and confirm that the structure_lost pixel includes the parameter sync-enabled with the value true.
4. Use the Crash App option in the Debug menu, then recompile the app and again simulate the structure_lost case. Verify that the debugBookmarksStructureLostAfterCrash pixel is sent.

### Impact and Risks

#### What could go wrong?
Some app startup paths may have changed, and we might now be crashing in places we didn’t before.

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/SyncDataProviders/Bookmarks/internal/BookmarkEntity+Syncable.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/SyncDataProviders/Bookmarks/internal/BookmarkEntity+Syncable.swift
@@ -109,8 +109,8 @@ extension BookmarkEntity {
                 decryptedUsing decrypt: (String) throws -> String,
                 metricsEvents: EventMapping<MetricsEvent>? = nil) throws {
         guard !syncable.isDeleted else {
-            if let syncableUUID = syncable.uuid, Constants.favoriteFoldersIDs.union([Constants.rootFolderID]).contains(syncableUUID) {
-                metricsEvents?.fire(.syncAttemptedToDeleteRoot(rootUUID: syncableUUID))
+            if parent == nil {
+                metricsEvents?.fire(.syncAttemptedToDeleteRoot(rootUUID: syncable.uuid ?? ""))
             }
             context.delete(self)
             return
--- a/SharedPackages/BrowserServicesKit/Sources/SyncDataProviders/Bookmarks/internal/BookmarkEntity+Syncable.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/SyncDataProviders/Bookmarks/internal/BookmarkEntity+Syncable.swift
@@ -17,6 +17,7 @@
 //
 
 import Bookmarks
+import Common
 import CoreData
 import DDGSync
 import Foundation
@@ -105,8 +106,12 @@ extension BookmarkEntity {
 
     func update(with syncable: SyncableBookmarkAdapter,
                 in context: NSManagedObjectContext,
-                decryptedUsing decrypt: (String) throws -> String) throws {
+                decryptedUsing decrypt: (String) throws -> String,
+                metricsEvents: EventMapping<MetricsEvent>? = nil) throws {
         guard !syncable.isDeleted else {
+            if let syncableUUID = syncable.uuid, Constants.favoriteFoldersIDs.union([Constants.rootFolderID]).contains(syncableUUID) {
+                metricsEvents?.fire(.syncAttemptedToDeleteRoot(rootUUID: syncableUUID))
+            }
             context.delete(self)
             return
         }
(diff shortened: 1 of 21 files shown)


SharedPackages/BrowserServicesKit/Sources/SyncDataProviders/Bookmarks/internal/BookmarkEntity+Syncable.swift
```diff
@@ -105,8 +106,12 @@ extension BookmarkEntity {
 
     func update(with syncable: SyncableBookmarkAdapter,
                 in context: NSManagedObjectContext,
-                decryptedUsing decrypt: (String) throws -> String) throws {
+                decryptedUsing decrypt: (String) throws -> String,
+                metricsEvents: EventMapping<MetricsEvent>? = nil) throws {
         guard !syncable.isDeleted else {
+            if let syncableUUID = syncable.uuid, Constants.favoriteFoldersIDs.union([Constants.rootFolderID]).contains(syncableUUID) {
```
bwaresiak: For optimization purposes, you can check is self has a parent - as root folders doesn't have one.

iOS/Core/BookmarksStateValidation.swift
```diff
@@ -95,10 +104,65 @@ public class BookmarksStateValidator: BookmarksStateValidation {
...
+    private var hasLaunchedSuccessfullyBefore: Bool {
+        guard let marker = BoolFileMarker(name: .hasSuccessfullyLaunchedBefore) else {
+            return false
         }
+        return marker.isPresent
     }
+
+    private func getRecentBookmarkError() -> (name: String, domain: String, code: Int)? {
+        let bookmarkErrorKey = "BookmarksValidator.lastBookmarkError"
+        let maxAgeHours: TimeInterval = 24 // 1 day
```
bwaresiak: I'd actually keep this in seconds no hours - as you use the TimeInterval type explicitly that is supposed to hold time periods in seconds.

iOS/DuckDuckGo/AppServices/CrashCollectionService.swift
```diff
@@ -60,6 +63,10 @@ final class CrashCollectionService {
                 } else {
                     DailyPixel.fireDaily(.dbCrashDetectedDaily(appIdentifier: appIdentifier))
                 }
+                
+                if isBookmarksStructureMissing {
+                    DailyPixel.fireDailyAndCount(pixel: .debugBookmarksStructureLostAfterCrash)
```
bwaresiak, line 69: Technicality: I think this should be sent only for main app crashes, and not for the extension ones

iOS/DuckDuckGo/AppConfiguration/PersistentStoresConfiguration.swift
```diff
@@ -58,11 +58,16 @@ final class PersistentStoresConfiguration {
         }
     }
 
-    private func loadAndMigrateBookmarksDatabase() throws {
-        let result = BookmarksDatabaseSetup().loadStoreAndMigrate(bookmarksDatabase: bookmarksDatabase)
-        if case .failure(let error) = result {
+    private func loadAndMigrateBookmarksDatabase() throws -> Bool {
+        var isMissingStructure: Bool = false
```
bwaresiak: I think more proper name would be something long the lines of "didRepairBookmarksStructure" here and in other places

iOS/Core/LegacyBookmarksStoreMigration.swift
```diff
@@ -185,12 +191,16 @@ public class LegacyBookmarksStoreMigration {
...
+            } catch {
+                throw BookmarksDatabaseError.couldNotPrepareBookmarksDBStructure(error)
+            }
+            
             do {
                 try destination.save(onErrorFire: .bookmarksMigrationCouldNotPrepareDatabaseOnFailedMigration)
             } catch {
-                Thread.sleep(forTimeInterval: 2)
-                fatalError("Could not write to Bookmarks DB")
+                throw BookmarksDatabaseError.couldNotWriteToBookmarksDB(error)
```
bwaresiak: Would that error contain managed object references that could potentially leak outside of context perform block scope?

bwaresiak (approved):

PR #1869, opened 2025-09-05
Fix Popup window titlebar

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1210796710882597?focus=true
- https://app.asana.com/1/137249556945/project/1202406491309510/task/1210544687082897?focus=true

### Description
- Fixes popup title bar issues
- Add popup window to the Windows menu and Dock menu item (inherit from NSWindow instead of NSPanel)

### Testing Steps

1. Open a popup window (Debug -> UI Triggers -> Show Pop up window)
2. Validate resizing, no full screen mode, no title bar context menu;
3. Activate/deactivate different pinned items from the regular window Navigation Bar context menu, validate they don‘t affect the Pop Up window
4. Only Downloads button and Saved login button should appear in the popup title bar when downloaded/password saved
5. Validate the popUp window is selectable in the Window menu and the app Dock context menu

### Impact and Risks

Medium

#### What could go wrong?

Window management breakage

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/MainWindow/MainWindowController.swift
+++ b/macOS/DuckDuckGo/MainWindow/MainWindowController.swift
@@ -54,7 +54,7 @@ final class MainWindowController: NSWindowController {
 
         assert(window == nil || [.unitTests, .integrationTests].contains(AppVersion.runType),
                "Window should not be set in non-test environment")
-        let popUp = mainViewController.tabCollectionViewModel.isPopup
+        let popUp = mainViewController.isInPopUpWindow
         let window = window ?? (popUp
             ? PopUpWindow(frame: frame)
             : MainWindow(frame: frame))
@@ -248,6 +248,11 @@ final class MainWindowController: NSWindowController {
             toolbarView.setAccessibilityEnabled(false)
             toolbarView.setAccessibilityElement(false)
             tabBarViewController.view.setAccessibilityParent(window)
+
+            // macOS 26 Glass Container prevents right clicks
+            if let glassContainer = toolbarView.subviews.first(where: { $0.className == "NSGlassContainerView" }) {
+                glassContainer.removeFromSuperview()
+            }
         }
 
         tabBarViewController.view.frame = newParentView.bounds
@@ -347,7 +352,7 @@ extension MainWindowController: NSWindowDelegate {
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) { [weak self] in
             guard let self else { return }
             mainViewController.disableTabPreviews()
-            mainViewController.mainView.isTabBarShown = false
+            mainViewController.mainView.setTabBarShown(false, animated: true)
             mainViewController.mainView.webContainerTopBinding = .navigationBar
             mainViewController.updateBookmarksBarViewVisibility(visible: false)
             moveTabBarView(toTitlebarView: false)
@@ -360,7 +365,7 @@ extension MainWindowController: NSWindowDelegate {
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) { [weak self] in
             guard let self else { return }
             mainViewController.enableTabPreviews()
-            mainViewController.mainView.isTabBarShown = true
+            mainViewController.mainView.setTabBarShown(true, animated: true)
             mainViewController.mainView.webContainerTopBinding = .tabBar
             mainViewController.updateBookmarksBarViewVisibility(visible: mainViewController.shouldShowBookmarksBar)
             window?.titlebarAppearsTransparent = true
@@ -524,22 +529,6 @@ fileprivate extension MainMenu {
             preferencesMenuItem.menu,
             manageBookmarksMenuItem.menu
         ].compactMap { $0 }
-    }
-
-}
-
-fileprivate extension NavigationBarViewController {
-
-    var controlsForUserPrevention: [NSControl?] {
-        return [homeButton,
-                optionsButton,
-                overflowButton,
-                bookmarkListButton,
-                passwordManagementButton,
-                addressBarViewController?.addressBarTextField,
-                addressBarViewController?.passiveTextField,
-                addressBarViewController?.addressBarButtonsViewController?.bookmarkButton
-        ]
     }
 
 }
--- a/macOS/DuckDuckGo/NavigationBar/View/AddressBarViewController.swift
+++ b/macOS/DuckDuckGo/NavigationBar/View/AddressBarViewController.swift
@@ -115,6 +115,14 @@ final class AddressBarViewController: NSViewController {
         }
     }
 
+    var isInPopUpWindow: Bool {
+        guard let navigationBarViewController = parent as? NavigationBarViewController else {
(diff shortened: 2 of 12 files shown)


macOS/DuckDuckGo/NavigationBar/View/NavigationBarViewController.swift
```diff
@@ -40,58 +40,59 @@ final class NavigationBarViewController: NSViewController {
...
+    @IBOutlet private var homeButtonSeparator: NSView!
+    @IBOutlet private var downloadsButton: MouseOverButton!
+    @IBOutlet private var shareButton: MouseOverButton!
+    @IBOutlet private var networkProtectionButton: NetworkProtectionButton!
+    @IBOutlet private var navigationButtons: NSStackView!
+    @IBOutlet private var addressBarContainer: NSView!
+    @IBOutlet private var daxLogo: NSImageView!
+    @IBOutlet private var addressBarStack: NSStackView!
+
+    @IBOutlet private var menuButtons: NSStackView!
```
rachelmcr: This could be `private(set)`, so it’s accessible from `WebExtensionManager` (currently causes a build error).

rachelmcr (changes requested):
There’s currently a build error related to the `menuButtons` access.

When I tested the build with that fixed, I ran into a couple issues with the elements that should be able to appear in a popup window:

1. After entering or registering login details (e.g. at https://fill.dev/form/login-simple or https://fill.dev/form/registration-email) the passwords shortcut flashes on screen and any toast/popover (offering or saying that login details are saved) either doesn’t appear or is only shown in a flash before disappearing.

https://github.com/user-attachments/assets/be6e3b7d-de24-4067-ac80-fb18cd3ccbd4



2. Another issue occurs when the download shortcut is/was displayed (after downloading a file) and then you e.g. enter login details to trigger the passwords shortcut to appear. If you then switch back to the main (non-popup) window, interacting with the passwords popover can crash the app, hitting the assertion failure in `NavigationBarViewController.isInPopUpWindow` even though you’re no longer interacting with the popup window.

I couldn’t find consistent steps to reliably reproduce the crash, but I repeatedly tried these steps and the app would crash although sometimes at a slightly different point:
* Open a popup window.
* Go to https://fill.dev/form/login-simple or https://fill.dev/form/registration-email and enter login details.
* Go back to the main window (with or without closing the popup window) and open the passwords shortcut.
* At some point after this while interacting with the passwords in the main window (clicking the passwords shortcut again to close it, or interacting with the popover e.g. to edit/delete a password — triggering `updatePasswordManagementButton()`) the app crashes on the assertion.

macOS/DuckDuckGo/MainWindow/MainWindowController.swift
```diff
@@ -248,6 +248,11 @@ final class MainWindowController: NSWindowController {
             toolbarView.setAccessibilityEnabled(false)
             toolbarView.setAccessibilityElement(false)
             tabBarViewController.view.setAccessibilityParent(window)
+
+            // macOS 26 Glass Container prevents right clicks
+            if let glassContainer = toolbarView.subviews.first(where: { $0.className == "NSGlassContainerView" }) {
+                glassContainer.removeFromSuperview()
+            }
```
rachelmcr, line 264: This shouldn’t be needed. It worked as a workaround in earlier macOS 26 betas, then stopped working as a workaround, and in the latest beta it isn’t needed (right clicks work without any workaround).

mallexxx: Yes, it‘s fixed on the prod build on current beta, but I‘m testing on macOS 26 beta 9 (latest) built with Xcode 26 beta 7 (with the new glass ui corners applied) and the menu is not working unless the glass view is removed

rachelmcr: Ah, you’re right — I thought this fix wasn’t needed in macOS 26 beta 9 but I just checked and realized that when I last tested with Xcode 26 beta 7 I still had this fix in place from before (it just wasn’t working in some of the intermediate betas). 👍

mallexxx:
@rachelmcr, thank you for your thorough testing;
the assertion failure was caused by the `NavigationBarViewController` leaking because of the Web Extensions manager
I‘ve fixed the blinking buttons and missing popover issues, as well as allowing to copy the title bar text from the PopUp window and fixing the regression making the address bar non-selectable with mouse
The PR ready for another pass now

macOS/DuckDuckGo/NavigationBar/View/AddressBarViewController.swift
```diff
@@ -217,17 +225,14 @@ final class AddressBarViewController: NSViewController {
...
-            return
-        }
-        if window.isPopUpWindow == true {
+        if isInPopUpWindow {
             addressBarTextField.isHidden = true
             inactiveBackgroundView.isHidden = true
             activeBackgroundViewWithSuggestions.isHidden = true
             activeOuterBorderView.isHidden = true
             activeBackgroundView.isHidden = true
+            passiveTextField.isSelectable = true
```
rachelmcr: This causes the mouse icon to change to a text cursor when hovering over the popup window title, but the text can’t actually be selected (I can only click and drag the window). Is it needed for any other reason?

rachelmcr (changes requested):
Looks good! I just see a few minor issues:

1. When I update a saved password, a toast message appears but clicking on “View” just dismisses the toast and passwords button. It should open the password item to view it.

https://github.com/user-attachments/assets/f16ba0cd-0a3c-450f-932d-34c642ef2f05


2. For some reason when a download is happening, the progress view is not aligned correctly with the downloads button.

Screenshot 2025-09-08 at 12 30 30

3. In the popup window, the mouse becomes a text cursor when hovering over the window title, but the text isn’t selectable (see comment).

macOS/DuckDuckGo/NavigationBar/View/AddressBarViewController.swift
mallexxx: since it uncovered some extra context menu handling, I moved this to a separate pr: https://github.com/duckduckgo/apple-browsers/pull/1911

macOS/DuckDuckGo/NavigationBar/View/AddressBarViewController.swift
```diff
@@ -217,17 +225,13 @@ final class AddressBarViewController: NSViewController {
     }
 
     override func viewWillAppear() {
-        guard let window = view.window else {
-            assert([.unitTests, .integrationTests].contains(AppVersion.runType),
-                   "AddressBarViewController.viewWillAppear: view.window is nil")
-            return
-        }
```
rachelmcr, line 224: It looks like removing this guard statement is causing some integration tests to fail because of a later assertion about the window being non-nil. Is there a reason we don’t want this check here? If so, we’ll need to update the later assertions (e.g. in `subscribeToAppearanceChanges()` and `subscribeToFirstResponder()`) so the integration tests can run.

PR #1872, opened 2025-09-05
PixelKit new approach to errors

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1211256910531201
CC: @diegoreymendez @bwaresiak @ayoy 

### Description

This PR modernizes PixelKit's error handling approach by consolidating the dual error system into a unified approach. The changes replace PixelKitEventV2 with PixelKitEvent and move error extraction to the protocol level using reflection.

Key Changes

- Removes PixelKitEventV2 protocol and consolidates into PixelKitEvent
- Adds reflection-based error extraction to PixelKitEvent protocol
- Updates all pixel events to use PixelKitEvent instead of PixelKitEventV2

Overall changes to PixelKit
- Passing an error as parameter to the `fire(...)` function is not allowed anymore
- The correct way to add an error (preferably DDGError) to a pixel is now to add a parameter to the enum:
```
enum SomePixelEvent: PixelKitEvent {
	case eventWithDDGError(DDGError)
	case eventWithError(Error)
}
```

### Testing Steps

1. PixelKit behaviour shold be unchanged
2. If a Pixel containing an error is not covered by unit tests, we should test it manually (which one??)

### Impact and Risks

We could break some grafana board

---
###### Internal references:


SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKitEvent.swift
```diff
@@ -17,10 +17,34 @@
...
+            // Check if the associated value is directly an Error
+            if let error = associated as? NSError {
+                return error
+            }
+
+            // If it's a tuple (multiple associated values), check each one
+            let associatedMirror = Mirror(reflecting: associated)
+            for child in associatedMirror.children {
+                if let error = child.value as? NSError {
+                    return error
```
samsymons: Do you think there's a possibility that someone using this API will have multiple `Error` associated values, and get tripped up by this code only using the first one? This wasn't an issue with the dedicated `var error` getter since it forced the developer to be explicit about it, but now that it's implicit it might get overlooked.

I was thinking we could benefit from an `assertionFailure` (or maybe just an error log?) if this sees multiple NSError instances attached to the enum case.

federicocappelli: You are right, the previous API was clear, but this isn't.
Assert added.

federicocappelli:
@samsymons  I think I found a solution that satisfies everyone. I re-added the `var error: NSError? { get }` to the PixelKitEvent protocol but it's automatically implemented by the extension, WDYT?

samsymons (approved):
Overall looks good to me from testing existing error types to confirm they still fire error parameters as expected, and reviewing the diff to check that no error types were affected otherwise.

Providing the reflection approach for free while allowing `error` to be overridden if needed makes sense too, and thanks for adding the assertion to catch enum cases that use multiple associated errors! If anyone does want to use multiple errors legitimately, they can override `error` directly, so it seems fairly user friendly.

PR #1873, opened 2025-09-05
iOS: Add SyncEnabled matching attribute for RMF

amddg44:
Task/Issue URL:
Tech Design URL:
CC:

### Testing Steps

1.
2.

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/RemoteMessagingConfigMatcherProvider.swift
+++ b/iOS/DuckDuckGo/RemoteMessagingConfigMatcherProvider.swift
@@ -26,6 +26,7 @@ import Bookmarks
 import RemoteMessaging
 import VPN
 import Subscription
+import DDGSync
 
 extension DefaultVPNActivationDateStore: VPNActivationDateProviding {}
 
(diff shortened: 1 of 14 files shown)


iOS/DuckDuckGo/RemoteMessagingConfigMatcherProvider.swift
```diff
@@ -87,7 +91,10 @@ final class RemoteMessagingConfigMatcherProvider: RemoteMessagingConfigMatcherPr
         var isDuckPlayerEnabled: Bool {
             appSettings.duckPlayerMode != .disabled
         }
-        
+        var isSyncEnabled: Bool {
+            syncService.authState != .inactive
```
samsymons, line 95: I was curious whether this should be an outright `== .active` check in case we run into timing issues where RMF thinks a user is using sync when really the service hasn't been set up yet, but we use this pattern in plenty of other places and the startup lifecycle code makes it impossible, so overall it looks good.

samsymons (approved):
LGTM, the attribute handling looks correct and testing it with both sync active and inactive states works as expected with the provided JSONBlob file - thanks!

PR #1879, opened 2025-09-08
New AddressBar force-choice screen

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1211288844785677?focus=true

### Description
Force-choice screen for the new address bar

### Testing Steps
1. See criteria of when to show/not show [here](https://app.asana.com/1/137249556945/project/1142021229838617/task/1211152753855410?focus=true)
2. Easiest way to test is going into the `NewAddressBarPickerDisplayValidator` and manually changing the checks to see if it behaves as expected.

### Impact and Risks

#### What could go wrong?
* Show the UI at the wrong moment
* Do not respect the checks to not show it again
* Do not respect the options chosen by the user


iOS/DuckDuckGo/AppLifecycle/AppStates/Foreground.swift
```diff
@@ -121,7 +122,7 @@ struct Foreground: ForegroundHandling {
         services.dbpService.resume()
         services.inactivityNotificationSchedulerService.resume()
         services.widePixelService.resume()
-
+        services.launchSourceManager.handleAppAction(launchAction)
```
dus7: My only concern is the placement of this call.

It's breaking the unwritten convention of a nice resume call list in this function. I'd probably move the `LaunchSourceManager` to be a dependency of `UIInteractionManager`, which (citing the code doc) _orchestrates the following operations: (...) 3. Handles immediate launch actions (if any) (...)_. 
With this the source storing can be done in a way that makes sure the source is available before the UI is being shown. 

cc @jaceklyp, any thoughts? Maybe I'm overthinking this?

iOS/DuckDuckGo/AppLifecycle/AppDependencies.swift
```diff
@@ -46,5 +46,6 @@ struct AppServices {
     let systemSettingsPiPTutorialService: SystemSettingsPiPTutorialService
     let inactivityNotificationSchedulerService: InactivityNotificationSchedulerService
     let widePixelService: WidePixelService
+    let launchSourceManager: LaunchSourceManaging
```
jaceklyp: these are AppServices, it was supposed to be added to AppDependencies { }

iOS/DuckDuckGo/AppLifecycle/AppStates/Foreground.swift
jaceklyp: tbh I wouldn't complicate this; `UIInteractionManager` has slightly different function;
instead, at the beginning of `Foreground` right after `configureAppearance()` I'd call:
`appDependencies.launchSourceManager.handleAppAction(launchAction)` so at least there's a vertical (visual) separation of concerns, let services do their thing and handle your launch "separately"

iOS/DuckDuckGo/AppLifecycle/AppDependencies.swift
Bunn: Done

iOS/DuckDuckGo/AppLifecycle/AppStates/Foreground.swift
Bunn: `appDependencies.launchSourceManager.handleAppAction(launchAction)` done :)

PR #1898, opened 2025-09-10
Add a workaround for invalid keyboard frame on iOS 26

dus7:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1211140989378405?focus=true
Tech Design URL:
CC:

### Description

Adds a layout reset for iOS 26 and later for invalid/missing keyboard frame updates when keyboard is shown and hidden rapidly.

### Testing Steps
On both iOS 18 and iOS 26 (using simulator is fine)
1. Set address bar position to bottom
2. Open SERP with any query
3. Tap on SERP Region filter and close the dialog.
4. OmniBar and keyboard will attempt to show and hide quickly.
5. Verify OmniBar position updates properly.
6. Verify find in page layout is not broken.
7. Open a few other tabs and browse to any URL and do a keyboard smoke check.

### Impact and Risks
Low: Minor visual changes, small bug fixes, improvement to existing features. Fix targets iOS 26 only.

#### What could go wrong?
Because of additional keyboard frame adjustments there may be some unexpected side effects, but there's nothing concrete coming to my mind. I treat that as very low risk since the adjustment is after the keyboard closes anyway. 

### Notes to Reviewer

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -680,6 +680,14 @@ class MainViewController: UIViewController {
     private func keyboardDidHide() {
         keyboardShowing = false
         didSendGestureDismissPixel = false
+
+        if #available(iOS 26, *) {
+            // Make sure the UI adjusts properly.
+            // Fix for a weird behavior on iOS 26, firing `keyboardWillChangeFrame` event
+            // with the same frame when keyboard is shown and hidded rapidly.
+            // https://app.asana.com/1/137249556945/project/414709148257752/task/1211140989378405
+            adjustUI(withKeyboardFrame: .zero)
+        }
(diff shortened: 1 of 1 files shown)


aataraxiaa (commented):
Code changes look good, but I see some issues on iOS 26 sim as follows:

https://github.com/user-attachments/assets/e91e15f9-ac98-4b0b-a43c-deb23d9edffb

1. Page seems to zoom in a bit after dismiss region dialog
2. App settings jumps in size when opening

See attached video.

Unsure if these are existing and should be fixed separately. 

<!-- Uploading "Simulator Screen Recording - iPhone 16, iOS 26 - 2025-09-10 at 14.00.31.mov"... —>

PR #1900, opened 2025-09-10
macOS: Add pixels for monitoring success/failure to read or write the encryption key in the keychain

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1211290138405080?focus=true
Tech Design URL:
CC: @diegoreymendez 

### Description

This implements pixels for monitoring when we succeed or fail to read or write the encryption key in the keychain during app startup. The success pixel is used to provide context for failures, and the failures report the underlying OSStatus error. We will use these pixels to monitor for any unexpected failures when we migrate this key to shared storage.

### Testing Steps

Success scenario:
1. Build and run the app.
2. Confirm the success pixel fires on app startup.

Failure scenario:
1. Force `EncryptionKeyStore.readKey` to throw an error, e.g. by adding `throw EncryptionKeyStoreError.readFailed(errSecInvalidItemRef)` to the beginning of the method.
2. Build and run the app.
3. Confirm the expected daily and count failure pixels fire with the expected parameters on app startup.

### Impact and Risks

### Quality Considerations

* Needs privacy triage for the new pixels (will request this before merging)

### Notes to Reviewer

Sample failure pixels:

```
👾[Daily and Count-Fired] m_mac_encryption_keystore_read_key_failed_daily ["appVersion": "1.156.0", "d": "DuckDuckGo_Privacy_Browser.EncryptionKeyStoreError", "e": "1", "pixelSource": "browser-dmg", "ud": "NSOSStatusErrorDomain", "ue": "-25243”]

👾[Daily and Count-Fired] m_mac_encryption_keystore_read_key_failed_count ["appVersion": "1.156.0", "d": "DuckDuckGo_Privacy_Browser.EncryptionKeyStoreError", "e": "1", "pixelSource": "browser-dmg", "ud": "NSOSStatusErrorDomain", "ue": "-25243”]
```

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -306,13 +306,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         } catch {
             Logger.general.error("App Encryption Key could not be read: \(error.localizedDescription)")
             fileStore = EncryptedFileStore()
-            switch error as? EncryptionKeyStoreError {
-            case .readFailed:
-                PixelKit.fire(DebugEvent(GeneralPixel.encryptionKeystoreReadKeyFailed, error: error), frequency: .dailyAndCount)
-            case .storageFailed:
-                PixelKit.fire(DebugEvent(GeneralPixel.encryptionKeystoreWriteKeyFailed, error: error), frequency: .dailyAndCount)
-            default:
-                break
+            if let pixelEvent = (error as? EncryptionKeyStoreError)?.debugPixel {
+                PixelKit.fire(pixelEvent, frequency: .dailyAndCount)
             }
         }
 
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -285,6 +285,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             didCrashDuringCrashHandlersSetUp.wrappedValue = false
         }
 
+        if AppVersion.runType.requiresEnvironment {
+            Self.configurePixelKit()
+        }
+
         do {
             keyValueStore = try KeyValueFileStore(location: URL.sandboxApplicationSupportURL, name: "AppKeyValueStore")
             // perform a dummy read to ensure that KVS is accessible
@@ -298,9 +302,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         do {
             let encryptionKey = AppVersion.runType.requiresEnvironment ? try keyStore.readKey() : nil
             fileStore = EncryptedFileStore(encryptionKey: encryptionKey)
+            PixelKit.fire(DebugEvent(GeneralPixel.encryptionKeystoreReadKeySucceeded), frequency: .dailyAndCount)
         } catch {
             Logger.general.error("App Encryption Key could not be read: \(error.localizedDescription)")
             fileStore = EncryptedFileStore()
+            switch error as? EncryptionKeyStoreError {
+            case .readFailed:
+                PixelKit.fire(DebugEvent(GeneralPixel.encryptionKeystoreReadKeyFailed, error: error), frequency: .dailyAndCount)
+            case .storageFailed:
+                PixelKit.fire(DebugEvent(GeneralPixel.encryptionKeystoreWriteKeyFailed, error: error), frequency: .dailyAndCount)
+            default:
+                break
+            }
         }
 
         bookmarkDatabase = BookmarkDatabase()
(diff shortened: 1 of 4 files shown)


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -285,6 +285,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             didCrashDuringCrashHandlersSetUp.wrappedValue = false
         }
 
+        if AppVersion.runType.requiresEnvironment {
+            Self.configurePixelKit()
+        }
```
diegoreymendez, line 290: Just curious: does this mean the pixel in line 297 will start firing for the first time now?

Might be worth flagging to whoever added it if that's the case.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -285,6 +285,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+
         do {
             keyValueStore = try KeyValueFileStore(location: URL.sandboxApplicationSupportURL, name: "AppKeyValueStore")
```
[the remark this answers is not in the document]
diegoreymendez: This is a valid concern, but I've looked at the code and I don't think there's a dependency here... could you double-check and mark resolved @rachelmcr?

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -298,9 +302,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         do {
             let encryptionKey = AppVersion.runType.requiresEnvironment ? try keyStore.readKey() : nil
             fileStore = EncryptedFileStore(encryptionKey: encryptionKey)
+            PixelKit.fire(DebugEvent(GeneralPixel.encryptionKeystoreReadKeySucceeded), frequency: .dailyAndCount)
```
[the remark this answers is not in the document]
diegoreymendez: For all intents and purposes this didn't seem to be considered a failure scenario before, and it should not be considered a failure scenario now.  I mean - even if this was the case - it's completely out of scope for what we're doing here.

If you agree @rachelmcr , could you mark resolved?

macOS/DuckDuckGo/Common/FileSystem/EncryptionKeys/EncryptionKeyStore.swift
```diff
@@ -19,26 +19,70 @@
...
+    var debugPixel: PixelKitEvent? {
+        switch self {
+        case .storageFailed:
+            return DebugEvent(GeneralPixel.encryptionKeystoreWriteKeyFailed, error: self)
+        case .readFailed:
+            return DebugEvent(GeneralPixel.encryptionKeystoreReadKeyFailed, error: self)
+        default:
+            return nil
         }
     }
```
diegoreymendez, line 87: This is ok, but I'd like to make a suggestion.  I don't think this should be blocking you though so feel free to take or leave - don't let it be a big distraction.

Have you considered implementing this in a way that the pixel definition is scoped and contained by this component?

I mean specifically using something like this instead of `DebugPixel`:

```swift
enum EncryptionKeyStorePixel: PixelKitEvent {
    case encryptionKeystoreReadKeySucceeded
    case encryptionKeystoreWriteKeyFailed(error: Error)
    case encryptionKeystoreReadKeyFailed(error: Error)
    
    var name: String {
        switch self {
            case .encryptionKeystoreReadKeySucceeded:
                return "encryption_keystore_read_key_succeeded"
            case .encryptionKeystoreWriteKeyFailed:
                return "encryption_keystore_read_key_succeeded"
            case .encryptionKeystoreReadKeyFailed:
                return "encryption_keystore_write_key_failed"
        }
    }

    var parameters: [String: String] {
        nil
    }

    var error: NSError? {
        switch self {
            case .encryptionKeystoreWriteKeyFailed(let error),
                     .encryptionKeystoreReadKeyFailed(let error):
                return error
            default:
                return nil
        }
    }
}
```

macOS/DuckDuckGo/Application/AppDelegate.swift
rachelmcr: Yes, I was planning to bring this up in asana once this change was merged. 👍

macOS/DuckDuckGo/Application/AppDelegate.swift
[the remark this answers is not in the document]
rachelmcr: Ah, yes, I did check this before making the change and I didn’t see any dependencies.

macOS/DuckDuckGo/Common/FileSystem/EncryptionKeys/EncryptionKeyStore.swift
rachelmcr: Ah I hadn’t considered scoping the pixel definition here (still getting used to how we define/handle pixels) but I like how contained it is. Thanks for the suggestion!

PR #2048, opened 2025-09-29
Hang Reporting I

frosty:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211150618310230/task/1211260525640633?focus=true
Tech Design URL:
CC:

### Description

This PR extends the existing watchdog functionality for use in hang reporting via a pixel. The actual connection to sending a pixel event will come in a follow-up PR. The changes in this PR cover:

- Extending the debug menu options for hang debugging to add:
    - Toggle 'crash on hang timeout', if required to aid debugging
    - Simulate varying lengths of hang
- The watchdog is now created in the app delegate instead of the view controller, so we only have a single watchdog per app.
- The watchdog now takes a minimum and maximum length of hang to watch for. If the hang exceeds the maximum length, it will be reported as a timeout and we’ll stop monitoring it.
- The watchdog now tracks the length of a hang once the minimum length is exceeded. Once (if) the app becomes responsive again, we’ll report the hang length via the ui_hang pixel (again, this will come in a follow-up PR).

### Testing Steps

1. The watchdog is disabled for debug builds by default, as pausing in the debugger will cause it to kick in. You can however enable it from the debug menu by choosing Debug > Hang Debugging > Toggle Hang Watchdog.
2. With the watchdog enabled, filter your debugger console output to category:hang-detection to just see the output from the watchdog.
3. Use the Debug > Hang debugging > Simulate hang menu to simulate hangs of various lengths. Ensure that you see the hang reported at approximately the correct length, and that a timeout is reported if you trigger a hang of longer than 10 seconds:
Screenshot 2025-09-29 at 14 16 46

4. You could also try simulating a hang using the Debug > UI Triggers > Append Tabs option to append a lot of tabs (e.g. 150) at once.

### Impact and Risks

There should be minor impact, as the watchdog runs quietly in the background pinging the UI thread. Certainly no directly user-facing changes.

#### What could go wrong?

- Not much! The watchdog is also enabled behind a feature flag to minimize its impact
- It can be affected by pausing the app in the debugger, which is why it’s disabled by default for debug builds
- I haven’t seen any negative impact on CPU or memory usage in my testing

### Quality Considerations

I’ve updated the existing tests and added a new one to cover the various state transitions we now have.

—
###### Internal references:

---

> [!NOTE]
> Refactors `Watchdog` to track hang states/durations with async API, integrates a single app-wide watchdog, and expands debug menu to toggle crashing and simulate hangs.
> 
> - **Watchdog (BrowserServicesKit)**:
>   - Refactor to `actor` with hang state machine (`responsive`/`hanging`/`timeout`) and duration tracking.
>   - New config: `minimumHangDuration`, `maximumHangDuration`, `checkInterval`, `crashOnTimeout` (toggle via `setCrashOnTimeout`).
>   - Async API: `start()`/`stop()`, `isRunning` state, detached monitoring/heartbeat tasks, improved logging; optional `killAppFunction`.
>   - Test-only publisher `hangStatePublisher` for state transitions.
> - **Tests**:
>   - Update to async/await and new API; add coverage for normal operation, timeouts, crashOnTimeout behavior, memory, concurrency, and state transition sequence.
> - **macOS app integration**:
>   - Instantiate a single `Watchdog` in `AppDelegate`; start it (non-DEBUG) behind feature flag `hangReporting`.
>   - Remove per-controller watchdog; route menu actions to `AppDelegate.watchdog`.
>   - Debug menu: add "Crash on timeout" toggle and "Simulate hang" submenu with durations; update menu state via `watchdog.isRunning` and `crashOnTimeout`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0d4e119427c36e1b32ae39d37df6ed080cd6034c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
```diff
@@ -92,20 +141,64 @@ public final class Watchdog {
...
+    }
+
+    private func killApp(timeout: TimeInterval) {
         // Log before crashing to help with debugging
         Self.logger.critical("Watchdog is terminating the app due to main thread hang")
 
         // Use fatalError to generate crash report with stack trace`
-        fatalError("Main thread hang detected by Watchdog (timeout: \(timeout)s). This crash is intentional to provide debugging information.")
+        fatalError("Main thread hang detected by Watchdog (timeout: \(maximumHangDuration)s). This crash is intentional to provide debugging information.")
+    }
```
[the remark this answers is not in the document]
frosty: The timeout property is mostly just provided for testing / mocking purposes.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
```diff
@@ -69,16 +111,23 @@ public final class Watchdog {
         monitoringTask?.cancel()
         monitoringTask = nil
 
+        heartbeatUpdateTask?.cancel()
+        heartbeatUpdateTask = nil
+
         Self.logger.info("Watchdog stopped monitoring")
     }
 
     private func startMonitoring() async {
```
diegoreymendez: I know this method isn't new, but could we rename this to `runMonitoringLoop`?  I think it'd be clearer.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
```diff
@@ -16,51 +16,93 @@
...
+    ///   - checkInterval: The interval at which the main thread is checked for hangs.
     @MainActor
-    convenience public init(timeout: TimeInterval = 10.0, checkInterval: TimeInterval = 2.0) {
-        self.init(killAppFunction: Self.killApp(afterTimeout:), timeout: timeout, checkInterval: checkInterval)
-    }
-
-    @MainActor
-    init(killAppFunction: @escaping (TimeInterval) -> Void, timeout: TimeInterval = 10.0, checkInterval: TimeInterval = 2.0) {
-        self.timeout = timeout
+    public init(minimumHangDuration: TimeInterval = 1.0, maximumHangDuration: TimeInterval = 10.0, checkInterval: TimeInterval = 0.25, killAppFunction: ((TimeInterval) -> Void)? = nil) {
```
diegoreymendez: Is checking every 0.25 seconds intentional?  Sounds quite a aggressive schedule considering the minimum hang duration is 1 second.

I realize we might watch to catch short hangs with precision, but my fear is being to aggressive here might introduce other performance issues and eventually bite us back - whereas maybe 0.5 sec or 1sec precision is much less likely to blow up.

If we keep at 0.25 seconds I'd suggest testing this thoroughly in older devices.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
```diff
@@ -92,20 +141,64 @@ public final class Watchdog {
 
             // Check if the heartbeat was actually updated
             let timeSinceLastCheck = await monitor.timeSinceLastHeartbeat()
+            handleHangDetection(timeSinceLastCheck: timeSinceLastCheck)
```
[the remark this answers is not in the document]
diegoreymendez: Cursor has a point.  We're accessing `heartbeatUpdateTask` from both the heartbeat closure (which runs in MainActor) and from the main run-loop (for cancelling it) we're opening ourselves up for race conditions.

Additionally right now we're accessing the heartbeat update task from `start` and `stop` which run on the main thread.

Even if it was true that the current implementation is unlikely to bring us trouble, this is very circumstantial and could change drastically on the next maintenance edit.

IMHO the best approach would be to convert `Watchdog` into an actor, and mostly avoid using the main thread.  The only exception we'd need is for `isRunning`, for which I'd add a `@MainActor` property that can be consulted synchronously.

diegoreymendez (commented):
Works for me - added some comments, the idea is not to block you but I do think we want to be careful due to the nature of these changes.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
frosty: Sure!

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
frosty: It was, just because I thought it would improve the precision of our timings – otherwise our detection would be up to half a second out. But if that’s not a huge issue, I’m totally happy to alter this. I agree that 0.25 does feel like faster than we need, really.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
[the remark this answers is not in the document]
frosty: Thanks, investigating this now!

frosty:
@diegoreymendez I’ve pushed some changes to switch the watchdog to an actor, as you recommended. I’ve also updated the tests accordingly. Everything seems to be working as expected – let me know what you think!

I’ve also increased the check interval to 0.5 seconds.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Watchdog/Watchdog.swift
```diff
@@ -16,69 +16,128 @@
...
+        crashOnTimeout = state
     }
 
     @MainActor
-    init(killAppFunction: @escaping (TimeInterval) -> Void, timeout: TimeInterval = 10.0, checkInterval: TimeInterval = 2.0) {
-        self.timeout = timeout
+    public private(set) var isRunning: Bool = false
+
+    @MainActor
+    private func setIsRunning(_ state: Bool) async {
```
diegoreymendez: Does this need to be async?

frosty: No, thanks for catching that!

PR #2070, opened 2025-10-01
Release pixel cleanup

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211470395820605?focus=true
Tech Design URL:
CC:

### Description

This PR removes the old macOS and iOS release pixels, and replaces them with the new format.

⚠️ The branch that this is targeted add sends a pixel when the build is _created_, not pushed live, but the macOS pixel on this branch is sent when pushed live to users. This should be consolidated.

### Testing Steps

1. Check that the release pixel format is correct

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> Remove deprecated release-annotation pixel action, drop its iOS usage, and switch macOS workflow to new send-pixel with updated pixel name.
> 
> - **CI/Workflows**:
>   - Remove deprecated action `./.github/actions/send-release-annotation-pixel` and its usage in `ios_tag_release_update_asana.yml`.
>   - macOS (`macos_publish_dmg_release.yml`): replace release pixel step with `./.github/actions/send-pixel`, gated on success, using `pixel-name: m_ci_apple_macos_release_version_\${{ steps.verify-tag.outputs.release-version }}`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 799d1f6d45f790d94bf277d5199d17aea0ed907f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.github/workflows/macos_publish_dmg_release.yml
```diff
@@ -286,14 +286,13 @@ jobs:
             echo "FILES_UPLOADED='No files uploaded.'" >> $GITHUB_ENV
           fi
 
-      - name: Send release annotation pixel
-        if: github.event.inputs.release-type != null && github.event.inputs.release-type != 'internal'
+      - name: Send Release Pixel
+        if: success() && github.event.inputs.release-type != null && github.event.inputs.release-type != 'internal'
```
bwaresiak, line 290: What I wonder is if we want to have it for internals as well:
1/ we can have only for public
2/ we can have for all
3/ we can have two pixels, one for internal and one for public releases

I’m sort of leaning towards 3, as it gives us most flexibility around how to use these pixels, but it’s a bt more tricky to send (e.g. we’d have internal, that then gets promoted to public).

samsymons: Yeah good question. If we do want internal pixels (and we may as well), then I agree option 3 is suitable. I think this will be doable by keeping this pixel as-is as the final release, but updating the other workflow file to send the internal pixel - let me take a go at that on Thursday.

PR #2379, opened 2025-10-31
[macOS] Add support to determine re-engagement prompt eligibility

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1211806562397682?focus=true
Tech Design URL:
CC: @alessandroboron 

### Description

The new re-engagement prompt for inactive users will only be shown to users with 7+ days of inactivity. To determine their eligibility, we need client-side storage of the user’s current and previous active day (to locally calculate days of inactivity). This ports the relevant logic from the iOS implementation so the client can calculate their eligibility to see the prompt:

* When the app becomes active, if the inactive user feature flag is enabled (internal only) and it’s the first time the user is active that day, the current and previous active day is stored in local storage.
* Adds a method to calculate the number of inactive days. This will be used to trigger the prompt locally.

Note: The active days are only stored client-side and used to determine eligibility for the local prompt.

This also encapsulates the AppDelegate properties related to the prompts in `DefaultBrowserAndDockPromptService`.

### Testing Steps

**Scenario 1: Inactive days**

1. Set internal user state if needed.
2. Run the app or bring it into the foreground.
3. Open the 'Debug Menu' → 'SAD/ATT Prompts’ and confirm that “Number of inactive days” is 0.
4. Override the date to 8 days in the future.
5. Open the 'Debug Menu' → 'SAD/ATT Prompts’ and confirm that “Number of inactive days” is 7.

**Scenario 2: Existing prompts**

1. Launch the App.
2. Open the 'Debug Menu' → 'SAD/ATT Prompts’ and check the popover show date.
3. Click ‘Override Today’s Date’ and change the date to match that.
4. Focus outside the window and focus again.
5. The SAD/ATT popover is prompted.
6. Click the ’Not Now’ button.
7. Open the 'Debug Menu' → 'SAD/ATT Prompts’ and check banner show date.
8. Click ‘Override Today’s Date’ and change the date to match that.
9. Focus outside the window and focus again.
10. The SAD/ATT banner is prompted.

### Impact and Risks

#### What could go wrong?

Could disrupt the existing prompts:
* Confirmed unit tests pass
* Manually tested existing prompts

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> Add local storage of current and previous active days for SAD/ATD re‑engagement and introduce `DefaultBrowserAndDockPromptService`, wiring it into app lifecycle, debug menu, and UI with new tests.
> 
> - **Default Browser & Dock Prompts**:
>   - **Service encapsulation**: Introduce `DefaultBrowserAndDockPromptService` to own prompt presenter, feature flagger, key-value store, and new user-activity manager; replace AppDelegate wiring and use `service.presenter` in `MainViewController`.
>   - **Active Days storage**: Add `UserActivity` model, `UserActivityManager`, `UserActivityStore`, and `UserActivityProvider` under `DefaultBrowserAndAddToDockPrompts/UserActivity`.
>   - **App lifecycle integration**: Call `defaultBrowserAndDockPromptService.applicationDidBecomeActive()` in `AppDelegate` to store active days when inactive-user flag is enabled.
>   - **Debug menu**: Update `DefaultBrowserAndDockPromptDebugMenu` to use the service, show "Number of inactive days", and reset/store active days via the new store/manager.
> - **Tests**:
>   - Add unit tests for `DefaultBrowserAndDockPromptUserActivityManager` and `DefaultBrowserAndDockPromptUserActivityStore`, plus a mock store.
> - **Project**:
>   - Add new source/test files and `UserActivity` group; update build settings and references.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d68c02fb83b3e9775fed1f2f9cff4b15de055222. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/DefaultBrowserAndAddToDockPrompts/DefaultBrowserAndDockPromptDebugMenu.swift
+++ b/macOS/DuckDuckGo/DefaultBrowserAndAddToDockPrompts/DefaultBrowserAndDockPromptDebugMenu.swift
@@ -30,6 +30,7 @@ final class DefaultBrowserAndDockPromptDebugMenu: NSMenu {
     private let defaultBrowserAndDockPromptFeatureFlagger = NSApp.delegateTyped.defaultBrowserAndDockPromptService.featureFlagger
     private let localStatisticsStore = LocalStatisticsStore()
     private let userActivityManager = NSApp.delegateTyped.defaultBrowserAndDockPromptService.userActivityManager
+    private let userActivityStore = NSApp.delegateTyped.defaultBrowserAndDockPromptService.userActivityManager.store
 
     private static let dateFormatter: DateFormatter = {
         let formatter = DateFormatter()
@@ -72,6 +73,7 @@ final class DefaultBrowserAndDockPromptDebugMenu: NSMenu {
         showDatePickerAlert { [weak self] date in
             guard let self, let date else { return }
             debugStore.simulatedTodayDate = date
+            userActivityManager.recordActivity()
         }
     }
 
@@ -80,6 +82,7 @@ final class DefaultBrowserAndDockPromptDebugMenu: NSMenu {
         store.popoverShownDate = nil
         store.bannerShownDate = nil
         store.isBannerPermanentlyDismissed = false
+        userActivityStore.save(DefaultBrowserAndDockPromptUserActivity(lastActiveDate: Date()))
         updateMenuItemsState()
     }
 
@@ -128,6 +131,7 @@ final class DefaultBrowserAndDockPromptDebugMenu: NSMenu {
         }
 
         func updateInactiveUserModalMenuInfo() {
+            userActivityManager.recordActivity()
             let inactiveDays = userActivityManager.numberOfInactiveDays()
             inactiveDaysMenuItem.title = "Number of inactive days: \(inactiveDays)"
         }


macOS/DuckDuckGo/DefaultBrowserAndAddToDockPrompts/UserActivity/DefaultBrowserAndDockPromptUserActivity.swift
```diff
@@ -0,0 +1,49 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// A value type that represents user activity data for the SAD & ATD prompt.
+///
+/// This struct measures when a user has been active by storing the last active days.
+public struct DefaultBrowserAndDockPromptUserActivity: Equatable, Sendable, Codable {
```
alessandroboron, line 24: @rachelmcr I wonder if we could somehow have the UserActivity manager in SharedPackage and have both platform use it.

That said the behaviour is slightly different because we don’t have the concept of number of active days for macOS so it may not be possible.

macOS/DuckDuckGo/DefaultBrowserAndAddToDockPrompts/DefaultBrowserAndDockPromptDebugMenu.swift
```diff
@@ -124,13 +127,21 @@ final class DefaultBrowserAndDockPromptDebugMenu: NSMenu {
             }
         }
 
+        func updateInactiveUserModalMenuInfo() {
+            let inactiveDays = userActivityManager.numberOfInactiveDays()
```
alessandroboron, line 135: When we tap “Reset Today’s Date and Prompts” should we reset number of inactive days too?

macOS/DuckDuckGo/DefaultBrowserAndAddToDockPrompts/UserActivity/DefaultBrowserAndDockPromptUserActivity.swift
rachelmcr: I did consider sharing this between the platforms, but as you mentioned only iOS uses active days. If both platforms share the UserActivity manager they’ll both have to handle active days throughout the implementation, which doesn’t make sense for macOS. There are enough small differences like that throughout the prompts that I think it’s worth keeping the implementations separate.

macOS/UnitTests/DefaultBrowserAndAddToDockPrompts/DefaultBrowserAndDockPromptUserActivityManagerTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        let secondLastActivityDate = Self.today
+        let lastActivityDate = secondLastActivityDate.advanced(by: .days(15))
+        storeMock.activityToReturn = .init(lastActiveDate: lastActivityDate, secondLastActiveDate: secondLastActivityDate)
+
+        // WHEN
+        let result = sut.numberOfInactiveDays()
+
+        // THEN
+        #expect(result == 15)
+    }
```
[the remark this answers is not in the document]
rachelmcr: This is a legitimate issue. In the method `numberOfInactiveDays()` we are counting the days since the previous activity date, but in that case we’re including one of the active days in the count. (For example, if `lastActivityDate` is today and `secondLastActivityDate` is yesterday, it will return 1 inactive day when I have been active every day.) We should exclude the active days from that count — so in the case of this test, when one active date is 15 days after the previous active date, the inactive days should be 14.

PR #2484, opened 2025-11-12
Disable animations while deep linking into iOS settings

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1211863103319361?focus=true
Tech Design URL:
CC:

### Description

This PR disables the navigation animation that plays when you're deep linking into the iOS settings.

There's a known issues with deep linking that was already present before this PR: if you deep link then navigate back within 1 second of settings being displayed, the UI will immediately push the deep linked view controller back onto the stack. This is a side effect of how `NavigationLink` is being used. I'll try to fix this in a future PR, though it's low priority (especially if we're going to move back to UIKit for settings eventually).

### Testing Steps

1. Run the app, then add the VPN widget to the home screen (you don't need to sign into the subscription to test this)
3. Tap the VPN widget
4. Confirm that the settings modal is displayed with the VPN or Subscription flows already visible, and that no push animation plays
5. Next, open settings to the root view controller, and then do the VPN widget test again - check that the VPN or Subscription flow is visible
6. Finally, smoke test settings to make sure that the non-deep linking path is unaffected.

#### What could go wrong?

Deep linking & settings navigation in general could break

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> Disables push animations when deep linking into Settings, plumbs explicit deepLinkTarget through segues, and removes artificial delay in Appearance subpage navigation.
> 
> - **iOS Settings navigation**:
>   - Disable push animation while deep linking by overriding `SettingsUINavigationController.pushViewController` to check `SettingsHostingController.isDeepLinking`.
>   - Expose `isDeepLinking` via `SettingsHostingController`.
>   - Update segues to pass `deepLinkTarget` into `launchSettings` for targets like `appearance`, `customizeAddressBarButton`, `customizeToolbarButton`, `subscriptionFlow()`, `restoreFlow`, `netP`, `aiChat`, and `privateSearch`.
>   - In `SettingsAppearanceView`, replace delayed dispatch with immediate main-queue navigation to subpages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8ef3175bbc460ef9ced6a4f3a9c1918bff595741. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/MainViewController+Segues.swift
+++ b/iOS/DuckDuckGo/MainViewController+Segues.swift
@@ -137,7 +137,7 @@ extension MainViewController {
             PrivacyDashboardViewController(coder: coder,
                                            privacyInfo: privacyInfo,
                                            entryPoint: entryPoint,
-                                           privacyConfigurationManager: ContentBlocking.shared.privacyConfigurationManager,
+                                           privacyConfigurationManager: self.privacyConfigurationManager,
                                            contentBlockingManager: ContentBlocking.shared.contentBlockingManager,
                                            breakageAdditionalInfo: self.currentTab?.makeBreakageAdditionalInfo())
         }
@@ -340,7 +340,7 @@ extension MainViewController {
                                                             daxDialogsManager: daxDialogsManager,
                                                             dbpIOSPublicInterface: dbpIOSPublicInterface)
 
-        let aiChatSettings = AIChatSettings(privacyConfigurationManager: ContentBlocking.shared.privacyConfigurationManager)
+        let aiChatSettings = AIChatSettings(privacyConfigurationManager: privacyConfigurationManager)
         let serpSettingsProvider = SERPSettingsProvider(aiChatProvider: aiChatSettings,
                                                         featureFlagger: featureFlagger)
 
@@ -361,6 +361,7 @@ extension MainViewController {
                                                   maliciousSiteProtectionPreferencesManager: maliciousSiteProtectionPreferencesManager,
                                                   themeManager: themeManager,
                                                   experimentalAIChatManager: ExperimentalAIChatManager(featureFlagger: featureFlagger),
+                                                  privacyConfigurationManager: privacyConfigurationManager,
                                                   keyValueStore: keyValueStore,
                                                   systemSettingsPiPTutorialManager: systemSettingsPiPTutorialManager,
                                                   runPrerequisitesDelegate: dbpIOSPublicInterface,
--- a/iOS/DuckDuckGo/SettingsAppearanceView.swift
+++ b/iOS/DuckDuckGo/SettingsAppearanceView.swift
@@ -125,10 +125,10 @@ struct SettingsAppearanceView: View {
     @ViewBuilder
     func addressBarButtonSetting() -> some View {
 
-        let destination = AddressBarCustomizationPickerView(isAIChatEnabled: viewModel.isAIChatEnabled, selectedAddressBarButton: viewModel.selectedAddressBarButton)
-            .applySettingsListModifiers(title: "",
-                                        displayMode: .inline,
-                                        viewModel: viewModel)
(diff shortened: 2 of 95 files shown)


iOS/DuckDuckGo/MainViewController+Segues.swift
```diff
@@ -225,25 +225,25 @@ extension MainViewController {
     func segueToDuckDuckGoSubscription() {
         Logger.lifecycle.debug(#function)
         hideAllHighlightsIfNeeded()
-        launchSettings {
+        launchSettings(completion: {
             $0.triggerDeepLinkNavigation(to: .subscriptionFlow())
-        }
+        }, deepLinkTarget: .subscriptionFlow())
```
samsymons, line 230: This looks redundant, but it's to handle two different cases:
1. When the deep link causes settings to be displayed for the first time, the `deepLinkTarget` parameter is used to show the flow
2. When settings is already visible, settings has already been launched so `deepLinkTarget` does nothing - instead, that is where the original `completion` block comes in, as it pushes the deep linked VC onto the stack

It's pretty gross, but it's how the API works and I'd love to avoid meaningful refactoring if possible 🤷

iOS/DuckDuckGo/SettingsAppearanceView.swift
```diff
@@ -46,9 +46,7 @@ struct SettingsAppearanceView: View {
     func navigateToSubPageIfNeeded() {
         deepLinkTarget = viewModel.deepLinkTarget
 
-        // This just needs to be longer than the deep link logic in the View Model which uses a timer 🙄
-        //  otherwise this immediately gets popped.
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.55) {
+        DispatchQueue.main.async {
```
samsymons, line 49: @brindy I was able to remove the delay since we're now passing in the deep link target at the time of calling `launchSettings`, rather than triggering the deep linking in the completion handler, which happens on a delay. This has worked well for me in tests across various OS versions, but do call out any weirdness you see, I'm not fully confident in the change yet.

iOS/DuckDuckGo/MainViewController+Segues.swift
```diff
@@ -454,6 +454,17 @@ class SettingsUINavigationController: UINavigationController {
...
 
+    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        // Settings uses NavigationLink for deep linking, but because we don't use it within a NavigationStack, it talks
+        // to the hosting navigation controller. It offers no control over navigation animation, so this workaround
+        // disables animation any time a view controller is pushed while deep linking is being processed.
+        if let settingsHostingController = self.viewControllers.first as? SettingsHostingController, settingsHostingController.isDeepLinking {
+            super.pushViewController(viewController, animated: false)
+        } else {
+            super.pushViewController(viewController, animated: animated)
+        }
```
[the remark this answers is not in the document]
samsymons: Tbh I prefer the original code here, type casting + nil coalescing + boolean logic in a single line is a bit much.

iOS/DuckDuckGo/SettingsAppearanceView.swift
brindy: Way better, thanks!

brindy (approved):
LGTM!

iOS/DuckDuckGo/SettingsHostingController.swift
```diff
@@ -26,6 +26,10 @@ class SettingsHostingController: UIHostingController<AnyView> {
     var viewModel: SettingsViewModel
     var viewProvider: SettingsLegacyViewProvider
 
+    public var isDeepLinking: Bool {
+        return viewModel.deepLinkTarget != nil
+    }
```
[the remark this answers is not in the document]
samsymons: This is true, but is difficult to resolve without more intense refactoring - we have a project scoped for that.

PR #2564, opened 2025-11-20
macOS: Tentative fix for Install update menu option not working

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206580121312550/task/1212005420435668?focus=true

### Description

We're trying to fix for the install update menu option not working

I'm adding inline notes to explain the rationale - but it basically boils down to identifying two potential points in the code that could be causing this issue.  One is avoided, the other now logs an error.

### Testing Steps

1. Change the build number to a lower number so you'll get an update offered.
2. Enable feature flag `autoUpdateInDEBUG`
3. Restart the app to make sure the flag is enabled right
4. Switch to automatic updates
5. Once the update is offered, go to DEBUG Menu > Updates > Expire current update
6. Click on the sandwich menu option to install the update... it should take you to release notes and re-download

### Impact and Risks

High.  Updates are a tricky area to modify.

#### What could go wrong?

The changes could be missing something.

### Quality Considerations

NA

### Notes to Reviewer

NA

---
###### Internal references:

### DIFF
--- a/macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
+++ b/macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
@@ -150,8 +150,6 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
         didSet {
             if oldValue != areAutomaticUpdatesEnabled {
                 updateWideEvent.areAutomaticUpdatesEnabled = areAutomaticUpdatesEnabled
-                // Cancel with .settingsChanged reason to distinguish from user-initiated
-                // cancellations. The 0.1s delay allows updater reconfiguration to complete.
                 updateWideEvent.cancelFlow(reason: .settingsChanged)
                 reconfigureUpdaterAndForceUpdateCheckAfterSmallDelay()
             }
@@ -357,7 +355,7 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
         updateWideEvent.cancelFlow(reason: .buildExpired)
         userDriver?.cancelAndDismissCurrentUpdate()
 
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) { [weak self] in
             guard let self else {
                 Logger.updates.log("Won't start new update check as self is gone")
                 return
@@ -409,11 +407,6 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
         Task { @MainActor in
             do {
                 let updater = try getOrMakeUpdater()
-
-                guard !updater.sessionInProgress else {
-                    Logger.updates.error("User-initiated update skipped as an update check is already in progress")
-                    return
-                }
 
                 Logger.updates.log("Checking for updates skipping rollout")
                 updater.checkForUpdates()
@@ -546,7 +539,7 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
     /// It may be possible to remove this method now that we're no longer nil-ing the updater. Unfortunately this delay
     /// was never documented so removing this during a bug fix isn't ideal.  We should review this for removal.
     private func reconfigureUpdaterAndForceUpdateCheckAfterSmallDelay() {
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) { [weak self] in
             self?.reconfigureUpdaterAndForceUpdateCheck()
         }
     }


macOS/Configuration/BuildNumber.xcconfig
```diff
@@ -1 +1 @@
-CURRENT_PROJECT_VERSION = 575
+CURRENT_PROJECT_VERSION = 570
```
diegoreymendez, line 1: 🔴🔴🔴🔴🔴🔴🔴🔴🔴🔴🔴🔴🔴 ROLL BACK BEFORE MERGING.  I'm only using this to make a test build.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -145,12 +145,6 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
             if newValue != areAutomaticUpdatesEnabled {
                 updateWideEvent.cancelFlow(reason: .settingsChanged)
                 userDriver?.cancelAndDismissCurrentUpdate()
-
-                if useLegacyAutoRestartLogic {
-                    updater = nil
-                } else {
-                    updater?.resetUpdateCycle()
-                }
```
diegoreymendez, line 153: Additionally instead of setting `updater = nil` I'm adjusting the code so that `configureUpdater()` below will always just recreate the updater.  This helps us prevent doing this in two steps separated by an async delayed call, which can be a good recipe for race conditions (that I suspect may be at play here).

Regarding the removal of `updater?.resetUpdateCycle()`, this was added for the new logic, but it doesn't do what we need it to do - this won't affect current UX as it's disabled.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -159,10 +153,7 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
                 // Cancel with .settingsChanged reason to distinguish from user-initiated
                 // cancellations. The 0.1s delay allows updater reconfiguration to complete.
                 updateWideEvent.cancelFlow(reason: .settingsChanged)
-                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
-                    _ = try? self?.configureUpdater()
-                    self?.checkForUpdateSkippingRollout()
-                }
+                reconfigureUpdaterAndForceUpdateCheckAfterSmallDelay()
```
diegoreymendez, line 154: Just moving the async call to its own method - make sure the logic is unchanged in regards to this.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -352,11 +343,6 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
 
         updateWideEvent.cancelFlow(reason: .buildExpired)
         userDriver?.cancelAndDismissCurrentUpdate()
-        if useLegacyAutoRestartLogic {
-            updater = nil
-        } else {
-            updater?.resetUpdateCycle()
-        }
```
diegoreymendez: Same as above... `configureUpdated` below takes care of recreating the updater.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -455,17 +441,9 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
         // Workaround to reset the updater state
         cachedUpdateResult = nil
 
-        if !useLegacyAutoRestartLogic, let userDriver {
-            userDriver.areAutomaticUpdatesEnabled = areAutomaticUpdatesEnabled
-        } else {
-            userDriver = UpdateUserDriver(internalUserDecider: internalUserDecider,
+        let userDriver = UpdateUserDriver(internalUserDecider: internalUserDecider,
                                           areAutomaticUpdatesEnabled: areAutomaticUpdatesEnabled)
```
diegoreymendez, line 459: Simplifying - we just recreate the user driver and pass the new autoupdate value.

The reason I'm doing this change is because there's no gain in trying to keep the old `userDriver` when we're resetting the updater.  This also allows me to remove one of the conditions that could result in calling `return nil` below.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -455,17 +441,9 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
...
-        }
-
-        guard let userDriver,
-              updater == nil else {
```
diegoreymendez, line 466: We don't care if there's already an updater.  We just re-create it.

This condition could result in keeping an updater pointing at the wrong user driver.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -533,15 +511,24 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
 
         updateWideEvent.cancelFlow(reason: .newCheckStarted)
         userDriver.cancelAndDismissCurrentUpdate()
-        if useLegacyAutoRestartLogic {
-            updater = nil
-        } else {
-            updater?.resetUpdateCycle()
-        }
 
+        reconfigureUpdaterAndForceUpdateCheckAfterSmallDelay()
```
diegoreymendez, line 536: GitHub did an awful job with the diff here.. .but we're just calling the new method instead of keeping the old async call.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdaterAvailabilityChecker.swift
```diff
@@ -1,35 +0,0 @@
-//
-//  SparkleUpdaterAvailabilityChecker.swift
```
diegoreymendez, line 2: Trying to simplify a bit, and this class is truly not necessary for a couple of reasons:
- `SPUUpdater` already implements `UpdaterAvailabilityChecking` in an extension.
- The tests for this class were offering zero value.

macOS/UnitTests/Updates/SparkleUpdaterAvailabilityCheckerTests.swift
```diff
@@ -1,111 +0,0 @@
...
-//  limitations under the License.
-//
-
-#if SPARKLE
-
-import XCTest
-import Sparkle
-@testable import DuckDuckGo_Privacy_Browser
-
-final class SparkleUpdaterAvailabilityCheckerTests: XCTestCase {
```
diegoreymendez, line 25: Zero value in these tests.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -243,7 +234,11 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
         // Clean up abandoned flows from previous sessions before starting any new checks
         self.updateWideEvent.cleanupAbandonedFlows()
 
-        _ = try? configureUpdater()
+        do {
+            _ = try makeUpdater()
+        } catch {
+            Logger.updates.error("Could not configure Sparkle Updater on initialization: \(error, privacy: .public)")
+        }
```
diegoreymendez, line 239: Just logging, the method is renamed and changed slightly.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -325,18 +318,28 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
         Task { @MainActor in
             // Handle expired builds first (critical path)
             guard !discardCurrentUpdateIfExpiredAndCheckAgain(skipRollout: false) else {
+                Logger.updates.log("Update discarded due to expired build")
                 return
             }
 
-            guard let updater, !updater.sessionInProgress else { return }
```
diegoreymendez, line 331: Similar logic, but not instead of bailing out if the updater is nil we:
1. Try to make an updater.
2. Logs an error if we ultimately have no updater to use and need to bail out.

PR #2568, opened 2025-11-20
[iOS + MacOS] Improvements on VPNConnectionWideEvent

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1212003395293972?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1211597559231999?focus=true

### Description
A few updates to the existing VPN wide-pixel events:
- Introduce a new `status_reason = “retried"` to distinguish:
  - partial_data caused by users abandoning the previous flow and restarting, vs.
  - partial_data caused by missing start/close interval.
- Add `browser-start` step errors for macOS.
- Add `onboarding_status` to provide more granular visibility on MacOS, especially for failure and unknown statuses.
- Explicit `is_setup` value that includes unknown when this information is unknown rather than defaulting to false
- Add a debug environment patch to `NetworkProtectionTunnelController.swift` to avoid data pollution in non-sandbox environment

### Testing Steps

1. Build and run the app. Verify VPN continues to work properly

#### Manual Testing
**iOS**
```
👾[Daily-Fired] m_ios_wide_vpn_connection_debug_daily ["app.form_factor": "phone", "app.name": "DuckDuckGo", "app.version": "7.196.0", "appVersion": "7.196.0", "context.name": "funnel_appsettings_ios", "feature.data.ext.controller_start_latency_ms": "63", "feature.data.ext.extension_type": "app", "feature.data.ext.is_setup": "no", "feature.data.ext.latency_ms": "67", "feature.data.ext.oauth_latency_ms": "1", "feature.data.ext.startup_method": "manual_by_main_app", "feature.data.ext.tunnel_start_latency_ms": "1", "feature.name": "vpn-connection", "feature.status": "SUCCESS", "global.platform": "iOS", "global.sample_rate": "1.0", "global.type": "app”]
```

**MacOS**
```
👾[Standard-Fired] m_mac_wide_vpn_connection_debug ["app.name": "DuckDuckGo VPN", "app.version": "1.166.0", "appVersion": "1.166.0", "context.name": "funnel_appsettings_macos", "feature.data.ext.browser_start_latency_ms": "32", "feature.data.ext.controller_start_latency_ms": "1709", "feature.data.ext.extension_type": "system", "feature.data.ext.is_setup": "yes", "feature.data.ext.latency_ms": "3624", "feature.data.ext.oauth_latency_ms": "251", "feature.data.ext.onboarding_status": "completed", "feature.data.ext.startup_method": "manual_by_main_app", "feature.data.ext.tunnel_start_latency_ms": "1606", "feature.name": "vpn-connection", "feature.status": "SUCCESS", "global.platform": "macOS", "global.sample_rate": "1.0", "global.type": "app”]
```

### Impact and Risks
Impact Level: Low
Changes affect internal instrumentation only.

#### What could go wrong?
Incorrect values in the new fields

### Quality Considerations
- Verified as many test cases as I can
- Ensured wide events is successfully fired

### Notes to Reviewer

---

> [!NOTE]
> Introduce enum-based is_setup, macOS onboarding_status, and a new retried status; update controllers to populate these fields and pixel schemas/tests accordingly.
> 
> - **Wide-event data model (`VPNConnectionWideEventData`)**:
>   - Replace `Bool` `isSetup` with enum `SetupState` (`yes|no|unknown`) and add optional `MacOSOnboardingStatus`.
>   - Add `StatusReason.retried`; make enums Codable; expose step key paths; keep `oauth` step.
>   - `pixelParameters()` emits new enum raw values and optional `onboarding_status`; add key `feature.data.ext.onboarding_status`.
> - **Controllers**:
>   - iOS: set `connectionWideEventData.isSetup` to `.yes` on new manager creation, otherwise `.no`.
>   - macOS (BothAppTargets): compute `isSetup` from onboarding state; sync `onboardingStatus` via `UserDefaults`; prefill browser/overall start times; mark incomplete pending flows as `retried`.
>   - macOS (IPC): initialize event with `isSetup: .unknown`, `onboardingStatus: .unknown`; record `browser_start` error on failures; pass through/discard wide-event data.
>   - Packet tunnel: for non-main-app starts set `isSetup: .no`, `onboardingStatus: .completed`; use `retried` for pending flows.
> - **Pixels/schema**:
>   - iOS/macOS JSON: add `feature.status_reason = retried`; change `feature.data.ext.is_setup` to `yes|no|unknown`.
>   - macOS JSON: add `feature.data.ext.onboarding_status` and include `browser_start` in step metrics.
>   - macOS app: use dedicated `UserDefaults` suite for `WideEvent` storage.
> - **Tests**:
>   - Update expectations to `is_setup = "unknown"` by default and keep per-step validations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55c434a0071864c10d8fc818f927cefcdc3b61e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionIPCTunnelController.swift
+++ b/macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionIPCTunnelController.swift
@@ -317,7 +317,7 @@ private extension NetworkProtectionIPCTunnelController {
         wideEvent.startFlow(data)
         data.overallDuration = WideEvent.MeasuredInterval.startingNow()
     }
-    
+
     func passthroughConnectionWideEventData() {
         guard isConnectionWideEventMeasurementEnabled, let data = self.connectionWideEventData else { return }
         vpnConnectionWideEventBrowserStartTime = data.browserStartDuration?.start


macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
```diff
@@ -1055,16 +1059,59 @@ extension NetworkProtectionTunnelController {
...
+            // This call happens before data prefill, so if we detect browser-written fields
+            // here, they come from cross-process leakage in non-sandbox mode.
+            // We discard them to simulate the real sandboxed behaviour.
+            if type(of: data).pixelName == "vpn_connection_debug" && data.browserStartDuration?.start != nil {
+                vpnConnectionWideEventBrowserStartTime = data.browserStartDuration?.start
+                vpnConnectionWideEventOverallStartTime = data.overallDuration?.start
+                wideEvent.discardFlow(data)
+                continue
+            }
+            #endif
```
diegoreymendez: I'm confused by this.  My understanding is user defaults should be process-specific unless we explicitly use an app group.

That said, assuming I'm wrong: my understanding is we want separated storage for each process... if that's the case, can we make the suite name for the wide pixel be different for each process?

I would favor resolving this fundamentally, rather than through a workaround - if we can avoid it.

hanyutang-sandra: @diegoreymendez 

>  My understanding is user defaults should be process-specific unless we explicitly use an app group.

Yes your understanding is correct that in production (macOS Browser App Store build) UserDefaults are process-specific unless using an App Group. The issue we're seeing only occurs because macOS debug build (macOS Browser) is not sandboxed by default, so both processes write to the same plist. 

> can we make the suite name for the wide pixel be different for each process?

And Yes, we can give each process its own suite name if we want full separation. However, this will not change anything for App Store or any sandboxed builds as those builds already isolate UserDefaults per process, so they don’t read each other’s data regardless of the suite name. I added this workaround to keep debug behavior consistent with the sandboxed behavior we ship but I agree that different suite name will solve this more cleanly. Will go ahead and do that just for VPN process

Lemme know if this makes sense.

macOS/DuckDuckGo/NetworkProtection/AppTargets/DeveloperIDTarget/NetworkProtectionIPCTunnelController.swift
```diff
@@ -155,6 +152,7 @@ extension NetworkProtectionIPCTunnelController: TunnelController {
 
             knownFailureStore.reset()
 
+            passthroughConnectionWideEventData()
```
hanyutang-sandra, line 155: Passing it earlier since now there is a possibility that the `NetworkProtectionTunnelController` started with data prefill happening but `try await withCheckedThrowingContinuation` has not returned. And always discard the wide event on success

macOS/DuckDuckGoVPN/DuckDuckGoVPNAppDelegate.swift
```diff
@@ -155,7 +155,8 @@ final class DuckDuckGoVPNAppDelegate: NSObject, NSApplicationDelegate {
         ),
         experimentManager: nil,
         for: FeatureFlag.self)
-    private lazy var wideEvent = WideEvent()
+    private let wideEventVPNAppStorageSuiteName: String = "com.duckduckgo.vpn.wideEvent"
```
hanyutang-sandra, line 158: Using a separate suite name to prevent data pollution in non-sandboxed debug builds

samsymons: Good catch, thanks!

samsymons (approved):
The tweak you made for @diegoreymendez's question makes sense to me, and this overall looks to work well so I'll approve it. Let's double check with @diegoreymendez before merging though.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
diegoreymendez: I think using a different suite name will make this definitely more maintainable.  Thanks!

diegoreymendez (approved):

PR #2577, opened 2025-11-20
WireGuard adapter error recovery

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1212009363425547?focus=true
Tech Design URL:
CC:

### Description

This PR adds `WireGuardAdapter` error recovery for iOS. This works by attempting to recover out of the temporary shutdown state as usual, but if that doesn't succeed then `Task.periodic` is used to keep retrying every 5 seconds, up to a limit of 20 attempts.

### Testing Steps

1. Check that the VPN on iOS and macOS successfully handles path changes (like toggling WiFi off to connect to cellular)

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes iOS WireGuard lifecycle behavior by adding an asynchronous retry loop for resuming from `.temporaryShutdown`, which can affect VPN connectivity and system network settings application under flaky networks.
> 
> **Overview**
> Improves iOS `WireGuardAdapter` resilience when resuming from `.temporaryShutdown` by introducing a delayed retry loop that repeatedly attempts to re-apply tunnel network settings and restart the WireGuard backend after failures.
> 
> Refactors backend bring-up into a shared `bringUpBackend` helper, adds epoch-based cancellation to prevent stale/in-flight recovery work from running, and ensures recovery is started/stopped appropriately on path updates as well as on `stop`/`snooze`.
> 
> Updates and extends `WireGuardAdapterTests` to cover repeated recovery failures/success, cancellation when the path becomes unsatisfiable, and that `update()` during temporary shutdown swaps the settings generator used for the eventual resume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e728d336ebc31710aea66adc36d1ec405f23268c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/VPN/Sources/VPN/WireGuardKit/WireGuardAdapter.swift
```diff
@@ -634,10 +648,104 @@ final class WireGuardAdapter: WireGuardAdapterProtocol {
...
+                    continuation.resume(returning: false)
+                    return
+                }
+
+                guard settingsGenerator === activeGenerator else {
+                    continuation.resume(returning: false)
+                    return
+                }
+
+                guard let lastKnownPathStatus = self.lastKnownPathStatus, lastKnownPathStatus.isSatisfiable else {
```
diegoreymendez: I'd consider skipping `lastKnownPathStatus.isSatisfiable`, as this kind of API sometimes caches status.  I don't know if this is the same as the old internet connectivity APIs Apple had, but Apple's suggestion used to be not to check those before testing connectivity - but test connectivity, and check those on failure.

SharedPackages/VPN/Sources/VPN/WireGuardKit/WireGuardAdapter.swift
```diff
@@ -634,10 +648,104 @@ final class WireGuardAdapter: WireGuardAdapterProtocol {
...
+                    continuation.resume(returning: false)
+                    return
+                }
```
diegoreymendez: Nit: Could we unify all the guards under a single `canAttemptRecovery` method?

SharedPackages/VPN/Sources/VPN/WireGuardKit/WireGuardAdapter.swift
samsymons: Good call, will update this and mark it ready for re-review once that's done. Thanks!

SharedPackages/VPN/Sources/VPN/WireGuardKit/WireGuardAdapter.swift
```diff
@@ -634,10 +632,58 @@ final class WireGuardAdapter: WireGuardAdapterProtocol {
...
+
+            if temporaryShutdownRecoveryFailed {
+                eventHandler.handle(.endTemporaryShutdownStateRecoveryFailure(error))
+            } else {
+                eventHandler.handle(.endTemporaryShutdownStateAttemptFailure(error))
+                temporaryShutdownRecoveryFailed = true
+            }
+
+            workQueue.asyncAfter(deadline: .now() + temporaryShutdownRecoveryDelay) { [weak self] in
+                self?.attemptResume(settingsGenerator: settingsGenerator)
```
samsymons: This will continue in a loop until it's either able to succeed or the VPN shuts down.

diegoreymendez: This is remembering the generator at schedule time, is this correct? ie: won't this affect what happens if the status was updated to a new `temporaryShutdown` with a new generator.

SharedPackages/VPN/Sources/VPN/WireGuardKit/WireGuardAdapter.swift
```diff
@@ -634,10 +632,58 @@ final class WireGuardAdapter: WireGuardAdapterProtocol {
...
+            logHandler(.error, "Failed to restart backend: \(error.localizedDescription)")
+
+            if temporaryShutdownRecoveryFailed {
+                eventHandler.handle(.endTemporaryShutdownStateRecoveryFailure(error))
+            } else {
+                eventHandler.handle(.endTemporaryShutdownStateAttemptFailure(error))
+                temporaryShutdownRecoveryFailed = true
+            }
+
+            workQueue.asyncAfter(deadline: .now() + temporaryShutdownRecoveryDelay) { [weak self] in
```
diegoreymendez, line 704: Is it possible we might schedule two in parallel if there's a status change before this fires?

diegoreymendez (commented):
Thanks @samsymons  for taking the time to push this.

Both inline questions trace back to the same gap: we don't keep a reference to the in-flight retry, so we can't cancel or replace it when conditions change.

My suggestion: make the retry a single cancellable task managed by `.temporaryShutdown`.  At a high level I mean:
- Entering `.temporaryShutdown` (from network-down or from update()) cancels any existing task and starts a new one. New generator gets picked up immediately; only one chain is ever in flight.
- The task is a loop: try `bringUpBackend`, sleep, repeat, (maybe) give up after N.  I think it's ok to leave it uncapped but if we do you might want to update the PR description as it says there's a cap of "20".
- Leaving `.temporaryShutdown` cancels the task. 
                                                                                                                                                                 
Feel free to ping me via MM if useful.

diegoreymendez (approved):
Works very well for me, let's push this out.

PR #2709, opened 2025-12-01
[macOS][Feature Flags] Remove SAD flag for active users

alessandroboron:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206329551987282/task/1211866474521433?focus=true
Tech Design URL:
CC:

### Description

This PR removes the feature flag for SAD active users. Even though we use remote settings for the prompts, the settings are defined in the parent feature, so it is safe to remove the subfeature from the codebase.

### Testing Steps
1. Smoke test showing SAD prompt for active users.
2. Put a breakpoint in `DefaultBrowserAndDockPromptFeatureFlagger` -> `getSettings(_ value:) -> Int` line 124.
Open Debug Menu and ensure that `remoteSettings[value.rawValue] as? Int` returns the expected value and it is not nil.

### Impact and Risks
<!— 

—>

#### What could go wrong?

---

> [!NOTE]
> Removes the active-user Default Browser/Add to Dock feature flag and references, keeping inactive-user flag and using remote settings for timing.
> 
> - **Feature Flags**:
>   - Remove `FeatureFlag.scheduledSetDefaultBrowserAndAddToDockPrompts` and its mappings (defaults, source).
>   - Drop `SetAsDefaultAndAddToDockSubfeature.scheduledDefaultBrowserAndDockPrompts`.
>   - Simplify provider/typealiases to only support inactive-user flag; remove active-user flag check/accessors.
> - **Prompt Logic**:
>   - `ActiveUser` prompt decider no longer gated by a feature flag; relies on timing settings only.
>   - Debug menu always builds core items; conditionally appends inactive-specific items based on `scheduledDefaultBrowserAndDockPromptsInactiveUser`.
> - **Tests**:
>   - Remove tests/mocks for the active-user flag; update tests to focus on inactive-user flag and timing settings (including remote/default values).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 20d7cca4a6327723d5a985fbc4d51af2b44e012d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


alessandroboron:
> Since we use remote settings with a remote parent feature, should we retain a corresponding parent feature flag in the app (e.g. in case we want to disable the remote parent feature)?

I believe we're only looking to adjust the remote settings and not disable the feature entirely. We’ve been testing it for a while and are happy with the conversion rate, CTR, and its functional behaviour. I don’t think we'll want to disable it.

The same goes for the SAD/ATT prompt for inactive users. Once we confirm that it works as expected and there are no issues, and if we decide to keep the feature, I see no reason to keep the feature flag in place.

PR #2783, opened 2025-12-04
macOS VPN: Mitigate VPN toggle issue

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203108348835387/task/1212293594200883?focus=true

### Testing Steps

**Preparation Steps**

Apply these temporary changes to simulate the issue:
- In VPNEnabledObserverThroughSession.swift:
a) Add counter variables:

```
   #if DEBUG
   private static var simulateNilSessionCounter = 0
   private static let simulateNilSessionEveryN = 3  // Fail every 3rd call
   #endif
```

b) Add simulation logic at the start of `handleNotification`:

```
   #if DEBUG
   Self.simulateNilSessionCounter += 1
   if Self.simulateNilSessionCounter % Self.simulateNilSessionEveryN == 0 {
       // Simulate activeSession() returning nil
       if subject.value != false {
           subject.send(false)
       }
       return
   }
   #endif
```

This forces the observer to report "disconnected" every 3rd notification, which on the first run is usually "Connected".

c) Start the VPN once to install the extension, then In macOS `NetworkProtectionTunnelController.swift` - comment out `activateSystemExtension` call as in DEBUG builds it allows the toggle to recover the connection by reinstalling the extension each time - we want to simulate release.

### Impact and Risks

Low.

#### What could go wrong?

I truly don't think there's any significant risk here.  Adding stop in the VPN start calls isn't great... but I think that's well limited.

### Quality Considerations

NA

### Notes to Reviewer

NA

---
###### Internal references:

---

> [!NOTE]
> On both iOS and macOS, starting while already connected now logs and stops the VPN to allow recovery; removed the macOS-specific “connectionAlreadyStarted” error.
> 
> - **Network Protection (start behavior)**:
>   - iOS `NetworkProtectionTunnelController.startWithError`: when `connection.status == .connected`, now logs and calls `stop()` to recover instead of no-op.
>   - macOS `NetworkProtectionTunnelController.start(isFirstAttempt:)`: when already connected, now logs and calls `stop()` to recover instead of throwing an error.
> - **Error handling (macOS)**:
>   - Removed `StartError.connectionAlreadyStarted` case and all associated codes/descriptions/usages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b766086de9807706e0a508dd109eccdb6ea01a6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
+++ b/macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
@@ -531,7 +531,6 @@ final class NetworkProtectionTunnelController: TunnelController, TunnelSessionPr
         case cancelled
         case noAuthToken
         case connectionStatusInvalid
-        case connectionAlreadyStarted
         case simulateControllerFailureError
         case startTunnelFailure(_ error: Error)
         case failedToFetchAuthToken(_ error: Error)
@@ -542,13 +541,6 @@ final class NetworkProtectionTunnelController: TunnelController, TunnelSessionPr
                 return nil
             case .noAuthToken:
                 return "You need a subscription to start the VPN"
-            case .connectionAlreadyStarted:
-#if DEBUG
-                return "[Debug] Connection already started"
-#else
-                return nil
-#endif
-
             case .connectionStatusInvalid:
 #if DEBUG
                 return "[DEBUG] Connection status invalid"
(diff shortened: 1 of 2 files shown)


macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionTunnelController.swift
```diff
@@ -542,13 +541,6 @@ final class NetworkProtectionTunnelController: TunnelController, TunnelSessionPr
                 return nil
             case .noAuthToken:
                 return "You need a subscription to start the VPN"
-            case .connectionAlreadyStarted:
```
diegoreymendez, line 545: No longer used as an error.

PR #2609, opened 2025-11-24
Defer creating Downloads directory until needed

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1209362951077111?focus=true

### Description
Implements lazy initialization for the iOS Downloads directory to prevent an empty "DuckDuckGo" folder from appearing in the Files app when no files have been downloaded.

**Changes:**
- **DownloadManager**: Removed eager directory creation from `init()`, added lazy creation before file operations. And added directory cleanup on launch
- **DownloadsDirectoryHandling**: Added `deleteDownloadsDirectoryIfEmpty()` to clean up empty directories
- **AIChat integration**: Added `aiChatViewControllerWillStartDownload()` delegate method to create directory before AIChat downloads
- **DownloadsDeleteHelper**: Added tracking for tmp directories
- **DownloadsListDataSource**: Now triggers directory cleanup if all downloads are deleted. Also tracks tmp directories.
---
### Testing Steps

**Verify directory not created on fresh install:**
1. Delete and reinstall the app (or clear app data)
2. Open Settings → Files app → On My iPhone
3. ✅ Verify "DuckDuckGo" folder does NOT appear

**Verify directory created on first download:**

1. In the DuckDuckGo app, navigate to a page with a downloadable file (e.g., https://file-examples.com/wp-content/storage/2017/02/zip_2MB.zip)
2. Download a file
3. Open Files app → On My iPhone
4. ✅ Verify "DuckDuckGo/Downloads" folder now appears with the downloaded file

**Verify temporary downloads don't create directory:**

1. Delete the app and reinstall (or clear app data)
2. Navigate to a page with a PDF file (e.g., https://file-examples.com/index.php/sample-documents-download/sample-pdf-download/)
3. Tap to preview the file (should auto-preview without saving)
4. Check Files app → On My iPhone
5. ✅ Verify "DuckDuckGo" folder does NOT appear

**Verify AIChat downloads create directory:**

1. Delete the app and reinstall (or clear app data)
2. Open AI Chat
3. ✅ Verify "DuckDuckGo" folder does NOT appear
4. Ask anything
5. Download the chat using the download button
6. Check Files app → On My iPhone
7. ✅ Verify "DuckDuckGo/Downloads" folder appears with the chat history file

**Verify empty directory cleanup on launch:**

1. Open the app
2. Make a download
3. Force quit the app
4. Open Files app, navigate to DuckDuckGo/Downloads
[...]

PR #2880, opened 2025-12-11
[iOS] [Hotfix] Fix potential data loss of downloads data

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211767540372501/task/1212341997874113?focus=true

### Description
- Changed downloadsDirectoryFiles property to throw errors instead of silently returning an empty array, enabling proper error handling when file system operations fail.
- Modified deleteDownloadsDirectoryIfEmpty() to fail-safe: if directory contents cannot be read, the directory is preserved rather than potentially deleted with files inside

### Testing Steps
Testing steps from #2609 

### Impact and Risks
Impact Level: Medium

### Notes to Reviewer

Nil return for download metadata: downloadMetaData now returns nil if filename generation fails, which may prevent some downloads from starting. This is safer than proceeding with an invalid filename

---
###### Internal references:

---

> [!NOTE]
> Make downloads directory listing throwable and fail-safe empty-directory deletion; update DownloadManager usage and tests/mocks.
> 
> - **Downloads handling**:
>   - Make `DownloadsDirectoryHandling.downloadsDirectoryFiles` throwable; implement guarded, throwing read with logging and directory-existence check.
>   - Update `deleteDownloadsDirectoryIfEmpty` to fail-safe (do not delete on read errors) and handle deletion errors.
>   - In `DownloadManager`, use computed `downloadsDirectoryFiles` that wraps the throwing call with `try?` and fallback to `[]`; use it when generating unique filenames; minor formatting tweaks.
> - **Tests**:
>   - Adjust unit/integration tests to `try` accessing `downloadsDirectoryFiles` and mark affected tests as `throws`.
>   - Update test mocks to conform to the throwing `downloadsDirectoryFiles` API.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit faa232d392cc34ecbd706659b9fc685b8cd1dd97. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/DownloadManager.swift
+++ b/iOS/DuckDuckGo/DownloadManager.swift
@@ -156,7 +156,6 @@ class DownloadManager: DownloadManaging {
 extension DownloadManager {
 
     private func convertToUniqueFilename(_ filename: String) -> String {
-        let downloadsDirectoryFiles: [URL] = (try? downloadsDirectoryHandler.downloadsDirectoryFiles) ?? []
         let downloadingFilenames = Set(downloadList.map { $0.filename })
         let downloadedFilenames = Set(downloadsDirectoryFiles.map { $0.lastPathComponent })
         let list = downloadingFilenames.union(downloadedFilenames)


iOS/DuckDuckGo/DownloadManager.swift
```diff
@@ -154,9 +157,10 @@ class DownloadManager: DownloadManaging {
 
 extension DownloadManager {
 
-    private func convertToUniqueFilename(_ filename: String) -> String {
+    private func convertToUniqueFilename(_ filename: String) throws -> String {
+        let downloadsDirectoryFiles: [URL] = (try? downloadsDirectoryHandler.downloadsDirectoryFiles) ?? []
```
dus7: This function has no throwing calls (because of `try?`), but handling this would mean further changes. Given the fact we didn't had any reports of data loss previously, and the issue we're fixing here is happening on startup and depends on timing (presumably) I think we should keep previous behavior here and convert this function to non throwing.
That will also allow to change other calls to non-throwing, only gracefully handling the initial failure when attempting to clear downloads directory.

iOS/DuckDuckGo/DownloadManager.swift
```diff
@@ -105,7 +105,10 @@ class DownloadManager: DownloadManaging {
     }
 
     func downloadMetaData(for response: URLResponse, suggestedFilename: String? = nil) -> DownloadMetadata? {
-        let filename = filename(forSuggestedFilename: suggestedFilename ?? response.suggestedFilename, mimeType: response.mimeType)
+        guard let filename = try? filename(forSuggestedFilename: suggestedFilename ?? response.suggestedFilename,
```
federicocappelli: I understand that this is enlarging the scope of the fix, and you may decide to postpone this change, but using nil as a response when we have a failure it's a risky approach.
I would suggest making this function throw, so the return value is clear:
- A value: all ok
- Nil: the file doesn't exist
- throw an error: something went wrong

iOS/DuckDuckGo/DownloadManager.swift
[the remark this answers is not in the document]
hassaanelgarem: Addressed in 689ee5edb655c0316f13e65ec83c22aebe0e87d9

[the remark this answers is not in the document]
federicocappelli: I agree with Cursor here, we rarely should mask errors and convert them in a nil, instead we should pass errors up the chain.

iOS/DuckDuckGo/DownloadManager.swift
```diff
@@ -243,6 +247,6 @@ extension NSNotification.Name {
 
 
 extension DownloadManager {
-    var downloadsDirectoryFiles: [URL] { downloadsDirectoryHandler.downloadsDirectoryFiles }
+    var downloadsDirectoryFiles: [URL] { (try? downloadsDirectoryHandler.downloadsDirectoryFiles) ?? [] }
```
federicocappelli, line 247: same here, an empty array means there are no files in the folder, not "something wrong happened"

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: I missed removing the optional try. Fixed here 689ee5edb655c0316f13e65ec83c22aebe0e87d9

As for the larger discussion, I get your point. I was trying to be extra careful this time around, but your reasoning makes sense as well.

iOS/DuckDuckGo/DownloadManager.swift
dus7: I completely agree here, but I'd argue that this change should be a follow up, and does not belong in a hotfix.

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: `downloadsDirectoryFiles` is used for reading, which doesn't present a data loss risk. So I chose to simplify this part.

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: This discussion is related: https://github.com/duckduckgo/apple-browsers/pull/2880#discussion_r2610993390
@dus7 thoughts?

hassaanelgarem: I agree on doing this as a follow-up. Same for `DownloadManager.downloadsDirectoryFiles`

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: Will address this in a separate PR 👍

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: Updated this change and will address this properly in a separate PR 👍

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: Updated this change and will address this properly in a separate PR 👍

iOS/DuckDuckGo/DownloadManager.swift
```diff
@@ -155,8 +156,9 @@ class DownloadManager: DownloadManaging {
 extension DownloadManager {
 
     private func convertToUniqueFilename(_ filename: String) -> String {
+        let downloadsDirectoryFiles: [URL] = (try? downloadsDirectoryHandler.downloadsDirectoryFiles) ?? []
```
miasma13: This is probably unnecessary as `downloadsDirectoryFiles` is defined with the same logic in the extension below in line  248

iOS/DuckDuckGo/DownloadsDirectoryHandling.swift
```diff
@@ -46,10 +46,20 @@ struct DownloadsDirectoryHandler: DownloadsDirectoryHandling {
...
+            do {
+                guard downloadsDirectoryExists() else {
+                    return []
+                }
+                let contents = try FileManager.default.contentsOfDirectory(at: downloadsDirectory,
+                                                                           includingPropertiesForKeys: nil,
+                                                                           options: .skipsHiddenFiles)
+                return contents.filter { !$0.hasDirectoryPath }
+            } catch {
+                Logger.general.debug("Could not read downloadsDirectoryFiles: \(error.localizedDescription)")
```
federicocappelli: Logger.general.error not debug

iOS/DuckDuckGo/DownloadsDirectoryHandling.swift
```diff
@@ -70,9 +80,21 @@ struct DownloadsDirectoryHandler: DownloadsDirectoryHandling {
...
         
-        // Check if directory is empty (no files)
-        if downloadsDirectoryFiles.isEmpty {
-            try? FileManager.default.removeItem(at: downloadsDirectory)
+        let files: [URL]
+        do {
+            files = try downloadsDirectoryFiles
+        } catch {
+            // Don't delete if we couldn't enumerate - fail safe
+            Logger.general.debug("Could not read downloads directory, not deleting: \(error.localizedDescription)")
```
federicocappelli: Logger.general.error not debug

iOS/DuckDuckGo/DownloadsDirectoryHandling.swift
```diff
@@ -70,9 +80,21 @@ struct DownloadsDirectoryHandler: DownloadsDirectoryHandling {
...
+            return
+        }
+        
+        guard files.isEmpty else { return }
+        
+        do {
+            try FileManager.default.removeItem(at: downloadsDirectory)
+        } catch {
+            Logger.general.debug("Could not delete downloads directory: \(error.localizedDescription)")
```
federicocappelli: Logger.general.error not debug

iOS/DuckDuckGo/DownloadManager.swift
hassaanelgarem: Addressed in faa232d392cc34ecbd706659b9fc685b8cd1dd97

iOS/DuckDuckGo/DownloadsDirectoryHandling.swift
hassaanelgarem: Addressed in d2ee3bcbd53b63a3cc9e8a4850e3ee11292ee05e
Same for the rest

dus7 (approved):

PR #2888, opened 2025-12-12
PIR: Implement click action delay optimization

edulpn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1212389011481642
Tech Design URL:
CC: @quanganhdo @samsymons 

### Description
This PR introduces changes to the delays executing when sending `click` actions to PIR web views. The legacy behaviour has 40s delays executing AFTER the `click` action is executed on the web view successfully. The optimized behaviour now has 3s delays executing BEFORE the `click` action is executed on the web view.  The two branches are switched by the use of the `clickActionDelayReductionOptimization` Privacy Config remote feature flag.

### Testing Steps
1. When first launching the app, the feature flag should be turned OFF, defaulting to the legacy behaviour
2. Make sure you are debugging the background process (DuckDuckGo Personal Information Removal).
3. Run a PIR scan and, ideally, an opt-out for the extracted profiles. Watch the debugger logs for `Executing click action delay AFTER click: 40s` (scan step click actions will show 0s instead)
4. Flip the feature flag manually (to ON) via the debug menu and re-run the scan and opt-out. Watch the debugger logs for `Executing click action delay BEFORE click: 3s` (scan step click actions will show 0s instead)

### Impact and Risks

#### What could go wrong?
More aggressive click action delays can lead to more PIR step failures

### Quality Considerations
Please focus on the delay implementation logic and feature flag structure.

### Notes to Reviewer

---

> [!NOTE]
> Implements a feature-flagged optimization that moves opt-out click delays before the click (3s) instead of after (40s), with configurable timings and full iOS/macOS flag plumbing plus tests.
> 
> - **DBP Core**:
>   - Add `DBPSubfeature.clickActionDelayReductionOptimization` and expose via `DBPFeatureFlagging` (`isClickActionDelayReductionOptimizationOn`).
>   - Extend `BrokerJobExecutionConfig` with configurable click await times: `optimizedClickAwaitTimeForOptOut` (3s), `legacyClickAwaitTimeForOptOut` (40s), `clickAwaitTimeForScan` (0s).
> - **Web Runners**:
>   - `BrokerProfileOptOutSubJobWebRunner`/`BrokerProfileScanSubJobWebRunner`: compute `clickAwaitTime` lazily from `executionConfig` and feature flag.
>   - `SubJobWebRunner`: when flag is ON and action is `ClickAction`, wait BEFORE click; otherwise wait AFTER click (legacy). Add logs for timing.
> - **Feature Flags (iOS/macOS)**:
>   - Add `FeatureFlag.dbpClickActionDelayReductionOptimization` and map to `DBPSubfeature.clickActionDelayReductionOptimization`.
>   - Wire flag through iOS `DBPService.DBPFeatureFlagger`, macOS `DBPFeatureFlagger`, and macOS debug view model.
> - **Tests/Mocks**:
>   - Update mocks to include new flag.
>   - Add tests verifying runner click delays (3s vs 40s) and scan delay (0s).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 48439c554a4b8bafc9e69cf30a51caa89851a362. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/QueueManaging/BrokerJobExecutionConfig.swift
+++ b/SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/QueueManaging/BrokerJobExecutionConfig.swift
@@ -43,10 +43,12 @@ public struct BrokerJobExecutionConfig {
         /// Number of concurrent jobs allowed during manual scans
         public static let defaultConcurrentJobsOnManualScans: Int = 6
 
-        /// Click await time when click delay optimization is enabled (delay before click)
-        public static let defaultOptimizedClickAwaitTime: TimeInterval = .seconds(3)
-        /// Click await time when click delay optimization is disabled (legacy delay after click)
-        public static let defaultLegacyClickAwaitTime: TimeInterval = .seconds(40)
+        /// Click await time when click delay optimization is enabled (delay before click) for opt-out step actions
+        public static let defaultOptimizedClickAwaitTimeForOptOut: TimeInterval = .seconds(3)
+        /// Click await time when click delay optimization is disabled (legacy delay after click) for opt-out step actions
+        public static let defaultLegacyClickAwaitTimeForOptOut: TimeInterval = .seconds(40)
+        /// Click await time for scan step actions
+        public static let defaultClickAwaitTimeForScan: TimeInterval = .seconds(0)
     }
 
     let intervalBetweenSameBrokerJobs: TimeInterval
@@ -54,8 +56,9 @@ public struct BrokerJobExecutionConfig {
     public let optOutJobTimeout: TimeInterval
     public let cssActionTimeout: TimeInterval
     public let cssActionCancellationCheckInterval: TimeInterval
-    public let optimizedClickAwaitTime: TimeInterval
-    public let legacyClickAwaitTime: TimeInterval
+    public let optimizedClickAwaitTimeForOptOut: TimeInterval
+    public let legacyClickAwaitTimeForOptOut: TimeInterval
+    public let clickAwaitTimeForScan: TimeInterval
 
     private let concurrentJobsDifferentBrokers: Int
     // https://app.asana.com/0/481882893211075/1206981742767469/f
@@ -76,8 +79,9 @@ public struct BrokerJobExecutionConfig {
                 cssActionCancellationCheckInterval: TimeInterval = Constants.defaultCssActionCancellationCheckInterval,
                 concurrentJobsDifferentBrokers: Int = Constants.defaultConcurrentJobsDifferentBrokers,
                 concurrentJobsOnManualScans: Int = Constants.defaultConcurrentJobsOnManualScans,
-                optimizedClickAwaitTime: TimeInterval = Constants.defaultOptimizedClickAwaitTime,
-                legacyClickAwaitTime: TimeInterval = Constants.defaultLegacyClickAwaitTime) {
+                optimizedClickAwaitTimeForOptOut: TimeInterval = Constants.defaultOptimizedClickAwaitTimeForOptOut,
+                legacyClickAwaitTimeForOptOut: TimeInterval = Constants.defaultLegacyClickAwaitTimeForOptOut,
+                clickAwaitTimeForScan: TimeInterval = Constants.defaultClickAwaitTimeForScan) {
         self.intervalBetweenSameBrokerJobs = intervalBetweenSameBrokerJobs
         self.scanJobTimeout = scanJobTimeout
         self.optOutJobTimeout = optOutJobTimeout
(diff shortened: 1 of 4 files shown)


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/BrokerProfileJobDependencies.swift
```diff
@@ -121,6 +121,7 @@ public struct BrokerProfileJobDependencies: BrokerProfileJobDependencyProviding
             emailConfirmationDataService: self.emailConfirmationDataService,
             captchaService: self.captchaService,
             featureFlagger: self.featureFlagger,
+            clickAwaitTime: featureFlagger.isClickActionDelayReductionOptimizationOn ? 3 : 40,
```
THISISDINOSAUR: I think we should move these numbers into the execution config

SharedPackages/DataBrokerProtectionCore/Tests/DataBrokerProtectionCoreTests/SubJobWebRunnerTests.swift
```diff
@@ -99,6 +99,76 @@ final class DataBrokerJobTests: XCTestCase {
         // Then
         XCTAssertEqual(sut.retriesCountOnError, 1)
     }
+
+    // MARK: - Click Delay Optimization Tests
+
+    func testWhenClickDelayOptimizationFlagIsOn_thenOptOutRunnerUsesOptimizedDelay() {
```
THISISDINOSAUR: I'm not sure I see the value in these tests

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/BrokerProfileJobDependencies.swift
THISISDINOSAUR: Also given BrokerProfileOptOutSubJobWebRunner has the feature flagger, I think it should decide what the click time should be itself rather than in the constructor here

samsymons: I agree with this overall, right now it's the test choosing the value using the feature flagger, and so all it ends up testing is that the value in `clickAwaitTime` gets set in the runner - not that the runner is using the correct value based on the feature flag.

But also I agree with @THISISDINOSAUR that the runner could use the feature flag directly and that we don't need a dedicated param in the constructor.

SharedPackages/DataBrokerProtectionCore/Tests/DataBrokerProtectionCoreTests/SubJobWebRunnerTests.swift
edulpn: Yeah, I wrote these mostly to make sure I was covering the full dependency injection chain, but noticed they are useless since they just check if the parameter is the same it’s passing. Forgot to delete, though 🤦

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/Operations/BrokerProfileJobDependencies.swift
edulpn: > I think we should move these numbers into the execution config

Create constants for high and low delays, right? I agree, I wasn’t aware we had `BrokerJobExecutionConfig`.

About the parameter, it looks like I’m creating it, but it already exists (comes from the `SubJobWebRunner` protocol) we just always use the default values which is exactly how we are configuring the delays currently (still using it to set 0s for scan sub job web runners). I guess I could change that, but that’d be undoing the dependency injection/protocol pre-implementation on `SubJobWebRunner`.

THISISDINOSAUR: "Create constants for high and low delays, right?"
Yep

Fwiw in general I think we need to be more deliberate about how we use default values in the PIR codebase (and in general give more thought to the function signatures). IMO the original doesn't really make sense, SubJobWebRunning having a parameter for clickAwaitTime, but then each subclass always having a default value passed through the init, that is then never changed is odd and IMO is what lead to the original mistake of the comment being incorrect on how long we actually wait.

Because of that, I think the only way it makes sense to keep the function parameter is if we actually make use of it and have the sub jobs figure out what the click action should be themselves, but...

If I was reworking the original, I'd probably change it so the subclasses (e.g. BrokerProfileOptOutSubJobWebRunner) have a computed value for clickAwaitTime and get rid of the function parameter

edulpn: So the suggestion is that we should re-work the original default parameter approach on this PR, then? Maybe this should be scoped separately, if that’s the case.

THISISDINOSAUR: Is is, but why separately? It's a fairly small change

edulpn: If you think it’s small enough of a change, sounds good to me. I’ll update the PR.

edulpn: I addressed the comments, I kept the `clickAwaitTime ` in the protocol and had it initialized as `0` for the scan sub job web runner and a lazy computed variable that uses the feature flagger directly on the opt-out sub job web runner.

THISISDINOSAUR (commented):
I'm getting "Executing click action delay BEFORE click: 3.0000s" before I turned on the feature flag override.

It looks like the default is true?
Screenshot 2025-12-17 at 2 45 00 PM

PR #2906, opened 2025-12-14
[iOS/macOS] Wide event POST endpoint support

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1212438492157495?focus=true
Tech Design URL:
CC:

### Description

This feature adds POST endpoint support for wide events.

### Testing Steps

1. Open `WideEventSending.swift` and modify the `defaultPOSTRequestHandler` function to send events while running in DEBUG mode (by default, they aren't sent)
2. Run the apps and go through some wide event flows, like restoring a subscription
3. Open [this Jupyter notebook](https://jupyter.duckduckgo.com/hub/user-redirect/lab/tree/ssymons/Wide%20Event%20POST%20Endpoint.ipynb) and refresh it to see if your events are present

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> Introduces feature‑flagged POST delivery for wide events and refactors sending behind a new sender.
> 
> - Adds `WideEventSending`/`DefaultWideEventSender` to build parameters, fire PixelKit daily/standard pixels, and (when enabled) POST JSON to `https://improving.duckduckgo.com/e`
> - `WideEvent` now requires a `featureFlagProvider` and delegates completion sending to the new `sender`; removes internal pixel name/dispatch logic; exposes `WideEventPixelKitEvent` publicly
> - Defines `WideEventFeatureFlagProviding` and adds `wideEventPostEndpoint` flags (iOS/macOS) with adapters in iOS/macOS apps, VPN tunnels, DBP agent, and subscription config; plumbs setting through VPN (`UserDefaults`, `VPNSettings`, `VPNSettingsSnapshot`, startup options)
> - Improves JSON parameter support (`jsonParameters` with `Encodable`, numeric `sample_rate`), adds `WideEventError.emptyPixelName`
> - Updates construction sites to pass `WideEvent(featureFlagProvider: ...)` and adds comprehensive unit tests for sender/flag behavior and VPN settings
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a826404380fcb1b6a7bc7375eec3ab84bd3b34f4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1212762411942336

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventSending.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventSending.swift
@@ -52,7 +52,12 @@ public final class DefaultWideEventSender: WideEventSending {
         featureFlagProvider: WideEventFeatureFlagProviding,
         onComplete: @escaping PixelKit.CompletionBlock
     ) {
-        let pixelName = Self.generatePixelName(for: T.pixelName)
+        guard let pixelName = Self.generatePixelName(for: T.pixelName) else {
+            Self.logger.warning("Cannot fire wide event: empty pixel name")
+            onComplete(false, WideEventError.emptyPixelName)
+            return
+        }
+
         let parameters = generateFinalParameters(from: data, status: status)
 
         guard !parameters.isEmpty else {
@@ -93,8 +98,10 @@ public final class DefaultWideEventSender: WideEventSending {
         return parameters
     }
 
-    private static func generatePixelName(for name: String) -> String {
-        assert(!name.isEmpty, "Pixel name should not be empty")
+    private static func generatePixelName(for name: String) -> String? {
+        guard !name.isEmpty else {
+            return nil
+        }
 
         #if os(macOS)
         return "m_mac_wide_\(name)"
--- a/SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventError.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventError.swift
@@ -26,6 +26,7 @@ public enum WideEventError: DDGError, LocalizedError {
     case invalidFlowState
     case storageError(Error)
     case invalidParameters(String)
+    case emptyPixelName
 
     public var description: String {
         switch self {
@@ -41,6 +42,8 @@ public enum WideEventError: DDGError, LocalizedError {
             return "Storage error: \(error.localizedDescription)"
         case .invalidParameters(let message):
             return "Invalid parameters: \(message)"
+        case .emptyPixelName:
+            return "Pixel name should not be empty"
         }
     }
 
@@ -64,12 +67,14 @@ public enum WideEventError: DDGError, LocalizedError {
             return 4
         case .invalidParameters:
             return 5
+        case .emptyPixelName:
+            return 6
         }
     }
 
     public var underlyingError: Error? {
         switch self {
-        case .flowNotFound, .typeMismatch, .invalidFlowState, .invalidParameters:
+        case .flowNotFound, .typeMismatch, .invalidFlowState, .invalidParameters, .emptyPixelName:
             return nil
         case .serializationFailed(let error), .storageError(let error):
             return error
@@ -90,6 +95,8 @@ public enum WideEventError: DDGError, LocalizedError {
             return (lhsError as NSError) == (rhsError as NSError)
         case (.invalidParameters(let lhsMessage), .invalidParameters(let rhsMessage)):
             return lhsMessage == rhsMessage
+        case (.emptyPixelName, .emptyPixelName):
+            return true
         default:
             return false
         }


SharedPackages/VPN/Sources/VPN/PacketTunnelProvider.swift
```diff
@@ -412,7 +412,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                 providerEvents: EventMapping<Event>,
                 settings: VPNSettings,
                 defaults: UserDefaults,
-                wideEvent: WideEventManaging = WideEvent(),
+                wideEvent: WideEventManaging = WideEvent(sendPOSTEnabled: { true }),
```
samsymons: This needs to be fixed by setting this on the app side before starting the VPN. I'll address this before merging, but in the mean time the core behavior can be reviewed.

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
```diff
@@ -129,6 +129,9 @@ final class AppDependencyProvider: DependencyProvider {
             featureFlagger = defaultFeatureFlagger
         }
 
+        let wideEvent = WideEvent(sendPOSTEnabled: { [featureFlagger] in featureFlagger.isFeatureOn(.wideEventPostEndpoint) })
```
hanyutang-sandra: Correct me if I’m missing anything. Is there a reason we’re passing this feature-flag check through the initializer rather than making it a computed property inside WideEvent class (Or even DefaultWideEventSender())?

SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventParameters.swift
```diff
@@ -20,52 +20,12 @@ import Foundation
 
 public protocol WideEventParameterProviding {
     func pixelParameters() -> [String: String]
-    func jsonParameters() throws -> String
+    func typedParameters() -> [String: Any]
```
hanyutang-sandra: nit: This acts as a type eraser in practice. Renaming it to something that reflects it (e.g. erasedParameters) maybe more clear?

SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventSending.swift
```diff
@@ -0,0 +1,258 @@
...
+        onComplete(true, nil)
+        return
+#else
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        request.httpBody = body
+        headers.forEach { request.setValue($1, forHTTPHeaderField: $0) }
+
+        URLSession.shared.dataTask(with: request) { _, response, error in
+            let success = (response as? HTTPURLResponse)?.statusCode == 200
```
hanyutang-sandra: nit: we are sure this is the only expected success code right? Which means no 201, 202 etc.

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
hanyutang-sandra: Or is it because of the hardcoded value in PacketTunnelProvider

SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventSending.swift
samsymons: That's the promise from the backend implementation, though to be safe I've updated it to treat all status codes in the 200 range as a success.

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
samsymons: That's so that we get the latest value every time we read it - the `WideEvent` instances live for a long time, even for the entire lifetime of the app, so if we pass it into the initializer then we'll cache the feature flag state and not be able to change it remotely easily. (This same problem hit us with the macOS hang rate pixel recently, and when we disabled it it took a long time for error pixels to calm down.)

SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventParameters.swift
samsymons: I ended up going back to `jsonParameters` for this function's name since I want to make it clear to developers what the purpose of it is, and I changed the type to `[String: Encodable]` for the same reason - let me know if that sounds reasonable to you or if you still prefer the original suggestion.

SharedPackages/BrowserServicesKit/Sources/PixelKit/WideEvent/WideEventSending.swift
```diff
@@ -0,0 +1,261 @@
...
+public protocol WideEventSending {
+    func send<T: WideEventData>(
+        _ data: T,
+        status: WideEventStatus,
+        sendPOSTEnabled: Bool,
+        onComplete: @escaping PixelKit.CompletionBlock
+    )
+}
+
+public final class DefaultWideEventSender: WideEventSending {
```
samsymons, line 31: This new class needs unit tests - I'll add them on Wednesday. It should also send pixels when it fails to encode the type, and a pixel if we receive a 400 code back (which is what the backend will send when validation fails). I'll add these too, and open a Privacy Triage.

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
hanyutang-sandra: I might be misunderstanding something here, but just wanted to double check: I was thinking of a computed property in WideEvent (not passing a boolean value through init or storing a boolean property) like:
```
var sendPOSTEnabled: Bool {
    featureFlagger.isFeatureOn(.wideEventPostEndpoint)
}
````
so that the state would be read everytime we call `sendPOSTEnabled`, and remote changes should still be observed for long-lived instances. 

The main reason I raised this is that the closure is passed through all call sites and adds a fair amount of changes. Centralizing it in WideEvent might reduce changes while still preserving this dynamic behavior. But I’m open if you prefer this way

Comment (samsymons) [reply iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift]: @hanyutang-sandra I see - overall I prefer your approach, the limitation is that `PixelKit` doesn't have direct access to the `PrivacyConfig` library so we can't access the feature flagger directly. I think no matter what we'll have to inject a value in the initializer.

But: we can replace this with the adapter pattern and remove the closure, which solves this problem for all future feature flag checks added to the WideEvent code - we won't need to keep expanding the initializer for each check.

Take a look at `WideEventFeatureFlagProviding` and see what you think. Thanks!

File: iOS/DuckDuckGo/AppServices/WideEventService.swift [hunk 6]
```diff
@@ -277,3 +277,18 @@ final class WideEventService {
         }
     }
 }
+
+struct WideEventFeatureFlagAdapter: WideEventFeatureFlagProviding {
+    private let featureFlagger: FeatureFlagger
+
+    init(featureFlagger: FeatureFlagger) {
+        self.featureFlagger = featureFlagger
```
samsymons, line 97: We now inject the feature flagger indirectly, allowing us to add new feature flags without modifying the `WideEvent` initializer any further. This is the same approach taken by some other features, like `SubscriptionPageFeatureFlagAdapter`.

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
hanyutang-sandra: Thanks for updating it. Given the limitations I think the current solution makes sense to me

PR #2919, opened 2025-12-15
[macOS] Fix leak in interactionEventsPublisher on PopupHandlingTabExtension

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207340338530322/task/1212423256533717?focus=true
Tech Design URL:
CC:

### Description
The featureFlagger stored property inside the closure given that it doesn’t have a capture list, so Swift implicitly captures self strongly to access that property. So featureFlagger is actually self.featureFlagger.

### Testing Steps
1. Check the app works normally

### Impact and Risks

#### What could go wrong?
Nothing specific.

### Quality Considerations
Nothing specific.

### Notes to Reviewer
Nothing specific.

---
###### Internal references:

---

> [!NOTE]
> Weakly capture and unwrap `featureFlagger` in the interaction events filter to prevent a strong self capture/leak.
> 
> - **macOS popup handling**:
>   - Weakly capture `featureFlagger` in `interactionEventsPublisher.filter` and unwrap with `guard let featureFlagger` to avoid retaining `self`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f0e94b7eb479ee43f87c8ec9e2f13282d2e24066. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/Tab/TabExtensions/PopupHandlingTabExtension.swift
+++ b/macOS/DuckDuckGo/Tab/TabExtensions/PopupHandlingTabExtension.swift
@@ -99,9 +99,10 @@ final class PopupHandlingTabExtension {
         self.isInPopUpWindow = isInPopUpWindow
 
         interactionEventsPublisher
-            .filter { [featureFlagger] event in
+            .filter { [weak featureFlagger] event in
                 Logger.navigation.debug("PopupHandlingTabExtension.interactionEventsPublisher.filter: event: \(String(describing: event))")
-                guard featureFlagger.isFeatureOn(.popupBlocking),
+                guard let featureFlagger,
+                      featureFlagger.isFeatureOn(.popupBlocking),
                       featureFlagger.isFeatureOn(.extendedUserInitiatedPopupTimeout) else { return false }
 
                 switch event {


macOS/DuckDuckGo/Tab/TabExtensions/PopupHandlingTabExtension.swift
```diff
@@ -99,7 +99,7 @@ final class PopupHandlingTabExtension {
         self.isInPopUpWindow = isInPopUpWindow
 
         interactionEventsPublisher
-            .filter { event in
+            .filter { [featureFlagger] event in
```
ayoy: @jotaemepereira shouldn't you use `[weak featureFlagger]` to capture it weakly?

But also, feature flagger is an AppDelegate-owned instance and I think it shouldn’t affect the memory graph if it’s held strongly by a tab extension (it doesn’t extend a tab or tab extension’s lifetime this way). Happy to merge this when it’s turned into  a weak reference :)

jotaemepereira: @ayoy You're right, since featureFlagger is app-scoped and doesn't reference the extension, capturing it strongly wouldn't create a retain cycle or extend the extension's lifetime. The issue was the implicit capture of self through self.featureFlagger, which I think makes a cycle.

That said, I've updated it to [weak featureFlagger]. I might be wrong, but I thought that if you did [featureFlagger], it would capture only featureFlagger instead of self, if you used it in the closure.

Let me know your thoughts.

ayoy: > I thought that if you did [featureFlagger], it would capture only featureFlagger instead of self, if you used it in the closure.

Ah I see what you meant. I believe the `featureFlagger` referenced in that context is the initializer parameter and not the property on self, but it doesn’t hurt capturing it weakly just in case 👍

macOS/DuckDuckGo/Tab/TabExtensions/PopupHandlingTabExtension.swift
```diff
@@ -99,9 +99,10 @@ final class PopupHandlingTabExtension {
         self.isInPopUpWindow = isInPopUpWindow
 
         interactionEventsPublisher
-            .filter { event in
+            .filter { [weak featureFlagger] event in
```
mallexxx, line 102: I think it‘s not needed here, since the feature flagger is a dependency, plus it's an app lifetime object, so it makes no point to make it weak

PR #3057, opened 2026-01-05
[iOS] Update the VPN to support the embedded Privacy Config

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1212663544984489?focus=true
Tech Design URL:
CC:

### Description

This PR updates the iOS VPN to support reading the embedded Privacy Config. The embedded config is used when the app fails to fetch one remotely. The embedded config is the exact same one used by the app itself, so the flag state will be in sync.

### Testing Steps

1. Override `public var currentConfig: Data` in `VPNPrivacyConfigurationManager` to always return the embedded config
2. Run the VPN and check that it doesn't crash (this will happen if it can't read the config, because we force unwrap it)

### Impact and Risks

High, could crash the VPN

#### What could go wrong?

The embedded config is force unwrapped, so if it isn't accessible it will cause the VPN to crash. The embedded config logic already has unit tests around it, so this is mitigated that way.

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> Aligns the VPN with the app’s embedded privacy configuration and enables proper ETag handling.
> 
> - `VPNPrivacyConfigurationManager` now sources `embeddedConfigData` from `AppPrivacyConfigurationDataProvider().embeddedData` instead of hardcoded JSON; continues to fall back to embedded when no fetched config
> - `ConfigurationStore.loadEmbeddedEtag` returns `AppPrivacyConfigurationDataProvider.Constants.embeddedDataETag` for `privacyConfiguration`
> - `AppPrivacyConfigurationDataProvider.loadEmbeddedAsData()` switches from force-unwrapped read to `do/try/catch` with `fatalError`; added public initializer
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit df71dae7defb662655cf3c93fc24169bc73bd818. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iOS/Core/AppPrivacyConfigurationDataProvider.swift
+++ b/iOS/Core/AppPrivacyConfigurationDataProvider.swift
@@ -44,8 +44,11 @@ final public class AppPrivacyConfigurationDataProvider: EmbeddedDataProvider {
     }
 
     static func loadEmbeddedAsData() -> Data {
-        let json = try? Data(contentsOf: embeddedUrl)
-        return json!
+        do {
+            return try Data(contentsOf: embeddedUrl)
+        } catch {
+            fatalError("Failed to load embedded privacy config: \(error.localizedDescription)")
+        }
     }
 
     public init() {}


iOS/Core/AppPrivacyConfigurationDataProvider.swift
```diff
@@ -47,4 +47,6 @@ final public class AppPrivacyConfigurationDataProvider: EmbeddedDataProvider {
         let json = try? Data(contentsOf: embeddedUrl)
         return json!
```
diegoreymendez: I know this isn't introduced here, but this code worries me a bit and I believe is the unwrap you pointed at.  I don't want to do crazy changes but could we, as part of this PR:

Log the error.
Consider calling fatalError with a clear message?  We'll still get the line this way, but a clear message could make it easier to track back issues to this snippet of code as the code changes across versions.

For clarity: I don't think my suggestions are a huge win, but just some defensive programming in case things ever go awry.

Something like this:

```
static func loadEmbeddedAsData() -> Data {
    do {
        return try Data(contentsOf: embeddedUrl)
    } catch {
        Logger.general.error("Failed to load embedded privacy config from \(embeddedUrl): \(error)")
        fatalError("Failed to load embedded privacy config: \(error)")
    }
}
```

samsymons: Agree, I'm happy to make that change here since it doesn't change any functionality but does gain us a bit of clarity with debugging - I'll push that before merging. Thanks!

PR #3097, opened 2026-01-08
[macOS] Enable new Next Steps card provider with feature flag (internal only)

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1212706738200505?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1212359353583683?focus=true

### Description

This enables the new Next Steps card provider from https://github.com/duckduckgo/apple-browsers/pull/3082 when the `nextStepsSingleCardIteration` feature flag is enabled.

Changing the feature flag does not change the Next Steps UI yet, but it will switch between the legacy and new card providers for the existing Next Steps widget on the New Tab Page.

### Testing Steps

1. Build and run the app.
2. Go to Debug menu → Feature Flag Overrides → Other and confirm that `nextStepsSingleCardIteration` is disabled (default).
3. Go to Debug menu → New Tab Page → Reset Next Steps
4. Relaunch the browser & open the new tab page.
5. Dismiss some of the cards.
6. Go to Debug menu → Feature Flag Overrides → Other and enable `nextStepsSingleCardIteration`.
7. Confirm the new tab page is refreshed and you can interact with the cards:
   * Any cards dismissed when the feature flag was disabled are still dismissed.
   * You can dismiss cards.
   * You can click the call to action and get the expected action.
6. Go to Debug menu → Feature Flag Overrides → Other and disable `nextStepsSingleCardIteration`. This will return you to the legacy (production) card provider. (Any cards dismissed while the feature flag was enabled will still be available, although if you actioned any cards you may not be eligible to see them anymore.)

### Impact and Risks

Medium: Could disrupt specific features or user flows

#### What could go wrong?

The wrong card provider could be used when the feature flag is disabled. There are unit tests in place to ensure the correct (active) card provider is used according to the feature flag setting.

### Notes to Reviewer

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> Enables feature-flagged switching of Next Steps cards provider and integrates it across New Tab Page setup.
> 
[...]

PR #3114, opened 2026-01-09
[macOS] Add support for new Next Steps cards

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1212718005054384?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1212359353583683?focus=true

### Description

This adds support for two new cards for the new Next Steps iteration: Personalize Your Browser and Sync. These cards are added to the new card provider (`NewTabPageNextStepsSingleCardProvider`):

* Personalize Your Browser:
   * Shown when the New Tab Page Customize panel hasn't been opened. (This requires a new persisted setting for whether it has been opened.)
   * Card action opens the Customize panel.
* Sync:
   * Shown when sync is not connected.
   * Card action launches the sync flow.

These new cards are not used with the production Next Steps experience (which relies on `NewTabPageNextStepsCardsProvider`). 

### Testing Steps

The new card provider is not enabled yet in this branch. (These cards will not work with Next Steps until the FE implementation is integrated.) Confirm the changes make sense and the production experience works as expected:

1. Build and run the app.
2. Go to Debug menu → Feature Flag Overrides → Other and confirm that `nextStepsSingleCardIteration` is disabled (default).
3. Go to Debug menu → New Tab Page → Reset Next Steps
4. Relaunch the browser & open the new tab page.
5. Confirm the Next Steps section appears with the available cards (no new cards).

### Impact and Risks

#### What could go wrong?

Passing new card types to the existing Next Steps widget can break the widget (causing no cards to be loaded). To avoid this going wrong:

* The existing card provider assembles the card list by mapping its expected features to the card types, and it does not have any cases that will return one of the new cards.
* Manual testing confirms the widget continues to work as expected with these new card types added to the app.
* A change is being made on the frontend to prevent breakage in the event that a new card type is passed to the current Next Steps widget. (The new card provider will be used with a new frontend widget.)

### Notes to Reviewer

These new cards will be fully manually tested once the FE implementation is ready to be tested and integrated. (The existing Next Steps widget doesn't currently work with new card IDs, and this branch isn't based on https://github.com/duckduckgo/apple-browsers/pull/3097 so enabling the feature flag will not enable these cards for manual testing.)

---
###### Internal references:

---

> [!NOTE]
> Introduces two new Next Steps cards for the new single-card flow and integrates required dependencies and telemetry.
> 
> - Adds `personalizeBrowser` and `sync` to `NewTabPageDataModel.CardID`; filters them out of legacy provider mapping
> - Implements actions: opens customization settings and starts device sync flow (`NewTabPageNextStepsCardsActionHandler`), using `NewTabPageNavigator` and optional `SyncDeviceFlowLaunching`
> - Updates `NewTabPageNextStepsSingleCardProvider` to surface new cards based on conditions: customization panel not opened (`AppearancePreferences.didOpenCustomizationSettings`) and sync inactive with flags enabled; respects legacy dismissal and persistence
> - Persists and marks customization-opened state (`AppearancePreferences` + UserDefaults key) and triggers when opening settings
> - Wires `syncService` through `AppDelegate`, `NewTabPageCoordinator`, `NewTabPageActionsManager`, and provider facade
> - Adds pixels for new cards (clicked/dismissed/shown) and a new sync touchpoint `nextStepsCard`
> - Adds unit tests for new actions, visibility, and provider behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ea7367bc10cf0016fe8ef51af01dcfa90025c04d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
```diff
@@ -198,6 +212,8 @@ private extension NewTabPageNextStepsSingleCardProvider {
             dismissedLegacySetting = !legacyPersistor.shouldShowImportSetting
         case .subscription:
             dismissedLegacySetting = !legacySubscriptionCardPersistor.shouldShowSubscriptionSetting
+        default:
```
alessandroboron, line 209: I suggest to explicitely use `.personalizeBrowser` and `.sync` instead of `default` so if in the future we add new cases the compiler wil compain and there’s no risk to miss this.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsCardsActionHandler.swift
```diff
@@ -28,6 +29,9 @@ protocol NewTabPageNextStepsCardsActionHandling {
     ///   - card: The identifier of the card for which to perform the action.
     ///   - completion: A closure to be called upon completion of the action, if needed to refresh the cards state.
     @MainActor func performAction(for card: NewTabPageDataModel.CardID, refreshCardsAction: (() -> Void)?)
+
+    /// Sets the sync launcher to use when performing sync actions.
+    func setSyncLauncher(_ syncLauncher: SyncDeviceFlowLaunching)
```
alessandroboron: Any particular reason why having both init injection AND a setter? I think we should stick to one pattern as having both makes lifecycle management and ownership unclear.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
```diff
@@ -234,8 +254,27 @@ private extension NewTabPageNextStepsSingleCardProvider {
...
             .store(in: &cancellables)
     }
+
+    /// Determines the visibility of the Sync feature card based on the sync service state.
+    ///
+    /// This also instantiates the device sync launcher and sets it in the card action handler if needed.
+    /// This is done at the point where the visibility is checked, and not passed as a dependency (except for testing),
+    /// to give time for the core Sync dependencies to be ready.
+    @MainActor
+    private var syncFeatureVisibility: SyncFeatureVisibility {
```
alessandroboron: I may miss something but couldn’t we change this to:

```
private var shouldShowSyncCard: Bool {
    if syncService?.featureFlags.contains(.all) == true,
       case .inactive = syncService?.authState {
       return true
    } else {
        return false
    }
}
```

Looking at the implementation of `SyncFeatureVisibility` there’s no need for `syncLauncher` when computing `shouldShowSyncFeature`. The launcher is only needed when the user actually clicks the sync card, not when determining whether to show it. 

The above implementation has multiple benefits:

1. No side effects within the computed property.
2. No need to re-call `cardActionHandler.setSyncLauncher(deviceSyncLauncher)` every time cards are refreshed.
3. We can drop all the `@MainActor` annotations and the `Task { @MainActor in }` wrappers in the sink closures that were only needed because `DeviceSyncCoordinator` has a `@MainActor init`.

For creating the launcher when actually needed, we could either:
- Inject `DeviceSyncCoordinator` into `cardActionHandler` at initialization time (if sync is already ready)
- Use a factory pattern to create it on-demand when the card action is triggered (maintains lazy initialization if that's truly needed and testability)

Let me know your thoughts! :)

rachelmcr: > Looking at the implementation of SyncFeatureVisibility there’s no need for syncLauncher when computing shouldShowSyncFeature. The launcher is only needed when the user actually clicks the sync card, not when determining whether to show it.

Yeah, I went back and forth a little bit about this. I originally created `DeviceSyncCoordinator` on demand when the card action was triggered, but changed it in 417f5b6 because it uses a failable initializer. If the sync dependencies aren't available (which they aren't when the class is initialized, so it can't be injected directly) the card could be visible but the action would silently fail.

I thought it would be safer to check that `DeviceSyncCoordinator` can be initialized when deciding whether to show the card, which I believe is the reason for having `SyncFeatureVisibility`.

That said, this is more for robustness than a real need at this point. When I was testing this out, sync was always ready by the time the cards were shown (and definitely by the time the action was taking place), and we do have quite a few places in the code where we call `DeviceSyncCoordinator()?.startDeviceSyncFlow` directly with the assumption it will work. So I could use a visibility check like you suggested and go back to something like this to perform the action (with an injected `syncLauncher` for use in tests):

```
func performSyncAction(completion: (() -> Void)?) {
        guard let syncLauncher = syncLauncher ?? DeviceSyncCoordinator() else {
                // Log error and return
        }
        syncLauncher.startDeviceSyncFlow(source: .nextStepsCard, completion: completion)
}
```

I'll double check the sync setup to make sure I'm not missing any scenarios where sync wouldn't be available, and if it seems reasonable safe I'll go back to that approach to simplify things.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
rachelmcr: I used `default` because in this context we're specifically only using it to handle cases that existed for the legacy card provider (so we don't show the card if it was dismissed in the legacy experience). Any new cases won't have a legacy setting, so they can be skipped by default. 

That said, I generally agree about using explicit cases in switch statements so I don't have a strong opinion about not doing that.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsCardsActionHandler.swift
rachelmcr: No particular reason; I included init injection mainly for ease of testing but that's a fair point. Based on [this discussion](https://github.com/duckduckgo/apple-browsers/pull/3114#discussion_r2684740194) I'll likely remove this setter entirely.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
rachelmcr: > If the sync dependencies aren't available (which they aren't when the class is initialized, so it can't be injected directly)

Actually it seems I was mistaken about this; the dependencies are ready at this point, so injecting DeviceSyncCoordinator directly in the action handler should be fine. I'll make that change.

rachelmcr:
@alessandroboron This is ready for another look. I left the late initialization for `DeviceSyncCoordinator()` when the action is performed (with direct injection as an option for testing), because if we inject it into cardActionHandler at initialization time it requires much more extensive changes for testability due to side effects on the actions that use `Tab`. This way is simpler and safer given that it waits until we need those sync dependencies to be ready.

PR #3358, opened 2026-01-29
[iOS/macOS] Update check for uncommitted .strings changes

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203301625297703/task/1213027435336043?focus=true

### Description

The "Translated Strings" PR check currently uses a script to check for any string changes that aren't extracted to .strings or .xcstrings files. However, this script does not catch the scenario where an existing string in a .strings file is updated and the updated .strings file is not committed after the app builds.

This PR updates the check to change how we check .strings files:

* Instead of using the script for checking **.strings** files:
   * After the build step finishes, use `git diff` to check for uncommitted changes to .strings files and save the output.
   * Use the output in the translated strings check to fail the check and notify of .strings files that need to be committed.
* No changes for checking **.xcstrings** files: String catalogs can't be checked after a build on the command line (.xcstrings files aren't updated) so we keep using the script. The script does catch if string catalogs don't include the changes from existing, updated strings, so the issue with the current .strings file check doesn't occur here.

### Testing Steps

1. Confirm that updating an existing string without committing the .strings file fails the check. Example failure: https://github.com/duckduckgo/apple-browsers/actions/runs/21486862730/job/61902031366?pr=3353
```
⚠️ Uncommitted changes to .strings files detected after build. Please commit these files:
iOS/DuckDuckGo/en.lproj/Localizable.strings
iOS/Widgets/en.lproj/Localizable.strings
```
2. Confirm that updating an existing string without committing the .xcstrings files continue to fail the check. Example failure: https://github.com/duckduckgo/apple-browsers/actions/runs/21486862791/job/61902882147?pr=3353
```
❌ Modified strings not updated in .xcstrings files:
   • SharedPackages/AIChat/Sources/AIChat/iOS/UserText.swift
     Key: aichat.download.show
   • macOS/DuckDuckGo/Common/Localizables/UserText.swift
     Key: about.app_name

💡 Please build the app and commit the corresponding .xcstrings files.
```

### Impact and Risks

### Notes to Reviewer

1. We need both .strings and .xcstrings checks for iOS; .strings files are mostly used in the project while string catalogs are used in some shared packages.
2. We currently only need the .xcstrings check for macOS. macOS does have a local package (`NetworkProtectionMac`) using .strings files but there isn't automated string extraction for it. I've still added the check so we're covered if we add this in the future.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes CI gating logic and job dependencies; mis-detection or a skipped `release-build` could incorrectly fail or skip translation checks and block merges.
> 
> **Overview**
> The iOS and macOS PR check workflows now capture any post-build diffs to `**/*.strings` during `release-build` (via `git diff --name-only`) and pass that list as a job output.
> 
> The `translation-check` job now depends on `release-build` and fails early when the build produced uncommitted `.strings` changes, while still running the existing Python verification for `.xcstrings`.
> 
> Also tweaks one iOS localized string value (`"in the last 7 days"` → `"in last 7 days"`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d04df5c21b69e3fb0ac80bdb29e0df2dc66beefe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/scripts/translation-pr-checks/verify_string_extraction.py
+++ b/scripts/translation-pr-checks/verify_string_extraction.py
@@ -3,7 +3,7 @@
 String Extraction Verification Script
 
 Verifies that NSLocalizedString calls in source code have been extracted to
-.xcstrings files.
+string files (.xcstrings, .strings).
 
 Usage:
     python3 verify_string_extraction.py --platform iOS
@@ -24,7 +24,9 @@ from typing import Dict, List, Optional, Set, Tuple
 from localization_utils import (
     get_base_branch,
     get_changed_files,
+    get_files_content_at_base,
     get_search_paths,
+    parse_strings_file,
     parse_xcstrings,
 )
 
@@ -162,13 +164,17 @@ def extract_string_value_for_key(content: str, key: str) -> Optional[str]:
 
 def find_string_files(paths: List[str]) -> Dict[str, List[str]]:
     """
-    Find .xcstrings files in the specified paths.
+    Find English string files in the specified paths.
+
+    For .xcstrings files: includes all (keys are language-agnostic)
+    For .strings files: only includes files in en.lproj directories
 
     Returns:
-        Dict with 'xcstrings' key mapping to list of file paths
+        Dict mapping file type (xcstrings, strings) to list of file paths
     """
     files = {
-        'xcstrings': []
+        'xcstrings': [],
+        'strings': []
     }
 
     for search_path in paths:
@@ -177,9 +183,15 @@ def find_string_files(paths: List[str]) -> Dict[str, List[str]]:
 
         for root, _, filenames in os.walk(search_path):
             for filename in filenames:
+                file_path = os.path.join(root, filename)
                 if filename.endswith('.xcstrings'):
-                    file_path = os.path.join(root, filename)
+                    # .xcstrings files contain all locales, keys are language-agnostic
                     files['xcstrings'].append(file_path)
+                elif filename.endswith('.strings'):
+                    # Only check English .strings files
+                    parent_dir = os.path.basename(root)
+                    if parent_dir == 'en.lproj':
+                        files['strings'].append(file_path)
 
     return files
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ios_pr_checks.yml
```diff
@@ -336,12 +339,20 @@ jobs:
...
+        echo "$changed_files" >> $GITHUB_OUTPUT
+        echo "EOF" >> $GITHUB_OUTPUT
+
   translation-check:
     name: Translated Strings
 
     runs-on: ubuntu-latest
 
-    needs: should-run-checks
+    needs: [should-run-checks, release-build]
```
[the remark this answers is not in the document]
rachelmcr: I believe this is okay. We skip the release-build job in these scenarios:
* Dependabot PRs
* Bumping the internal release

These shouldn't change strings, so the translated strings check shouldn't be needed. We're only concerned about the translation checks in PRs where developers add or update strings.

I'd rather skip the check entirely if we don't have all the dependencies, for clarity, than end up in a situation where only some of the steps are run and the check passes when it shouldn't.

.github/workflows/ios_pr_checks.yml
```diff
@@ -336,12 +339,20 @@ jobs:
         pixel-params: |
           runner: ${{ env.RUNS_ON }}
 
+    - name: Check for changed strings files
+      id: changed-strings-files
+      run: |
+        changed_files=$(git diff --name-only HEAD -- ':(glob)**/*.strings' || true)
+        echo "changed_files<<EOF" >> $GITHUB_OUTPUT
+        echo "$changed_files" >> $GITHUB_OUTPUT
+        echo "EOF" >> $GITHUB_OUTPUT
```
ayoy, line 349: Is this needed to preserve newline characters in `changed_files`? Could you include a comment explaining this?

Alternatively you could store base64-encoded value in the output, but I’m not strong on it if it’s used in this 1 scenario only:

```
changed_files_base64="$(git diff --name-only HEAD -- ':(glob)**/*.strings' || true)"
echo "$changed_files_base64" | base64 >> $GITHUB_OUTPUT
        
[…]

env:
  CHANGED_FILES_BASE64: ${{ needs.release-build.outputs.changed_files_base64 }}
run: |
  CHANGED_FILES="$(base64 -d <<< "$CHANGED_FILES_BASE64")"
```

rachelmcr: Yes, it's to preserve newline characters. I'm happy to switch to using a base64-encoded value if we'd rather standardize on that approach. Either way I'll add a comment to explain why it's needed.

ayoy: I default to using base64 encoding for fragile variables. Before DMG code signing moved to fastlane, we had 20+ GHA secrets for code signing, stored as base64 strings :) as long as this works here (and it does) adding a comment should be fine :)

scripts/translation-pr-checks/verify_string_extraction.py
```diff
@@ -207,22 +195,9 @@ def check_key_in_xcstrings(key: str, file_path: str) -> bool:
...
-    except FileNotFoundError:
-        return False
-
-    strings = parse_strings_file(content)
-    return key in strings
-
 def find_key_in_string_files(key: str, string_files: Dict[str, List[str]]) -> Optional[str]:
     """
-    Find which string file contains the given key.
+    Find which .xcstrings file contains the given key.
```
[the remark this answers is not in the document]
rachelmcr: I thought I accounted for this but I just checked and it's a real issue. I'll reinstate the .strings check in the script so we only fail that step if a new string is missing from both .strings and .xcstrings files.

PR #3373, opened 2026-01-30
[macOS] Prevent excessive recursion during crash processing

dus7:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201037661562251/task/1212836028823773?focus=true
Tech Design URL:
CC:

### Description

Applies solution from https://github.com/duckduckgo/apple-browsers/pull/3370 for macOS.

Two-way attempt to prevent crash during crash processing, caused by "Thread stack size exceeded due to excessive recursion":

* Turn off keys sorting, as this is most likely when the crash happens.
* Limit call stack depth to prevent excessive recursion (static limit set to 250)

### Testing Steps
1. Verify CI is green.

### Impact and Risks

#### What could go wrong?
* Issues while processing crashes without sorted keys.
* Limited data about crashes because the call stack is truncated

Both are mitigated by using failsafe feature flag for each solution.

### Notes to Reviewer

---

> [!NOTE]
> **Medium Risk**
> Changes crash-report processing behavior by default (disabling key sorting and optionally truncating call stacks), which could affect crash payload fidelity and crash-upload prompting paths despite being guarded by feature flags.
> 
> **Overview**
> Reduces the chance of stack-overflow during macOS App Store crash processing by **disabling crash payload key sorting** and **optionally limiting call stack tree depth** (default limit `250`).
> 
> Adds two new remote-configurable failsafe flags (plumbed via `PrivacyConfig`/`FeatureFlags`) and uses them in `AppDelegate` when starting `crashCollection.startAttachingCrashLogMessages(callStackDepthLimit:sortKeys:)`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 653819ba1aa216632b76125723246f2cbd725a98. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


macOS/LocalPackages/FeatureFlags/Sources/FeatureFlags/FeatureFlag.swift
```diff
@@ -569,6 +581,10 @@ extension FeatureFlag: FeatureFlagDescribing {
             return .disabled
         case .wideEventPostEndpoint:
             return .remoteReleasable(.subfeature(MacOSBrowserConfigSubfeature.wideEventPostEndpoint))
+        case .crashCollectionDisableKeysSorting:
+            return .remoteReleasable(.subfeature(MacOSBrowserConfigSubfeature.disableKeysSorting))
+        case .crashCollectionLimitCallStackTreeDepth:
+            return .remoteReleasable(.subfeature(MacOSBrowserConfigSubfeature.limitCallStackTreeDepth))
```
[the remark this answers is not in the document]
dus7: https://github.com/duckduckgo/apple-browsers/pull/3373/commits/2096d50ca85bc5a28d85c98c863e018827484ae9

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1184,7 +1184,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         applyPreferredTheme()
 
 #if APPSTORE
-        crashCollection.startAttachingCrashLogMessages { [weak self] pixelParameters, payloads, completion in
+        let sortKeys = !featureFlagger.isFeatureOn(.crashCollectionDisableKeysSorting)
+        let isCallStackLimitingEnabled = featureFlagger.isFeatureOn(.crashCollectionLimitCallStackTreeDepth)
+        let callStackDepthLimit: Int? = isCallStackLimitingEnabled ? 250 : nil
```
mallexxx, line 1189: Having the reports where the recursion overflows the stack at around 715 frames, with 6 calls per a nested object, it gives us the depth of around 120 nested objects;
Should we limit the depth to 100 or even less? Can sane call stacks contain such a number of nested objects at all?
Or maybe we should even make the limit adjustable with an optional config value, so if we keep receiving the crashes we could adjust the number on the fly?

Image

dus7: > Can sane call stacks contain such a number of nested objects at all?

No idea - probably not. I wanted to play on a “safe” side here to not loose any potential crash information, as this was a fix that went to internal right before the release. FWIW, this is not happening anymore on iOS after the fix was applied.

> Should we limit the depth to 100 or even less.

We could, but since the existing fix did the trick, maybe the limiting is not required at all. With 250 nested objects it’s ~2-3 calls per object for 715 frames (similar count was on iOS) and it looks like it’s within the limit for a non-main thread when the sorting is off.

I think the adjustable config value is a good idea, but not sure it’s required immediately.

PR #3390, opened 2026-02-02
App updater refactoring

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199230911884351/task/1213067106906950?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1199230911884351/task/1212999214064716?focus=true
https://app.asana.com/1/137249556945/project/481882893211075/task/1211728521308688
CC: @ayoy @diegoreymendez 

### Description

- 2nd PR refactoring Sparkle and App Store update controllers before moving AppUpdater to local package ([macOS/LocalPackages/AppUpdater](https://github.com/duckduckgo/apple-browsers/pull/3369))
- Eliminate conditional compilation (#if SPARKLE/#if APPSTORE) in package code by using runtime checks with ApplicationBuildType
- Convert SPARKLE_ALLOWS_UNSIGNED_UPDATES compile-time flag to ApplicationBuildType runtime checks (isDebugBuild/isReviewBuild)

#### Prepare for testing
1. Launch the app
2. Enable the `autoUpdateInDEBUG` flag (if testing automatic updates)
3. Go to Debug Menu > Updates > Set up Sparkle testing environment...
4. When the alert comes up asking you for a key, retrieve it from Bitwarden > Private key for testing Sparkle updates
5. Paste it in the dialog - it should take a bit to create a folder on your desktop and open a readme file
6. Follow the instructions from the readme file to start a local updates server
7. Once you have the feed URL (shown in the CLI utility), paste it in Debug Menu > Updates > Set custom feed URL...
8. Restart the app for the changes to take effect

#### Test: Automatic Updates
1. Make sure you're on automatic updates in App > Settings > About
2. Once you launch the app you should NOT see any update indicator immediately (indicator shows after 1 minute in debug builds, 2 days in release)
3. Go into DDG > Settings > About and verify there's an update ready to install with a CTA that reads "Restart to Update"
4. Click "Restart to Update" or quit manually, and verify the app restarts and updates

#### Test: Manual Updates
1. Make sure you're on manual updates in App > Settings > About
2. Once you launch the app you should see the update indicator immediately
3. Go into DDG > Settings > About and verify there's an update ready to install with a CTA that reads "Restart to Update"
4. Click "Restart to Update" or quit manually, and verify the app restarts and updates

#### Test: Package Tests
1. Run AppUpdater package tests: `cd macOS/LocalPackages/AppUpdater && swift test`
2. Verify all tests pass (174 tests)
3. Run main app unit tests that use AppUpdater: `xcodebuild test -scheme "macOS Browser" -only-testing:Unit\ Tests/ReinstallUserDetectionTests`
4. Verify ReinstallUserDetectionTests pass

### Impact and Risks

**Impact**: Medium - This is a significant refactoring that affects update functionality, but changes are isolated to the AppUpdater package and maintain backward compatibility.

#### What could go wrong?
- Update checks might fail if ApplicationBuildType is not properly initialized
- Custom feed URL functionality might not work in debug/review builds if buildType checks are incorrect
- Reinstall detection might not work if isSparkleBuild is not set correctly
- Package tests might not run in CI if workflow configuration is incorrect

**Mitigation**: All changes are behind existing feature flags and runtime checks. Package tests are added to CI to catch regressions early.

### Quality Considerations

- **Edge cases**: ApplicationBuildType checks handle debug/review builds correctly; reinstall detection properly checks for Sparkle builds
- **Performance**: No performance impact - runtime checks replace compile-time flags
- **Monitoring**: Existing update tracking (WideEvent) continues to work
- **Documentation**: Package structure is self-documenting with clear module separation
- **Privacy/Security**: No privacy/security impact - this is internal refactoring

### Notes to Reviewer
- Failing UI tests to be fixed in https://github.com/duckduckgo/apple-browsers/pull/3320
- Focus on verifying that conditional compilation has been properly eliminated in the package code
- Check that ApplicationBuildType runtime checks match the previous compile-time flag behavior
- Verify that all moved files maintain their original functionality
- Ensure package tests cover the same scenarios as before the move

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches core update-checking and release-notes flows across both Sparkle and App Store builds, so regressions could affect update availability, UI indicators, or restart/update behavior despite being largely a refactor.
> 
> **Overview**
> Refactors macOS update-controller initialization to use runtime `ApplicationBuildType` + `UpdateControllerFactory` instantiation, making `AppDelegate.updateController` optional and removing most `#if SPARKLE/#if APPSTORE` branching from app code paths.
> 
> Reworks release-notes integration to be Sparkle-only via new `ReleaseNotesUserScriptProvider` and a Sparkle-created navigation responder/user script (replacing the previous tab extension approach), and adjusts menu/about/update-state code to cast to `SparkleUpdateController` only when needed.
> 
> Hardens and modularizes update components: `AppStoreUpdateController` becomes publicly constructible via a factory extension with dependency injection, new shared models (`ReleaseMetadata`, `PendingUpdateInfo`) are introduced, legacy updater-related `GeneralPixel` cases are removed, and the SPM dependency `content-scope-scripts` is bumped to `13.8.0`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit daf922aea5f4e4c91fe4cb52193545f22c5bd577. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


macOS/DuckDuckGo/Updates/Sparkle/SimplifiedSparkleUpdateController.swift
```diff
@@ -263,27 +263,33 @@ final class SimplifiedSparkleUpdateController: NSObject, SparkleUpdateController
...
+                         featureFlagger: FeatureFlagger,
+                         eventMapping: EventMapping<UpdateControllerEvent>?,
+                         notificationPresenter: UpdateNotificationPresenting,
+                         keyValueStore: ThrowingKeyValueStoring,
+                         buildType: ApplicationBuildType?,
+                         wideEvent: WideEventManaging?) {
+        // Sparkle builds require these parameters
+        guard let buildType, let wideEvent else {
+            fatalError("SimplifiedSparkleUpdateController requires buildType and wideEvent")
+        }
```
diegoreymendez: The protocol signature declares `buildType` and `wideEvent` as optional, but this implementation crashes with fatalError when they're nil.

This seems to be because we're forcing AppStore and Sparkle to share the same initializer signature even though they have different requirements. I'd really like us to considering factory methods that let each implementation specify its own requirements.

Specifically, I think it would be better to avoid a shared initializer in the protocol.  As these implementations continue to diverge, they'll likely need even more different initialization parameters or flows.  Keeping initializers separate would let each implementation maintain a clear contract about what it actually requires.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -270,35 +290,45 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
...
+                  featureFlagger: FeatureFlagger,
+                  eventMapping: EventMapping<UpdateControllerEvent>?,
+                  notificationPresenter: any UpdateNotificationPresenting,
+                  keyValueStore: any Persistence.ThrowingKeyValueStoring,
+                  buildType: ApplicationBuildType?,
+                  wideEvent: WideEventManaging?) {
+        // Sparkle builds require these parameters
+        guard let wideEvent else {
+            fatalError("SparkleUpdateController requires wideEvent")
+        }
```
diegoreymendez: `wideEvent` should become non-optional rather than crash, or stay optional and allow being `nil`.

macOS/DuckDuckGo/Updates/AppStore/AppStoreUpdateController.swift
```diff
@@ -16,64 +16,86 @@
...
-         featureFlagger: FeatureFlagger = NSApp.delegateTyped.featureFlagger,
-         internalUserDecider: InternalUserDecider = NSApp.delegateTyped.internalUserDecider,
-         appStoreOpener: AppStoreOpener = DefaultAppStoreOpener()) {
-        self.updateCheckState = updateCheckState
+    /// Protocol-conforming initializer for production use.
+    public required init(internalUserDecider: InternalUserDecider,
+                         featureFlagger: FeatureFlagger,
+                         eventMapping: EventMapping<UpdateControllerEvent>?,
+                         notificationPresenter: any UpdateNotificationPresenting,
+                         keyValueStore: any Persistence.ThrowingKeyValueStoring,
```
diegoreymendez: Some of these parameters are not used in the AppStore controller.  Probably tied to my other comment that we're trying to make all initializers fit in a protocol, which takes away flexibility and makes it harder to maintain the different controllers for little gain.

macOS/DuckDuckGo/Updates/AppStore/AppStoreUpdateController.swift
```diff
@@ -16,64 +16,86 @@
...
+                         buildType: ApplicationBuildType?,
+                         wideEvent: WideEventManaging?) {
```
diegoreymendez: We don't need this parameter in the AppStore controller either.  This is a dependency that's only necessary in Sparkle.

macOS/DuckDuckGo/Updates/AppStore/AppStoreUpdateController.swift
```diff
@@ -16,64 +16,86 @@
...
-         internalUserDecider: InternalUserDecider = NSApp.delegateTyped.internalUserDecider,
-         appStoreOpener: AppStoreOpener = DefaultAppStoreOpener()) {
-        self.updateCheckState = updateCheckState
+    /// Protocol-conforming initializer for production use.
+    public required init(internalUserDecider: InternalUserDecider,
+                         featureFlagger: FeatureFlagger,
+                         eventMapping: EventMapping<UpdateControllerEvent>?,
+                         notificationPresenter: any UpdateNotificationPresenting,
+                         keyValueStore: any Persistence.ThrowingKeyValueStoring,
+                         buildType: ApplicationBuildType?,
```
diegoreymendez: AppStore doesn't need this param either.

diegoreymendez (commented):
Hey @mallexxx, I've taken a first pass looking at the code.

I have some feedback I'd like us to address before moving forward.

My concern is that we're coupling the initializers across all controllers through the shared protocol signature. This removes the flexibility for each controller to define its own initialization requirements. It also makes the method signatures less correct—parameters are marked optional to accommodate different implementations, but then we use assertions when they're actually required. As implementations diverge or new controller types are added, these constraints will make it harder to adapt individual controllers without affecting all of them.

Could we take a few steps back on this specific change?

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -1,7 +1,7 @@
 //
 //  UpdateController.swift
 //
-//  Copyright © 2021 DuckDuckGo. All rights reserved.
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
```
diegoreymendez: Copyright year shouldn't be updated.

macOS/DuckDuckGo/Updates/AppStore/AppStoreUpdateController.swift
```diff
@@ -1,7 +1,7 @@
 //
 //  AppStoreUpdateController.swift
 //
-//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
```
diegoreymendez: I noticed we're updating the copyright year in many files - we shouldn't.  Can you roll all of these back?

macOS/DuckDuckGo/Updates/Sparkle/UpdateWideEventData.swift
```diff
@@ -1,7 +1,7 @@
 //
 //  UpdateWideEventData.swift
 //
-//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
```
diegoreymendez: Just marking this too.

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -16,12 +16,116 @@
...
+              .updateApplicationFailure(let r1, let r2, let r3, let r4, let r5, let r6, let r7, let r8, let r9, let r10)):
+            return l1 == r1 && l2 == r2 && l3 == r3 && l4 == r4 && l5 == r5 && l6 == r6 && l7 == r7 && l8 == r8 && l9 == r9 && l10 == r10
+        case (.updateApplicationUnexpected(let l1, let l2, let l3),
+              .updateApplicationUnexpected(let r1, let r2, let r3)):
+            return l1 == r1 && l2 == r2 && l3 == r3
+        default:
+            return false
+        }
+    }
+}
```
diegoreymendez: Could we move back to just using `PixelKit` directly?

This enum seems to be necessary due to the introduction of `EventMapping` - but it's not clear to me it's giving us a ton of value.  The same applies to the code in `UpdateControllerMappings`.

We could entirely skip this if we added `PixelKit` as a dependency to the new package, and it would give us the following advantages:
- We'll have less risk in these changes.
- We'll have less code to maintain.
- The relevant pixels could be moved and scoped to the appropriate package.
- Less maintenance: adding a pixel is trivial.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1382,6 +1368,30 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         SyncDiagnosisHelper(syncService: syncService).diagnoseAccountStatus()
     }
 
+    @MainActor
+    private func initializeUpdateController() {
```
jotaemepereira, line 1420: Why is the reasoning for removing the compilation flags in favor of runtime checks here in the main application? I prefer that we rely on compilation flags rather than runtime objects we create.

macOS/DuckDuckGo/Common/Utilities/ApplicationBuildType.swift
```diff
@@ -16,43 +16,36 @@
 //  limitations under the License.
 //
 
-protocol ApplicationBuildType {
```
jotaemepereira, line 19: It seems that this protocol was moved to the Updates folder, it should live in the same folder as the implementation on the `/Common/Utilities/`

macOS/DuckDuckGo/DefaultBrowserAndAddToDockPrompts/DefaultBrowserAndDockPromptCoordinator.swift
```diff
@@ -97,12 +97,12 @@
 // - `isBannerPermanentlyDismissed`: User clicked "Never Ask Again"
 // - `debugSetDefaultAndAddToDockPromptCurrentDateKey`: Simulated date override (debug only)
 
-import Combine
-import SwiftUI
-import SwiftUIExtensions
 import BrowserServicesKit
+import Combine
```
jotaemepereira, line 100: The changes introduced in this PR are substantial; let’s try to avoid changes caused by moving imports around.

macOS/DuckDuckGo/NavigationBar/View/MoreOptionsMenu.swift
```diff
@@ -496,25 +493,25 @@ final class MoreOptionsMenu: NSMenu, NSMenuDelegate {
...
         updateMenuItem = menuItem
         addItem(menuItem)
 
-        #if SPARKLE
+#if SPARKLE
         if let releaseNotes = NSApp.mainMenuTyped.releaseNotesMenuItem.copy() as? NSMenuItem {
             addItem(releaseNotes)
         }
-        #endif
+#endif
```
jotaemepereira: The same goes for these changes to remove spacing around compilation flags. Let’s keep the PR focused on the core changes; let’s avoid extra noise.

macOS/DuckDuckGo/NetworkProtection/AppTargets/BothAppTargets/NetworkProtectionDebugMenu.swift
```diff
@@ -236,11 +236,7 @@ final class NetworkProtectionDebugMenu: NSMenu {
     ///
     @objc func disableLoginItem(_ sender: Any?) {
         Task { @MainActor in
-            do {
-                try await debugUtilities.disableLoginItems()
-            } catch {
-                await NSAlert(error: error).runModal()
-            }
+            debugUtilities.disableLoginItems()
```
jotaemepereira: How is this change related to the PR goal? If it is not, let’s remove it from here.

macOS/DuckDuckGo/Preferences/Model/AboutPreferences.swift
```diff
@@ -193,8 +189,7 @@ final class AboutPreferences: ObservableObject, PreferencesTabOpening {
 
 #if SPARKLE
     private var isAtRestartCheckpoint: Bool {
-        guard let updateController = updateController as? any SparkleUpdateControllerProtocol else { return false }
```
jotaemepereira, line 196: I’m confused here, so now that we use the new controller, should we still use the compilation flag for the `isAtRestartCheckpoint` var?

macOS/DuckDuckGo/Preferences/Model/AppearancePreferencesMock.swift
```diff
@@ -25,8 +25,7 @@ import PrivacyConfigTestsUtils
 
 extension AppearancePreferences {
     static let mock = AppearancePreferences(
-        // swiftlint:disable:next force_try
-        persistor: AppearancePreferencesUserDefaultsPersistor(keyValueStore: try! MockKeyValueFileStore()),
+        persistor: AppearancePreferencesUserDefaultsPersistor(keyValueStore: MockKeyValueFileStore()),
```
jotaemepereira: This seems like another unintended change. Let’s keep the PR focused on the main goal.

macOS/DuckDuckGo/Statistics/ATB/ReinstallingUserDetecting.swift
```diff
@@ -16,6 +16,7 @@
 //  limitations under the License.
 //
 
+import Common
```
jotaemepereira: This seems not needed, let’s remove it.

macOS/DuckDuckGo/Tab/UserScripts/UserScripts.swift
```diff
@@ -52,9 +52,7 @@ final class UserScripts: UserScriptsProvider {
     let youtubePlayerUserScript: YoutubePlayerUserScript?
     let specialErrorPageUserScript: SpecialErrorPageUserScript?
     let onboardingUserScript: OnboardingUserScript?
-#if SPARKLE
-    let releaseNotesUserScript: ReleaseNotesUserScript?
-#endif
+    let releaseNotesUserScript: Subfeature? /*ReleaseNotesUserScript*/
```
jotaemepereira, line 54: I’m really confused about the use of the compilation flags. I’m seeing some places where we still use them, but not in others.

macOS/DuckDuckGo/Tab/TabExtensions/TabExtensions.swift
```diff
@@ -334,15 +335,18 @@ extension TabExtensionsBuilder {
...
 
-#if SPARKLE
-        add {
-            ReleaseNotesTabExtension(scriptsPublisher: userScripts.compactMap { $0 }, webViewPublisher: args.webViewFuture)
-        }
-#else
-        add {
-            ReleaseNotesTabExtension()
+        // Release notes tab extension - only available for Sparkle builds
+        if let updateController = dependencies.updateController {
```
jotaemepereira: This reads much more difficult than the old code. With the compilation flags, it is easier to understand that it was used for Sparkle only; we need to depend on a comment and unwrapping a controller.

macOS/DuckDuckGo/Tab/UserScripts/UserScripts.swift
```diff
@@ -168,9 +166,18 @@ final class UserScripts: UserScriptsProvider {
             youtubePlayerUserScript = nil
         }
 
-#if SPARKLE
-        releaseNotesUserScript = ReleaseNotesUserScript(keyValueStore: UserDefaults.standard)
-#endif
+        // Release notes user script - only available for Sparkle builds
```
jotaemepereira, line 169: The same comment here. The compilation flags code were easier to read, and we had the advantages of being sure that code was not part of either Sparkle of AppStore, now we depend on a object that we are creating, which at the same time is adding more complexity to read the code.

macOS/DuckDuckGo/Updates/AppStore/AppStoreOpener.swift
```diff
@@ -16,11 +16,11 @@
 //  limitations under the License.
 //
 
-import Foundation
 import AppKit
+import Foundation
 
 /// Protocol for opening the App Store
-protocol AppStoreOpener {
+public protocol AppStoreOpener {
```
jotaemepereira: ❓ This is going to be public given that it will be move to the package, right?

macOS/DuckDuckGo/Updates/AppStore/AppStoreUpdateController.swift
```diff
@@ -138,10 +175,20 @@ final class AppStoreUpdateController: NSObject, UpdateController {
     }
 
     /// For App Store builds given that we cannot run an update. We just check for a new by going to the App Store.
-    func runUpdate() {
+    public func runUpdate() {
         openUpdatesPage()
     }
 
+    // MARK: - Sparkle-only methods - no-op for App Store builds.
```
jotaemepereira: 🔴 This makes the class harder to understand. Previously, we had a protocol with operations that both Sparkle and AppStore needed; now we have only one, and concrete classes need to adhere even though they do not need the operations.

`UpdateController` protocol should only contain the shared contract.

macOS/DuckDuckGo/Updates/Sparkle/ApplicationUpdateDetector.swift
```diff
@@ -45,6 +40,18 @@ final class ApplicationUpdateDetector {
         set { try? settings.set(newValue, for: \.previousBuild) }
     }
 
+    public static func isApplicationUpdated(currentVersion: String? = nil,
```
jotaemepereira: Why do we need a static function now for this? Can the ApplicationUpdateDetector live somewhere and use the non-static isApplicationUpdated function instead of creating a new detector every time from the static function?

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
```diff
@@ -175,10 +192,6 @@ final class SparkleUpdateController: NSObject, SparkleUpdateControllerProtocol {
     var lastUpdateCheckDate: Date? { updater?.lastUpdateCheckDate }
     var lastUpdateNotificationShownDate: Date = .distantPast
 
-    private var shouldShowUpdateNotification: Bool {
```
jotaemepereira, line 178: Why was this deleted?

macOS/DuckDuckGo/Updates/UpdateController.swift
mallexxx: Good call; this required moving the pixel definitions from GeneralPixel but overall made it cleaner

macOS/DuckDuckGo/Updates/Sparkle/SimplifiedSparkleUpdateController.swift
```diff
@@ -146,21 +146,21 @@ final class SimplifiedSparkleUpdateController: NSObject, SparkleUpdateController
...
+    public var lastUpdateNotificationShownDate: Date = .distantPast
 
-#if SPARKLE_ALLOWS_UNSIGNED_UPDATES
     private var customFeedURL: String? {
-        get { try? settings.debugSparkleCustomFeedURL }
-        set { try? settings.set(newValue, for: \.debugSparkleCustomFeedURL) }
-    }
-#endif
-
-    private var shouldShowUpdateNotification: Bool {
```
jotaemepereira, line 159: Why was this removed?

macOS/DuckDuckGo/Updates/ApplicationBuildTypeProtocol.swift
```diff
@@ -0,0 +1,27 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// Protocol for determining application build type characteristics.
+public protocol ApplicationBuildType {
```
jotaemepereira: As I mentioned before, this should live in Common/Utilities

macOS/DuckDuckGo/Updates/UpdateController.swift
```diff
@@ -16,12 +16,116 @@
...
-import AppKit
+import Common
+import FeatureFlags
+import Foundation
+import Persistence
 import PixelKit
+import PrivacyConfig
+import Subscription
+
+public enum UpdateControllerEvent: Equatable {
```
jotaemepereira: Why is this needed for the pixels? If I recall correctly, we didn’t have this before. Do we need to adhere to == every time we change or add a new pixel? Let’s use PixelKit as a dependency of the package so we do not need to do this kind of thing.

macOS/DuckDuckGo/InfoPlist.xcstrings
```diff
@@ -63,7 +63,7 @@
     },
     "NSAppDataUsageDescription" : {
       "comment" : "Privacy - Other Application Data Usage Description",
-      "extractionState" : "stale",
+      "extractionState" : "extracted_with_value",
```
jotaemepereira, line 66: This seems an unintended change

macOS/LocalPackages/NewTabPage/Tests/NewTabPageTests/UserDefaultsNewTabPageProtectionsReportSettingsPersistorTests.swift
```diff
@@ -26,7 +26,7 @@ final class UserDefaultsNewTabPageProtectionsReportSettingsPersistorTests: XCTes
 
     override func setUp() async throws {
         try await super.setUp()
-        keyValueStore = try MockKeyValueFileStore()
+        keyValueStore = MockKeyValueFileStore()
```
jotaemepereira: Similar comment as before, let’s try to keep the PR focused on its goal.

macOS/DuckDuckGo/Updates/UpdateController.swift
mallexxx: updated

macOS/DuckDuckGo/Application/AppDelegate.swift
mallexxx: Compilation flags are being removed as they are unfortunately incompatible with SPM in scope of moving towards complete modularization (https://app.asana.com/1/137249556945/project/481882893211075/task/1211728521308688)

macOS/DuckDuckGo/Updates/AppStore/AppStoreOpener.swift
mallexxx: actually that shouldn‘t be exposed

macOS/DuckDuckGo/Tab/UserScripts/UserScripts.swift
mallexxx: the compilation flags are to be removed, this PR isn‘t aimed to complete it yet.

macOS/DuckDuckGo/Updates/Sparkle/SparkleUpdateController.swift
mallexxx: This was an unused leftover after #3483: the check is now made in `UpdateController.isUpdateNotificationAllowed`

macOS/DuckDuckGo/Updates/Sparkle/ApplicationUpdateDetector.swift
mallexxx: leftover, removed

macOS/DuckDuckGo/Tab/TabExtensions/TabExtensions.swift
mallexxx: Updated the conditional instantiation for better clarity: moved factory to `SparkleUpdateController` protocol
In fact, the `ReleaseNotesTabExtension` doesn't need to be a `TabExtension`, it only needs to conform to `NavigationResponder`, so I‘ve dropped all the unneded boilerplate and conditionals around it so it could be fully moved to the SparkleAppUpdater package

macOS/DuckDuckGo/Updates/Sparkle/SimplifiedSparkleUpdateController.swift
```diff
@@ -446,15 +454,14 @@ final class SimplifiedSparkleUpdateController: NSObject, SparkleUpdateController
...
     }
 
-    func openUpdatesPage() {
-        DispatchQueue.main.async {
-            Application.appDelegate.windowControllersManager.showTab(with: .releaseNotes)
-        }
+    public func openUpdatesPage() {
+        // Empty method kept for protocol conformance.
+        // Opening release notes page is implemented via UpdateNotificationPresenting in the app layer.
     }
```
diegoreymendez, line 467: Making this a no-op has two issues:

1. We keep a method with a concrete semantic purpose ("open the updates page") that silently does nothing. That's confusing for future readers and fragile if someone adds a new call site expecting it to work.
2. There are three existing call sites that now silently break:
   - `SparkleUpdateController.swift:569` — non-legacy path in runUpdateFromMenuItem()
   - `SparkleUpdateController.swift:581` — legacy path, force-update-check case
   - `SimplifiedSparkleUpdateController.swift:520` — simplified runUpdateFromMenuItem()

Instead of moving the logic to the app layer via UpdateNotificationPresenting and leaving a no-op behind, could we inject an openUpdatesPage callback (like we already do with `isOnboardingFinished`) and call through to it? That way the controllers still perform the navigation without knowing how, and we don't risk silently changing behavior.

macOS/DuckDuckGo/Updates/Sparkle/SimplifiedSparkleUpdateController.swift
```diff
@@ -146,21 +146,21 @@ final class SimplifiedSparkleUpdateController: NSObject, SparkleUpdateController
...
     private var customFeedURL: String? {
-        get { try? settings.debugSparkleCustomFeedURL }
-        set { try? settings.set(newValue, for: \.debugSparkleCustomFeedURL) }
-    }
-#endif
-
-    private var shouldShowUpdateNotification: Bool {
-        Date().timeIntervalSince(lastUpdateNotificationShownDate) > .days(7)
+        get {
+            guard allowUnsignedUpdates else { return nil }
```
diegoreymendez: The flag `SPARKLE_ALLOWS_UNSIGNED_UPDATES` is actually poorly named.  It's a mistake in a previous PR of my own.

Can we rename these to `allowCustomUpdateFeed`?

macOS/DuckDuckGo/Updates/Sparkle/SimplifiedSparkleUpdateController.swift
```diff
@@ -233,63 +232,71 @@ final class SimplifiedSparkleUpdateController: NSObject, SparkleUpdateController
     // MARK: - Feature Flags support
 
     private let featureFlagger: FeatureFlagger
-    private let buildType: ApplicationBuildType
+    private let allowUnsignedUpdates: Bool
```
diegoreymendez: The flag `SPARKLE_ALLOWS_UNSIGNED_UPDATES` is actually poorly named.  It's a mistake in a previous PR of my own.

Can we rename these to `allowCustomUpdateFeed`?

macOS/DuckDuckGo/Updates/Sparkle/SimplifiedSparkleUpdateController.swift
```diff
@@ -233,63 +232,71 @@ final class SimplifiedSparkleUpdateController: NSObject, SparkleUpdateController
...
-        buildType: ApplicationBuildType,
-        featureFlagger: FeatureFlagger,
-        userPreference: Bool
-    ) -> Bool {
-        if buildType.isDebugBuild {
-            return featureFlagger.isFeatureOn(.autoUpdateInDEBUG) && userPreference
-        } else if buildType.isReviewBuild {
-            return featureFlagger.isFeatureOn(.autoUpdateInREVIEW) && userPreference
-        } else {
+    public static func resolveAutoDownloadEnabled(allowUnsignedUpdates: Bool,
```
diegoreymendez: This looks like a nice change to simplify, thanks.

The flag `SPARKLE_ALLOWS_UNSIGNED_UPDATES`, which this was derived from, is actually poorly named.  It's a mistake in a previous PR of my own.

Can we rename these to `allowCustomUpdateFeed`?

macOS/DuckDuckGo/Updates/UpdateValidationResult.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  UpdateValidationResult.swift
```
diegoreymendez: I believe this is now dead code.

macOS/DuckDuckGo/Tab/UserScripts/UserScripts.swift
```diff
@@ -168,9 +166,16 @@ final class UserScripts: UserScriptsProvider {
...
+        if let updateController = Application.appDelegate.updateController as? any SparkleUpdateController {
+            releaseNotesUserScript = updateController.makeReleaseNotesUserScript(
+                pixelFiring: PixelKit.shared,
+                keyValueStore: UserDefaults.standard,
+                releaseNotesURL: .releaseNotes
+            )
+        } else {
+            releaseNotesUserScript = nil
+        }
```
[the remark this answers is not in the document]
mallexxx: initializeUpdateController is called from `AppDelegate.applicatIonWillFinishLaunching`, state restoration happens in `stateRestorationManager.applicationDidFinishLaunching`

PR #3446, opened 2026-02-04
[macOS] Fix crashes in NewTabPageNextStepsCardsProviderFacade

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1213119725978272?focus=true

### Description

This attempts to fix 2 low-frequency, normal-severity crashes reported in Sentry related to threading and possible deallocation in `NewTabPageNextStepsCardsProviderFacade`.

The fixes address:

* **Thread-safety:** Publishers could be used from background threads. The fix adds `.receive(on: DispatchQueue.main)` so publisher emissions from `activeProvider` are received on the main queue.
* **Provider lifetime:** Card providers were recreated when the feature flag changed, so the old provider could be deallocated while still referenced by `$activeProvider.flatMap`. The fix creates both providers once at init and only switches the active one, keeping both alive for the facade’s lifetime.
* **Publisher stability:** Publishers were recreated on each access, which could cause timing and threading issues. The fix uses lazy stored publishers so each subscription chain is created once and reused.

These issues were introduced in an overzealous attempt to address potential retain cycles when the facade was introduced. However, I discovered the cause was actually a `NotificationCenter` subscription in `NewTabPageNextStepsSingleCardProvider`, fixed with 3821f0f.

### Testing Steps

I couldn't find steps to reliably reproduce the crash, so confirm the card providers continue to work as expected:
1. Update the BSK `Package.swift` file in this branch with this change and reset package caches (this adds the required frontend changes when the feature flag is enabled):
```diff
-        .package(url: "https://github.com/duckduckgo/content-scope-scripts.git", exact: "12.30.0"),
+        .package(url: "https://github.com/duckduckgo/content-scope-scripts.git", branch: "rachelmcr/next-steps-list-testing"),
```
2. Build and run the app.
3. Go to Debug menu → New Tab Page → Reset Next Steps.
4. Confirm the current (production) Next Steps cards are shown as expected.
5. Go to Debug → Feature Flag Overrides → Other → enable `nextStepsListWidget`.
6. Confirm the new (feature-flagged) Next Steps cards are shown, starting with "Personalize."
7. Dismiss some of the cards.
8. On the "Take back your privacy!" card, select "Set as Default" and confirm you can set the browser as default and the card is dismissed. (This relies on the notification subscription updated in 3821f0f.)
9. Go to Debug → Feature Flag Overrides → Other → disable `nextStepsListWidget`.
10. Confirm the current Next Steps cards are shown, with no changes from before the feature flag was enabled (except hiding the "Set as Default" card if the app is still set as your default browser).

### Impact and Risks

#### What could go wrong?

* Primary impact is on reliability of the existing (production) Next Steps experience. Unit tests cover the facade behavior, and fixes address issues with thread safety and deallocation.
* Unit tests also confirm the behavior when enabling/disabling the feature flag (internal only).

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches New Tab Page state/publisher wiring and introduces a new Combine scheduling dependency; mis-scheduling or switching logic could cause missed UI updates or flaky behavior, though changes are localized and covered by updated tests.
> 
> **Overview**
> Fixes low-frequency crashes in the New Tab Page Next Steps cards facade by making provider switching and publisher delivery safer.
> 
> `NewTabPageNextStepsCardsProviderFacade` now **constructs both card providers once** and switches `activeProvider` between them (instead of recreating providers), **delivers feature-flag and `$activeProvider` updates on an injectable scheduler**, and **uses lazy stored `cardsPublisher`/`isViewExpandedPublisher` with `switchToLatest`** to avoid stale subscriptions and background-thread emissions. This PR also adds the `combine-schedulers` SwiftPM dependency and adjusts unit tests to use an immediate scheduler and await provider-switch expectations; `NewTabPageNextStepsSingleCardProvider` tweaks its key-window notification handling to async-dispatch refreshes to avoid deallocation/queue-retention issues in tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de7a602ceeb47040048e705e2955124dde99d458. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
+++ b/macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
@@ -311,7 +311,10 @@ private extension NewTabPageNextStepsSingleCardProvider {
     }
 
     func observeKeyWindowChanges() {
-        // We schedule this async (not receiving it directly on the main queue) to avoid a retain cycle
+        // Async dispatch allows the default browser setting to propagate after being changed in the system dialog.
+        // We schedule this in the sink block (not receiving it directly on the main queue) to avoid the main queue
+        // holding a reference to the block and preventing full deallocation in integration tests that end immediately
+        // after opening the New Tab Page, which would require flushing the queue.
         NotificationCenter.default.publisher(for: NSWindow.didBecomeKeyNotification)
             .sink { _ in
                 DispatchQueue.main.async { [weak self] in


macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsCardsProviderFacade.swift
```diff
@@ -89,25 +98,29 @@ final class NewTabPageNextStepsCardsProviderFacade: NewTabPageNextStepsCardsProv
...
         $activeProvider
-            .flatMap { provider in
+            .receive(on: DispatchQueue.main)
+            .map { provider in
                 provider.cardsPublisher
             }
+            .switchToLatest()
             .eraseToAnyPublisher()
-    }
+    }()
```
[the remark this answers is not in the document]
rachelmcr: This is not the cause of the reported crash (the subscriptions in `NewTabPageNextStepsCardsClient` ensure `@MainActor` for the values emitted, and these crashes didn't occur with the provider publishers before the facade was introduced) although we could change this for safety.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsCardsProviderFacade.swift
```diff
@@ -38,41 +41,47 @@ final class NewTabPageNextStepsCardsProviderFacade: NewTabPageNextStepsCardsProv
...
 
         featureFlagger.updatesPublisher
+            .compactMap { [weak self] in
+                self?.featureFlagger.isFeatureOn(.nextStepsListWidget)
+            }
+            .removeDuplicates()
+            .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in
                 guard let self else { return }
                 activeProvider = getActiveProvider()
```
alessandroboron: I guess we could use the value computed in the compactMap wiithout calling again getActiveProvider? 

activeProvider = isOn ? singleCardProvider : legacyCardsProvider

macOS/UnitTests/NewTabPage/NewTabPageNextStepsCardsProviderFacadeTests.swift
```diff
@@ -115,6 +119,8 @@ final class NewTabPageNextStepsCardsProviderFacadeTests: XCTestCase {
         let cancellable = facade.isViewExpandedPublisher.sink { value in
             receivedValues.append(value)
         }
+        // Allow subscription to inner publisher
+        RunLoop.main.run(until: Date().addingTimeInterval(0.1))
```
alessandroboron: These could lead to flaky tests, especially in CI. Perhaps we could inject a scheduler in the class and mock the scheduler in tests.

In iOS for some Combine tests we used pointfree [CombineSchedulers](https://github.com/pointfreeco/combine-schedulers). We could have a default parameter injected in the class which is the mainQueue scheduler and then we could replace it in tests with a TestScheduler which fires immediately. 

what do you think?

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
```diff
@@ -311,10 +311,12 @@ private extension NewTabPageNextStepsSingleCardProvider {
     }
 
     func observeKeyWindowChanges() {
+        // We schedule this async (not receiving it directly on the main queue) to avoid a retain cycle
```
alessandroboron: Hey @rachelmcr I don’t think I’m understanding this fix. Could you explain me more about it?

In the previous code self was captured weakly so that should already break the cycle. What’s the issue with ` .receive(on: DispatchQueue.main)`? Thanks

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsCardsProviderFacade.swift
rachelmcr: Yeah, we could. I was using that method to consolidate the logic around which provider to set, but this would be more direct. 👍

macOS/UnitTests/NewTabPage/NewTabPageNextStepsCardsProviderFacadeTests.swift
rachelmcr: That's a great approach, thanks for the suggestion! I wasn't too happy with this call but wasn't aware of that framework; it's much more reliable.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
rachelmcr: Sure! I was made aware of the issue by some tests failing because not all objects were deallocated, even though self was captured weakly here. Even when I removed the facade and used this provider directly, this subscription in particular was causing the failure.

As I understand it, because of how `receive(on:)` scheduling works, if a notification like this one is fired right as an object is trying to deallocate, the `DispatchQueue` can hold a reference to the execution block in its internal queue until the next run loop cycle, even if the block has a weak reference to self. (So it isn't an issue with self being retained, but the execution block itself that was scheduled.) I tried some other approaches to ensure it wasn't self/the subscription in self causing the issue (e.g. using an explicit `deinit` to cancel the `cancellables`) and confirmed that this wasn't the source of the issue — the tests still fail unless the queue is given an opportunity to perform the scheduled block.

This change (removing `receive(on:)` and scheduling the async task in the `sink` block) ensures that reference can't be held. That said, it's incredibly unlikely to be an issue in real user interactions so I could keep `receive(on:)` here and just update the tests to make sure the main queue completes all the scheduled work, e.g. like this:

```swift
let expectation = XCTestExpectation(description: "Finish scheduled tasks in main queue")
DispatchQueue.main.async {
    expectation.fulfill()
}
wait(for: [expectation], timeout: 1.0)
```

Alternately I could probably use `CombineSchedulers` here too, just need to check that I can add the dependency in the right place for the failing tests.

rachelmcr: I looked into injecting a scheduler for this, but given that the failing tests (integration tests in `AddressBarTests`) don't directly use the facade or its client, injecting the dependency requires a significant refactor of the `NewTabActionsManager` initialization in two different places. It got complicated very quickly and I'm somewhat concerned about introducing potential issues with our new tab actions handling just to inject this dependency in tests.

The issue seems to occur primarily in tests that set `TestRunHelper.allowAppSendUserEvents` to true, and only if the test opens a window and quickly closes it again (e.g. only making assertions on the opened window, without performing other actions). That makes sense because any user events that would trigger the key window notification would be recognized and the notification posted, but any scheduled execution blocks wouldn't then have time to be performed.

Since it isn't likely a real user issue, I'm inclined to go back to using `receive(on:)` and flush the main queue in the test teardown after setting `TestRunHelper.allowAppSendUserEvents` to false. That has some potential for flakiness but I think it's a reasonable tradeoff here. I made that change in e62713e but let me know what you think.

rachelmcr: I had to add the same change as above to `PinnedTabsManagerProviderTests`, too. The failing tests were similar to the `AddressBarTests` described above: opening a window and ending the test quickly, with a subscription received on the main queue after the test ended executing a sink block and causing an assertion failure.

alessandroboron: Thanks for the explanation @rachelmcr! I agree that an extensive refactor to inject a scheduler isn’t the right call here.

One concern: could new tests that set `TestRunHelper.allowAppSendUserEvents = true` and open/close a window quickly still hit this if they don’t call `waitForMainQueueToFlush()?`. I’m not sure how we’d enforce that.

If we expect to add this flush in multiple test classes over time, is it worth reconsidering keeping the original fix (removing .receive(on:)) instead?

rachelmcr: Yeah, I thought this over and 1) given that it's used on the new tab page so it affects unrelated tests that open new windows, 2) given that the only way to enforce it that I can think of is to flush the queue more aggressively (e.g. for every integration test class run) I think I'll go back to the original fix. It's not the most "right" way to queue work like this but has fewer side effects and works fine. I'll include some additional documentation to explain the fix more clearly.

PR #3468, opened 2026-02-05
[iOS] [Fire Button] Add Pixels and Reporter Service for Data Clearing flow

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213203351971108

### Description
Adds a new `DataClearingPixels` enum + `DataClearingPixelsReporter` to emit **end-to-end, duration, residue, and error** pixels for the Fire/data-clearing flow, including retrigger-within-20s and user-action-before-completion signals.

For individual actions instrumented in this flow:
https://app.asana.com/1/137249556945/task/1212930441623249

### Testing Steps

1. Compile and run the app, ensure that the app runs and key flows work as expected
2. [Manual Fire] Tap the fire button, ensure:

Overall duration Pixel fired as expected 
```
👾[Standard-Fired] m_fire_clearing_completion ["appVersion": "7.206.0", "duration": "2912", "option": "tabs,data,ai_chats", "pixelSource": "phone", "scope": "all", "source": "browsing", "trigger": “manual_fire”]
```
Individual pixels fired as expected 
```
👾[Standard-Fired] m_fire_burn_url_cache_duration ["appVersion": "7.206.0", "duration": "0", "pixelSource": "phone”]
```
```
👾[Standard-Fired] m_fire_burn_url_cache_has_residue ["appVersion": "7.206.0", "pixelSource": "phone”]
```
[Auto Fire] Turn on Auto Fire in Settings. Ensure auto fire pixels are as expected
```
👾[Standard-Fired] m_fire_clearing_completion ["appVersion": "7.206.0", "duration": "4296", "option": "tabs,data", "pixelSource": "phone", "scope": "all", "source": "autoClear", "trigger": “auto_clear_on_launch”]
```
[Retrigger] Rapidly trigger fire. Ensure rapid retrigger pixel fired as expected
```
👾[Standard-Fired] m_fire_retrigger_in_20s ["appVersion": "7.206.0", "pixelSource": "phone"]
```
[User action] Clicks on the screen before fire is complete. Ensure pixel fired as expected
```
👾[Standard-Fired] m_fire_user_action_before_completion ["appVersion": "7.206.0", "pixelSource": "phone”]
```

### Impact and Risks
**Medium Risk**
Adds new analytics reporting across the fire/data-clearing pipeline (tabs, website data, history, AI chat) and changes pixel schemas/IDs, which could affect telemetry correctness and introduces small async residue-check tasks during clearing.

#### What could go wrong?

---

> [!NOTE]
> **Medium Risk**
> Touches the core Fire/data-clearing pipeline and history/web cache clearing to add new telemetry and residue checks, which could affect clearing timing and analytics correctness if misfired or too noisy.
> 
> **Overview**
> Adds a new `DataClearingPixels`/`DataClearingPixelsReporter` instrumentation layer and new pixel definitions to measure the Fire/data-clearing flow end-to-end, including completion timing (with option/trigger/scope/source), retrigger-within-20s, and “user action before completion”.
> 
> Wires the reporter into `FireExecutor` and related clearing paths to emit per-step duration/residue/error pixels (tabs, URL cache, website data via `WebCacheClearingReporter`, browsing history, and AI chat history), including adding a transparent overlay in `MainViewController` to detect taps during manual burns. Updates history-clearing pixel hooks (`DataClearingPixelsHandling` gains scoped duration + default no-ops; `HistoryCoordinator`/`HistoryManager` renamed handler property and now report tab vs all scopes) and extends the legacy `clearDataInDefaultPersistence` debug pixel to include `scope`, with new unit tests for the reporter.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5e4aae9f9a4b98c6a9b954f83eeb3687ef9daac9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/Fire/Statistics/DataClearingPixels.swift
+++ b/iOS/DuckDuckGo/Fire/Statistics/DataClearingPixels.swift
@@ -42,7 +42,6 @@ enum DataClearingPixels {
     // MARK: - Website Data
 
     case burnWebsiteDataHasResidue(step: String, scope: String?)
-    case burnWebsiteDataError(Error)
 
     // MARK: - History
 
@@ -82,8 +81,6 @@ extension DataClearingPixels: PixelKitEvent {
 
         case .burnWebsiteDataHasResidue:
             return "m_fire_burn_website_data_has_residue"
-        case .burnWebsiteDataError:
-            return "m_fire_burn_website_data_error"
 
         case .burnHistoryDuration:
             return "m_fire_burn_history_duration"
@@ -128,7 +125,7 @@ extension DataClearingPixels: PixelKitEvent {
             
         case .retriggerIn20s, .userActionBeforeCompletion,
              .burnURLCacheHasResidue,
-             .burnTabsError, .burnHistoryError, .burnWebsiteDataError, .burnAIChatHistoryError:
+             .burnTabsError, .burnHistoryError, .burnAIChatHistoryError:
             return nil
         }
     }
@@ -136,7 +133,6 @@ extension DataClearingPixels: PixelKitEvent {
     var error: NSError? {
         switch self {
         case .burnTabsError(let error),
-             .burnWebsiteDataError(let error),
              .burnHistoryError(let error),
              .burnAIChatHistoryError(let error):
             return error as NSError
--- a/iOS/DuckDuckGoTests/Fire/DataClearingPixelsReporterTests.swift
+++ b/iOS/DuckDuckGoTests/Fire/DataClearingPixelsReporterTests.swift
@@ -202,18 +202,18 @@ final class DataClearingPixelsReporterTests: XCTestCase {
         let testError = NSError(domain: "WebKit", code: 500)
         
         // When
-        sut.fireErrorPixel(DataClearingPixels.burnWebsiteDataError(testError))
+        sut.fireErrorPixel(DataClearingPixels.burnHistoryError(testError))
         
         // Then
         XCTAssertEqual(mockPixelFiring.actualFireCalls.count, 1)
         XCTAssertEqual(mockPixelFiring.actualFireCalls.first?.frequency, .dailyAndStandard)
         
-        if case .burnWebsiteDataError(let error) =
+        if case .burnHistoryError(let error) =
             mockPixelFiring.actualFireCalls.first?.pixel as? DataClearingPixels {
             XCTAssertEqual((error as NSError).domain, testError.domain)
             XCTAssertEqual((error as NSError).code, testError.code)
         } else {
-            XCTFail("Expected burnWebsiteDataError pixel")
+            XCTFail("Expected burnHistoryError pixel")
         }
     }
     
--- a/iOS/PixelDefinitions/pixels/definitions/data_clearing_pixels.json5
+++ b/iOS/PixelDefinitions/pixels/definitions/data_clearing_pixels.json5
@@ -143,14 +143,6 @@
         ]
     },
 
-    "m_fire_burn_website_data_error": {
-        "description": "Fired when an error occurs while clearing website data",
-        "owners": ["hanyutang-sandra"],
-        "triggers": ["other"],
-        "suffixes": ["daily_standard", "platform", "form_factor"],
-        "parameters": ["appVersion", "pixelSource", "errorDomain", "errorCode"]
-    },
-
     "m_fire_burn_history_duration": {
         "description": "Fired to report the duration of history clearing",
         "owners": ["hanyutang-sandra"],


iOS/Core/WebCacheManager.swift
```diff
@@ -247,12 +275,24 @@ extension WebCacheManager {
...
                 dataRecords(record.displayName)
             }
             await dataStore.removeData(ofTypes: Self.safelyRemovableWebsiteDataTypes, for: removableRecords)
+            
+            Task {
+                let remainingRecords = await dataStore.dataRecords(ofTypes: Self.safelyRemovableWebsiteDataTypes)
+                if remainingRecords.contains(where: { record in dataRecords(record.displayName) }) {
+                    clearingReporter?.onResidue(ClearingStep.clearDataForSafelyRemovableDataTypes.rawValue, scope.description)
+                }
+            }
```
jaceklyp: Why is it not in the same runloop? You don't have to spawn Tasks {} as it is already an asynchronous function. Jumping out of a runloop can give us false positives, as some items might be recreated in the cache.

iOS/Core/WebCacheManager.swift
```diff
@@ -283,6 +334,17 @@ extension WebCacheManager {
         for cookie in cookiesToRemove {
             await cookieStore.deleteCookie(cookie)
         }
+        
+        Task {
+            let remainingCookies = await cookieStore.allCookies()
```
jaceklyp: same here

iOS/Core/WebCacheManager.swift
```diff
@@ -54,6 +54,23 @@ public protocol WebsiteDataManaging {
 
 }
 
+public struct WebCacheClearingReporter {
+    public var onResidue: (String, String?) -> Void
+
+    public init(
+        onResidue: @escaping (String, String?) -> Void
+    ) {
```
jaceklyp: it's a bit unfortunate formatting, let's do the init with args on the same line

iOS/Core/HistoryManager.swift
```diff
@@ -51,6 +51,8 @@ public class HistoryManager: HistoryManaging {
         }
         return dbCoordinator
     }
+    private let burnTabsPixelsHandling: DataClearingPixelsHandling?
+    private let burnHistoryPixelsHandling: DataClearingPixelsHandling?
```
jaceklyp: I think these should be Handlers, suffix Handling is a good one for the protocol

iOS/Core/HistoryManager.swift
```diff
@@ -132,6 +139,7 @@ public class HistoryManager: HistoryManaging {
         do {
             try await tabHistoryCoordinator.removeVisits(for: tabIDs)
         } catch {
+            burnTabsPixelsHandling?.fireErrorPixel(error)
```
jaceklyp: shouldn't it be burnHistoryPixelsHandler?

iOS/Core/HistoryManager.swift
```diff
@@ -143,8 +151,11 @@ public class HistoryManager: HistoryManaging {
     @MainActor
     public func removeBrowsingHistory(tabID: String) async {
         do {
+            let startTime = CACurrentMediaTime()
             try await dbCoordinator.burnVisits(for: tabID)
+            burnHistoryPixelsHandling?.fireDurationPixel(from: startTime, scope: "tab")
```
jaceklyp: could the scope be part of the enum case?

jaceklyp: I think external argument "from" is slightly awkward grammatically;
personally I'd prefer fireDurationPixel(startTime: startTime...)

iOS/Core/WebCacheManager.swift
```diff
@@ -54,6 +54,23 @@ public protocol WebsiteDataManaging {
...
+        self.onResidue = onResidue
+    }
+}
+
+private enum ClearingStep: String {
+    case clearDataForSafelyRemovableDataTypes = "clear_data_for_safely_removable_data_types"
+    case clearFireproofableDataForNonFireproofDomains = "clear_fireproofable_data_for_non_fireproof_domains"
+    case clearCookiesForNonFireproofedDomains = "clear_cookies_for_non_fireproofed_domains"
+    case removeObservationsData = "remove_observations_data"
```
jaceklyp: there's inconsistency in naming in one case it is ForNonFireproofDomains and in the second it is ForNonFireproofedDomains, let's make it consistent

iOS/DuckDuckGo/Fire/Statistics/DataClearingPixels.swift
```diff
@@ -0,0 +1,151 @@
...
+                params["scope"] = scope
+            }
+            return params
+            
+        case .burnTabsHasResidue(let scope):
+            return ["scope": scope]
+            
+        case .retriggerIn20s, .userActionBeforeCompletion,
+              .burnURLCacheHasResidue,
+             .burnTabsError, .burnHistoryError, .burnWebsiteDataError, .burnAIChatHistoryError:
```
jaceklyp: let's fix the formatting

iOS/DuckDuckGo/Fire/Statistics/DataClearingPixelsHandler.swift
```diff
@@ -0,0 +1,51 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import History
+
+struct DataClearingBurnTabsPixelsHandler: DataClearingPixelsHandling {
+    
+    let dataClearingPixelsResporter: DataClearingPixelsReporter
```
jaceklyp: there's a typo "Resporter"

iOS/DuckDuckGo/Fire/Statistics/DataClearingPixelsReporter.swift
```diff
@@ -0,0 +1,111 @@
...
+                scope: request.scope.description,
+                source: request.source.rawValue
+            ),
+            frequency: .standard
+        )
+    }
+
+    @MainActor
+    func fireRetriggerPixelIfNeeded() {
+        let now = timeProvider()
```
jaceklyp, line 60: why do we use "timeProvider()" can't we explicitly use CACurrentMediaTime()?

iOS/DuckDuckGo/Fire/Statistics/DataClearingPixelsReporter.swift
```diff
@@ -0,0 +1,111 @@
...
+        if let lastFire = lastFireTime, (now - lastFire) <= retriggerWindow {
```
jaceklyp: you can just write if let lastFireTime, (now - lastFireTime) ...

iOS/DuckDuckGo/Fire/Statistics/DataClearingPixelsReporter.swift
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+// MARK: - Private Helpers
+
+private extension DataClearingPixelsReporter {
+
+    private func prepareDuration(from startTime: CFTimeInterval, to endTime: CFTimeInterval) -> Int {
+        Int((endTime - startTime) * 1000)
+    }
```
jaceklyp: I'd make this function name more explict and perhaps without endTime argument:

private func elapsedMilliseconds(since startTime: CFTimeInterval) -> Int {
    Int((timeProvider() - startTime) * 1000)
}

iOS/DuckDuckGo/Fire/FireExecutor.swift
```diff
@@ -42,17 +42,46 @@ struct FireRequest {
         static let data = Options(rawValue: 1 << 1)
         static let aiChats = Options(rawValue: 1 << 2)
         static let all: Options = [.tabs, .data, .aiChats]
+        
+        private static let descriptions: [(Options, String)] = [(.tabs, "tabs"), (.data, "data"), (.aiChats, "ai_chats")]
```
jaceklyp, line 46: can we make descriptions part of the Options' enum instead of creating this dict?

so var description: String {
   switch self {
   case .tabs: return "tabs", etc.
   }
}

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -274,6 +274,10 @@ class MainViewController: UIViewController {
     func setWebExtensionManager(_ manager: WebExtensionManaging?) {
         self.webExtensionManager = manager
     }
+    
+    // A transparent bunring overlay appplied during firing
```
jaceklyp: bunring typo

iOS/Core/HistoryManager.swift
hassaanelgarem: FYI: This isn't 100% related to the fire button. The function is also called when a tab is closed.

iOS/DuckDuckGo/Fire/FireExecutor.swift
```diff
@@ -272,14 +309,20 @@ class FireExecutor: FireExecuting {
     private func burnTabs(scope: FireRequest.Scope, domains: [String]?) {
         switch scope {
         case .all:
+            let startTime = CACurrentMediaTime()
```
hassaanelgarem, line 312: We use this pattern throughout:
```swift
let starttime = CACurrentMediaTime()
// do something
dataClearingPixelsReporter.fireDurationPixel(pixel, from: startTime, args…)
```

This is fine, but it makes it harder to follow along. Especially when there are multiple start times in the same function. How about we encapsulate this logic using something like this:

* Add a new component `TimedOperation`
* Add a function to `DataClearingPixelsReporter` that creates an operation and sets the time. Then use it this way:
```swift
let burnHistoryTimedOperation = dataClearingPixelsReporter.startOperation(pixel, args…)
// do something
burnHistoryTimedOperation.finish() // or dataClearingPixelsReporter.finish(burnHistoryTimedOperation)
```
This would make the code flow more naturally, in my opinion. It's not a huge improvement though, so feel free to ignore.

P.S. We can take it a step further and do something like this:
```swift
dataClearingPixelsReporter.performAndMonitor { 
// do something
}
```

iOS/DuckDuckGo/Fire/FireExecutor.swift
```diff
@@ -343,7 +396,9 @@ class FireExecutor: FireExecuting {
         }
 
         self.forgetTextZoom()
+        let clearHistoryStartTime = CACurrentMediaTime()
```
hassaanelgarem: Since we already injected the pixels reporter into HistoryManager, I suggest moving this there, similar to tab history burning.

iOS/Core/WebCacheManager.swift
```diff
@@ -54,6 +54,23 @@ public protocol WebsiteDataManaging {
 
 }
 
+public struct WebCacheClearingReporter {
+    public var onResidue: (String, String?) -> Void
```
hassaanelgarem: [Suggestion] Can we use a typealias in place of String to make this more readable?

```suggestion
    public typealias ClearingStep = String
    public typealias Scope = String
    
    public var onResidue: (ClearingStep, Scope?) -> Void
```

iOS/Core/WebCacheManager.swift
```diff
@@ -247,12 +275,24 @@ extension WebCacheManager {
         switch scope {
         case .all:
             await dataStore.removeData(ofTypes: Self.safelyRemovableWebsiteDataTypes, modifiedSince: Date.distantPast)
+            Task {
+                if await !dataStore.dataRecords(ofTypes: Self.safelyRemovableWebsiteDataTypes).isEmpty {
+                    clearingReporter?.onResidue(ClearingStep.clearDataForSafelyRemovableDataTypes.rawValue, scope.description)
```
hassaanelgarem: In my testing, this is always getting triggered. Do we know why?

iOS/Core/WebCacheManager.swift
```diff
@@ -247,12 +275,24 @@ extension WebCacheManager {
...
             let allRecords = await dataStore.dataRecords(ofTypes: WKWebsiteDataStore.allWebsiteDataTypes())
             let removableRecords = allRecords.filter { record in
                 dataRecords(record.displayName)
             }
             await dataStore.removeData(ofTypes: Self.safelyRemovableWebsiteDataTypes, for: removableRecords)
+            
+            Task {
+                let remainingRecords = await dataStore.dataRecords(ofTypes: Self.safelyRemovableWebsiteDataTypes)
+                if remainingRecords.contains(where: { record in dataRecords(record.displayName) }) {
+                    clearingReporter?.onResidue(ClearingStep.clearDataForSafelyRemovableDataTypes.rawValue, scope.description)
```
hassaanelgarem: Same for the single tab scope, this always triggers

iOS/DuckDuckGo/Fire/FireExecutor.swift
```diff
@@ -325,7 +372,13 @@ class FireExecutor: FireExecuting {
     
     @MainActor
     private func burnAllData() async {
+        let urlSessionClearStartTime = CACurrentMediaTime()
         URLSession.shared.configuration.urlCache?.removeAllCachedResponses()
+        dataClearingPixelsReporter.fireDurationPixel(DataClearingPixels.burnURLCacheDuration, from: urlSessionClearStartTime)
+        dataClearingPixelsReporter.fireResiduePixelIfNeeded(DataClearingPixels.burnURLCacheHasResidue) {
```
hassaanelgarem, line 378: This is always firing for me as well, even on a fresh install with no browsing activity.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -3958,6 +3962,7 @@ extension MainViewController {
 extension MainViewController: FireExecutorDelegate {
     
     func willStartBurning(fireRequest: FireRequest) {
+        showBurningOverlay()
```
hassaanelgarem: Should this be shown even for auto clear?

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Also changed the original function naming to align with it

iOS/DuckDuckGo/Fire/Statistics/DataClearingPixelsReporter.swift
hanyutang-sandra: This is for unit test so that we can inject it to test the retrigger logic without using `sleep()`

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Not sure. It does not trigger for me

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: This is true. This method does trigger for me a lot of times. I suspect there’re something either purposely not removed or immediately created.

iOS/Core/WebCacheManager.swift
hanyutang-sandra: So that the residue checking can run asynchronously in the background without blocking the cleaning pipeline which might further delay the cleaning time. I can move it back if we are okay with it

iOS/Core/WebCacheManager.swift
hanyutang-sandra: So that the residue checking can run asynchronously in the background without blocking the cleaning pipeline which might further delay the cleaning time. I can move it back if we are okay with it

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: Can you clarify it more? `Options` is an `OptionSet` not an `enum`, so it’s possible it’s a combination of multiple values like [.tabs, .data], [.aiChats, .data] …. I can do something like if creating a dict is not the best solution
```
var description: String {
        var components: [String] = []
        if contains(.tabs) { components.append("tabs") }
        if contains(.data) { components.append("data") }
        if contains(.aiChats) { components.append("ai_chats") }
        return components.joined(separator: ",")
    }
```

iOS/Core/HistoryManager.swift
hanyutang-sandra: Yeah that’s true, but I don’t know a good way to differentiate it. imo it’s fine given it’s a single pixel with error information.

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: I initially considered the trailing closure solution but imo this pattern works well for wrapping single method calls, but breaks down for complex orchestration with closures like the overall duration measurement. I also thought about using micros but felt it’s overkill and make things overly complicated this time. If there’s a phase 2 for this project, I’m probably looking more into a micro-based solution

iOS/DuckDuckGo/MainViewController.swift
hanyutang-sandra: imo it makes sense to show it still as it helps us understand if clearing is too slow that users are getting impatient and try to interact

jaceklyp: I don't think it should, we already show an overlay for this and have a pixel that tracks that today.

iOS/Core/HistoryManager.swift
hanyutang-sandra: Personally I like the from label more since it follows the official Swift API Design Guidelines and matches Apple's own APIs like Date(timeIntervalSince:) and calendar.date(from:).

iOS/Core/WebCacheManager.swift
hanyutang-sandra: This triggers for me as well, The check might be too strict. Lemme look into it

iOS/Core/HistoryManager.swift
jaceklyp: It’s not quite the same as the Date(timeIntervalSince:) or Calendar.date(from:) examples. In those cases, since and from naturally describe a point in time, so the wording makes sense.

With fireDurationPixel(from:), however, from can be misread as referring to a location or source, rather than a starting time.

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Also loosened up a little bit

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Loosened up a little bit

iOS/DuckDuckGo/MainViewController.swift
hanyutang-sandra: ok removed it

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -4050,6 +4056,34 @@ extension MainViewController: FireExecutorDelegate {
...
+        let overlay = UIView(frame: window.bounds)
+        overlay.backgroundColor = .clear
+        
+        let tap = UITapGestureRecognizer(target: self, action: #selector(burningOverlayTapped))
+        tap.cancelsTouchesInView = false
+        overlay.addGestureRecognizer(tap)
+        
+        window.addSubview(overlay)
+        window.bringSubviewToFront(overlay)
+        burningOverlayView = overlay
```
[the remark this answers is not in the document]
hanyutang-sandra: fixed

iOS/DuckDuckGo/TabPreviewsSource.swift
```diff
@@ -98,10 +102,17 @@ class DefaultTabPreviewsSource: TabPreviewsSource {
...
             for previewUrl in previews {
-                try? FileManager.default.removeItem(at: previewUrl)
+                do {
+                    try FileManager.default.removeItem(at: previewUrl)
+                } catch {
+                    dataClearingPixelsReporter.fireErrorPixel(DataClearingPixels.burnTabsError(error))
+                }
             }
+        } catch {
+            dataClearingPixelsReporter.fireErrorPixel(DataClearingPixels.burnTabsError(error))
```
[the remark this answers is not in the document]
hanyutang-sandra: This is known. It should be fine as we are just getting error information here, which is useful regardless of the calling path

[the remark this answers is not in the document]
jaceklyp: Sending this in other cases will skew the metrics. How do you plan to distinguish when it’s actually supposed to be sent?

iOS/Core/WebCacheManager.swift
jaceklyp: but how do you plan to address what I mentioned:
"Jumping out of a runloop can give us false positives, as some items might be recreated in the cache."

iOS/Core/HistoryManager.swift
jaceklyp: Same question as elsewhere: sending this in other cases will skew the metrics. How do you plan to distinguish when it’s actually supposed to be sent?

iOS/Core/WebCacheManager.swift
jaceklyp: @hassaanelgarem is it still triggered for you?

iOS/DuckDuckGo/Fire/FireExecutor.swift
jaceklyp: what's your plan for that?

iOS/Core/WebCacheManager.swift
```diff
@@ -54,6 +54,23 @@ public protocol WebsiteDataManaging {
...
+    ) {
+        self.onResidue = onResidue
+    }
+}
+
+private enum ClearingStep: String {
+    case clearDataForSafelyRemovableDataTypes = "clear_data_for_safely_removable_data_types"
+    case clearFireproofableDataForNonFireproofDomains = "clear_fireproofable_data_for_non_fireproof_domains"
+    case clearCookiesForNonFireproofedDomains = "clear_cookies_for_non_fireproofed_domains"
+    case removeObservationsData = "remove_observations_data"
```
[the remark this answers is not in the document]
jaceklyp: ?

iOS/DuckDuckGo/Fire/FireExecutor.swift
```diff
@@ -290,7 +333,11 @@ class FireExecutor: FireExecuting {
             // didFinishBurning(fireRequest:) manually clears data after burn is complete
             // Close the tab and append a new empty tab, reusing existing one if exists
             tabManager.closeTabAndNavigateToHomepage(viewModel.tab, clearTabHistory: false)
-            
+            dataClearingPixelsReporter.fireDurationPixel(DataClearingPixels.burnTabsDuration, startTime: startTime, scope: scope.description)
```
jaceklyp, line 336: How do you differentiate between burning a single tab and burning all tabs? If there’s no differentiation, maybe we should focus only on the “burn all tabs” case. I assume the time to burn a single tab versus multiple tabs will be significantly different, and we probably don’t want them grouped into the same bucket.

jaceklyp (changes requested):

iOS/DuckDuckGo/Fire/FireExecutor.swift
```diff
@@ -490,7 +488,7 @@ class FireExecutor: FireExecuting {
         case .all:
             await burnAllAIHistory(trigger: request.trigger)
         }
-        dataClearingPixelsReporter.fireDurationPixel(DataClearingPixels.burnAIChatHistoryDuration, from: startTime, scope: request.scope.description)
+        dataClearingPixelsReporter.fireDurationPixel(DataClearingPixels.burnAIChatHistoryDuration, startTime: startTime, scope: request.scope.description)
```
hassaanelgarem: This is getting fired even if we don't burn AI history. There are a couple of scenarios where we don't:
* Scope is tab, and there's no chat to burn
* Scope is all but AI chats clearing is disabled by the user

I think if we fire this in these scenarios, the data will be skewed, as these scenarios will naturally have low durations

iOS/DuckDuckGo/Fire/FireExecutor.swift
hassaanelgarem: I think we need to understand the actual expectation before releasing this, otherwise the pixel will be redundant.

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Yes, it's a tradeoff between speed and residue check accuracy. I opted for speed because the Task spawns immediately after clearing complete. Again like I said above, if you feel strongly that accuracy is more important than avoiding the slight performance impact, I'm happy to move it inline

iOS/Core/WebCacheManager.swift
hassaanelgarem: This still triggers for me. But it doesn't trigger for fresh tabs
- Create a new tab
- Navigate to any website
- Burn this tab
- The residue pixel is triggered

However:
- Create a new tab
- Immediately burn this tab
- The residue pixel doesn't get triggered

iOS/Core/WebCacheManager.swift
hassaanelgarem: This still triggers for me. But it doesn't trigger when I do a subsequent burn.
* Have some browsing activity
* Burn all tabs
* This gets triggered
* Immediately burn all tabs again
* This doesn't get triggered

iOS/Core/WebCacheManager.swift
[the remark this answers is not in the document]
hanyutang-sandra: Yeah it’s completion because this is one of the steps. We just choose to not instrument it this time given product guidance. I can remove it if you prefer

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: They are differentiated with `scope`

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: We don't know if it's 100% or edge cases without production data. I will push it out and see if it truly happens 100% of the time across our user base. If so we can either:
- Look into the removal method and see if the residue happens due to it
- OR, remove the residue check and accept that there will be residue for it

What is your recommendation here?

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: Does `burnAIHistory(request: FireRequest)` get called in these scenarios?

iOS/Core/WebCacheManager.swift
jaceklyp: Actually, there is 0 guarantee that Task will be executed immediately after clearing completes. It may be delayed depending on thread pool availability, system load, and task priority. Meanwhile, caches can be lazily recreated during this delay, leading to false positives in residue detection.

On the other hand, the performance impact from inline measurement is negligible. The accuracy benefit far outweighs the minimal performance cost.

For SLI instrumentation accuracy, we should prioritize correctness over theoretical performance degradation - unless you have a strong proof it has a significant impact on the whole thing.

hanyutang-sandra: Ok moving it inside

iOS/Core/WebCacheManager.swift
hanyutang-sandra: imo this is a loose enough residue check. 

I could further loosen it up by comparing the count of data records before and after burn, and consider it good if the after.count < before.count. But I don’t think this will show anything valuable

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Similar to above, this is a loose enough residue check. And further loosen it up will reduce its value. It’s entirely possible that a burn does leave residue in which case this instrumentation does what it should do. Or it could be false positive, but I can’t see if there’re any logical mistakes

iOS/Core/HistoryManager.swift
hanyutang-sandra: What do you suggest here?

iOS/Core/WebCacheManager.swift
jaceklyp: but have you debugged/investigated what's actually left in the store to better understand the situation?

iOS/DuckDuckGo/Fire/FireExecutor.swift
hassaanelgarem: > Scope is tab, and there's no chat to burn

Get's called in this scenario yes

iOS/Core/WebCacheManager.swift
jaceklyp: same question: have you debugged/investigated what's actually left in the store to better understand the situation?

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: Do you know how these can be filtered out?

iOS/DuckDuckGo/Fire/FireExecutor.swift
hassaanelgarem: I suggest investigating the expectation. I assume currentDiskUsage and currentMemoryUsage have thresholds they don't fall below. Once we do, we can define the residue logic correctly.

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Internal domain remains for iOS, although it is explicitly filtered out on MacOS. I’m going to filter it out for residue check

iOS/Core/WebCacheManager.swift
hanyutang-sandra: Answered above

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: Somehow this API cannot reliably clean cache on disk. I’m checking cache in memory only to avoid false positives

iOS/DuckDuckGo/Fire/FireExecutor.swift
hanyutang-sandra: From `shouldBurnAIHistory`:
- Scope is tab, and there's no chat to burn -> This is a problem. I will move the pixel firing after await inside
- Scope is all but AI chats clearing is disabled by the user -> This is handled by `shouldAllowAIChatsBurn` if I’m not mistaken.

iOS/DuckDuckGo/TabPreviewsSource.swift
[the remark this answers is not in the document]
hanyutang-sandra: I don’t think it’s reasonable to just pass in a param for the fire check. The error information is valuable on it's own regardless of the path trggering it

iOS/Core/HistoryManager.swift
```diff
@@ -325,14 +345,17 @@ extension HistoryManager {
 
         let context = database.makeContext(concurrencyType: .privateQueueConcurrencyType)
         let dbCoordinator = HistoryCoordinator(historyStoring: HistoryStore(context: context, eventMapper: HistoryStoreEventMapper()))
+        dbCoordinator.dataClearingPixelsHandler = burnHistoryPixelsHandler
```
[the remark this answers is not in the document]
hanyutang-sandra: Like above, imo the error information itself is valuable. Regardless if it’s triggered by fire or not. In the dashboard if this error increases, regardless of what triggers it, it worths looking into anyways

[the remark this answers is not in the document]
hanyutang-sandra: Again, if the error increases, we need to investigate regardless of what triggered it. The underlying clearing infrastructure issue affects users the same way whether it's from burn, auto-clear, or settings.

Alternatively to know where it comes from, I need to pass burn context through the call chain, and this adds complexity for little benefit since all clearing failures are worth measuring

iOS/DuckDuckGo/TabPreviewsSource.swift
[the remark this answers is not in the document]
hanyutang-sandra: Again, if the error increases, we need to investigate regardless of what triggered it. The underlying clearing infrastructure issue affects users the same way whether it's from burn, auto-clear, or settings.

Alternatively to know where it comes from, I need to pass burn context through the call chain, and this adds complexity for little benefit since all clearing failures are worth measuring

iOS/Core/HistoryManager.swift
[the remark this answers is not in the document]
jaceklyp: Hanyu, you're right in that part that finding out above any error is valuable, but again, let me rephrase my previous question:
- How do you plan to differentiate when this is legitimately triggered by burning logic? If it also fires in other cases, it will for sure distort the metrics. The project wasn't about increasing visibility on any errors, but SLI related to the fire logic.

iOS/DuckDuckGo/UICoordination/MainCoordinator.swift
```diff
@@ -260,7 +260,9 @@ final class MainCoordinator {
         switch HistoryManager.make(isAutocompleteEnabledByUser: provider.appSettings.autocomplete,
                                    isRecentlyVisitedSitesEnabledByUser: provider.appSettings.recentlyVisitedSites,
                                    openTabIDsProvider: { tabsModel.tabs.map { $0.uid } },
-                                   tld: provider.storageCache.tld) {
+                                   tld: provider.storageCache.tld,
+                                   burnTabsPixelsHandler: DataClearingBurnTabsPixelsHandler(),
+                                   burnHistoryPixelsHandler: DataClearingBurnHistoryPixelsHandler()) {
```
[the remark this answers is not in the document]
hanyutang-sandra: Underneath it’s using the shared pixelFiring and the struct is lightweight. Doing this avoid passing things and around and makes it cleaner

PR #3474, opened 2026-02-05
Cursor rules for pixel registry definitions

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213143587385064
Tech Design URL:
CC:

### Description

This PR adds a Cursor rules file for pixel registry definitions.

### Testing Steps

1. Check the included pixel and its definition to verify that it's correct - that pixel was added as a test case, after which I asked Cursor to use the new rules file to write a definition for it
2. Check that the rules are accurate and there's nothing missing

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Documentation-only change under `.cursor/rules/` with no runtime or build impact beyond developer guidance.
> 
> **Overview**
> Adds a new Cursor rule file, `.cursor/rules/pixel-definitions.mdc`, documenting the expected schema and conventions for iOS/macOS pixel definition `.json5` files (required fields, triggers, parameter/suffix dictionary usage, and wide-event structure).
> 
> Also includes guidance for deriving parameters/suffixes from Swift `PixelKit` call sites and for validating/formatting definitions via the existing npm scripts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2c000a5df749ede1b229b8e986cecf0f6b793892. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ladamski (commented):
Note: using Opus 4.6 high

1. Looks like we should tell it to ignore TEMPLATE files:
```
Both TEMPLATE.json5 files reference "expires": "2025-01-30".
These should be reviewed: either remove the expired pixels, extend expiry, or promote to permanent.
```

2. Maybe provide more context on what a suffix is (ie. "a suffix is normally defined separately and appended to the pixel name, but in cases where a single suffix value is always present, a "daily" pixel may for example just include "_d" at the end of the pixel name")
```
macOS Missing daily Suffix on m_mac_daily_active_user_d
In macOS/daily_pixels.json5, the pixel m_mac_daily_active_user_d has no suffixes field despite having _d in its name (suggesting a daily suffix). Other daily pixels in the same file do include "suffixes": ["daily"]. This may be intentional (the _d is baked into the pixel name) but is worth verifying.
```
Also as a note, its not really clear how to get it to follow the same workflow as would be in cursorbot review.  I tried this prompt, and the results were "mixed":
```
Review current pixel definitions in iOS and Mac directories per .cursor/rules/pixel-definitions.mdc
```
After I asked `Did you actually consider the guidance in .cursor/rules/pixel-definitions.mdc?`, I basically got a "Yeah, nah, would you like me to actually do that?"
🤷‍♂️

PR #3485, opened 2026-02-06
[iOS] Enable the VPN error UI

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213114656619106
Tech Design URL:
CC:

### Description

This PR enables the VPN error UI, using new copy for the `TunnelError` cases that the packet tunnel provider can return.

### Testing Steps

1. This branch contains hardcoded failure logic, so just start the VPN a few times to cycle through each of the messages to confirm that it looks good - if you cycle enough, the VPN will connect, so verify that the banner goes away
2. Also, check that the copy is correct

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes VPN connection error handling and display logic, including cross-process error code mapping and async fetching of disconnect errors; issues could lead to incorrect/missing error UI or misclassified disconnect reasons.
> 
> **Overview**
> **Enables the VPN error banner for users** by translating tunnel disconnect failures into specific, localized messages (auth failure, server connection failure, subscription expired, and unknown).
> 
> Adds a stable `TunnelError.Code` enum in `PacketTunnelProvider` for cross-process serialization, introduces `VPNConnectionError` to map `NSError`/fallback strings to user-facing copy, and updates the status view model to fetch `NETunnelProviderSession.fetchLastDisconnectError` asynchronously and clear errors on successful connect.
> 
> Updates the error view to render markdown (for bold emphasis in the fallback help text) and adds the new localized VPN error strings across `Localizable.strings` files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a16d1be46390a4942ef04d066597e194bbb2fc3b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


SharedPackages/VPN/Sources/VPN/PacketTunnelProvider.swift
```diff
@@ -508,9 +508,29 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            throw TunnelError.settingsMissing
+        case 3:
+            throw TunnelError.tokenReset
+        case 4:
+            throw TunnelError.vpnAccessRevoked(NSError(domain: "test", code: 0))
+        case 5:
+            // No error - allow VPN to connect successfully to test that error UI clears
+            break
+        default:
+            break
```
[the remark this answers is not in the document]
samsymons: This is intentional for testing.

iOS/DuckDuckGo/VPNConnectionError.swift
```diff
@@ -0,0 +1,123 @@
...
+    /// This uses pattern matching on known error message strings from TunnelError.errorDescription.
+    /// This approach is less robust than NSError-based initialization, so prefer using `init(nsError:)`
+    /// when possible.
+    ///
+    /// - Parameter errorMessage: The error message string.
+    /// - Returns: A `VPNConnectionError` case, or `nil` if the error should not be displayed.
+    ///
+    init?(errorMessage: String) {
+        switch errorMessage {
+        case let msg where msg.contains("Missing auth token"):
```
samsymons, line 97: This is a bit of a hack in cases where we can't fetch the actual error object from the OS (due to availability checks), so I'm just matching against a few hardcoded error messages to return the new copy. Happy to change the approach - we would be able to clean this up once the proposal to drop iOS 15 and 16 is approved.

SharedPackages/VPN/Sources/VPN/PacketTunnelProvider.swift
```diff
@@ -525,9 +542,29 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
     }
 
+    // TODO: Remove this debug code before merging
     private func runDebugSimulations(options: StartupOptions) throws {
-        if options.simulateError {
-            throw TunnelError.simulateTunnelFailureError
+        // Force error on every startup for testing error UI
+        // Cycles through 5 error types, then 1 success case to verify UI clears
+        let errorIndex = UserDefaults.standard.integer(forKey: "vpn.debug.errorIndex")
+        UserDefaults.standard.set((errorIndex + 1) % 6, forKey: "vpn.debug.errorIndex")
```
diegoreymendez: Being super extra careful we might want to flag the PR as "do not merge".

🔴🔴🔴🔴🔴🔴 Revert / adjust before merging.

iOS/DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -379,23 +381,57 @@ final class NetworkProtectionStatusViewModel: ObservableObject {
...
+                    return
                 }
 
-                return ErrorItem(title: "Failed to Connect", message: errorMessage)
+                Task {
+                    let errorItem = await self.createErrorItem(fallbackMessage: errorMessage)
+                    await MainActor.run {
+                        self.error = errorItem
+                    }
+                }
```
diegoreymendez: I think there's some merit to the comment above as it's possible this task might override `self.error` at a bad time.  I chatted through this with Claude and he suggested the following approach for your consideration.

---

# CLAUDE SUGGESTION

**Potential race condition in `setUpErrorPublishers`** (`NetworkProtectionStatusViewModel.swift:394`)

The sink spawns an unstructured `Task` to fetch the error via `fetchLastDisconnectError`. If `.connected` arrives while that Task is in-flight, it sets `self.error = nil` (line 246), but then the stale Task completes and overwrites it with an `ErrorItem`. The user sees an error banner after a successful reconnect.

**Suggested fix** — store the Task and cancel it when clearing the error:

```swift
// Add a property:
private var errorTask: Task<Void, Never>?
```

In the sink (line 394), store and cancel:
```swift
self.errorTask?.cancel()
self.errorTask = Task {
    let errorItem = await self.createErrorItem(fallbackMessage: errorMessage)
    guard !Task.isCancelled else { return }
    await MainActor.run {
        self.error = errorItem
    }
}
```

In the `.connected` handler (line 246):
```swift
case .connected:
    self?.isNetPEnabled = true
    self?.errorTask?.cancel()
    self?.error = nil
```

samsymons: That changes makes sense, I've applied it - thanks!

PR #3531, opened 2026-02-10
[macOS] Memory interval pixel reporter implementation

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213200269801331
Tech Design URL:
CC:

### Description
Implements MemoryUsageIntervalReporter, which fires the m_mac_memory_usage pixel at startup and scheduled intervals (1h, 2h, 4h, 8h, 24h). Each pixel includes context parameters: browser_memory_mb, windows, tabs, architecture, and sync_enabled.

The reporter subscribes to the `.memoryUsageReporting` feature flag via Combine and starts/stops monitoring accordingly. 
Disabling and re-enabling the flag resets the session. A background Task.detached polls every 60 seconds, checking elapsed time and firing unfired triggers. Context is collected on @MainActor at the moment of firing.

Debug menu items added: "Fire Interval Pixel Now..." (fire a specific trigger with the current context) and integration with the existing simulated memory actions.

Builds on #3507 (interval pixel definition).

### Testing Steps
1. Enable the memoryUsageReporting feature flag
2. Launch the app and verify in the debug console that the startup m_mac_memory_usage pixel fires with parameters (trigger=startup, browser_memory_mb, windows, tabs, architecture, sync_enabled)
3. Use Debug menu > "Fire Interval Pixel Now..." > select a trigger (e.g., 1h) > verify the pixel fires with the selected trigger and current context (windows, tabs, arch and sync_enabled)
4. Use Debug menu > "Simulate Memory Usage Report" > enter a value (e.g., 5000 MB) > "Fire Interval Pixel Now..." > verify browser_memory_mb reflects the simulated value
5. Disable the feature flag, re-enable it, and verify the startup pixel fires again (session reset)

### Impact and Risks
Low -- behind feature flag, only fires analytics pixels, no user-facing changes.

#### What could go wrong?
- Pixels could fire more than once per trigger per session if cancellation is not handled. Mitigated with Task.isCancelled guards and NSLock-protected deduplication set.
- Feature flag toggle during mid-iteration could fire stale triggers. Mitigated with cancellation check inside the trigger loop.

### Quality Considerations
- Context parameters fall back to "unknown" if dependencies (WindowControllersManager, sync service) are nil

### Notes to Reviewer
- memoryUsageIntervalReporter in AppDelegate is a var Optional because the syncServiceProvider closure captures self, which is unavailable before super.init(). A comment explains this.
- `fireTriggerNow` intentionally bypasses feature flag and deduplication checks for debug use.
- There is an edge case that if the feature flag is turned on while browsing, it will fire a `startup` pixel; this case should be low given that the feature flag is turned on by default.

—
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Introduces a new background polling task with locking and feature-flag lifecycle that can affect pixel volume and runtime behavior if mis-fired, though it’s gated behind `.memoryUsageReporting` and is analytics-only.
> 
> **Overview**
> **Adds interval-based memory reporting.** Introduces `MemoryUsageIntervalReporter` to fire the `m_mac_memory_usage` pixel at `startup`, `1h`, `2h`, `4h`, `8h`, and `24h`, deduped per session and automatically started/stopped via Combine updates to the `.memoryUsageReporting` feature flag.
> 
> **Refactors context collection + wiring/testing.** `MemoryReportingContext.collect` now takes a lightweight `WindowContext` and `isSyncEnabled` boolean instead of pulling from `WindowControllersManager`/`DDGSyncing`; `AppDelegate` wires the new reporter, and the Debug menu gains a `Fire Interval Pixel Now...` action. Adds comprehensive unit tests covering trigger timing, deduplication/session reset, and the production feature-flag publisher path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8bf4eb10fca1707ead795649d5038c63fcdc747e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/UnitTests/AppHealth/MemoryUsageIntervalReporterTests.swift
+++ b/macOS/UnitTests/AppHealth/MemoryUsageIntervalReporterTests.swift
@@ -268,22 +268,6 @@ final class MemoryUsageIntervalReporterTests: XCTestCase {
         XCTAssertEqual(totalStartupFires, 2, "Startup pixel should fire again after OFF->ON toggle (new session)")
     }
 
-    // MARK: - No Start Time
-
-    func testWhenNotStarted_ThenDoesNotFirePixels() async {
-        // Given
-        mockMemoryUsageMonitor.currentPhysFootprintMB = 1024
-        mockFeatureFlagger.enabledFeatureFlags = [.memoryUsageReporting]
-        sut = makeSUT()
-        // Note: startMonitoringForTesting() NOT called
-
-        // When
-        await sut.checkIntervalsNow()
-
-        // Then
-        XCTAssertTrue(mockPixelFiring.actualFireCalls.isEmpty)
-    }
-
     // MARK: - Memory Bucketing in Context
 
     func testWhenMemoryChanges_ThenContextReflectsCurrentMemory() async {


macOS/DuckDuckGo/AppHealth/MemoryUsageIntervalReporter.swift
```diff
@@ -0,0 +1,248 @@
...
+    /// The interval between checks, in seconds.
+    static let defaultCheckInterval: TimeInterval = 60
+
+    // MARK: - Dependencies
+
+    private let memoryUsageMonitor: MemoryUsageMonitoring
+    private let featureFlagger: FeatureFlagger
+    private let pixelFiring: PixelFiring?
+    private let windowControllersManager: WindowControllersManagerProtocol?
+    private let syncServiceProvider: () -> DDGSyncing?
```
ayoy: We shouldn’t need to couple Sync and memory reporting here. This could be `private let isSyncEnabled: () -> Bool`.

Similarly, we could decouple WindowControllersManagerProtocol into windows and tabs accessor. Then you would provide the logic in closures inside AppDelegate, and MemoryReportingContext would only bucket the value as appropriate.

(think of AppHealth as a standalone Swift module that doesn’t have access to the rest of the app code - I think if it’s easy enough to decouple code, we should always be doing it)

macOS/DuckDuckGo/AppHealth/MemoryUsageIntervalReporter.swift
```diff
@@ -0,0 +1,248 @@
...
+        lock.withLock { self.startTime = startTime }
+    }
+
+    /// For testing and debug menu: triggers an immediate interval check.
+    func checkIntervalsNow() async {
+        await checkAndFireIntervals()
+    }
+
+    /// Clears fired triggers, allowing all triggers to fire again.
+    func resetFiredTriggers() {
```
ayoy, line 227: since this is only called from unit tests, you could wrap this function in `#if DEBUG`

macOS/DuckDuckGo/Menus/MainMenuActions.swift
```diff
@@ -680,6 +680,35 @@ extension AppDelegate {
         alert.runModal()
     }
 
+    @objc func fireIntervalPixelNow(_ sender: Any?) {
+        let reporter = NSApp.delegateTyped.memoryUsageIntervalReporter
```
ayoy: You can remove this line and use `memoryUsageIntervalReporter` directly.

Also please update `clearSimulatedMemory` and `startMemoryReporterImmediately` functions above to not use `NSApp.delegateTyped` :)

macOS/DuckDuckGo/AppHealth/MemoryUsageIntervalReporter.swift
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Starts a background task that periodically checks and fires interval pixels.
+    private func startIntervalChecking() {
+        let interval = checkInterval
+        monitoringTask = Task.detached(priority: .utility) { [weak self] in
+            // Fire startup + any already-elapsed triggers immediately
+            await self?.checkAndFireIntervals()
+
+            while !Task.isCancelled {
+                try? await Task.sleep(nanoseconds: NSEC_PER_SEC * UInt64(interval))
```
ayoy, line 142: Just curious, have you checked how this will work when the laptop is put to sleep?

EDIT: Nevermind, given that we’re comparing dates it looks like overdue pixels will be sent after waking up. I guess this is fine for now.

macOS/DuckDuckGo/AppHealth/MemoryUsageIntervalReporter.swift
jotaemepereira: Great idea. I will refactor the code 💪🏼

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1055,6 +1059,23 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
 
+        memoryUsageIntervalReporter = MemoryUsageIntervalReporter(
+            memoryUsageMonitor: memoryUsageMonitor,
+            featureFlagger: featureFlagger,
+            pixelFiring: PixelKit.shared,
+            windowContext: WindowContext(
+                tabs: windowControllersManager.allTabCollectionViewModels.reduce(0) { $0 + $1.allTabsCount },
+                windows: windowControllersManager.mainWindowControllers.count
+            ),
+            // Strong capture of self is intentional — AppDelegate lives for the entire app lifecycle.
```
ayoy: This comment doesn’t seem to be true or needed :)

PR #3807, opened 2026-03-03
[iPad] Tableview and Keyboard improvements on duck.ai

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211654189969294/task/1213510221952206?focus=true

### Description
Add Shift+Enter keyboard shortcut to the address bar that switches to AI Chat mode (or inserts a newline if already in AI Chat mode). Fix the chat history list on iPad by removing the default table separators and using a custom cell separator that hides on the last row. Also remove the history section header separator for iPad.

### Testing Steps
  1. On iPad, open the AI Chat history list and verify cell dividers appear between rows but not after the last row.
  2. Focus the address bar and press Shift+Enter — verify it switches to AI Chat mode. Press Shift+Enter again and verify it inserts a newline in  the AI Chat text view.
  3. On search mode, write any query, check if the suggestions panel are "floating" same way the duck.ai suggestions panel is. i.e: No grouped background container in this view.

---

> [!NOTE]
> **Medium Risk**
> Mostly iPad-only UI/interaction changes, but it alters suggestion tray sizing/styling and keyboard command handling, which could cause layout regressions across different iPad modes and content sizes.
> 
> **Overview**
> Adds an iPad-only **Shift+Enter** keyboard shortcut in `DefaultOmniBarViewController` to switch into duck.ai mode (or insert a newline when already in duck.ai).
> 
> Updates the iPad AI chat history list to remove default `UITableView` separators and use a custom per-cell separator that is hidden on the last row.
> 
> Introduces an iPad feature-flagged *plain list* autocomplete presentation: `AutocompleteView` can hide the last row separator for the final populated section, and reports actual SwiftUI content height via a preference key; `AutocompleteViewController` forwards that height to the tray and skips its legacy height calculation when plain style is enabled.
> 
> Adjusts `SuggestionTrayViewController`/`MainViewController` so the tray can switch between floating vs non-floating styling (corner radius/shadow/top inset/initial height) and changes the large-width tray width padding to match the plain-style iPad duck.ai toggle experience.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 42f4c28162a7e8907c35ead79a2710864deb2266. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AutocompleteViewController.swift
+++ b/iOS/DuckDuckGo/AutocompleteViewController.swift
@@ -60,7 +60,6 @@ class AutocompleteViewController: UIHostingController<AutocompleteView> {
     private let productSurfaceTelemetry: ProductSurfaceTelemetry
 
     private var task: URLSessionDataTask?
-    private weak var cachedScrollView: UIScrollView?
     private var lastReportedContentHeight: CGFloat = 0
 
     private var isUsingUnifiedPrediction: Bool {
@@ -126,25 +125,17 @@ class AutocompleteViewController: UIHostingController<AutocompleteView> {
 
         view.backgroundColor = UIColor(designSystemColor: .background)
 
+        model.onContentHeightChanged = { [weak self] height in
+            guard let self, height > 0, height != self.lastReportedContentHeight else { return }
+            self.lastReportedContentHeight = height
+            self.presentationDelegate?.autocompleteDidChangeContentHeight(height: height)
+        }
+
         queryDebounceCancellable = $query
             .debounce(for: .milliseconds(Self.debounceDelayMS), scheduler: RunLoop.main)
             .sink { [weak self] query in
                 self?.requestSuggestions(query: query)
             }
-    }
-
-    override func viewDidLayoutSubviews() {
-        super.viewDidLayoutSubviews()
-        guard model.usePlainListStyle else { return }
-
-        let scrollView = cachedScrollView ?? Self.findScrollView(in: view)
-        cachedScrollView = scrollView
-
-        guard let scrollView,
-              scrollView.contentSize.height > 0,
-              scrollView.contentSize.height != lastReportedContentHeight else { return }
-        lastReportedContentHeight = scrollView.contentSize.height
-        presentationDelegate?.autocompleteDidChangeContentHeight(height: scrollView.contentSize.height)
     }
 
     override func viewDidAppear(_ animated: Bool) {
@@ -259,13 +250,8 @@ class AutocompleteViewController: UIHostingController<AutocompleteView> {
     private func updateHeight() {
         guard let lastResults else { return }
 
-        // For plain list style, height is measured from the underlying scroll view's
-        // content size in viewDidLayoutSubviews to avoid hardcoded cell height mismatches.
-        if model.usePlainListStyle {
-            lastReportedContentHeight = 0
-            view.setNeedsLayout()
-            return
-        }
+        // For plain list style, height is reported by SwiftUI via ContentHeightPreferenceKey.
+        if model.usePlainListStyle { return }
 
         let messageHeight = model.isMessageVisible ? 196 : 0
         let sectionPadding = 12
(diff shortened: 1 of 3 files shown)


iOS/DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -261,6 +281,18 @@ class AutocompleteViewController: UIHostingController<AutocompleteView> {
             .autocompleteDidChangeContentHeight(height: CGFloat(height))
     }
 
+    private static func findScrollView(in view: UIView) -> UIScrollView? {
```
aataraxiaa: **[High]** This recursively walks UIKit's internal view tree to find SwiftUI's scroll view. SwiftUI doesn't guarantee this structure across OS versions — if it returns `nil` (or finds the wrong scroll view), `autocompleteDidChangeContentHeight` is never called and the suggestion tray stays at height 0 (invisible).

The stable alternative is a SwiftUI `PreferenceKey` with a `GeometryReader` inside `AutocompleteView` to measure content height and surface it to the UIKit layer via the existing delegate.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/ChatHistory/AIChatHistoryListViewController.swift
```diff
@@ -259,3 +267,27 @@ extension AIChatHistoryListViewController: UITableViewDelegate {
...
+        + AIChatHistoryListViewController.Constants.iconTextSpacing
+
+    private lazy var customSeparator: UIView = {
+        let view = UIView()
+        view.backgroundColor = UIColor(designSystemColor: .lines)
+        view.translatesAutoresizingMaskIntoConstraints = false
+        contentView.addSubview(view)
+        NSLayoutConstraint.activate([
+            view.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: Self.separatorLeadingInset),
+            view.trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
```
aataraxiaa, line 283: **[Medium]** `UIScreen.main` is deprecated since iOS 16 and can return the wrong scale on multi-scene iPads.

Use `traitCollection.displayScale` instead, or lazily resolve it when the view has a window: `window?.screen.scale ?? 1`.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/ChatHistory/AIChatHistoryListViewController.swift
```diff
@@ -259,3 +267,27 @@ extension AIChatHistoryListViewController: UITableViewDelegate {
         return 0
     }
 }
+
+private final class AIChatHistoryCell: UITableViewCell {
+    private static let separatorLeadingInset: CGFloat = AIChatHistoryListViewController.Constants.horizontalInset
```
aataraxiaa, line 272: **[Low]** This happens to equal `Constants.horizontalInset + Constants.iconSize + Constants.iconTextSpacing` (16 + 16 + 12), but isn't derived from them — they could drift independently.

Consider referencing the constants directly so it stays in sync.

iOS/DuckDuckGo/SuggestionTrayViewController.swift
```diff
@@ -210,19 +211,31 @@ class SuggestionTrayViewController: UIViewController {
...
 
-        containerView.layer.cornerRadius = 24
-        containerView.layer.masksToBounds = true
- 
-        backgroundView.layer.cornerRadius = 24
-        backgroundView.backgroundColor = UIColor(designSystemColor: .background)
-        backgroundView.clipsToBounds = false
-        backgroundView.applyActiveShadow()
+        if useFloatingStyle {
+            containerView.layer.cornerRadius = 24
```
aataraxiaa, line 215: **[Low]** The non-floating path (below) sets `.cornerCurve = .continuous` on both container and background views, but the floating path here doesn't. The result is floating style uses legacy circular corners while the new plain style gets the modern squircle. Intentional, or worth aligning?

PR #3956, opened 2026-03-11
[mac][Subscription] No subscription menu after failed Stripe purchase

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1213433154508757?focus=true

### Description

After a failed Stripe purchase, the subscription menu would disappear on macOS because `getSubscription(forceRefresh:)` threw `SubscriptionEndpointServiceError.noData` and callers treated it as a fatal error.

**Root cause:** The backend returns "no data" when a user has no active subscription (e.g. after a failed purchase). This was surfaced as a thrown error, forcing every caller to distinguish a network-layer error from a normal business condition.

**Fix:** Changed `getSubscription(forceRefresh:)` to return `DuckDuckGoSubscription?` — returning `nil` when the backend has no subscription instead of throwing. Updated all callers across iOS and macOS to handle the optional return.

Key changes:
- `SubscriptionManager` protocol: return type changed from `DuckDuckGoSubscription` to `DuckDuckGoSubscription?`
- Implementation catches `SubscriptionEndpointServiceError.noData` internally, clears cache, posts `.subscriptionDidChange`, and returns `nil`
- All `try await` callers updated with `guard let`/`if let` for the nil case
- All `try?` callers with direct member access updated to use optional chaining: `(try? await ...)?.property`
- Extracted `SubscriptionCachingService` as a protocol; `DefaultSubscriptionCachingService` reimplemented as a Swift `actor` (replacing `struct` + `DispatchQueue.sync`) for correct mutual exclusion and async/await compatibility
- `SubscriptionRequestCoalescer` actor: uses a `FetchMode` enum key (`cached` / `forceRefresh`) instead of `Bool` for self-documenting deduplication
- `clearSubscriptionCache()` synchronous guarantee restored: `reset()` is `nonisolated` on the actor (valid because `UserDefaultsCache` is thread-safe), so no `Task` wrapper is needed
- `WideEventService` (iOS + macOS): `checkForCurrentEntitlements()` uses `forceRefresh: true` to avoid cold-cache mis-attribution during post-purchase wide-event completion at app launch
- `.subscriptionDidChange` notification handler in `PreferencesSidebarModel`: uses `forceRefresh: true` so external posters (e.g. `DefaultSubscriptionFlowsExecuter`, `SubscriptionPagesUseSubscriptionFeature`) always see fresh state
- Removed dead code from `UnifiedToggleInputCoordinator`: unused `subscriptionManager` stored property and `resolveSubscriptionState()` method
- Added unit tests: nil-return path, no-token notification suppression, coalescer cancellation propagation, and fresh-work-after-cancellation

### Testing Steps
A full manual subscription regression test is needed

In Particular:
- Build the app using the alpha schema
- Connect to US VPN
- Strat a stripe purchase
- Reach the payment page
- Close the tab
- The Subscription menu should be visible and a new subscription can be purchased
- Close the menu / restart the app
- The Subscription menu should be visible and a new subscription can be purchased

### Impact and Risks

High — affects subscription visibility and purchase flows on both platforms.

#### What could go wrong?

- The cache-clearing on `nil` could cause unnecessary re-fetches. Mitigated by only clearing when transitioning from a cached subscription to nil.
- The actor-based caching service changes the concurrency model. Mitigated by keeping `isPresent` and `reset()` `nonisolated` to preserve synchronous access contracts.

---
###### Internal references:

---

> [!NOTE]
> **High Risk**
> Changes the subscription retrieval contract and caching/concurrency behavior across iOS/macOS, which can affect purchase flows, entitlement gating, and UI visibility if edge cases regress. Risk is elevated due to new actor-based cache, notification semantics, and widespread call-site updates.
> 
> **Overview**
> Treats “no subscription on backend” as a normal state by changing `SubscriptionManager.getSubscription` to return `DuckDuckGoSubscription?` (with `forceRefresh`) and updating callers to handle `nil` instead of `SubscriptionEndpointServiceError.noData`.
> 
> Moves subscription caching and feature enrichment responsibilities out of `SubscriptionEndpointService` into `DefaultSubscriptionManager`, introducing an actor-based `SubscriptionCachingService` plus a `SubscriptionRequestCoalescer` to dedupe concurrent fetches, and adjusts `.subscriptionDidChange` posting/handling (including nil `userInfo` to represent cleared state).
> 
> Updates purchase/restore flows, settings/remote-messaging/AI-tier resolution, and wide-event entitlement checks to use the new APIs (including targeted `forceRefresh`), and adds/updates unit tests and mocks to cover nil/no-token/transient-error fallbacks and coalescer cancellation behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afde3d564e31ec8452a9eb8436507930f2dee3ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/Core/PixelEvent.swift
```diff
@@ -2709,11 +2707,7 @@ extension Pixel.Event {
         case .ddgSubscriptionManagementRemoval: return "m_privacy-pro_settings_remove-from-device_click"
         case .subscriptionSuccessfulSubscriptionAttribution: return "m_subscribe"
         case .subscriptionKeychainAccessError: return "m_privacy-pro_keychain_access_error"
-            // Auth
-        case .subscriptionAuthV2GetTokensError2: return "m_privacy-pro_auth_v2_get_tokens_error2"
```
federicocappelli, line 3005: unused pixel

macOS/DuckDuckGo/Preferences/Model/PreferencesSidebarModel.swift
```diff
@@ -401,39 +401,46 @@ final class PreferencesSidebarModel: ObservableObject {
...
+                isPaidAIChatEnabled: entitlementStatus.paidAIChat,
+
+                isNetworkProtectionRemovalAvailable: isNetworkProtectionAvailable,
+                isPersonalInformationRemovalAvailable: isDataBrokerProtectionAvailable,
+                isIdentityTheftRestorationAvailable: isIdentityTheftRestorationAvailable || isIdentityTheftRestorationGlobalAvailable,
+                isPaidAIChatAvailable: featureFlagger.isFeatureOn(.paidAIChat) && isPaidAIChatAvailable)
+        } else {
+            return PreferencesSidebarSubscriptionState(shouldHideSubscriptionPurchase: shouldHideSubscriptionPurchase)
+        }
     }
```
federicocappelli, line 469: This is the main change, if a subscription is not present in the local cache, then we show the purchase option directly by setting all variables in PreferencesSidebarSubscriptionState to false (except for the shouldHideSubscriptionPurchase)

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -46,10 +46,11 @@ public protocol SubscriptionManager: SubscriptionTokenProvider, SubscriptionAuth
...
 
-    /// Retrieve the purchased subscription
-    /// - Parameter cachePolicy: The cache policy, `remoteFirst` or `cacheFirst`
-    /// - Returns: A `DuckDuckGoSubscription` if available, throws `SubscriptionEndpointServiceError.noData` if the subscription is not available or any other errors if the process failed at any point.
-    @discardableResult func getSubscription(cachePolicy: SubscriptionCachePolicy) async throws -> DuckDuckGoSubscription
+    /// Retrieve the purchased subscription.
+    /// - Parameter forceRefresh: When `true`, skips the cache and fetches the subscription from the remote backend. When `false`, returns the cached subscription if available, otherwise fetches remotely.
+    /// - Returns: The subscription if available, or `nil` if no subscription exists on the backend.
+    /// - Throws: `SubscriptionManagerError.noTokenAvailable` if the user is not authenticated. `SubscriptionManagerError.noLocalSubscription` if the token is invalid and no cached subscription is available.
+    @discardableResult func getSubscription(forceRefresh: Bool) async throws -> DuckDuckGoSubscription?
```
federicocappelli, line 54: Main changes here:
1. show only `forceRefresh` and not the confusing cache policy
2. Allow to return nil if the subscription is not really there

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -90,6 +91,9 @@ public protocol SubscriptionManager: SubscriptionTokenProvider, SubscriptionAuth
     /// Removes the subscription cache, this will trigger a remote fetch the next time `getSubscription(...)` is called
     func clearSubscriptionCache()
 
+    /// Ingests a subscription by enriching it with tier features, caching it, and posting a change notification.
+    func ingestSubscription(_ subscription: DuckDuckGoSubscription) async throws
```
federicocappelli: moved from SubscriptionEndpointService

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -292,55 +303,77 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
 
     @discardableResult
-    public func getSubscription(cachePolicy: SubscriptionCachePolicy) async throws -> DuckDuckGoSubscription {
-
-        // NOTE: This is ugly, the subscription cache will be moved from the endpoint service to here and handled properly https://app.asana.com/0/0/1209015691872191
-
-        guard isUserAuthenticated else {
-            throw SubscriptionEndpointServiceError.noData
-        }
+    public func getSubscription(forceRefresh: Bool = false) async throws -> DuckDuckGoSubscription? {
```
federicocappelli, line 359: this is where now we manage the subscription caching

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -353,12 +386,22 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
         return await tierOptionsProvider.subscriptionTierOptions(includeProTier: includeProTier, currentSubscription: subscription)
     }
 
     public func clearSubscriptionCache() {
-        subscriptionEndpointService.clearSubscription()
+        subscriptionCachingService.reset()
+    }
+
+    /// Enriches a subscription with tier features fetched from the backend.
+    private func enrichSubscriptionWithFeatures(_ subscription: DuckDuckGoSubscription) async throws -> DuckDuckGoSubscription {
```
federicocappelli, line 475: moved from SubscriptionEndpointService

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -390,14 +433,14 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
 
     public func getCustomerPortalURL() async throws -> URL {
         guard isUserAuthenticated else {
-            throw SubscriptionEndpointServiceError.noData
+            throw SubscriptionManagerError.noTokenAvailable
```
federicocappelli, line 513: important change, SubscriptionManager now throws only SubscriptionManagerError

SharedPackages/BrowserServicesKit/Sources/Subscription/Storage/SubscriptionCachingService.swift
```diff
@@ -0,0 +1,85 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import Common
+import os.log
+
+/// Manages caching of `DuckDuckGoSubscription` with thread-safe access and expiration logic.
+public protocol SubscriptionCachingService {
```
federicocappelli, line 24: This contains all subscription caching; the business logic is unchanged.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -274,73 +284,114 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
+    /// When no cache exists, throws `fallbackError` so callers propagate the appropriate error.
+    private func cachedSubscriptionOrThrow(_ fallbackError: Error) throws -> DuckDuckGoSubscription {
+        guard let cached = subscriptionCachingService.get() else {
+            throw fallbackError
         }
+        if cached.isActive { pixelHandler.handle(pixel: .subscriptionIsActive) }
+        return cached
+    }
 
-        var subscription: DuckDuckGoSubscription
```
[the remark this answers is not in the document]
federicocappelli: as intended

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
federicocappelli: TODO: Add is logged in check before anything

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -274,73 +326,123 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
-        subscriptionEndpointService.getCachedSubscription() != nil
+        subscriptionCachingService.isPresent
+    }
+
+    public func ingestSubscription(_ subscription: DuckDuckGoSubscription) async throws -> DuckDuckGoSubscription {
+        subscriptionCachingService.reset()
+        let enrichedSubscription = try await enrichSubscriptionWithFeatures(subscription)
+        await subscriptionCachingService.set(enrichedSubscription)
+        NotificationCenter.default.post(name: .subscriptionDidChange, object: self, userInfo: [UserDefaultsCacheKey.subscription: enrichedSubscription])
+        return enrichedSubscription
```
miasma13, line 446: **ingestSubscription clears the cache before enrichment can fail**

  reset() runs before enrichSubscriptionWithFeatures — a network call to getSubscriptionTierFeatures that can throw on transient errors. If it does, we exit
  with the cache cleared even though the caller has a valid subscription in hand.

  Callers are confirmPurchase (line 710) and getSubscriptionFrom(lastTransactionJWSRepresentation:) (line 445) — both fire right after the user has just been
   confirmed as subscribed. After a failure here:

  - isSubscriptionPresent() returns false
  - Sidebar takes the no-subscription branch in PreferencesSidebarModel.swift:448
  - iOS Settings VMs reset to no-sub
  - User just paid, app shows no subscription, until next successful getSubscription(forceRefresh: true)

  Same regression class this PR fixes, different trigger.

  Fix: enrich first, then cache+notify; on enrichment failure, cache the raw subscription so isSubscriptionPresent() stays truthful. Drop the leading reset()
   — set overwrites.

  public func ingestSubscription(_ subscription: DuckDuckGoSubscription) async throws -> DuckDuckGoSubscription {
      let enriched: DuckDuckGoSubscription
      do {
          enriched = try await enrichSubscriptionWithFeatures(subscription)
      } catch {
          await subscriptionCachingService.set(subscription)
          NotificationCenter.default.post(name: .subscriptionDidChange, object: self, userInfo: [UserDefaultsCacheKey.subscription: subscription])
          throw error
      }
      await subscriptionCachingService.set(enriched)
      NotificationCenter.default.post(name: .subscriptionDidChange, object: self, userInfo: [UserDefaultsCacheKey.subscription: enriched])
      return enriched
  }

  Test idea: stub getSubscriptionTierFeatures to throw, call confirmPurchase, assert isSubscriptionPresent() is true.

macOS/DuckDuckGo/Preferences/Model/PreferencesSidebarModel.swift
```diff
@@ -409,61 +409,64 @@ final class PreferencesSidebarModel: ObservableObject {
...
+                } catch SubscriptionManagerError.noTokenAvailable {
+                    // Token gone — treat as signed out
+                    updatedState = PreferencesSidebarSubscriptionState(shouldHideSubscriptionPurchase: shouldHideSubscriptionPurchase)
+                } catch {
+                    Logger.general.error("Failed to refresh subscription state, using degraded state: \(error, privacy: .public)")
+                    updatedState = PreferencesSidebarSubscriptionState(
+                        hasSubscription: subscriptionManager.isSubscriptionPresent(),
+                        shouldHideSubscriptionPurchase: shouldHideSubscriptionPurchase
+                    )
+                }
```
miasma13, line 446: **Sidebar flips to "subscription present, no features" on any transient error**

  The generic catch builds a fresh state with hasSubscription from the cache but every is*Enabled/is*Available flag defaulting to false (see
  PreferencesSidebarSubscriptionState.swift:40-49). That state flows into loadSections and decides sub-menu visibility.

  Behavior change introduced by this PR — pre-PR the catch only logged and left currentSubscriptionState untouched.

  Repro: user has VPN/PIR/ITR, currentSubscriptionFeatures(forceRefresh: true) hits a transient 500 → sidebar flips to "Subscription Settings visible,
  sub-items hidden" until the next successful refresh. Same UX symptom class as the bug this PR fixes.

  Fix: keep current state on transient errors. noTokenAvailable is definitive (cache is also cleared in fetchSubscription before it's thrown), so the
  signed-out fallback there is fine; only the generic catch is the problem.

  } catch SubscriptionManagerError.noTokenAvailable {
      updatedState = PreferencesSidebarSubscriptionState(shouldHideSubscriptionPurchase: shouldHideSubscriptionPurchase)
  } catch {
      Logger.general.error("Failed to refresh subscription state, keeping current: \(error, privacy: .public)")
      return
  }

  Test idea: stub currentSubscriptionFeatures(forceRefresh:) to throw a non-noTokenAvailable error; assert currentSubscriptionState is unchanged.

miasma13 (approved):
Tested the happy path and the fix seems to be correct. 
Just these 2 points worth investigating  found by Claude during review.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
federicocappelli: Improvement implemented, thanks!

macOS/DuckDuckGo/Preferences/Model/PreferencesSidebarModel.swift
federicocappelli: Implemented, thanks!

iOS/DuckDuckGo/AppServices/WideEventService.swift
```diff
@@ -76,7 +76,7 @@ actor WideEventService {
 
     private func checkForCurrentEntitlements() async -> Bool {
         do {
-            let entitlements = try await subscriptionManager.currentSubscriptionFeatures()
+            let entitlements = try await subscriptionManager.currentSubscriptionFeatures(forceRefresh: true)
```
diegoreymendez, line 79: This is flipping the behavior, is this intentional?

macOS/DuckDuckGo/Application/WideEventService.swift
```diff
@@ -71,7 +71,7 @@ actor WideEventService {
 
     private func checkForCurrentEntitlements() async -> Bool {
         do {
-            let entitlements = try await subscriptionManager.currentSubscriptionFeatures()
+            let entitlements = try await subscriptionManager.currentSubscriptionFeatures(forceRefresh: true)
```
diegoreymendez, line 74: This is flipping the behavior, is this intentional?

iOS/DuckDuckGo/UnifiedToggleInput/UTIModelStore.swift
```diff
@@ -162,8 +162,9 @@ final class UTIModelStore {
 
     nonisolated func resolveSubscriptionState() async -> SubscriptionState {
         do {
-            let subscription = try await subscriptionManager.getSubscription(cachePolicy: .cacheFirst)
-            guard subscription.isActive, let tier = subscription.tier else {
+            guard let subscription = try await subscriptionManager.getSubscription(),
+                  subscription.isActive,
+                    let tier = subscription.tier else {
```
diegoreymendez: Nit: Indentation looks off here.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
```diff
@@ -215,59 +215,53 @@ final class SubscriptionSettingsViewModel: ObservableObject {
...
     }
 
-    private func fetchAndUpdateSubscriptionDetails(cachePolicy: SubscriptionCachePolicy, loadingIndicator: Bool) async -> Bool {
+    @MainActor
+    private func fetchAndUpdateSubscriptionDetails(forceRefresh: Bool, loadingIndicator: Bool) async -> Bool {
         Logger.subscription.log("Fetch and update subscription details")
         guard subscriptionManager.isUserAuthenticated else { return false }
 
-        if loadingIndicator { self.displaySubscriptionLoader(true) }
+        state.isLoadingSubscriptionInfo = loadingIndicator
```
diegoreymendez, line 236: Is it intentional we're now setting this when `loadingIndicator` is `false` where we previously would not?

Looking at callers I think this was intentionally only managed by UI callers setting this to true.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1517,85 +1517,74 @@ extension SettingsViewModel {
 // MARK: Subscriptions
 extension SettingsViewModel {
 
+    /// Fetches the current subscription from the backend and updates `state.subscription` and the cache.
+    /// Handles three outcomes: active subscription (populates entitlements), nil/no subscription, and no token (unauthenticated).
     @MainActor
     private func setupSubscriptionEnvironment() async {
-        // Create a temporary subscription state to batch all updates
```
diegoreymendez, line 1522: My 2 cents, break this up - a minimalistic bug fix reduces most of the risk introduced by so many behavioral changes.

The changes in the file are quite involved and hard for me to validate as they are - the PR description alone does not explain what I'm seeing here.

diegoreymendez (commented):
Added several comments.  I recommend breaking this up to fix the bug with minimalistic changes and making other improvements step by step, properly documenting why things are being changed.

diegoreymendez:
PS: if you prefer to keep the PR unified, I'd consider adding significant documentation to the larger behavioral changes to explain the "why" so that a reviewer will quickly understand how the change is relevant / necessary.

iOS/DuckDuckGo/AppServices/WideEventService.swift
federicocappelli: This is intentional, from the PR description:
- `WideEventService` (iOS + macOS): `checkForCurrentEntitlements()` uses `forceRefresh: true` to avoid cold-cache mis-attribution during post-purchase wide-event completion at app launch

macOS/DuckDuckGo/Application/WideEventService.swift
federicocappelli: Same as above

iOS/DuckDuckGo/UnifiedToggleInput/UTIModelStore.swift
federicocappelli: fixed, thanks

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionSettingsViewModel.swift
federicocappelli: This is a bug fix, the old code was showing the spinning indicator forever in case of an error thrown in the following do-catch.

iOS/DuckDuckGo/SettingsViewModel.swift
federicocappelli: I understand, the problem is that 2 fundamental underlying changes drive all of these changes:
- Now `subscriptionManager.getSubscription()` can return nil, tha was not the case before and it changes a lot.
- We don't check the auth token directly anymore with `subscriptionManager.getAccessToken()`, but we rely on the `subscriptionManager.getSubscription()` answer for populating the model. This insulates the model from the underlying auth logic.

This isn't a bug fix; it's a change in behaviour due to the underlying framework changes. Splitting these changes into multiple parts isn't feasible but I can explain this step by step if you want.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1517,85 +1517,74 @@ extension SettingsViewModel {
...
+            // 3. Fetch subscription from backend (returns nil if no subscription exists)
+            guard let subscription = try await subscriptionManager.getSubscription() else {
+                // 3a. No subscription on backend — reset subscription fields and exit early
+                Logger.subscription.debug("No subscription data available")
+                applyNoSubscriptionState(&updatedSubscriptionState)
+                DailyPixel.fireDailyAndCount(pixel: .settingsSubscriptionAccountWithNoSubscriptionFound)
+                state.subscription = updatedSubscriptionState
+                subscriptionStateCache.set(state.subscription)
+                return
+            }
```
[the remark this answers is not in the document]
federicocappelli: So guard let subscription = try await subscriptionManager.getSubscription() else { ... fire pixel ... } is semantically equivalent to catching SubscriptionEndpointServiceError.noData directly — which is exactly what the old code did. The abstraction boundary changed but the trigger condition didn't.

The reviewer's scenario — "backend is genuinely unreachable and returned noData erroneously on a prior call that cleared the cache, so the current call fires the pixel" — doesn't hold because:

If the backend is unreachable on the current call, getSubscription() throws a network error, which falls to the generic catch (keep state, no pixel).
The state of the cache from a prior call is irrelevant to whether the pixel fires on the current call. If the prior call cleared the cache and the current call hits the backend and gets a genuine noData, the pixel should fire — the backend is saying the subscription doesn't exist.

federicocappelli:
@diegoreymendez I'm unfortunately running out of time with this, a couple of points for you:
- If you want/can, we can zoom so I can explain live all changes
- Did you have any time to test the VPN behaviour with these changes?

iOS/DuckDuckGo/SettingsViewModel.swift
diegoreymendez: Why not feasible?  This is marked as a high risk change and I see multiple changes that are really not necessary to resolve the core issue.

The core bug is that the API conflates a business state ("user has no subscription") with an error ("fetch failed"). The fix lives at the manager/endpoint boundary, which is ok.

Additional changes, no strict ties to the bug:
- A latent concurrency hazard: `DispatchQueue.sync` inside a struct-based cache. Touches every call site that observes the cache.
- An efficiency optimization: concurrent `getSubscription` calls each do their own network round-trip. Fix is a coalescer. Introduces new cancellation semantics.
- A diagnostic cleanup: splitting `Logger.subscription` into `Logger.subscription` + `Logger.subscriptionTokensManagement`.

diegoreymendez:
I tested in debug and in an alpha build and in both cases this is how the menu looks after the test.  I'll assume this is intended as I can purchase a new subscription from settings, even though it's not what I'd have expected to see.

Screenshot 2026-05-19 at 14 03 05

I'm approving so you can keep moving, but some disclaimers:

1. This seems rather large for a bug fix, especially considering the PR is flagged as high risk.  [I recommended breaking this up](https://github.com/duckduckgo/apple-browsers/pull/3956#discussion_r3236200189).
2. I'm struggling to understand the full extent of the changes looking at what's documented in the PR, so I focused on validating the steps provided in the PR description.

diegoreymendez (approved):

federicocappelli:
@diegoreymendez I must admit I don't remember the hamburger button menu states, but trying in main I see:
Before purchase:
Screenshot 2026-05-19 at 15 02 52
After purchase:
Screenshot 2026-05-19 at 15 07 55
So I'm assuming what you are seeing is correct.

Another state exists in Main, I've seen a "try for free" label and a menu without sub-menu, but I can't reproduce those states reliably. I think this is already a bit messy in production, we can have a look after this PR is merged.

Splitting the PR:
The major changes you are considering "unrelated" are made for fixing bugs or unwanted behaviours caused by the original bug fix, they are indeed not unrelated (a part the logs improvements) and the PR is not mergeable without them.

PR #3957, opened 2026-03-11
Avoid showing browser UI before startup onboarding

afterxleep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1176956903599313/task/1211347552141574
Tech Design URL: N/A
CC: N/A

### Description
- Prevent the first-launch browser UI from flashing before startup onboarding by placing a temporary cover over the already-attached main UI only when onboarding is pending
- Use the actual `LaunchScreen.storyboard` as that cover so the transition matches the real launch screen across device sizes instead of approximating it with a flat color
- Keep the existing tab/home attachment and appearance lifecycle unchanged; onboarding is still presented from the normal startup path
- Centralize startup onboarding decision logic for normal, developer override, and UI-test override paths and add regression coverage

### Testing Steps
1. Install or reset the app so startup onboarding is required, then launch and confirm the launch screen remains visible until onboarding is on screen with no browser or web view flash underneath
2. Complete or skip onboarding and confirm the app transitions into the normal main UI without a blank intermediate state
3. Relaunch after onboarding is marked seen and confirm startup behaves normally with no onboarding cover

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- If the launch screen storyboard fails to instantiate at runtime, startup falls back to a plain background cover and exact visual matching is lost
- If onboarding presentation timing changes later without updating the cover removal callback, the cover could linger briefly or the startup flash could return
- This still touches the first-launch presentation path, so simulator or device verification matters more than unit coverage alone

### Quality Considerations
- Reuses the existing launch screen storyboard instead of duplicating launch-screen layout and assets in code
- Adds unit coverage for `StartupOnboardingDecision`, including incomplete developer and UI-test override cases

---

> [!NOTE]
> **Medium Risk**
> Touches the app startup/onboarding presentation path and view lifecycle, so regressions could cause onboarding not to appear, the cover to linger, or orientation/prompt timing issues. Changes are localized and include new unit tests to reduce risk.
> 
> **Overview**
> Prevents the initial browser UI from briefly appearing before first-launch onboarding by introducing a `StartupOnboardingCover` that overlays the main view with `LaunchScreen.storyboard` while startup onboarding is pending, then removes it after the onboarding controller is presented.
> 
> Centralizes the startup onboarding decision into `StartupOnboardingDecision` (including developer/UI-test overrides), gates onboarding presentation to run once, and updates related logic (e.g., orientation and sync recovery prompt “onboarding complete” checks) to use the new pending-onboarding state. Adds unit coverage for the new decision logic and updates an onboarding intro expectation test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 991df8dd308fb8163a4c652e4f10d3e109035df9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -119,6 +135,11 @@ class MainViewController: UIViewController {
 
     var autoClearInProgress = false
     var autoClearShouldRefreshUIAfterClear = true
+    private var hasLoadedInitialView = false
+    private var isStartupOnboardingPending = false
```
alessandroboron, line 139: Question: Could we achieve the same behaviour using just the object references themselves as state indicators? What I mean is:
1. Replace `isStartupOnboardingPending` with `startupOnboardingCoverView != nil` instead
2. Replace `hasPresentedStartupOnboarding` with a reference to OnboardingIntroViewController? instead. We can set this in `segueToDaxOnboarding()`, and check it in `startOnboardingFlowIfNotSeenBefore()`. This prevents re-presentation on subsequent `viewDidAppear()` calls using the actual view controller reference rather than a separate boolean.

That way, there are fewer variables to update. 
Would this simplification work, or am I missing edge cases that the current flag-based approach handles?

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -557,6 +578,62 @@ class MainViewController: UIViewController {
         mobileCustomization.delegate = self
     }
 
+    private func configureStartupPresentation() {
```
alessandroboron, line 586: nice to have: Would be nice to extract this into a dedicated `StartupOnboardingCover` class that encapsulates the cover lifecycle. Something like

final class StartupOnboardingCover {
      init(parentViewController: UIViewController)
      func attach()
      func bringToFront()
      func detach()
      var isAttached: Bool
  }

That way we keep the MainViewController “simpler”.

iOS/DuckDuckGo/MainViewController.swift
afterxleep: Its a bit more complex, so I’ll keep the flags..

The cover is only a temporary visual mask and we intentionally remove it as soon as onboarding is presented.  But we still need needsToShowOnboardingIntro() to report onboarding as incomplete until markOnboardingSeen() runs in MainViewController, because that state is also used by orientation and sync-recovery.  

 I also wouldn’t replace hasPresentedStartupOnboarding with an onboarding controller reference. That would couple “did we already trigger presentation from this startup path?” to UIKit object lifetime (ontroller reference can be nil because it was dismissed, deallocated, or presentation failed).  The boolean is explicit and cheaper

PR #4078, opened 2026-03-18
[macOS] More Options menu performance improvements

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1213736217542108?focus=true
Tech Design URL:
CC:

### Description

Eliminates a main-thread hang caused by eager bookmark menu rebuilds under memory pressure.

A spindump from a user with a 7.66 GB process footprint showed the main thread blocked in `[NSMenu setItemArray:]` → `_recursivelyNoteChangedIsInMainMenu:` (a recursive AppKit internal walk) triggered by `bookmarkManager.listPublisher`. Under memory pressure, each node in the menu tree caused a page fault and decompression cycle, compounding into a 127-second hang.

The fix defers all menu population to `NSMenuDelegate.menuNeedsUpdate(_:)`, which AppKit calls just before a menu is displayed — never proactively. Three surfaces are covered:

  - **MainMenu bookmarks/favorites** — dirty flags store pending data on publisher fire; `menuNeedsUpdate` rebuilds only when the user opens the menu
  - **Bookmark folder submenus** — new `LazyBookmarkFolderMenuDelegate` populates each folder on first hover, recursing lazily into nested folders
  - **MoreOptionsMenu submenus** — `FeedbackSubMenu`, `ZoomSubMenu`, `BookmarksSubMenu`, `LoginsSubMenu`, `HelpSubMenu` each defer item building to first hover

  All changes are gated behind a `lazyMenuRebuild` feature flag (enabled by default, supports local override).

### Testing Steps
 1. Open the app with a large bookmark collection (100+ bookmarks, several nested folders). You can download 30k bookmarks from [here](https://github.com/duckduckgo/iOS/blob/main/DuckDuckGoTests/bookmarks_30k.html)
 2. Open the **Bookmarks** menu from the menu bar — verify it populates correctly
 3. Hover over a bookmark folder — verify the submenu populates on hover
 4.  Click the **⋯ options button** in the navigation bar — verify all submenus (Feedback, Zoom, Passwords, Help) open and populate correctly
 5. Add or delete a bookmark, then open the Bookmarks menu again — verify the updated list is shown
 6. Check that when opening the ‘More Options’ menu, the spinning ball doesn’t not hang the system
 7. To test the flag-off path: enable local override in Internal Settings to disable `lazyMenuRebuild`, repeat steps 2–5 and verify the same behaviour
 8. With the feature flag off, you will see that when opening the More Options Menu the main thread is blocked.

### Impact and Risks
**Low** — this is a performance improvement to existing menu behaviour. The user-visible result is identical; menus populate on hover rather than proactively.

#### What could go wrong?
  - **Submenu appears empty on hover** — `NSMenu.delegate` is a weak reference; if the delegate is deallocated before the menu opens, `menuNeedsUpdate` is never called and the menu shows only the placeholder. Mitigated by `MainMenu.folderDelegates` retaining top-level delegates and `LazyBookmarkFolderMenuDelegate.childDelegates` retaining nested ones for the lifetime of each rebuild cycle.
  - **Stale data after bookmark change** — if `listPublisher` fires while a menu is open, the dirty flag is set but `menuNeedsUpdate` won't re-fire for the already-open menu. The menu will show stale content until it is closed and reopened. Acceptable for a first iteration.
  - **Flag-off regression** — the old eager path is preserved verbatim behind the flag. If issues are found in production the flag can be disabled remotely without a release.

### Quality Considerations
  - `isPopulated` on each submenu means item-building runs at most once per `MoreOptionsMenu` instance. Since `MoreOptionsMenu` is created fresh on each button click, this is equivalent to the previous behaviour.
  - Folder delegates are cleared (`folderDelegates.removeAll()`) on every full bookmark rebuild, preventing memory accumulation across bookmark list changes.
  - No changes to action handlers, pixels, accessibility identifiers, or keyboard equivalents.

### Notes to Reviewer
  - `LazyBookmarkFolderMenuDelegate` is a new class at the top of `MainMenu.swift` — worth a close read since it owns the recursive lazy folder population.
  - The `childDelegates` property on `LazyBookmarkFolderMenuDelegate` (not in the original design) is intentional: it retains sub-delegates since `NSMenu.delegate` is weak.

—
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes menu-building timing to defer population until `NSMenuDelegate.menuNeedsUpdate(_:)`, which could surface as empty/stale menus if delegates aren’t retained or update flags misfire. Feature-flag gating reduces blast radius but it’s enabled by default, so regressions would affect common UI paths.
> 
> **Overview**
> Improves macOS menu performance by introducing the `lazyMenuRebuild` feature flag and deferring menu population to `NSMenuDelegate.menuNeedsUpdate(_:)` instead of rebuilding eagerly on data publishers.
> 
> In `MainMenu`, bookmark/favorites updates now store pending view models and “dirty” flags, rebuild only when the user opens the menu, and lazily populate bookmark folder submenus via a new `LazyBookmarkFolderMenuDelegate` (with explicit delegate retention).
> 
> In `MoreOptionsMenu`, `FeedbackSubMenu`, `ZoomSubMenu`, `BookmarksSubMenu`, `LoginsSubMenu`, and `HelpSubMenu` now build their items on first hover/open (placeholder item + delegate) when the flag is on, while preserving the old eager path when off; unit tests were added/updated to cover the lazy behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1fd83ae8643243ac1212e4c3a5bfd822893b8e0e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/Menus/MainMenu.swift
+++ b/macOS/DuckDuckGo/Menus/MainMenu.swift
@@ -176,6 +176,7 @@ final class MainMenu: NSMenu {
     let appAboutDDGMenuItem = NSMenuItem(title: UserText.aboutDuckDuckGo, action: #selector(AppDelegate.openAbout))
 
     private let featureFlagger: FeatureFlagger
+    private let isLazyMenuRebuild: Bool
     private let dockCustomizer: DockCustomization?
     private let defaultBrowserPreferences: DefaultBrowserPreferences
     private let aiChatMenuConfig: AIChatMenuVisibilityConfigurable
@@ -215,6 +216,7 @@ final class MainMenu: NSMenu {
          duckAIChromeButtonsVisibilityManager: DuckAIChromeButtonsVisibilityManaging = LocalDuckAIChromeButtonsVisibilityManager()) {
 
         self.featureFlagger = featureFlagger
+        self.isLazyMenuRebuild = featureFlagger.isFeatureOn(.lazyMenuRebuild)
         self.internalUserDecider = internalUserDecider
         self.appearancePreferences = appearancePreferences
         self.privacyConfigurationManager = privacyConfigurationManager
@@ -246,7 +248,7 @@ final class MainMenu: NSMenu {
         subscribeToBookmarkList(bookmarkManager: bookmarkManager)
         subscribeToFavicons(faviconManager: faviconManager)
 
-        if featureFlagger.isFeatureOn(.lazyMenuRebuild) {
+        if isLazyMenuRebuild {
             bookmarksMenu.delegate = self
             favoritesMenu.delegate = self
         }
@@ -629,7 +631,7 @@ final class MainMenu: NSMenu {
             .receive(on: DispatchQueue.main)
             .sink { [weak self] loaded in
                 guard let self, loaded else { return }
-                if self.featureFlagger.isFeatureOn(.lazyMenuRebuild) {
+                if self.isLazyMenuRebuild {
                     self.bookmarkFaviconsNeedUpdate = true
                 } else {
                     self.updateFavicons(in: bookmarksMenu)
@@ -661,7 +663,7 @@ final class MainMenu: NSMenu {
             }
             .sink { [weak self] favorites, topLevel in
                 guard let self else { return }
-                if self.featureFlagger.isFeatureOn(.lazyMenuRebuild) {
+                if self.isLazyMenuRebuild {
                     Task { @MainActor [weak self] in
                         guard let self else { return }
                         self.pendingFavoriteViewModels = favorites
@@ -669,9 +671,9 @@ final class MainMenu: NSMenu {
                         self.bookmarksMenuNeedsRebuild = true
                     }
                 } else {
-                    Task { @MainActor in
-                        self.updateBookmarksMenu(favoriteViewModels: favorites,
-                                                 topLevelBookmarkViewModels: topLevel)
+                    Task { @MainActor [weak self] in
+                        self?.updateBookmarksMenu(favoriteViewModels: favorites,
+                                                  topLevelBookmarkViewModels: topLevel)
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -1330,6 +1411,26 @@ extension NSMenu {
...
+        if bookmarksMenuNeedsRebuild {
+            updateBookmarksMenu(
+                favoriteViewModels: pendingFavoriteViewModels,
+                topLevelBookmarkViewModels: pendingTopLevelViewModels
+            )
+            bookmarksMenuNeedsRebuild = false
+            bookmarkFaviconsNeedUpdate = false
+        } else if bookmarkFaviconsNeedUpdate {
+            updateFavicons(in: menu)
+            bookmarkFaviconsNeedUpdate = false
```
diegoreymendez: Should this be two separate flags?

`bookmarksMenu` and `favoritesMenu` share this single bool — if the user opens one, the flag clears and the other never gets its favicon refresh.

macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -623,8 +689,12 @@ final class MainMenu: NSMenu {
     // Nested recursing functions cause body length
     @MainActor
     func updateBookmarksMenu(favoriteViewModels: [BookmarkViewModel], topLevelBookmarkViewModels: [BookmarkViewModel]) {
+        let isLazy = featureFlagger.isFeatureOn(.lazyMenuRebuild)
```
diegoreymendez: Should we store a value for this flag once, and honor it until the app quits?

My concern with the dynamic checks is they'll likely break if the flag changes at runtime.

jotaemepereira: Good callout. Addresed

diegoreymendez (approved):
Could you please open a new task for the remaining issue where bookmarks are slow to load, and maybe reference it here for traceability?

This is approved as it makes the situation a bit better, but I think we need to give visibility and address the remaining problem which is quite noticeable.

PR #4132, opened 2026-03-22
[iOS] Cancel HTTPS upgrade task when burning

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1213736310619061?focus=true
CC: @miasma13 @edulpn 

### Description
- Fix a crash (SIGKILL) occurring when the fire button triggers data clearing while an HTTPS upgrade navigation is still in-flight, causing WebKit to access invalid internal state during a navigation policy decision
- Store the HTTPS upgrade `Task` in a property so it can be explicitly cancelled during `prepareForDataClearing()`, and add a cancellation guard before processing the upgrade result

### Testing Steps

1. Open several tabs with HTTP sites that would be upgraded to HTTPS (e.g., `http://example.com`)
2. While pages are loading / navigating, tap the fire button to clear data
3. Verify no crash occurs and data is cleared successfully
4. Verify normal HTTPS upgrades still work correctly after clearing and opening new tabs

### Impact and Risks

#### What could go wrong?
- If the cancellation check is hit during a legitimate (non-fire) navigation, the HTTPS upgrade for that navigation would be silently skipped. This is mitigated by the fact that the `httpsUpgradeTask` property is only nilled out in `prepareForDataClearing()`, which also removes the navigation delegate, so no further navigations would occur on that tab anyway.

### Notes to Reviewer
The fix is minimal and scoped to the race condition between fire/data clearing and the async HTTPS upgrade task

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Low risk, localized to `TabViewController` HTTPS-upgrade flow; primary concern is inadvertently skipping an upgrade if cancellation happens outside the intended data-clearing path.
> 
> **Overview**
> Prevents a crash during “Fire”/data clearing by tracking the in-flight HTTPS upgrade `Task` and cancelling it in `prepareForDataClearing()`.
> 
> The HTTPS upgrade path now stores the task in `httpsUpgradeTask` and guards on `Task.isCancelled` before applying an upgrade result, avoiding WebKit navigation policy callbacks after the tab has been torn down.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e5304ca621742294f950bb860381a07e8403f5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


miasma13 (approved):
All worked as expected, was not able to detect any regression. 👍

PR #4187, opened 2026-03-25
Integrate freemium DBP banner into promo queue

jozsef-vesza:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1213554887556414
Tech Design URL:
CC: @selman92 

### Description

This PR adds the following changes for the Freemium PIR banner:
* Promo cycling: it will be displayed for 7 days, then go on a 28-day cooldown.
* Integrates with the new Promo service to avoid clashes with other promotions.

**Key changes:**
- `FreemiumDBPPromotionViewCoordinator` no longer owns an `isHomePagePromotionVisible` concept. Instead it exposes raw eligibility signals (`isFeatureAvailable`, `isDismissed`, `isDisplayWindowExpired`) that the delegate composes into `isEligible`.
- `FreemiumDBPPromoDelegate` conforms to `PromoDelegate` with a `CheckedContinuation`-based `show()` that suspends until user interaction or `hide()`.
- Display window scheduling matches Windows: 7-day display → 28-day cooldown → repeat. Explicit user action (close or proceed) permanently stops the cycle.
- Exposed `isFeatureFlagEnabled` on the `FreemiumDBPFeature` protocol for synchronous eligibility checks.
- Delegate registration is deferred until feature availability settles, giving the PromoService's 1-second buffer window the best chance of including correct eligibility.
- A fast-path in `isEligible` bypasses the async product availability check during an active display window, using only the synchronous feature flag.

### Testing Steps

**Setup:**
1. Debug → Freemium → Override Feature Flag → ON
2. Debug → Freemium → Override Storefront → USA

**Banner shows on NTP:**
1. Relaunch the app -> banner should appear (may need a second fresh NTP on first-ever show due to async timing)

**Close permanently dismisses:**
1. Close the banner
2. Open a new fresh NTP -> banner should not come back
3. Debug → Promo Queue → freemium banner shows "dismissed"

**Proceed permanently dismisses:**
1. Repeat setup steps, relaunch
2. Click proceed on the banner -> banner should not come back

**7-day window → 28-day cooldown:**
1. Repeat setup steps, relaunch, let banner show (don't interact)
2. Debug → Promo Queue → Advance Simulated Date by 1 Week
3. Relaunch → open fresh NTP → banner should NOT show
4. Debug → Promo Queue → freemium banner shows "on cooldown"

**Cooldown expires → new cycle:**
1. From previous test, Debug → Promo Queue → Advance Simulated Date by 1 Month
2. Relaunch → open fresh NTP → banner should show again

### Impact and Risks

Medium: Changes how the freemium DBP banner is scheduled and displayed. The banner is a key part of the freemium conversion funnel.

#### What could go wrong?

- **Banner not showing for eligible users on first NTP after launch:** Freemium eligibility depends on an async product availability check that may not settle before the first trigger is evaluated. Mitigated by delayed delegate registration and a fast-path that bypasses the async check during an active display window. Worst case: banner shows on the second fresh NTP.
- **NTP reuse:** The `.newTabPageAppeared` trigger only fires for fresh NTPs, not reused ones. If the first NTP misses due to async timing and the user reopens the same NTP, the banner won't appear until a fresh NTP is created (e.g. after navigating to a website first). This is a pre-existing promo system limitation, not introduced by this PR.

### Quality Considerations

- Added tests to verify the new behavior
- Added debug tooling: "Advance by 1 Week" and "Advance by 1 Month" options in the Promo Queue debug menu, and wired the shared simulated date store into the coordinator

### Notes to Reviewer

N/A

---

> [!NOTE]
> **Medium Risk**
> Changes promo scheduling/eligibility and session timeout handling in `PromoService`, which can affect when/if promotions appear or persist across restarts (including freemium DBP conversion surfaces). Adds some Xcode project entries with `(null)` build files that may cause build configuration issues if not intended.
> 
> **Overview**
> **Integrates the Freemium DBP NTP banner into the promo queue.** Adds a new `PromoServiceFactory.freemiumDBP` promo with a **7‑day display timeout** and **28‑day cooldown**, and registers a new `FreemiumDBPPromoDelegate` from the NTP actions manager to control banner visibility and report results back to `PromoService`.
> 
> **Refactors the freemium DBP banner coordinator to be promo-service driven.** `FreemiumDBPPromotionViewCoordinator` drops its own visibility/dismissal state, now exposing raw signals (`isFeatureAvailable`, legacy dismissal/scan results, callbacks) plus explicit `refreshViewModel()`/`clearViewModel()` for the delegate-controlled show/hide lifecycle; `FreemiumDBPFeature` gains synchronous `isFeatureFlagEnabled` for eligibility checks.
> 
> **Improves debug/test tooling and timeout correctness.** `PromoDebugMenu` can advance simulated time by week/month and now triggers `PromoService.reconcileActivePromoTimeoutsAfterSimulatedDateAdvance()`, while `PromoService` tracks a logical `timeoutDeadline` so active promos can be expired when simulated time jumps; unit tests are added/updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1690afb6232fe12e10bd028c05d70c06495f4f1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/NewTabPage/NewTabPageActionsManagerExtension.swift
+++ b/macOS/DuckDuckGo/NewTabPage/NewTabPageActionsManagerExtension.swift
@@ -159,6 +159,8 @@ extension NewTabPageActionsManager {
             syncService: syncService,
             dockCustomization: dockCustomization
         )
+        let buildType = StandardApplicationBuildType()
+
         if let promoService {
             let coordinator = freemiumDBPPromotionViewCoordinator
 
@@ -167,7 +169,6 @@ extension NewTabPageActionsManager {
             // to fast-path eligibility during an active display window,
             // avoiding the wait for async product availability checks.
             let dateProvider: () -> Date
-            let buildType = StandardApplicationBuildType()
             if buildType.isDebugBuild || buildType.isReviewBuild {
                 let debugDateStore = DebugSimulatedDateStore(keyValueStore: keyValueStore)
                 dateProvider = { debugDateStore.simulatedDate ?? Date() }
@@ -187,7 +188,6 @@ extension NewTabPageActionsManager {
             promoService.setDelegate(for: PromoServiceFactory.nextSteps.id, delegate: nextStepsDelegate)
         }
 
-        let buildType = StandardApplicationBuildType()
         let environment: NewTabPageConfigurationClient.Environment = (buildType.isDebugBuild || buildType.isReviewBuild) ? .development : .production
 
         self.init(scriptClients: [
--- a/macOS/UnitTests/Freemium/DBP/FreemiumDBPPromoDelegateTests.swift
+++ b/macOS/UnitTests/Freemium/DBP/FreemiumDBPPromoDelegateTests.swift
@@ -120,15 +120,6 @@ final class FreemiumDBPPromoDelegateTests: XCTestCase {
         _ = await task.value
     }
 
-    func testShow_returnsIgnoredWhenDisplayWindowExpired() async {
-        var history = PromoHistoryRecord(id: "test")
-        history.lastShown = Date().addingTimeInterval(-.days(8))
-
-        let result = await sut.show(history: history, force: false)
-
-        XCTAssertEqual(result, .ignored(cooldown: .days(28)))
-    }
-
     func testShow_suspendsAndResumesWithActionedOnProceed() async {
         let task = Task {
             await sut.show(history: PromoHistoryRecord(id: "test"), force: false)


macOS/DuckDuckGo/Freemium/DBP/FreemiumDBPPromotionViewCoordinator.swift
```diff
@@ -28,13 +28,37 @@ import Common
...
+    /// Whether the freemium DBP feature is currently available (includes async checks like product availability).
+    @Published private(set) var isFeatureAvailable: Bool = false
+
+    /// Whether the feature flag is enabled (synchronous — no async dependencies).
+    var isFeatureFlagEnabled: Bool {
+        freemiumDBPFeature.isFeatureFlagEnabled
+    }
+
+    /// Whether the user has dismissed the promotion.
+    @Published private(set) var isDismissed: Bool = false
```
rachelmcr: I'm running into an issue when I test the promo dismissal flow, where step 3 (Debug → Promo Queue → freemium banner shows "dismissed") doesn't work. Here's what's happening:

With `PromoService`, it's best not to independently track and use promo dismissal for eligibility checks. The service itself will keep track of when the promo is dismissed and ensure the promo isn't shown again.

Otherwise, using `isDismissed` as part of the eligibility check can make the promo become ineligible to be shown within `PromoService`. That makes the queue withdraw the promo, which is a different flow from the user dismissing the promo:

* When a promo becomes ineligible to be shown (e.g. because the feature becomes unavailable), the service calls `handleEligibilityLost(promoId:)` and a `.noChange` result is recorded. That allows the promo to be shown again if it's eligible, or other promos can be immediately shown with no cooldown.
* When a promo is explicitly dismissed by the user, the promo's `show()` result is recorded.
* When a promo becomes ineligible at the same time as an explicit user dismissal, either the `.noChange` result or the intended dismissal result (i.e. `.ignored()`) could be recorded in history.

If you need to account for historical dismissals in the promo eligibility (e.g. if the user dismissed the promo before it was added to the queue) that's fine, but you shouldn't update that value when the user dismisses the promo now that it's in the queue. The service will keep track of the dismissal for you and won't try to show the promo again.

Let me know if any of that is unclear! I'll update the queue documentation to clarify that point.

macOS/DuckDuckGo/NewTabPage/Features/NewTabPageFreemiumDBPBannerProvider.swift
```diff
@@ -73,3 +74,80 @@ extension NewTabPageDataModel.FreemiumPIRBannerMessage {
...
+        if coordinator.displayWindowStartDate == nil {
+            coordinator.displayWindowStartDate = coordinator.dateProvider()
+        }
+
+        coordinator.updateDisplayWindowExpiredState()
+        if coordinator.isDisplayWindowExpired {
+            coordinator.displayWindowStartDate = nil
+            coordinator.updateDisplayWindowExpiredState()
+            return .ignored(cooldown: .days(28))
+        }
```
rachelmcr: Instead of tracking `coordinator.isDisplayWindowExpired` independently, you can use the `PromoHistoryRecord` to check the promo's display history and determine if it needs to go into cooldown.

In this case, you can use `history.lastShown` to get the date the promo was initially shown, and check if the display window (7 days) has passed, for example:

```swift
if let promoShownDate = history.lastShown,
    Date().timeIntervalSince(promoShownDate) >= .days(7) {
    return .ignored(cooldown: .days(28))
}
```

macOS/DuckDuckGo/Promotions/Promos/PromoServiceFactory+FreemiumDBP.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// Freemium DBP banner promo. Delegate is set when NTP is built.
+    static let freemiumDBP = Promo(
+        id: "freemium-dbp-ntp-banner",
+        triggers: [.newTabPageAppeared],
+        initiated: .app,
+        promoType: PromoType(.remoteMessage),
+        context: .newTabPage,
+        coexistingPromoIDs: [],
+        respectsGlobalCooldown: false,
+        setsGlobalCooldown: false,
```
rachelmcr: If you need to set these values, could you document why?

The global cooldown helps the queue prevent too many promos from being shown in the same session: Not respecting the global cooldown means this promo could be shown immediately after another promo is dismissed, and not setting the global cooldown means another promo could be shown immediately after this one is dismissed. Not that these can't be overridden, but I'm not sure it's needed for this promo.

macOS/DuckDuckGo/Promotions/PromoDebugMenu.swift
```diff
@@ -141,6 +141,14 @@ final class PromoDebugMenu: NSMenu {
         advanceDayItem.setAccessibilityIdentifier(AccessibilityIdentifiers.PromoQueue.advanceSimulatedDate1Day)
         addItem(advanceDayItem)
 
+        let advanceWeekItem = NSMenuItem(title: "Advance Simulated Date by 1 Week", action: #selector(advanceSimulatedDateByWeek), keyEquivalent: "")
+        advanceWeekItem.target = self
+        addItem(advanceWeekItem)
+
+        let advanceMonthItem = NSMenuItem(title: "Advance Simulated Date by 1 Month", action: #selector(advanceSimulatedDateByMonth), keyEquivalent: "")
+        advanceMonthItem.target = self
+        addItem(advanceMonthItem)
```
rachelmcr, line 150: Nice, thanks for adding these! We could also reuse the "Override today's date..." menu item from `DefaultBrowserAndDockPromptDebugMenu` if we want to more easily support a variety of date changes, but this works too.

macOS/DuckDuckGo/Freemium/DBP/FreemiumDBPPromotionViewCoordinator.swift
jozsef-vesza: I think I get it: only use the value for checking historical dismissals, but defer dismiss handling to the `PromoService`. I can make that update. The only tricky bit I just remembered, is that users need to become eligible again if free scan results come in (we display a different variant of the banner). IIRC I can use `refreshEligibility` for that.

rachelmcr: Yeah, `refreshEligibility` can be used to do any work to refresh/recompute eligibility criteria before it's evaluated again.

macOS/DuckDuckGo/NewTabPage/Features/NewTabPageFreemiumDBPBannerProvider.swift
jozsef-vesza: This makes sense. The only caveat is that I'm currently using that date to trigger the 'fast path' on subsequent launches. Will see if I can use data from the promo service for that.

rachelmcr: You can use `PromoHistoryProviding.historyPublisher(for:)` to subscribe to the promo service's history for the promo, and use it as needed.

macOS/DuckDuckGo/Promotions/Promos/PromoServiceFactory+FreemiumDBP.swift
jozsef-vesza: Yeah, I don't believe it's needed. Will remove 👍

macOS/DuckDuckGo/NewTabPage/Features/NewTabPageFreemiumDBPBannerProvider.swift
jozsef-vesza: After a bit further thought, I don't think `history.lastShown` can work for a visibility window. If I understand correctly, `PromoHistoryRecord`'s `lastShown` will be updated every time `show(...)` is called on a promo delegate. 

So for something like a visibility window I'm seeing the following options:
* Add `firstShown` to `PromoHistoryRecord`
* Keep the separate tracking for now

I'm leaning towards the former as it might help other, future promos too.

rachelmcr: There are two scenarios with showing a promo:

* If `show(...)` is called when the promo is triggered (i.e. the promo wasn't already visible), `lastShown` is recorded.
* If `show(...)` is called because the promo was visible and is being restored (i.e. on app restart), `lastShown` is not recorded. In this case, `lastShown` continues to be when the promo was initially triggered.

That way, the promo knows when it was initially triggered for the user, and `lastShown` will measure how long the promo has been on screen for the user. Other scenarios:

* The user dismisses the promo (permanently dismissed)
* The user acts on the promo (permanently dismissed)
* The promo eligibility changes (retracted due to ineligibility) — in this case, if it becomes eligible again and is triggered, it will be shown and `lastShown` will update to the new date it is re-shown.

I believe that's what this promo needs, since it's meant to be displayed for 7 days at a time. That said, if for whatever reason that's not what this promo needs, I agree that enhancing `PromoHistoryRecord` to include what's needed is a good approach. 👍

macOS/DuckDuckGo/NewTabPage/Features/NewTabPageFreemiumDBPBannerProvider.swift
```diff
@@ -73,3 +74,124 @@ extension NewTabPageDataModel.FreemiumPIRBannerMessage {
...
+    }
+
+    @MainActor
+    func show(history: PromoHistoryRecord, force: Bool) async -> PromoResult {
+        resumeContinuation(with: .noChange)
+
+        // Honor legacy dismissals from before queue integration.
+        if coordinator.hasLegacyDismissal && coordinator.firstScanResults == nil && !force {
+            return .ignored()
+        }
```
rachelmcr: A slightly cleaner approach would be to include this in the `isEligible` check, so the promo is never eligible to be shown if it has been dismissed in the past. That way the promo queue doesn't try to show it and it won't impact other promos in the queue.

macOS/DuckDuckGo/NewTabPage/Features/NewTabPageFreemiumDBPBannerProvider.swift
```diff
@@ -73,3 +74,124 @@ extension NewTabPageDataModel.FreemiumPIRBannerMessage {
...
+
+        // Display window: if the promo has been shown for 7+ days, enter cooldown.
+        if let promoShownDate = history.lastShown,
+           dateProvider().timeIntervalSince(promoShownDate) >= .days(7) {
+            return .ignored(cooldown: .days(28))
+        }
```
rachelmcr: Apologies, it just occurred to me that I should have suggested a simpler way to handle this:

In the `Promo` struct, for `PromoType`, you can set a custom timeout for the promo. That way the promo can timeout on its own and start the cooldown period, and you don't need this cooldown check here. That also makes sure the promo is dismissed immediately once the 7 day window ends, separate from the queue calling this `show(...)` method.

See my comment below for a suggested `PromoType` definition to handle that.

macOS/DuckDuckGo/Promotions/Promos/PromoServiceFactory+FreemiumDBP.swift
```diff
@@ -0,0 +1,33 @@
...
+import Foundation
+
+extension PromoServiceFactory {
+
+    /// Freemium DBP banner promo. Delegate is set when NTP is built.
+    static let freemiumDBP = Promo(
+        id: "freemium-dbp-ntp-banner",
+        triggers: [.newTabPageAppeared],
+        initiated: .app,
+        promoType: PromoType(.remoteMessage),
```
rachelmcr: Here you can set a custom timeout on the promo so it handles entering the cooldown without any additional work needed:


```suggestion
        promoType: PromoType(.remoteMessage, customTimeoutInterval: .days(7), customTimeoutResult: .ignored(cooldown: 28)),
```

macOS/DuckDuckGo/Promotions/Promos/PromoServiceFactory+FreemiumDBP.swift
```diff
@@ -0,0 +1,33 @@
...
+import Foundation
+
+extension PromoServiceFactory {
+
+    /// Freemium DBP banner promo. Delegate is set when NTP is built.
+    static let freemiumDBP = Promo(
+        id: "freemium-dbp-ntp-banner",
+        triggers: [.newTabPageAppeared],
+        initiated: .app,
+        promoType: PromoType(.remoteMessage, customTimeoutInterval: .days(7), customTimeoutResult: .ignored(cooldown: 28)),
```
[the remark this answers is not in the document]
rachelmcr: Sorry, this was a typo on my part; I'll push a fix.

PR #4296, opened 2026-04-01
[htang][Claude]Add wideEvent-starter Claude skill

hanyutang-sandra:
## Summary
- Adds a Claude Code skill for end-to-end wide event creation in the apple-browsers repo
- Supports both JSON5 (pixels/definitions/) and JSON (wide_events/definitions/) definition formats
- Includes guides for Swift WideEventData class + XCTestCase creation
- Uses auth-v2 token refresh as the consistent example across all references

## Skill Structure
```
.claude/skills/wideEvent-starter/
├── SKILL.md                              — Entry point: MCP fetch → format choice → spawn subagent
└── references/
    ├── definition/                       — Action 1: Wide Event Definition
    │   ├── json5-guide.md                — JSON5 pixels/ format + template
    │   ├── json-guide.md                 — JSON wide_events/ format + template
    │   ├── parameter-formats.md          — Shared parameter syntax (both formats)
    │   └── examples/
    │       ├── json5-auth-v2-token-refresh.json5
    │       └── json-auth-v2-token-refresh.json
    └── data-class/                       — Action 2: Data Class & Tests
        ├── guide.md                      — WideEventData protocol + class structure
        ├── test-guide.md                 — Test class boilerplate + runner
        └── examples/
            ├── AuthV2TokenRefreshWideEventData.swift
            └── AuthV2WideEventTests.swift
```

## How it works
1. Skill triggers when user mentions "wide event", "wide event definition", or "wide event data class"
2. Main conversation fetches the wide event format from Asana (MCP call — subagents can't do MCP)
3. Asks user which format (JSON5 or JSON)
4. Spawns a subagent with the format context to create the definition and/or data class

## Test plan
- [ ] Trigger skill by saying "I need to create a new wide event for X"
- [ ] Verify it fetches the wide event format from Asana
- [ ] Verify it asks which format (JSON5 or JSON)
- [ ] Verify subagent creates a definition following the correct guide
- [ ] Verify data class action creates Swift class + tests

---

> [!NOTE]
> **Low Risk**
> Adds only Claude skill documentation/reference materials (guides + examples) without changing production app code. Risk is limited to developer workflow/consistency issues if the guidance is incorrect.
> 
> **Overview**
> Introduces a new Claude Code skill, `wideEvent-starter`, that standardizes the end-to-end workflow for creating *wide events* in the apple-browsers repo, including an MCP-driven step to fetch the canonical format from Asana and a handoff to a subagent.
> 
> Adds reference docs for both definition formats—**legacy JSON5** under `pixels/definitions/` and **nested JSON** under `wide_events/definitions/`—including templates, platform notes, validation commands, and a shared parameter-format reference.
> 
> Includes complete example artifacts (AuthV2 token refresh) for a wide event definition in both formats, plus a sample `WideEventData` Swift class and corresponding XCTest cases (including latency bucketing assertions) to serve as copy/paste starting points.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 83f7396b78b1dd035f1c467c5aa3bb7ae5d7b778. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.claude/skills/wideEvent-starter/SKILL.md
+++ b/.claude/skills/wideEvent-starter/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: wideEvent-starter
+description: End-to-end wide event creation for apple-browsers — from wide event definition (JSON5 or JSON format) to Swift data class and tests. Use this skill whenever the user wants to define a new wide event, create a wide event definition, create a wide event data class, add a wide event to PixelDefinitions, or provides an Asana link to a wide event task. Also trigger when the user mentions "wide event", "wide event definition", "wide event data class", or wants to add a definition to pixels/definitions/ or wide_events/definitions/ for a wide event, even if they don't explicitly say "definition".
+---
+
+# Wide Event Starter Skill
+
+This skill is instructions and reference material — not an agent itself. Subagents don't have access to MCP servers (like Asana), so the main conversation must handle MCP calls first, then pass the results to a subagent.
+
+**When this skill triggers, follow this sequence:**
+
+1. **Main conversation: Gather context** — Do the MCP calls and user questions yourself (Steps 1–2 below). Subagents can't access MCP servers.
+2. **Spawn subagent: Do the work** — Once you have the wide event format and the user's preferences, spawn a subagent briefed with:
+   - The user's request (what wide event they need, platform, format preference)
+   - The wide event format (the full text fetched from Asana or pasted by the user)
+   - Any Asana task details (if the user provided a task link, fetch it and include the task description)
+   - The skill path: `~/.claude/skills/wideEvent-starter/`
+   - Instructions to read the relevant guide and follow Actions 1–2 below
+
+The subagent writes definitions, creates Swift classes, runs tests, and reports back. If it needs more user input, it returns to the main conversation.
+
+Wide event definitions (both JSON5 and JSON formats) are validated and used to generate a JSON Schema that the backend checks incoming events against. If the definition is wrong or incomplete, the backend will **drop the event** — so correctness matters.
+
+Two actions, usable in sequence or independently:
+
+| Action | What it does | Reference |
+|--------|-------------|-----------|
+| **Wide Event Definition** | Create the JSON5 or JSON definition file | `references/definition/` |
+| **Data Class & Tests** | Create the Swift WideEventData class + XCTestCase | `references/data-class/` |
+
+**Full flow:** Wide Event Definition → Data Class & Tests (+ SwiftLint) → Git Commit. At each transition, ask what the user wants next.
+
+**Partial flow:** Figure out what exists and what the user is asking for, then run the relevant action.
+
+---
+
+## Before Spawning the Subagent (main conversation)
+
+### Step 1 — Fetch the Wide Event Format
+
+Run once per session, before writing the first definition. If the format has already been resolved in this conversation, reuse what's in context.
+
+**Canonical format task ID:** `1211144309209145` ([Wide event format](https://app.asana.com/1/137249556945/project/1212203965891161/task/1211144309209145))
+
+1. **Fetch from Asana:**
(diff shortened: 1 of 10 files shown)


.claude/skills/wideEvent-starter/references/definition/examples/json-auth-v2-token-refresh.json
```diff
@@ -0,0 +1,69 @@
+{
+    "ios-authv2-token-refresh": {
```
samsymons, line 2: What do you think about instead pointing Claude at the `PixelDefinitions` directory directly if it wants to see an example? My main concern is that this file will get out of date as the wide event schema evolves, whereas we know the definitions in the main directories will get updates as otherwise validation will stop working.

.claude/skills/wideEvent-starter/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+
+**If the task description seems incomplete or ambiguous:** Ask the user for clarification rather than guessing.
+
+### Step 3 — Choose Definition Format
+
+Ask the user:
+
+> Which format would you like?
+> **(a)** JSON5 under `pixels/definitions/` (legacy flat-parameter format)
+> **(b)** JSON under `wide_events/definitions/` (new nested-object format)
```
samsymons, line 62: I think we should enforce it creating both, since we are planning to move to the JSON format for all events and deprecate the pixel path, but aren't ready to make the cutover yet. It will be easy to make that change if developers have been creating definitions for both, which should be cheap to do thanks to this skill.

.claude/skills/wideEvent-starter/references/definition/json-guide.md
```diff
@@ -0,0 +1,195 @@
+# JSON Wide Event Definition Guide (wide_events/)
+
+This guide covers the **new JSON nested-object format** stored under `wide_events/definitions/`.
```
samsymons, line 3: I like the content of this though I worry we're now getting duplication with the existing `pixel-definitions.mdc` file in the Cursor rules directory. That file also contains guidance around wide events, though it's not as detailed as this.

I think we could see if there are any pieces from the rules file that would be worth including here (probably not, this looks comprehensive) and then clean up the wide event section of the rules file.

samsymons:
Adding @ladamski for a Privacy Team review, since this skill touches wide event definitions.

.claude/skills/wideEvent-starter/references/definition/examples/json-auth-v2-token-refresh.json
hanyutang-sandra: imo it’s a tradeoff. `PixelDefinitions/` is actively maintained and updated, but it consists of many files and each read requires quite a lot of tokens (ofc. if we don’t care about token efficiency this is less of a problem) and some delays.

Another concern I have is that different files use slightly different patterns or format, so it may confuse the agent which one is the best example to follow, and depending on which file it picks up and learns from, the output is not as consistent each time.

But yeah using examples here means that it may get outdated and stale, so there is a maintenance cost

.claude/skills/wideEvent-starter/SKILL.md
hanyutang-sandra: Yeah make sense. 

On the topic of both formats (and related to productions wide event), it’s less favorable to maintain both long term. Eventually we should probably migrate towards option B

.claude/skills/wideEvent-starter/references/definition/json-guide.md
hanyutang-sandra: imo the most ideal config is to have shared references .md files for both cursor and Claude Code (and other coding agent should anyone wants to use - codex for example). And add pointers at tool-specific files to point to them. For example:

On Claude Code SKILL.md
`Read references/definition/json-guide.md…`

On Cursor .mdc
`Read references/definition/json-guide.md…`

This way we can maintain an org-wide standard while adapting to different agent of choice.

PR #4308, opened 2026-04-01
[iOS] Implement filter logic to only display final destination URL

edulpn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1213973295528872?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1213882661942199
CC: @brindy @hassaanelgarem @SabrinaTardio @mallexxx @samsymons 

### Description

Hides intermediate redirect URLs from the iOS address bar during server redirect chains, matching existing macOS behavior. When a user clicks a SERP ad, the address bar no longer flashes bing.com (or other intermediate domains) before the final destination loads.

This is also address bar spoofing protection: a redirect through a trusted domain could briefly display a misleading URL. Firefox iOS and macOS DuckDuckGo both use this approach (filtering by committed security origin).

### Testing Steps
- Redirect filtering on SERP ads
    - Go to `duckduckgo.com`, search for a broad term ("shoes", "insurance")
    - Click a sponsored result (marked with "Ad" badge)
    - Verify: the address bar should NOT flash `bing.com` during the redirect. It should go straight to the final destination.
- Cross-origin redirect chain (`privacy-test-pages`)
    - Navigate to `https://privacy-test-pages.site/security/address-bar-spoofing/index.html`
    - Click "Redirect Chain (cross-origin, 3s delay)"
    - This will redirect users to `first-party.site`, which redirects back to `privacy-test-pages.site` after a 3s delay for easiness of verification.
    - Verify: during the 3-second delay, the address bar should NOT show `first-party.site`. It should stay on privacy-`test-pages.site` until the final destination loads.
- User-interactive navigation cases: These can also be done with the ad link from step 1 to verify that we don’t over filter addresses the user inputs/interacts with on purpose.
    - Type/copy-paste `https://edulpn.github.io/` in the address bar and press Enter. Verify the address bar updates immediately.
    - From `https://edulpn.github.io/`, type/copy-paste `https://privacy-test-pages.site/security/address-bar-spoofing/index.html` in the address bar. Tap Back, then Forward. Verify the address bar shows the correct URL after each navigation.
    - On `https://privacy-test-pages.site/security/address-bar-spoofing/index.html`, click any test page link. Verify the address bar updates normally.
    - Navigate to any page and refresh. Verify the page reloads and the address bar shows the correct URL.

### Impact and Risks

#### What could go wrong?
- Address bar shows stale URL: If the filter incorrectly blocks an update, the address bar could show the previous page's URL until `didCommit` fires. 
    - Mitigated by: user-initiated navigations bypass the filter entirely, and the fail-safe feature flag allows remote disable.
- Address bar shows intermediate redirect URL: If a new navigation entry point is added without tagging it as user-initiated, it would be filtered (address bar updates at commit) rather than leaking a redirect URL. This is the safe failure direction.

### Quality Considerations
- **Security**: This is security hardening. Closes an address bar spoofing vector matching macOS behavior.
- **Privacy**: No change to data collection.
- **Edge cases**: Same-document navigation (`#links`) works because the security origin matches. Custom URL schemes (`about:`, `duck://`) always pass through. When no committed origin exists, updates are blocked (matching macOS).

### Notes to Reviewer
- `load(urlRequest:)` intentionally does NOT set `isUserInitiatedNavigation` because it's also called from `decidePolicyFor` for web-driven navigation reissuance (link rewriting, referrer trimming, GPC). The 5 tagged entry points are: `load(url:)`, `load(backForwardListItem:)`, `reload()`, `goBack()`, `goForward()`.
- The `legacySameHostFallback` parameter in `shouldUpdateAddressBar` preserves exact legacy KVO behavior when the flag is off. This can be removed when the flag is removed.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes tab navigation URL-update behavior to suppress some web-driven URL updates until navigation commit, which could cause the address bar to appear stale in edge cases despite the failsafe feature flag.
> 
> **Overview**
> Prevents intermediate cross-origin redirect URLs from briefly appearing in the iOS address bar by gating address-bar updates on committed `SecurityOrigin` during web-driven navigations.
> 
> Adds `AddressBarURLFilter` and integrates it into `TabViewController` for user navigation/reload/back/forward and server-redirect handling, behind a new failsafe feature flag `filterAddressBarUpdates` wired to `iOSBrowserConfigSubfeature.filterAddressBarUpdates`.
> 
> Introduces unit tests for the filter logic and a Maestro security test covering address-bar spoofing/redirect-chain behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 207139cde2a9a3d36e0ffb875b0a96ee638cf0e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/WebProgressWorker.swift
```diff
@@ -20,44 +20,46 @@
...
+
     weak var progressBar: ProgressView? {
         didSet(oldValue) {
             oldValue?.hide()
-            
+
             if isLoading {
-                self.progressBar?.show(initialProgress: currentProgress)
+                // Cap initial progress below 1.0 so finishAndHide can always animate to completion
+                let cappedProgress = min(currentProgress, 0.9)
```
dus7: I’d suggest to cap this at a value which will not cause the progress bar to be artificially visible. I’ve tried capping that at 0.01 and it worked - the initial state was not visible and the animation was able to finish. I’d suggest to use that value instead, so there’s no visible change to how progress bar behaves.

edulpn: I also tested with different values (both 0.1 and 0.99) and it doesn’t really matter given the fact that the bar is always (as far as I was able to debug) not visible at this point and will only be set at 0.9 if `currrentProgress` is higher than that. 
At 0.01, the progress bar would always appear nearly empty when attached late, then have to animate all the way from 1% to 100%. For a page that’s already almost done loading, that would look "wrong”.
0.9 makes sense because it shows the bar near completion (which it is), and the remaining 10% gives finishAndHide enough room to animate to 100% and trigger the hide. Visually it looks like the page is almost done, which is accurate.
Did you notice any jarring behaviour with 0.9?

dus7 (approved):
LGTM

iOS/DuckDuckGo/AddressBarURLFilter.swift
```diff
@@ -0,0 +1,72 @@
...
+        if newURL.isCustomURLScheme() {
+            return true
+        }
+
+        guard let committed = committedSecurityOrigin, !committed.isEmpty else {
+            return false
+        }
+
+        return newURL.securityOrigin == committed
+    }
```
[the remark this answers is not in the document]
edulpn: This was a stray from changes requested from the TD.

iOS/DuckDuckGo/WebProgressWorker.swift
dus7: For the record: we’ve agreed to not include a fix for this bug as part of this PR as it’s a pre-existing issue and proposed change turned out to be just a way to mask the issue, not fixing the root cause.

PR #4381, opened 2026-04-08
Migrate .feature() flag sources to .subfeature()

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200194497630846/task/1213820379759753
Tech Design URL:
CC:

### Description

This PR does two things:

1. **Migrate** all 20 `.feature(PrivacyFeature)` flag usages to `.subfeature(PrivacySubfeature)` across iOS, macOS, and SharedPackages.
2. **Fix** a parent kill-switch bypass that the migration would otherwise expose.

#### 1. Migration

Using `.feature()` as a feature flag source silently disables rollout, targeting, and cohort support — developers don't realize this until after shipping. This migration ensures all flags go through `.subfeature()`, which supports the full feature flag lifecycle.

Changes:
- Add 13 new subfeature enums and `featureEnabled` case to 3 existing enums in `PrivacyFeature.swift`
- Migrate all `.feature()` references in iOS (11), macOS (8), and SharedPackages (1) feature flag definitions
- Update all tests and doc comments referencing `.feature()`
- Fix `defaultValue` regressions for `adBlockingExtension` (iOS + macOS) and `delayedWebviewPresentation` (macOS) so they match the live privacy config

Part 1 of 2: Migration + bug fix → Remove `PrivacyConfigFeatureLevel`

#### 2. Parent kill-switch bypass fix

Without this fix, when a subfeature is *absent* from privacy config, `stateFor(subfeature:)` returned `.disabled(.featureMissing)` without consulting the parent. Callers then fell through to the flag's `defaultValue`, so a flag declared with `defaultValue: .enabled` would turn ON even if the parent feature was killed in config.

Both public `stateFor` overloads in `AppPrivacyConfiguration.swift` now resolve the parent state first. If the parent isn't `.enabled`, we return the parent's disabled reason directly and never touch `defaultValue`.

**Behavior table** — *parent off ⇒ children off, regardless of `defaultValue`; only exception is `state: "internal"` + internal user, which keeps the existing fallback*:

| Parent state | User | Subfeature in config | Result |
|---|---|---|---|
| `disabled` | any | missing | `.disabled(.disabledInConfig)` — overrides `defaultValue` |
| `internal` | non-internal | missing | `.disabled(.limitedToInternalUsers)` — overrides `defaultValue` |
| `internal` | internal | missing | `.disabled(.featureMissing)` → falls through to `defaultValue` (unchanged) |
| `enabled` | any | missing | `.disabled(.featureMissing)` → falls through to `defaultValue` (unchanged) |

Three tests in `AppPrivacyConfigurationTests` pin each row above. The `internal` rows mirror the live config for `adBlockingExtension` (parent `webExtensions` is `internal`), which is why this matters for the migrated flags in this PR.

### Testing Steps

1. Build iOS and macOS targets — verify compilation succeeds
2. Run `DefaultFeatureFlaggerTests` — verify all tests pass
3. Run `AppPrivacyConfigurationTests` — verify all tests pass (including the three new parent-kill-switch tests)
4. Grep for `.feature(.` in Swift code — confirm zero remaining usages

### Impact and Risks

Low for the migration itself: each flag's `defaultValue` is set to match the live privacy config, so users see the same behavior whether resolution comes from JSON or fallback.

The kill-switch fix is a behavior change in one specific case: parent disabled (or `internal` for non-internal users) + subfeature missing from config. Previously this fell through to `defaultValue`; now it correctly returns disabled. This is the desired behavior.

#### What could go wrong?

A flag whose parent was disabled in config but relied on `defaultValue: .enabled` to remain on for users will now be disabled. This is the bug being fixed — any flag in that state was already misconfigured.

### Quality Considerations

- `defaultValue` for each migrated flag now matches the live privacy config (verified against `staticcdn.duckduckgo.com/trackerblocking/config/v4/{ios,macos}-config.json`)
- New unit tests cover all four parent-state × user-type combinations for the missing-subfeature case

### Notes to Reviewer

- The `.feature` case on `PrivacyConfigFeatureLevel` is intentionally left in place — it will be removed in the follow-up PR after this lands
- Each new subfeature enum has a single `featureEnabled` case as the migration target

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes core feature-flag resolution and privacy-config subfeature state behavior, which can alter rollout/kill-switch outcomes for flags missing from config. Risk is mitigated by targeted unit tests covering the new parent-override cases and updated defaults.
> 
> **Overview**
> Migrates remote feature-flag sources from `PrivacyConfigFeatureLevel.feature` to `PrivacyConfigFeatureLevel.subfeature` across iOS, macOS, and shared packages, introducing `featureEnabled` subfeatures (and several new one-case subfeature enums) to represent top-level toggles while preserving intended defaults (e.g. `adBlockingExtension` stays `internalOnly`).
> 
> Fixes a privacy-config bypass where a *missing* subfeature could fall back to `defaultValue` even if its parent feature was disabled/internal, by making both `AppPrivacyConfiguration.stateFor(subfeature:)` overloads consult and enforce the parent state first; unit tests and mocks are updated/expanded to lock in the new behavior and updated sources.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1031116958c4bc01a863e224eb5fc03c80eaad80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/LocalPackages/FeatureFlags/Sources/FeatureFlags/FeatureFlag.swift
```diff
@@ -401,19 +401,19 @@ extension FeatureFlag: FeatureFlagDescribing {
...
+            Config(source: .remoteReleasable(.subfeature(AdBlockingExtensionSubfeature.featureEnabled)), category: .webExtensions)
         case .forceDarkModeOnWebsites:
             Config(source: .remoteReleasable(.subfeature(ForceDarkModeOnWebsitesSubfeature.featureRollout)), category: .webExtensions)
         case .syncSeamlessAccountSwitching:
             Config(source: .remoteReleasable(.subfeature(SyncSubfeature.seamlessAccountSwitching)), category: .sync)
         case .tabCrashDebugging:
             Config(source: .disabled)
         case .delayedWebviewPresentation:
-            Config(source: .remoteReleasable(.feature(.delayedWebviewPresentation)))
+            Config(defaultValue: .internalOnly, source: .remoteReleasable(.subfeature(DelayedWebviewPresentationSubfeature.featureEnabled)))
```
diegoreymendez: Are we planning to duplicate these settings?  I think we'll need to resolve that part before merging this PR as otherwise this would result in feature availability changes.

Additionally, we'll likely need to keep both the top level feature for old app versions and the new subfeature in parallel (max-version and min-version may help document the change).

iOS/Core/FeatureFlag.swift
```diff
@@ -623,13 +623,13 @@ extension FeatureFlag: FeatureFlagDescribing {
...
         case .webExtensions:
-            Config(source: .remoteReleasable(.feature(.webExtensions)))
+            Config(defaultValue: .enabled, source: .remoteReleasable(.subfeature(WebExtensionsSubfeature.featureEnabled)))
         case .embeddedExtension:
             Config(source: .remoteReleasable(.subfeature(WebExtensionsSubfeature.embeddedExtension)))
         case .forceDarkModeOnWebsites:
             Config(source: .remoteReleasable(.subfeature(ForceDarkModeOnWebsitesSubfeature.featureRollout)))
         case .adBlockingExtension:
-            Config(source: .remoteReleasable(.feature(.adBlockingExtension)))
+            Config(source: .remoteReleasable(.subfeature(AdBlockingExtensionSubfeature.featureEnabled)))
```
diegoreymendez: Same comment that we should push these changes to remove config now.

In this case we should do it for both iOS and macOS.

diegoreymendez (commented):
Looking good, but we should push privacy config changes before we merge this PR.

diegoreymendez:
After chatting via DM with @SabrinaTardio I now understand there's no need to add the new remote values as the parent feature is the original feature we're moving.

Only thing that keeps flagged are these flips (likely due to how long it took me to review):

```
  🚨 Three flips found                                                                                                          
                                                 
  1. macOS delayedWebviewPresentation (macOS/LocalPackages/FeatureFlags/Sources/FeatureFlags/FeatureFlag.swift:416)                                                  
    - Parent delayedWebviewPresentation is enabled in macOS CDN → all users currently get it.
    - Post-PR defaultValue: .internalOnly → only internal users.                                                                                                     
    - Regression: every production user loses this feature. Either the author intended to gate it (rollout?) or this should be .enabled.                             
  2. iOS adBlockingExtension (iOS/Core/FeatureFlag.swift:632)                                                                                                        
    - Parent is internal in iOS CDN → internal users currently get it.                                                                                               
    - Post-PR no defaultValue (i.e. .disabled) → nobody gets it.                                                                                                     
    - Regression: internal testers lose access.                                                                                                                      
  3. macOS adBlockingExtension (macOS/LocalPackages/FeatureFlags/Sources/FeatureFlags/FeatureFlag.swift:408)                                                         
    - Same as #2 on macOS CDN. Regression: internal testers lose access. 
```

iOS/Core/FeatureFlag.swift
```diff
@@ -641,13 +641,13 @@ extension FeatureFlag: FeatureFlagDescribing {
         case .onboardingRebranding:
             Config(source: .remoteReleasable(.subfeature(iOSBrowserConfigSubfeature.onboardingRebranding)))
         case .webExtensions:
-            Config(source: .remoteReleasable(.feature(.webExtensions)))
+            Config(defaultValue: .enabled, source: .remoteReleasable(.subfeature(WebExtensionsSubfeature.featureEnabled)))
```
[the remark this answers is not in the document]
SabrinaTardio: It seems to be live on https://staticcdn.duckduckgo.com/trackerblocking/config/v4/ios-config.json

diegoreymendez (approved):

PR #4433, opened 2026-04-10
Add optional 'channel' pixel parameter

samhouts:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1213914821996600?focus=true
Tech Design URL: N/A
CC: @ayoy 

### Description

Introduce an optional PixelKit `channel` parameter to tag non-production pixels, matching the existing Windows behavior where non-production pixel requests include a `channel` query parameter.

**Channel values:**

- `"canary"` — internal users (logged in via the internal user flow), on any build type
- `"dev"` — alpha or review builds where the user is *not* an internal user
- omitted — production builds with non-internal users (the default)

**PixelKit (shared package):**

- Add `channel` parameter key constant to `PixelKit.Parameters`
- Store `channel` on the `PixelKit` instance, accept it in `setUp`/`init`
- Inject `channel` into fired pixel parameters when present (unconditionally, like `test=1` — not gated behind `PixelKitStandardParameter` since it identifies the build context, not the event)
- Include `channel` in the debug init log
- Add `channel` to `standardPixelParameters` in test utilities so it's filtered during assertions

**ApplicationBuildType:**

- Add `channelName(isInternalUser:)` method via protocol extension
- Internal users always get `"canary"` (highest priority)
- Non-internal users on alpha/review builds get `"dev"`
- All others get `nil` (channel omitted)
- Protocol extension avoids breaking existing mock conformers

**Call sites — pass channel into PixelKit.setUp:**

- Main browser (AppDelegate) — moved `EncryptedFileStore` and `internalUserDecider` initialization earlier in startup so `isInternalUser` is available when configuring PixelKit; `configurePixelKit` and `setUpPixelKit` now accept `isInternalUser`
- VPN agent (DuckDuckGoVPNAppDelegate) — uses `DefaultInternalUserDecider` with `UserDefaults.appConfiguration`
- DBP agent (DuckDuckGoDBPBackgroundAgentAppDelegate) — uses `DefaultInternalUserDecider` with the shared app-group `UserDefaults`
- Packet tunnel provider (MacPacketTunnelProvider) — uses `DefaultInternalUserDecider` with `UserDefaults.appConfiguration`
- Transparent proxy provider (MacTransparentProxyProvider) — uses `DefaultInternalUserDecider` with `UserDefaults.appConfiguration`

**Tests:**

- `ApplicationBuildTypeChannelTests` — 6 tests covering all branches of `channelName(isInternalUser:)`: internal user, alpha build, review build, production build, and priority tests (internal user on alpha/review gets `"canary"`, not `"dev"`)
- `PixelKitTests` — Channel present when set, absent when nil, coexists with other standard parameters

**Pixel definitions:**

- Add `channel` entry to macOS `params_dictionary.json5` with enum `["canary", "dev"]`

### Testing Steps

1. Run PixelKit unit tests — the three channel tests should pass
2. Run macOS unit tests — the six `ApplicationBuildTypeChannelTests` should pass
3. Build a production build, log in as an internal user, fire any pixel → verify `channel=canary` is present in the request parameters
4. Build a production build as a non-internal user, fire any pixel → verify `channel` is absent from the request
5. Build a debug build, fire any pixel → verify pixels are dry-run (not sent) and `channel` is absent from the logged parameters (only `test=1` should appear)

> **Note:** Alpha and review builds currently have `dryRun: true`, so `channel=dev` won't be sent over the network until a future PR enables pixel sending for those builds. The infrastructure is in place and the value is correctly computed — it's just not transmitted yet.

### Impact and Risks

Low — Adds a query parameter to existing pixel requests. For production users who are not internal, behavior is completely unchanged (parameter is omitted). For internal users, `channel=canary` is added to existing pixel requests. No user-facing UI changes.

#### What could go wrong?

- **Wrong channel value sent:** Mitigated by centralizing the logic in `channelName(isInternalUser:)` with comprehensive unit tests covering all 6 input combinations.
- **Channel sent on production builds for non-internal users:** `channelName` returns `nil` when `isInternalUser` is false and the build is neither alpha nor review. PixelKit only includes the parameter when non-nil. Unit test verifies omission.
- **Breaking existing test mocks:** `channelName(isInternalUser:)` is added as a protocol extension with a default implementation, not as a protocol requirement — existing `ApplicationBuildTypeMock` and other conformers compile without changes.
- **Internal user state unavailable in auxiliary targets:** VPN, DBP agent, and network extension targets use `DefaultInternalUserDecider` initialized with the shared app-group `UserDefaults`, matching the existing pattern for reading `isInternalUser` across process boundaries. The main browser app moves `internalUserDecider` initialization earlier in startup (before PixelKit setup) so the flag is available at configuration time; no functional change to other initialization that depends on it.

### Quality Considerations

- **Edge cases:** Alpha/review builds have `dryRun: true`, so `channel=dev` is currently a no-op. This is intentional — the channel value is ready for when a future PR enables pixel sending for those builds.
- **Performance:** Negligible — one boolean check + string comparison at startup, one dictionary insertion per pixel fire.
- **Monitoring:** No new pixels or dashboards needed. This enriches existing pixel requests with a parameter that can be used for filtering in existing dashboards.
- **Privacy:** The channel values are a fixed set (`"canary"`, `"dev"`) — not user input, no PII, and identical for all users of a given build context. Does not enable user linking. The `isInternalUser` flag is only used locally to select the channel string; the flag itself is never transmitted. Consistent with the already-triaged `pixelSource` parameter.

### Notes to Reviewer

- The channel parameter is injected unconditionally when non-nil (in `PixelKit.swift`'s private `fire` method), unlike `pixelSource` which requires events to opt in via `PixelKitStandardParameter`. This is intentional — `channel` identifies the build context and should appear on every pixel, matching the `test=1` pattern and Windows behavior.
- This is a macOS-only change. iOS does not currently have non-production build channels, so no iOS changes are needed. If iOS gains channels in the future, the PixelKit infrastructure is already in place.
- The design was updated based on @ayoy's feedback: rather than deriving channel from build type alone (which was moot since alpha/review builds don't send pixels), the primary signal is now `isInternalUser`. Alpha/review builds get `channel=dev` as a secondary signal, ready for when `dryRun` is disabled for those builds in a follow-up PR.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes pixel request payloads across multiple macOS targets and adjusts startup/setup code paths, plus updates SwiftPM/Xcode project dependencies (e.g., adding `PrivacyConfig`).
> 
> **Overview**
> Adds a new optional `channel` pixel parameter to `PixelKit` (stored on the instance, accepted via `setUp`/`init`, and injected into every fired pixel when non-nil), and updates test utilities and `PixelKitTests` to account for/validate the new parameter.
> 
> On macOS, introduces `ApplicationBuildType.channelName(isInternalUser:)` to derive `"canary"` vs `"dev"` vs omitted, then threads this value into `PixelKit.setUp` from the main app, VPN agent, DBP background agent, and network extension providers (with a small startup reordering in `AppDelegate` to compute internal-user status before PixelKit setup). Also updates pixel parameter definitions (`params_dictionary.json5`) and project/dependency files (adds `PrivacyConfig` to multiple targets and bumps `content-scope-scripts` to `14.0.0`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85a33a224343ae7e3655c504fba3b1b7b0fd334f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1940,11 +1940,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     private static func setUpPixelKit(dryRun: Bool) {
         let source = NSApp.isSandboxed ? "browser-appstore" : "browser-dmg"
+        let channel = StandardApplicationBuildType().channelName
         let userAgent = UserAgent.duckDuckGoUserAgent()
 
         PixelKit.setUp(dryRun: dryRun,
```
ayoy, line 1982: This `dryRun` flag controls if pixels are at all being sent from the app.

It currently resolves to `true` for Debug, Review and Alpha builds, so neither of these builds sends pixels.

macOS/DuckDuckGo/Common/Utilities/ApplicationBuildType.swift
```diff
@@ -27,6 +27,18 @@ protocol ApplicationBuildType {
     var isAlphaBuild: Bool { get }
 }
 
+extension ApplicationBuildType {
+    /// Returns the channel name for non-production builds, or nil for production/debug.
+    /// Alpha builds are internal-only (canary); review builds are public beta / TestFlight (preview).
```
ayoy: Review builds are not the same thing as TestFlight, "public beta" or even "internal release".

Review builds are a way of making a release build from an arbitrary branch to be able to run it alongside the regular release app. They are never public, only used internally, and they're also not updated by Sparkle.

We distribute official releases internally for a week before they are published, but currently there's no indication in the code if a given build is in internal testing. That would be the regular production build but we'd have to analyze Sparkle appcast XML file to learn if the given app run is "preview" or not (based on the app version its channel attribute in the appcast file).

ayoy (commented):
@samhouts thanks for this! I'm dropping a comment related to review builds as I'm not sure if that's what you wanted to monitor. Also currently none of the debug, alpha and review builds actually send pixels so we'd have to change that too. Let me know if you have further questions!

macOS/DuckDuckGo/Application/AppDelegate.swift
samhouts: Okay, got it. I won't change that for now. I will set it up to send the channel with those pixels if we ever change it, though.

macOS/DuckDuckGo/Common/Utilities/ApplicationBuildType.swift
samhouts: Ah, okay, that makes sense. I changed the approach to just look for internal log in instead, and we'll just consider those canary whether it's preview or release. That should be sufficient for our current investigation.

ayoy: Cool, that makes sense. We could assume that internal users are on the version that hasn't been released publicly yet.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1940,7 +1940,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     private static func setUpPixelKit(dryRun: Bool) {
         let source = NSApp.isSandboxed ? "browser-appstore" : "browser-dmg"
-        let channel = StandardApplicationBuildType().channelName
+        let channel = StandardApplicationBuildType().channelName(isInternalUser: UserDefaults.appConfiguration.isInternalUser)
```
ayoy: The correct way of checking internal user state is via DefaultInternalUserDecider (user defaults is not enough here). This makes it problematic, because internal user decider is initialized after setting up PixelKit.

We should move `internalUserDecider` initialization in AppDelegate.init to before `configurePixelKit()` – it should be safe as internal user decider doesn't depend on pixels.

Then you can either:
1. pass `isInternalUser` as value here, via configurePixelKit and setUpPixelKit. The drawback is that if a user became internal during the app session, the change wouldn't be reflected.
2. pass `() -> Bool` closure to PixelKit, that retrieves the value for each pixel event. This would use a correct value all the time, but would be more expensive.

I'd go with 1.

macOS/DuckDuckGo/NetworkProtection/NetworkExtensionTargets/NetworkExtensionTargets/MacPacketTunnelProvider.swift
```diff
@@ -685,11 +685,15 @@ final class MacPacketTunnelProvider: PacketTunnelProvider {
         source = "vpnAppExtension"
 #endif
 
+        let appConfigDefaults = UserDefaults(suiteName: Bundle.main.appGroup(bundle: .appConfiguration))
+        let isInternalUser = appConfigDefaults?.bool(forKey: "isInternalUser") == true
+        let channel = StandardApplicationBuildType().channelName(isInternalUser: isInternalUser)
```
ayoy: Ideally it should use `DefaultInternalUserDecider` from `PrivacyConfig` module:

1. Add `import PrivacyConfig` at the top.
2. Add PrivacyConfig to dependencies of all targets that use this file.
3. Rewrite the code as follows:

```suggestion
        let internalUserDecider = DefaultInternalUserDecider(store: UserDefaults.appConfiguration)
        let channel = StandardApplicationBuildType().channelName(isInternalUser: internalUserDecider.isInternalUser)
```

macOS/DuckDuckGo/NetworkProtection/NetworkExtensionTargets/NetworkExtensionTargets/MacTransparentProxyProvider.swift
```diff
@@ -56,9 +56,13 @@ final class MacTransparentProxyProvider: TransparentProxyProvider {
             loadSettingsFromProviderConfiguration: loadSettingsFromStartupOptions)
 
 #if !NETP_SYSTEM_EXTENSION
+        let appConfigDefaults = UserDefaults(suiteName: Bundle.main.appGroup(bundle: .appConfiguration))
+        let isInternalUser = appConfigDefaults?.bool(forKey: "isInternalUser") == true
+        let channel = StandardApplicationBuildType().channelName(isInternalUser: isInternalUser)
```
ayoy: The same here wrt using `DefaultInternalUserDecider`

macOS/DuckDuckGoDBPBackgroundAgent/DuckDuckGoDBPBackgroundAgentAppDelegate.swift
```diff
@@ -36,11 +36,15 @@ final class DuckDuckGoDBPBackgroundAgentApplication: NSApplication {
 
     override init() {
         Logger.dbpBackgroundAgent.log("🟢 Starting: \(NSRunningApplication.current.processIdentifier, privacy: .public)")
+        let appConfigDefaults = UserDefaults(suiteName: Bundle.main.appGroup(bundle: .appConfiguration))
+        let isInternalUser = appConfigDefaults?.bool(forKey: "isInternalUser") == true
+        let channel = StandardApplicationBuildType().channelName(isInternalUser: isInternalUser)
```
ayoy: And the same here, but you'll need to keep `appConfigDefaults` as-is because `UserDefaults.appConfiguration` is unavailable for this target.

macOS/DuckDuckGo/Common/Utilities/ApplicationBuildType.swift
```diff
@@ -44,3 +44,17 @@ struct StandardApplicationBuildType: ApplicationBuildType {
...
 
 }
+
+extension ApplicationBuildType {
+
+    /// Returns the pixel channel name based on internal-user status and build type.
+    ///
+    /// - `"canary"` — internal users (logged in via the internal user flow)
+    /// - `"dev"` — alpha or review builds that are not internal users
+    /// - `nil` — production builds (channel parameter is omitted)
```
ayoy: We could clarify that `nil` is for production builds for non-internal users.

ayoy (changes requested):
@samhouts thanks for making changes. Just a few things to get it right. The most important thing is that the internal user store in the main app target is not UserDefaults based, but it uses a custom file store.

samhouts:
> @samhouts thanks for making changes. Just a few things to get it right. The most important thing is that the internal user store in the main app target is not UserDefaults based, but it uses a custom file store.

Thank for the review!! I think it might be ready now!!

macOS/DuckDuckGoDBPBackgroundAgent/DuckDuckGoDBPBackgroundAgentAppDelegate.swift
```diff
@@ -36,11 +37,15 @@ final class DuckDuckGoDBPBackgroundAgentApplication: NSApplication {
 
     override init() {
         Logger.dbpBackgroundAgent.log("🟢 Starting: \(NSRunningApplication.current.processIdentifier, privacy: .public)")
+        let appConfigDefaults = UserDefaults(suiteName: Bundle.main.appGroup(bundle: .appConfiguration))
+        let internalUserDecider = DefaultInternalUserDecider(store: appConfigDefaults ?? .standard)
+        let channel = StandardApplicationBuildType().channelName(isInternalUser: internalUserDecider.isInternalUser)
```
ayoy: If appConfigDefaults is nil, the internal user decider won't be able to retrieve the state. Let's default to nil channel in such case. EDIT: I did it myself :)

ayoy (approved):
LGTM, thanks @samhouts!

I took the liberty of adding a small change on top to make it clear that we use nil channel in DBP when UserDefaults fail to initialize, and I updated project settings to include PrivacyConfig (it compiled anyway because it was pulled implicitly but better declare the dependency for clarity).

I've tested with internal and non-internal user and the channel is set properly. Thanks again!

macOS/DuckDuckGoDBPBackgroundAgent/DuckDuckGoDBPBackgroundAgentAppDelegate.swift
samhouts: Awesome, thank you!!

PR #4436, opened 2026-04-10
[iOS] Ship review changes for for fire mode

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214011006395194?focus=true
cc @SabrinaTardio 

### Description
This PR addresses the following issues:
- [Issue 1](https://app.asana.com/1/137249556945/task/1213996221739083): Sidebar button in chats wasn't working in fire tabs. Addressed by adjusting the logic of `isFrontendReady` to not depend on sync being enabled.
- [Issue 2](https://app.asana.com/1/137249556945/task/1213996221739086): Burning the last open chat in normal mode didn't open a new chat

### Testing Steps

1. Make sure both fireMode and fireButtonRefinements flags are enabled.
2. Verify that the sidebar button works on fire tabs
3. Have 1 tab open in normal mode and have it be a chat. Burn it. Verify a new chat is opened.

### Impact and Risks

#### What could go wrong?
N/A

### Quality Considerations
N/A

### Notes to Reviewer
Issue 2 was happening because burning a chat triggers a replacement of the current tab with a new chat tab. The old logic has an optimization that was the root cause. The optimization assumes that when there's only 1 normal tab, the auto-created blank tab "serves the same purpose" as newTab. This is true when newTab is a blank/home tab, but not when newTab has a link (like an AI chat URL). The newTab is silently discarded.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches tab replacement logic for the last normal tab and adjusts AI Chat frontend-readiness detection, which could affect tab lifecycle and queued UI actions if edge cases are missed.
> 
> **Overview**
> Fixes AI Chat sidebar/toggle actions in fire tabs by marking `AIChatContentHandler` as frontend-ready on most JS messages (excluding `getAIChatNativeConfigValues`), rather than waiting on a specific sync/history message.
> 
> Fixes burning/replacing the last normal-mode chat tab by only relying on the auto-created blank tab when the replacement tab is also blank; if the replacement has a URL (e.g., AI Chat), it now removes without auto-insert and explicitly inserts the new tab to avoid dropping it or ending up with two tabs. Updates unit tests to reflect the new readiness signal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f523f414dbdc31ac9673deb9d2ae854974ae51bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AIChat/AIChatContentHandler.swift
```diff
@@ -266,7 +266,7 @@ extension AIChatContentHandler: AIChatUserScriptDelegate {
             delegate?.aiChatContentHandlerDidReceivePageContextRequest(self)
         }
 
-        if message == .setAIChatHistoryEnabled {
+        if message == .getAIChatNativeConfigValues {
```
SabrinaTardio: Unfortunately this does not work because the FE is not fully set up yet when we receive this message, that is why I was using the next message, but I thought it was only about showing chat history so I assumed it always happened. 
However if we do if we do message != .getAIChatNativeConfigValues (so that it use the first useful message aka any after this)
seems to work in both cases. 
It’s not ideal but we could use it while I get the FE to have the url parameter for toggling the page.

hassaanelgarem: @SabrinaTardio it does work 👍 
Addressed in 9e4999c3dfebc115d0b6fab37dd9d30de2c210c7

dus7 (approved):

PR #4461, opened 2026-04-16
[iOS] Relax PIR auth gates for freemium users

jozsef-vesza:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210938375848299
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1210938375848290
CC: @THISISDINOSAUR 

### Description

Allows activated freemium users to run scans (but not opt-outs) through all iOS PIR execution paths, mirroring the macOS pattern.

**What changed:**
- `validateRunPrerequisites()`: freemium short-circuit: activated free users skip entitlement check
- `appDidBecomeActive()`: lets activated freemium users through to scan-only path
- `handleBGProcessingTask()`: routes free users to `startScheduledScanOperationsIfPermitted`
- `dashboardDidOpen()`: routes free users to scan-only using cached auth state
- `makeContinuedProcessingOptOutPlan()`: returns empty plan for free users (skips opt-out phase)

**Gating:**
1. `isFreemiumPIREnabled` feature flag: remote-configurable via privacy config, defaults to disabled ([remote config PR](https://github.com/duckduckgo/privacy-configuration/pull/4973))
2. `FreemiumDBPUserStateManaging.didActivate`: defaults to false via no-op `DisabledFreemiumDBPUserStateManager`
3. `canRunFreemiumScans` computed property centralizes both checks

**Dependencies:**
- `FreemiumDBPUserStateManager` concrete implementation wires in the real state manager. This will come in  future PR.

### Testing Steps

Green CI for now. Production behavior will be testable once `FreemiumDBPUserStateManager` is in place.

### Impact and Risks

Low: All new paths are gated behind a feature flag that defaults to disabled. No behavior change until the flag is explicitly enabled via privacy config AND the real `FreemiumDBPUserStateManager` is wired in.

#### What could go wrong?

- **Opt-out leak:** A free user could trigger opt-outs if a routing path was missed. Mitigated by: `canRunFreemiumScans` check at all 5 consumer sites, plus unit tests covering the flag on/off matrix for every execution path.
- **Feature flag bypass:** If `currentRunIsFreeScan` cache is stale, downstream consumers could route incorrectly. Mitigated by: `canRunFreemiumScans` is checked alongside `currentRunIsFreeScan` at every routing site (defense-in-depth).

### Quality Considerations

- Added new unit tests in `DataBrokerProtectionIOSManagerAuthGateTests` covering all 5 execution paths with authenticated/freemium/flag-off variations
[...]

PR #4471, opened 2026-04-17
[iOS] FreemiumDBPUserStateManager implementation

jozsef-vesza:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1210938375848294
Tech Design URL: 
CC: @THISISDINOSAUR 

### Description

Adds the concrete `DefaultFreemiumDBPUserStateManager` on iOS and wires it into the PIR flow. Completing this subtask activates the auth-gate relaxation in #4461: `didActivate` now flips to `true` at profile-save, so freemium users can actually run scans.

Production changes (stacked on top of `jozsef/freemium-pir-auth-gate`):

- **Protocol expansion.** `FreemiumDBPUserStateManaging` adds read/write methods.
- **`DefaultFreemiumDBPUserStateManager`** UserDefaults-backed default implementation.
- **`BrokerProfileJobEventsHandler`**  will now update the state manager when profile is saved.
- **`handleScanCompletion()` helper on `DataBrokerProtectionIOSManager`.** Deduplicates the scan-completion logic across both scan paths and records the freemium scan result. Skips the write if the DB read fails, so a transient error can't permanently stamp the wrong result.

Out of scope: `recordSubscriptionUpgradeIfNeeded()` exists on the protocol and manager but has no production caller yet. This will come later when working on pixels.

### Testing Steps

Green CI, there are no entry points yet.

### Impact and Risks

Low: No entry points + guarded by feature flag.

#### What could go wrong?

- DB read error on the first scan. Freemium result is recorded only once, so a throw during the very first scan could lock in a wrong "no matches" result forever. Handled by skipping the write on error; the next successful scan still gets to record.
- Concurrent scan completions across both paths. Two scan completions could race to record the result. The state manager serializes all reads and writes through a lock, so the first one wins.

### Quality Considerations

- **Persistence.** UserDefaults `.dbp` suite, key namespace `ios.browser.freemium.dbp.*` — iOS-scoped so there can be no collision with macOS's `macos.browser.freemium.dbp.*` keys.
- **Concurrency.** All reads and writes pass through one `NSLock`. Lock acquisition happens *after* the async auth check, so the lock is never held across a suspension point — avoids the Swift 6 concurrency checker error.
- **Cross-platform.** iOS intentionally diverges from `macOS/LocalPackages/Freemium/FreemiumDBPUserStateManager.swift`: simpler enum (`FreemiumFirstScanResult`) instead of macOS's `FreemiumDBPMatchResults` struct; drops macOS-only notification/dismissal flags. Spec call: RMF on iOS can't template counts, so the struct adds no value there.
- **Testability.** 34 new tests: 20 state-manager unit tests, 5 handler integration tests, 9 scan-completion integration tests including cross-path and DB-error cases.

### Notes to Reviewer

- **Stacked on #4461.** This PR targets `jozsef/freemium-pir-auth-gate`. Review that one first; I'll retarget this to `main` once it merges.
- **Subscription-upgrade caller is intentionally deferred** to the attribution-pixel subtask.

---

> [!NOTE]
> **Medium Risk**
> Introduces new UserDefaults-backed freemium state and wires it into scan completion and profile-save events; mistakes here could mis-record first-scan outcomes or affect scan gating despite feature-flag checks.
> 
> **Overview**
> Implements iOS freemium PIR user-state persistence via a new `DefaultFreemiumDBPUserStateManager` and expands `FreemiumDBPUserStateManaging` to track activation, first profile-save timestamp, first-scan result, and (future) upgrade timestamp, with serialized writes and first-write-wins semantics.
> 
> Wires this state into the PIR flow: `DBPService` now constructs/injects the manager, `BrokerProfileJobEventsHandler` records activation on `.profileSaved`, and `DataBrokerProtectionIOSManager` centralizes scan-completion handling to both fire the existing “matches found” event and (only on normal completion) persist the first-scan match result, skipping writes when `database.hasMatches()` fails.
> 
> Adds/updates test support and coverage, including configurable `hasMatches()` behavior in DB mocks and new unit/integration tests for state persistence, event handling, and both scan-completion paths (including interrupted and DB-error scenarios).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e69d88cd52379297bd7613b65352403d4133a350. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/ManagingAndCoordinating/DataBrokerProtectionIOSManager.swift
+++ b/iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/ManagingAndCoordinating/DataBrokerProtectionIOSManager.swift
@@ -923,23 +923,14 @@ private extension DataBrokerProtectionIOSManager {
 
 private extension DataBrokerProtectionIOSManager {
 
-    /// Applies side-effects tied to a scan-operations completion callback. Used from both
-    /// `startImmediateScanOperations()` and `coordinatorIsReadyForScanOperations()`.
-    ///
-    /// - `.firstScanCompletedAndMatchesFound` always fires when the DB reports matches. This
-    ///   is pre-branch behavior: the completion block fired it on both normally-finished and
-    ///   interrupted runs, and downstream notification scheduling depends on it.
-    /// - The freemium `firstScanResult` write is gated on `didCompleteNormally`. It's a
-    ///   branch-introduced side-effect and is first-write-wins, so an interrupted run's
-    ///   mid-scan DB snapshot must not permanently lock the wrong value. A transient DB read
-    ///   error is also handled — if `hasMatches()` throws, nothing fires or persists and a
-    ///   later successful scan captures the correct outcome.
-    func applyScanCompletionSideEffects(didCompleteNormally: Bool) async {
+    /// Handles common completion work for immediate scan operations.
+    /// The queue also runs completion for interrupted scans; only normal completions may persist first-write-wins freemium state.
+    func handleScanOperationsCompletion(scanCompletedNormally: Bool) async {
         guard let hasMatches = try? database.hasMatches() else { return }
         if hasMatches {
             eventsHandler.fire(.firstScanCompletedAndMatchesFound)
         }
-        guard didCompleteNormally else { return }
+        guard scanCompletedNormally else { return }
         await freemiumDBPUserStateManager.recordFirstScanResultIfNeeded(hasMatches: hasMatches)
     }
 
@@ -955,25 +946,21 @@ extension DataBrokerProtectionIOSManager {
         }
 
         await checkForEmailConfirmationData()
-        // `JobQueueManager` invokes this completion block on both normal finish AND on
-        // interruption (e.g. when `appDidBecomeActive` kicks off a replacement run mid-scan).
-        // The errorHandler tells us which one; thread that signal into the completion so we
-        // can gate the branch-introduced freemium write on real completion. Pre-existing
-        // side-effects (`.firstScanCompletedAndMatchesFound`) stay unconditional.
-        var didCompleteNormally = false
+        // Completion also runs for interrupted scans; the error handler is the normal-finish signal.
+        var scanCompletedNormally = false
         queueManager.startImmediateScanOperationsIfPermitted(
             showWebView: false,
             jobDependencies: jobDependencies,
             errorHandler: { [weak self] errors in
                 if errors?.oneTimeError == nil {
-                    didCompleteNormally = true
+                    scanCompletedNormally = true
                     self?.eventsHandler.fire(.firstScanCompleted)
                 }
             }
         ) { [weak self] in
             guard let self else { return }
             Task {
-                await self.applyScanCompletionSideEffects(didCompleteNormally: didCompleteNormally)
+                await self.handleScanOperationsCompletion(scanCompletedNormally: scanCompletedNormally)
                 DispatchQueue.main.async {
                     backgroundAssertion.release()
                 }
@@ -1065,23 +1052,21 @@ extension DataBrokerProtectionIOSManager: DBPContinuedProcessingDelegate {
         }
 
         await checkForEmailConfirmationData()
-        // See startImmediateScanOperations for the rationale. Continued-processing's
-        // `.scanPhaseCompleted` emission is explicitly unchanged — that's pre-branch behavior;
-        // revisiting how the coordinator handles interrupted scans is a separate concern.
-        var didCompleteNormally = false
+        // Same completion semantics as `startImmediateScanOperations()`.
+        var scanCompletedNormally = false
         queueManager.startImmediateScanOperationsIfPermitted(
             showWebView: false,
             jobDependencies: jobDependencies,
             errorHandler: { [weak self] errors in
                 if errors?.oneTimeError == nil {
-                    didCompleteNormally = true
+                    scanCompletedNormally = true
                     self?.eventsHandler.fire(.firstScanCompleted)
                 }
             }
         ) { [weak self] in
             guard let self else { return }
             Task {
-                await self.applyScanCompletionSideEffects(didCompleteNormally: didCompleteNormally)
+                await self.handleScanOperationsCompletion(scanCompletedNormally: scanCompletedNormally)
                 DispatchQueue.main.async {
                     Task { [weak self] in
                         await self?.continuedProcessingCoordinator.didEmit(event: .scanPhaseCompleted)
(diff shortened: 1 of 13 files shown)


iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/ManagingAndCoordinating/DefaultFreemiumDBPUserStateManager.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public func recordFirstScanResultIfNeeded(hasMatches: Bool) async {
+        guard await !isUserAuthenticated() else { return }
+        persistFirstScanResultIfAbsent(hasMatches: hasMatches)
+    }
+
+    public func recordSubscriptionUpgradeIfNeeded() async {
+        // By contract, the caller drives this from a real purchase-success / transition
+        // signal, so we do NOT check isUserAuthenticated here. See spec §3.
```
quanganhdo: How about 

```suggestion
    public func recordSubscriptionUpgradeIfEligible() async {
        // By contract, the caller drives this from a real purchase-success / transition
        // signal, so we do NOT check isUserAuthenticated here. See spec §3.
```

because of the difference?

iOS/LocalPackages/DataBrokerProtection-iOS/Tests/DataBrokerProtection-iOSTests/DefaultFreemiumDBPUserStateManagerTests.swift
```diff
@@ -0,0 +1,261 @@
...
+
+    // MARK: - upgradeToSubscriptionTimestamp
+
+    func test_upgradeToSubscriptionTimestamp_returnsPersistedValue() {
+        let date = Date(timeIntervalSince1970: 1_700_000_000)
+        userDefaults.set(date, forKey: "ios.browser.freemium.dbp.upgrade.to.subscription.timestamp")
+        let sut = makeSUT()
+        XCTAssertEqual(sut.upgradeToSubscriptionTimestamp, date)
+    }
+
```
quanganhdo: Quite superficial to me.

iOS/LocalPackages/DataBrokerProtection-iOS/Tests/DataBrokerProtection-iOSTests/DataBrokerProtectionIOSManagerScanCompletionTests.swift
```diff
@@ -0,0 +1,203 @@
...
+    private func awaitCondition(
+        timeout: TimeInterval = 1.0,
+        _ predicate: () -> Bool
+    ) async {
+        let deadline = Date().addingTimeInterval(timeout)
+        while Date() < deadline {
+            if predicate() { return }
+            try? await Task.sleep(nanoseconds: 10_000_000)
+        }
+    }
```
quanganhdo, line 65: Can’t we do XCTestExpectation?

jozsef-vesza: It isn’t a straightforward move here. The behavior we test completes inside an unstructured Task that is kicked off after the queue completion runs, and there isn’t a single completion signal that covers all branches of this flow. For some cases we can only observe the resulting side effects, which is why the test currently waits on those rather than a dedicated expectation.

A possible fix is to introduce a wrapper around the state manager with some test hooks. I don’t think that extra test-only scaffolding is worth adding in this PR, so I’d leave it as is. WDYT?

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/ManagingAndCoordinating/DataBrokerProtectionIOSManager.swift
```diff
@@ -940,12 +959,12 @@ private extension DataBrokerProtectionIOSManager {
...
         ) { [weak self] in
-            if let hasMatches = try? self?.database.hasMatches(), hasMatches {
-                self?.eventsHandler.fire(.firstScanCompletedAndMatchesFound)
-            }
-
-            DispatchQueue.main.async {
-                backgroundAssertion.release()
+            guard let self else { return }
+            Task {
+                await self.handleScanCompletion()
```
quanganhdo: LLM flagged this:

> Freemium user saves profile → scan starts → user backgrounds the app (checks a notification, reads a message, whatever) → foregrounds → `appDidBecomeActive` triggers a fresh `startImmediateScanOperations` → the in-flight scan's completion runs `handleScanCompletion` on a mid-scan database snapshot → `firstScanResult` locks to whatever's in `database.hasMatches()` at that instant, forever. Any downstream freemium UI/pixels that key off `firstScanResult` will be wrong for that user for the lifetime of the install.

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/ManagingAndCoordinating/DefaultFreemiumDBPUserStateManager.swift
```diff
@@ -0,0 +1,123 @@
...
+
+    public func recordProfileSavedIfNeeded() async {
+        guard await !isUserAuthenticated() else { return }
+        persistProfileSaved()
+    }
+
+    public func recordFirstScanResultIfNeeded(hasMatches: Bool) async {
+        guard await !isUserAuthenticated() else { return }
+        persistFirstScanResultIfAbsent(hasMatches: hasMatches)
+    }
```
quanganhdo, line 88: I wonder if we should also gate these behavior under freemium feature flag (just in case), as they are meant to persist forever.

quanganhdo (changes requested):

iOS/LocalPackages/DataBrokerProtection-iOS/Sources/DataBrokerProtection-iOS/ManagingAndCoordinating/DefaultFreemiumDBPUserStateManager.swift
jozsef-vesza: Fair, fixed in e69d88cd52379297bd7613b65352403d4133a350. TLDR: we only record scan results if the scan was completed normally (and not interrupted)

quanganhdo (approved):
Thanks for addressing the feedback. I don’t love the name `scanCompletedNormally`, maybe something like `scanNotInterrupted` would be more meaningful?. 

Otherwise the changes are guarded behind ff for now so it should be good.

PR #4483, opened 2026-04-20
Fix downloads pixel assertion: fast-fail on real createFile errors

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1213975525679527?focus=true
Tech Design URL: N/A
CC: N/A

### Description

`defaultDownloadLocation` used `fm.createFile` to reserve a placeholder filename. `createFile` returns only a `Bool`, so any failure — disk full, read-only volume, no permission — was always converted to `CocoaError(.fileWriteFileExists)`, causing `withNonExistentUrl` to exhaust its retry loop and ultimately fire a `mac_debug_assertion_failure` pixel even for persistent, unrecoverable errors.

**Changes:**

- `FileDownloadManager`: replaced `fm.createFile` with `Data().write(to:options:.withoutOverwriting)` so real `NSError` codes (`ENOSPC`, `EROFS`, `EACCES`, `ENOENT`) are available. A TOCTOU race (another process claimed the name between the existence check and the write) still retries with the next available filename; any other write error is treated as unrecoverable — `pixelAssertionFailure(error:)` is fired and `defaultDownloadLocation` throws, causing `chooseDestination` to call `task.failDestination(with:)` immediately
- `WebKitDownloadTask`: added `failDestination(with:)` to propagate the real underlying error directly into the `.failedToCompleteDownloadTask` failure path and mark the download as non-retryable
- `PixelKit+Assertion.swift`: `pixelAssertionFailure` now accepts an optional `Error` forwarded to `DebugEvent` for breadcrumbs in telemetry; also fixed to call the module-level `assertionFailure` (which respects the `customAssertionFailure` hook in DEBUG builds) instead of `Swift.assertionFailure` directly
- `FileDownloadManagerTests`: write errors are injected via the `reservePlaceholderFile` closure; tests cover persistent failure → download immediately `.failed`, and TOCTOU race → retries with next available name
- `DownloadListCoordinatorTests`: two new integration tests — cleanup of temp files when a retry fails non-retryably (disk full before writing begins), and cleanup of temp files when a failed download is explicitly removed

### Testing Steps
- Open Disk Utility app; select the system APFS volume, click **+** and add a new APFS partition with a Storage Quota of **10 MB**
- In DDG **Settings → General**, set the Downloads directory to the newly created partition
- Open https://www.thinkbroadband.com/download and start a large file download; wait for it to fail with a disk-full error
- Click **Restart** repeatedly, validate Restart button appears and no assertions raised.
- Start another download and verify it is listed in the Downloads popup with a correct error message, no Restart button should appear for it on the already-full partition.
- Smoke test a download to a regular location, validate it completes normally.

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- The TOCTOU retry path relies on `Data().write(to:options:.withoutOverwriting)` throwing `EEXIST` when the file was claimed by another process — this is a standard POSIX/Foundation guarantee and is safe
- The fix does not change behaviour when placeholder creation succeeds, or in the user-prompt path (`requestDestinationFromUser`)

### Quality Considerations
- Both DMG and App Store (sandboxed) builds are affected equally; `Data.write(to:)` and the underlying `open(O_EXCL)` go through the same kernel access enforcement
- The "overwrite" flow (WebKitDownloadTask replacing the placeholder via `replaceItemAt`) is unaffected — the fix only changes what happens when placeholder creation fails

---

> [!NOTE]
> **Medium Risk**
> Touches core download destination, failure, and cleanup paths where wrong retry or cleanup could lose files or block retries, though behavior when writes succeed is largely unchanged.
> 
> **Overview**
> Fixes **macOS download destination setup** so persistent filesystem errors (disk full, read-only volume, etc.) failOn longer exhaust filename retries and fire misleading debug assertion pixels.
> 
> **`FileDownloadManager`** now reserves default-download placeholders with **`Data().write(to:options:.withoutOverwriting)`** instead of **`createFile`**, preserving real **`NSError`** codes. TOCTOU “file already exists” still retries via **`withNonExistentUrl`**; other write failures **throw** and **`chooseDestination`** calls new **`WebKitDownloadTask.failDestination(with:)`** so the task fails immediately with the underlying error (non-retryable). Adds injectable **`reservePlaceholderFile`** for unit tests.
> 
> **`WebKitDownloadTask`** aligns placeholder creation with the same atomic write, rolls back on hide-file failure, centralizes temp/destination cleanup in **`cleanupNonRetryableDownload`**, and tightens **retry vs non-retry** (e.g. presenter/setup failures and final move errors can be retryable when already downloading). **`decideDestination`** also treats **`.downloading`** as a resume-as-replacement path. UI/list behavior: **Restart** only when **`error.isRetryable`**, and **`DownloadListCoordinator`** shows instant failures in the list with a better filename fallback.
> 
> **`pixelAssertionFailure`** optionally forwards an **`Error`** to telemetry and uses module-level **`assertionFailure`** for test hooks. New unit/integration tests cover disk-full failure, TOCTOU rename retry, and temp file cleanup on non-retryable retry failure and on remove.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a8e74ded03d1edf8343455ec1b9e12f72685b25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mallexxx:
🤖 Executor: This PR is still active and awaiting review. Keeping alive.

macOS/DuckDuckGo/FileDownload/Services/DownloadListCoordinator.swift
```diff
@@ -378,11 +378,14 @@ final class DownloadListCoordinator {
             }
 
             if item == nil,
-                case .failure(let failure) = result, !failure.isCancelled,
-                let fileName = task.selectedDestinationURL?.lastPathComponent {
```
mallexxx, line 382: fix instantly failing downloads not appearing in the download manager

macOS/DuckDuckGo/FileDownload/View/DownloadsViewController.swift
```diff
@@ -481,7 +481,7 @@ extension DownloadsViewController: NSMenuDelegate {
             case #selector(removeDownloadAction(_:)):
                 menuItem.isHidden = !(item.state.progress == nil)
             case #selector(restartDownloadAction(_:)):
-                menuItem.isHidden = !(item.state.error != nil)
+                menuItem.isHidden = !(item.state.error?.isRetryable == true)
```
mallexxx, line 484: match the "Restart" menu item appearance to the Restart button (show only for `retryable` downloads)

ayoy (approved):
LGTM, works flawlessly now, thanks @mallexxx!

PR #4495, opened 2026-04-20
[macOS] Duck.ai last model pick-up consistency

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1214148331342596?focus=true
Tech Design URL:
CC:

### Description
Unifies the Duck.ai model selection between the native omnibar and the NTP (webview) omnibar so that changing the model in either one is reflected in the other.

Both omnibars previously tracked their own picks:

 | Omnibar | UserDefaults key |
 |---|---|
 | Native | `aichat.omnibar.selected-model-id` (`AIChatPreferencesPersisting`) |
 | NTP | `newTabPageSelectedModelId` (`NewTabPageOmnibarConfigProvider`) |

Changes:
- `NewTabPageOmnibarConfigProvider.selectedModelId` now reads/writes through the shared `AIChatPreferencesPersisting` (same key as native).
- Added `selectedModelIdPublisher` to the provider, backed by `UserDefaults.didChangeNotification` (received on the main queue to avoid a Swift exclusive-access crash in the struct's setter).
- `NewTabPageOmnibarClient` subscribes to that publisher and calls `notifyConfigUpdated()` so JS receives the new model via `omnibar_onConfigUpdate` when the native side changes it.
- One-time migration: on init, if the legacy `newTabPageSelectedModelId` key exists, copy it into the shared key (only when the shared key is empty) and drop the legacy key.

### Testing Steps
1. Quit the app. Seed UserDefaults to simulate an upgrading NTP-only user:
```sh
  BID=com.duckduckgo.macos.browser.debug   # adjust for your build
  defaults delete "$BID" "aichat.omnibar.selected-model-id" 2>/dev/null
  defaults write "$BID" newTabPageSelectedModelId -string "<some-valid-model-id>"
```
2. Launch the app → open the NTP. The model picker should show the migrated model. Confirm the legacy key is gone: `defaults read "$BID" newTabPageSelectedModelId` → "does not exist".
3. Focus the address bar (native Duck.ai omnibar). The same model should be selected.
 4. Change the model in the **native** omnibar (e.g. pick Maverick). Dismiss the popover → the NTP picker reflects the new model.
5. Change the model in the **NTP** omnibar (e.g. pick GPT-5.2). Reopen the native omnibar → it shows the new model.
6. Quit and relaunch → the last selection persists on both sides.

### Impact and Risks
**Medium.** Scoped to Duck.ai omnibar model selection UX. No impact on auth, data, or navigation. The worst realistic case is the wrong default model on first launch after update for a narrow user slice.

#### What could go wrong?
- **Users who only picked in NTP lose their selection.** Mitigated by the one-time migration that copies the legacy key into the shared key before dropping it.
- **Users who picked in both omnibars:** the native value wins (single source of truth). Acceptable — the legacy NTP value is dropped, but the user still has *a* valid recent pick, and they can re-select.
- **Migration running twice** and clobbering a freshly-picked model. Migration removes the legacy key on first run, so subsequent launches are no-op. Regression test: `testMigration_runsOnlyOnce`.

### Quality Considerations
- **Edge cases:** (a) legacy key present + shared empty → migrate; (b) legacy key present + shared populated → native wins, legacy dropped; (c) neither present → no-op; (d) legacy dropped, user picks again → next launch preserves. All four are covered by tests in `NewTabPageOmnibarConfigProviderTests`.

### Notes to Reviewer
 - Please sanity-check the migration's `inout AIChatPreferencesPersisting` pattern in `migrateLegacySelectedModelIdIfNeeded(...)`. It's a `static` helper called from `init` before the notification observer is wired, so the migration write does not kick the publisher (no phantom emission to JS on launch).

---

> [!NOTE]
> **Medium Risk**
> Changes persistence and cross-component propagation for the selected AI model, including a one-time UserDefaults migration; regressions could cause incorrect/default model selection or stale short-name caching across omnibars.
> 
> **Overview**
> Unifies Duck.ai model selection between the native omnibar and the NTP omnibar by routing both through a shared `AIChatPreferencesPersistor` instance and notifying the NTP webview when the native selection changes.
> 
> `AIChatPreferencesPersistor` is now a reference type that exposes `selectedModelIdPublisher` (distinct-write only), `NewTabPageOmnibarConfigProvider` reads/writes `selectedModelId`/`selectedModelShortName` via that persistor, and a one-time migration copies the legacy NTP-only model key into the shared store before deleting it. `NewTabPageOmnibarClient` now triggers `omnibar_onConfigUpdate` on model changes and avoids clobbering cached short names when the model id is echoed unchanged. Tests and stubs were updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0ec0a07cd9e8592b2a66fe3d05d07f48bf8669e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/AIChat/Sources/AIChat/Shared/AIChatPreferencesPersistor.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/Shared/AIChatPreferencesPersistor.swift
@@ -16,16 +16,22 @@
 //  limitations under the License.
 //
 
+import Combine
 import Foundation
 import Persistence
 
 public protocol AIChatPreferencesPersisting {
     var selectedModelId: String? { get set }
+    /// Emits the new value whenever `selectedModelId` changes through this persistor instance.
+    /// Consumers that need cross-component sync must share the same instance.
+    var selectedModelIdPublisher: AnyPublisher<String?, Never> { get }
     /// The short display name of the last selected model, used to show the button before models are fetched.
     var selectedModelShortName: String? { get set }
 }
 
-public struct AIChatPreferencesPersistor: AIChatPreferencesPersisting {
+/// Reference type so that a single instance can be shared across components (e.g. the native address-bar
+/// omnibar and the New Tab Page omnibar) and both observe the same `selectedModelIdPublisher`.
+public final class AIChatPreferencesPersistor: AIChatPreferencesPersisting {
 
     enum Key: String {
         case selectedModelId = "aichat.omnibar.selected-model-id"
@@ -33,6 +39,7 @@ public struct AIChatPreferencesPersistor: AIChatPreferencesPersisting {
     }
 
     private let keyValueStore: ThrowingKeyValueStoring
+    private let selectedModelIdSubject = PassthroughSubject<String?, Never>()
 
     public init(keyValueStore: ThrowingKeyValueStoring = UserDefaults.standard) {
         self.keyValueStore = keyValueStore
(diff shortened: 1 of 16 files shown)


tomasstrba (approved):
Works as expected ✔️ 

One recommendation from me:
`UserDefaults.didChangeNotification` fires on every defaults write app-wide; the lastObservedModelId dedupe saves the publisher but the sink still wakes on every write. A possible solution would be to move the observation into AIChatPreferencesPersisting:                                                                                                                                      
                             
```
  public protocol AIChatPreferencesPersisting {
      var selectedModelId: String? { get set }                                                                                                                                                
      var selectedModelIdPublisher: AnyPublisher<String?, Never> { get }
      var selectedModelShortName: String? { get set }                                                                                                                                         
  } 
```

PR #4504, opened 2026-04-21
Integrate Fire Mode into Unified Toggle Input

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201392122292466/task/1213726694425609?focus=true
Tech Design URL:
CC:

### Description

Plumbs Fire Mode support into the new Unified Toggle Input (UTI) flow so the UTI matches the legacy OmniBar editing-state behavior on fire tabs.

Highlights:
- Integrates Fire Mode colors into UTI surfaces (card background, text cursor, submit button, fire-tab empty state).
- Adds the Fire Tab empty state to the UTI content container (via `DaxLogoManager`), matching the legacy flow.
- Rebuilds `AIChatHistoryManager` on fire-tab switches so the fire tab uses `NilSuggestionsReader` (no history fetched/displayed) and non-fire tabs get the real reader.
- Hardens `DaxLogoManager` install: `anchorView` is now optional (fire-tab-only use case), and an assertion guards the non-fire top-bar path.
- Suppresses the favorites overlay on fire tabs so it doesn't sit behind the fire empty state.
- Adds a `fireModeCardBackground` single-use color matching Figma's Surface Canvas (`#1C1C1C` dark, `#3D3D3D` light) for the UTI card and the OmniBar search area in fire mode.
- Adds coordinator tests for fire-tab propagation.

### Testing Steps

1. Enable UTI and open the app.
2. Open a regular tab. Start editing the omnibar and confirm colors match pre-fire-mode baseline (blue accent, light/dark neutral card).
3. Open a fire tab. Start editing the omnibar and confirm:
    - UTI card, OmniBar search area, cursor, submit, and voice-chat button use the fire-mode orange accent.
    - Card background is Surface Canvas (`#1C1C1C` in dark mode), not near-black.
    - Fire-tab empty state (the dax-logo area) renders and does not let favorites/remote messages show through.
4. Switch between fire and non-fire tabs several times and confirm:
    - Chat history only appears on non-fire tabs (no leakage of prior chats on fire tabs).
    - Colors update correctly in both directions.
5. Run the UTI coordinator tests.

### Impact and Risks

Medium: changes are scoped to UTI on fire tabs plus the shared `fireModeCardBackground` color token. The color token is only used on the UTI card and the OmniBar search-area container in fire mode; the legacy `SwitchBarTextEntryViewController` still uses `fireModeBackground` so its behavior is unchanged.

#### What could go wrong?

- Fire-mode color tokens not resolving correctly in light/dark OS → verified by keeping the UI trait on `cardView` and forcing `.dark` only on content subviews.
- Chat history reader not swapping on tab switches → addressed by rebuilding the manager in `refreshFireMode`; covered by existing + new coordinator tests.
- `DaxLogoManager` install asserting on legitimate paths → assertion is scoped to non-fire top-bar case with a nil anchor and has an early return for release builds.

### Quality Considerations

- Escape hatch plumbing into UTI is deliberately deferred to a separate branch; this PR is scoped to Fire Mode parity.
- Added unit tests for the new `updateIsFireTab` flow on `UnifiedToggleInputCoordinator`.
- `fireModeCardBackground` was introduced instead of mutating `fireModeBackground` to keep the blast radius minimal (legacy search-bar surfaces untouched).

### Notes to Reviewer

- `fireModeBackground` is still used in one place (`SwitchBarTextEntryViewController`). Let me know if you'd like me to consolidate the two tokens in a follow-up.
- Visual comparison screenshots are in the Asana task.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches multiple UTI UI surfaces and state propagation when switching fire/non-fire tabs, which could cause visual regressions or stale view/controller state if rebuild/teardown paths are missed.
> 
> **Overview**
> Adds Fire Mode support to the Unified Toggle Input (UTI): new `fireModeCardBackground` color token is introduced and used for the UTI card and fire-mode OmniBar search-area background.
> 
> UTI now propagates fire-tab state through the coordinator (`updateIsFireTab`) and refreshes UTI subviews (text cursor/tints, toolbar/submit button styling, and card appearance), including a shared `CircularButton` component for consistent circular-button behavior.
> 
> Fire tabs now render a dedicated empty state via `DaxLogoManager` (with optional `anchorView`, safe-area inset syncing, opaque background, and teardown/rebuild support), suppress favorites overlay behind it, and rebuild chat history management on fire-tab switches; omnibar show/hide dismiss animation is hardened to avoid stale completion cleanup. Tests are added to cover fire-tab propagation and manager rebuild behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b912aa4d19365efed5b8c4f5dadeb7ddeaa73a9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputView.swift
```diff
@@ -324,6 +324,25 @@ final class UnifiedToggleInputView: UIView {
...
+        applyFireModeAppearance(isFireTab: fireMode)
+        textEntryView.refreshFireMode(fireMode: fireMode)
+        toolsToolbar.refreshFireMode(fireMode: fireMode)
+    }
+
+    private func applyFireModeAppearance(isFireTab: Bool) {
+        cardView.backgroundColor = isFireTab
+            ? UIColor(singleUseColor: .fireModeCardBackground)
+            : UIColor(singleUseColor: .unifiedToggleInputCardBackground)
+        // cardView keeps the OS trait so `fireModeBackground` picks its light variant in light OS; content subviews force `.dark` so their dynamic colors resolve against the dark surface.
```
pikorddg: Cosmetic: Didn’t you mean `fireModeCardBackground` in this comment?

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputContentContainerViewController.swift
```diff
@@ -198,6 +196,20 @@ final class UnifiedInputContentContainerViewController: UIViewController {
         daxLogoManager.setForcedHidden(hidden)
     }
 
+    func refreshFireMode(fireMode: Bool) {
+        rebuildDaxLogoManager(isFireTab: fireMode)
+        rebuildChatHistoryManager()
+    }
+
+    private func rebuildDaxLogoManager(isFireTab: Bool) {
```
pikorddg, line 209: I think this might be related to a UI bug I noticed (LP2). An agentic review flagged this spot with the following feedback:

> `DaxLogoManager` keeps `forcedHidden: Bool = false` as internal state. When `rebuildDaxLogoManager` runs, it tears down the old instance and creates a fresh one — and the new instance starts with `forcedHidden = false` regardless of what the previous instance had.
>
> `setDaxLogoForcedHidden(_:)` on `UnifiedInputContentContainerViewController` is the only external entry point that sets this flag, and it forwards directly to the manager without caching the value locally. So the sequence
>
> 1. Caller sets `forcedHidden = true`
> 2. Fire-mode toggle happens → `refreshFireMode` → `rebuildDaxLogoManager`
> 3. New manager has `forcedHidden = false`
>
> leaves the Dax logo visible even though the caller's intent was to keep it hidden. Two possible fixes:
>
> - Pass the current `forcedHidden` value into `DaxLogoManager`'s init so the new instance picks it up, or
> - Cache `forcedHidden` on `UnifiedInputContentContainerViewController` and reapply it after the rebuild (via `setDaxLogoForcedHidden` on the new manager).
>
> Given that `forcedHidden` isn't exposed anywhere else, option 2 is probably cleaner — it keeps the source of truth in the VC rather than scattering it across short-lived manager instances.

iOS/DuckDuckGoTests/UnifiedToggleInput/UnifiedToggleInputCoordinatorTests.swift
```diff
@@ -616,6 +616,20 @@ final class UnifiedToggleInputCoordinatorTests: XCTestCase {
         waitForExpectations(timeout: 0.1)
     }
 
+    // MARK: - Fire Tab
```
pikorddg, line 619: Here’s agentic review of tests. Note that point number two may cover LP2 issue.

> The two new tests cover `handler.isFireTab` propagation, but the rest of `updateIsFireTab`'s contract is unverified. A few gaps worth filling:
>
> **1. Idempotency guard (`UnifiedToggleInputCoordinator.swift:540`)** — `guard viewController.handler.isFireTab != isFireTab else { return }` has no test. Calling `updateIsFireTab(false)` on a fresh `sut` (where `isFireTab` is already `false`) should *not* trigger a rebuild. If someone removes this guard in the future, both existing tests still pass and we start tearing down/rebuilding `DaxLogoManager` and `AIChatHistoryManager` on every tab switch.
>
> **2. `DaxLogoManager` rebuild (`UnifiedInputContentContainerViewController.swift:204-211`)** — the whole `tearDown` → new instance → `installDaxLogoView` → `applyRequestedContentInset` → `updateDaxVisibility` flow is untested. This may be the mechanism behind the UI bug (Dax logo visible) — a test that asserts the new manager's `isFireTab` matches, and that the old one is torn down, would have caught it.
>
> **3. `NilSuggestionsReader` ↔ `SuggestionsReader` swap (`UnifiedInputContentContainerViewController.swift:467-485`)** — this is the core feature of the PR ("fire tabs don't fetch or show chat history") and has no test. If the branch in `makeAIChatHistoryManager` were accidentally inverted, fire tabs would leak chat history and nothing in CI would fail. Worth a test that injects a spy reader and verifies fire-tab managers never call the real suggestions path.

pikorddg:
### Fire colors
Regarding

> fireModeBackground is still used in one place (SwitchBarTextEntryViewController). Let me know if you'd like me to consolidate the two tokens in a follow-up.

I'd say we should unify it, but since we're still treating UTI as a work in progress and it might change, I think it's fine to keep two definitions for now.

### UI issues

LP1/ Scenario: open a regular tab and then a fire tab, UTI has the regular color for both Search and [Duck.ai](http://duck.ai/). After the app goes from background to foreground → colors are updated to the correct ones for the fire tab

#### Before
| Search | Duck |
|--------|--------|
| Image | Image |

#### After
| Search | Duck |
|--------|--------|
| IMG_0260 | IMG_0261 |

----

LP2/ Search shows the "Fire Tab" icon, while [Duck.ai](http://duck.ai/) shows the Dax Logo → after going background → foreground, both screens switch to the Fire Tab logo → Figma doesn't cover this state, but either way the behavior is inconsistent — logo display should be uniform across screens
See screenshots in LP1

----

LP3/ I know we discussed this in the context of my hatch PR, so I don't see it as in scope for this PR — just flagging that the jumping Dax Logo issue (when switching between Search and Duck.ai is present here as well, so we know the scope of the problem for the future.

----

LP4/ Figma seems inconsistent with the X button — sometimes showing it inside the UTI, sometimes outside (I'd consider that out of scope for this PR). More importantly though, it appears in both a [light](https://www.figma.com/design/cTVuHJaq4QWVeToHoxcoie/%F0%9F%94%A5-Fire-Mode-Tabs--iOS-Android---Copy-?node-id=6001-26406&m=dev) and a [dark](https://www.figma.com/design/cTVuHJaq4QWVeToHoxcoie/%F0%9F%94%A5-Fire-Mode-Tabs--iOS-Android---Copy-?node-id=6001-31210&m=dev) variant. Which one is correct? The light X next to the dark UTI looks inconsistent to me — might be worth consulting  with designer.


----

LP5/ The voice button in bottom mode is too bright (top mode looks fine). Compare with current main.

| This PR | main |
|--------|--------|
| IMG_0266 | IMG_0267 |

jaceklyp:
LP1, LP2, LP5 should be addressed
LP3 out of scope
LP4 X inline button thanks to my other PR should be merged now and be more consistent with new designs

pikorddg (approved):
Looks and works great! 🥳

PR #4512, opened 2026-04-22
Fix broken tab rendering when duplicating a tab

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1213659555956031/task/1214175883656703

### Description

Duplicating a tab created the copy as `AnyTab.unloaded`, which caused `select(at:)` to take the materialize path. This fired `didReplaceTabAt` → `reloadItems(at:)` before the collection view had been told about the insert, leaving the tab bar in an inconsistent state (tabs visually stacking on top of each other).

The fix inserts the duplicate as a loaded `Tab` instead. This avoids the materialize path during selection, so the collection view receives the insert and select notifications in the correct order. The source tab is still not materialized (preserving the intent of 5125f8176a).

### Testing Steps

1. Open a few tabs in the browser.
2. Right-click on a tab and select "Duplicate Tab".
3. Verify the duplicated tab appears correctly next to the original tab in the tab bar.
4. Verify tabs don't visually stack or overlap.
5. Repeat with pinned and unpinned tabs.

### Impact and Risks

Low — fixes a visual bug in tab duplication. The change is scoped to a single method (`duplicateTab(at:)`) in `TabCollectionViewModel`.

#### What could go wrong?

The duplicated tab is now inserted as a loaded `Tab` rather than an unloaded one, which means it will use slightly more memory upfront. However, since `shouldLoadInBackground: true` is set, this matches the expected behavior for a tab the user explicitly duplicated.

### Quality Considerations

- Edge case: duplicating a tab in a popup window still takes the redirect path and is unaffected by this change.
- No performance concern — duplicating a tab is a user-initiated, one-at-a-time action.
- No privacy/security implications.

### Notes to Reviewer

The key insight is that `AnyTab.unloaded` triggered the materialize-on-select path, which sent a reload notification before the insert notification reached the collection view. Switching to a loaded `Tab` sidesteps this entirely.

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Low Risk**
> Small, localized change to `duplicateTab(at:)` that only affects how duplicated tabs are inserted/selected; primary risk is minor memory/runtime behavior differences from creating a loaded tab copy.
> 
> **Overview**
[...]

PR #4530, opened 2026-04-23
[macOS] Add UITest for BUG-02: Duplicate Tab not appearing in tab bar

mallexxx:
Task/Issue URL: https://app.asana.com/0/0/1214147700081496

### Description

Adds a regression UI test for BUG-02 (duplicate tab not appearing in the tab bar).

The bug itself was already fixed on main by [#4512](https://github.com/duckduckgo/apple-browsers/pull/4512) (insert a loaded `Tab` instead of `AnyTab.unloaded` so the materialise-on-select path doesn't fire `didReplaceTabAt` before the collection view knows about the new item). Per @ayoy's review:

> we have fixed this in #4512. We could still use the test, though :)

This PR has been rebased to keep only the test contribution; the previously-included fix and unrelated keychain/NetworkProtection changes have been dropped.

### Testing Steps

1. Check out the branch.
2. Run the macOS `UI Tests` scheme — `DuplicateTabTests.test_duplicateTab_appearsInTabBar_viaWindowMenu` should pass.
3. To confirm it actually catches the regression, temporarily revert #4512 (`TabCollectionViewModel.duplicateTab(at:)` back to inserting `AnyTab.unloaded`) — the test should fail.

### Impact and Risks

Low — test-only change. Adds one UI test and four `project.pbxproj` entries.

#### What could go wrong?

The test depends on the local `tests-server` (via `UITests.simpleServedPage(titled:)`) and the `Duplicate Tab` window-menu item. If either changes, the test will need to follow.

### Quality Considerations

- One assertion only (`tabs.count == initialCount + 1`) — the second `isHittable` test was dropped because the `isHittable` autoclosure tripped a Swift WMO parse failure on this SDK; the count assertion is sufficient to prove both the bug and the fix.
- Uses the local served page rather than an external URL so the test doesn't depend on network reachability in CI.

### Notes to Reviewer

[executor] automated PR. The diff vs `main` is now exactly: `macOS/UITests/DuplicateTabTests.swift` (new) plus four pbxproj entries that wire it into the `UI Tests` target.


ayoy:
@mallexxx we have fixed this in https://github.com/duckduckgo/apple-browsers/pull/4512. We could still use the test, though :)

PR #4533, opened 2026-04-23
[macOS] Fix race condition between extension load and sync on startup

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200848783441532/task/1214156699597063?focus=true
Tech Design URL: N/A
CC:

### Description

Fixes a race condition between `loadInstalledExtensions()` and `syncEmbeddedExtensions()` during app startup that causes `WebExtensionLoaderError.extensionNotFound` errors (visible as `m.mac.web.extension.load.error` pixel spikes since macOS 1.186.1).

The root cause is that `WebExtensionFeatureFlagHandler` wraps its callbacks in `Task { @MainActor in }`, allowing `syncEmbeddedExtensions()` to interleave with `loadInstalledExtensions()` at `await` suspension points. When sync detects a version bump, it uninstalls the old extension (deleting files from disk), while the loader still holds stale identifiers pointing to those deleted files.

The fix:
- Adds an `isLoadingInstalledExtensions` flag to `WebExtensionManager` that prevents `syncEmbeddedExtensions` from running during load
- Introduces `loadAndSyncExtensions(enabledTypes:)` that performs both operations as an atomic sequence
- Updates both macOS and iOS call sites to use the new atomic method
- Extracts `enabledEmbeddedExtensionTypes()` helper on both platforms to reduce duplication

### Testing Steps

1. Launch the app with web extensions enabled (autoconsent, Dark Reader, or ad blocking)
2. Verify extensions load without errors in the console
3. Toggle Dark Reader / ad blocking on and off — verify extensions sync correctly
4. Fire (clear data) and verify extensions reload properly afterward

### Impact and Risks

**Low** — This is a targeted fix in the web extensions startup flow. The guard is conservative (skips sync during load, which is safe because sync always follows load).

#### What could go wrong?

- If a feature flag publisher fires `syncEmbeddedExtensions` while load is in progress, the sync will be skipped. This is intentional and safe because `loadAndSyncExtensions` calls sync immediately after load completes.
- If a standalone `syncEmbeddedExtensions` call (e.g., from Dark Reader toggle) arrives during load, it returns early. The next explicit sync trigger will pick up the change.

### Quality Considerations

- The fix lives in the shared `WebExtensions` package and protects both macOS and iOS
- No protocol or test mock changes required — new API is on the concrete `WebExtensionManager` class only
- The iOS Fire path (`MainViewController`) which calls `loadInstalledExtensions()` alone is unaffected
- Task cancellation checks are preserved between load and sync

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes the web extension startup flow on both macOS and iOS by serializing load+sync and skipping sync during load; risk is moderate because it touches extension lifecycle timing and could delay or miss a sync trigger if not followed up properly.
> 
> **Overview**
> Fixes a startup race where `syncEmbeddedExtensions` could uninstall/upgrade embedded extensions while `loadInstalledExtensions` was still loading them from disk.
> 
> `WebExtensionManager` now tracks `isLoadingInstalledExtensions`, makes `syncEmbeddedExtensions` a no-op during load, and adds `loadAndSyncExtensions(enabledTypes:)` to run load→sync as an atomic sequence.
> 
> macOS `AppDelegate` and iOS `MainCoordinator` are updated to use the new load+sync entry point, and both extract an `enabledEmbeddedExtensionTypes()` helper to compute which embedded extensions should be installed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0383a7066cd626dbc98b4a1f84a55a562080d05a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
+++ b/SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
@@ -58,7 +58,7 @@ open class WebExtensionManager: NSObject, WebExtensionManaging, WebExtensionInst
     /// Used to prevent `syncEmbeddedExtensions` from interleaving during load,
     /// which can cause race conditions where sync deletes files that load still needs.
     @MainActor
-    public private(set) var isLoadingInstalledExtensions = false
+    private(set) var isLoadingInstalledExtensions = false
 
     /// Pixel firing for analytics.
     let pixelFiring: WebExtensionPixelFiring
--- a/SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
+++ b/SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
@@ -392,6 +392,7 @@ open class WebExtensionManager: NSObject, WebExtensionManaging, WebExtensionInst
     @MainActor
     public func loadAndSyncExtensions(enabledTypes: Set<DuckDuckGoWebExtensionType>) async {
         await loadInstalledExtensions()
+        guard !Task.isCancelled else { return }
         await syncEmbeddedExtensions(enabledTypes: enabledTypes)
     }
 
--- a/SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
+++ b/SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
@@ -54,6 +54,12 @@ open class WebExtensionManager: NSObject, WebExtensionManaging, WebExtensionInst
     /// The scriptlet configuration, if scriptlet support is enabled.
     private let scriptletConfiguration: ScriptletConfiguration?
 
+    /// Tracks whether `loadInstalledExtensions()` is currently running.
+    /// Used to prevent `syncEmbeddedExtensions` from interleaving during load,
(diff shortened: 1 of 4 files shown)


SharedPackages/WebExtensions/Sources/WebExtensions/Manager/WebExtensionManager.swift
```diff
@@ -54,6 +54,12 @@ open class WebExtensionManager: NSObject, WebExtensionManaging, WebExtensionInst
     /// The scriptlet configuration, if scriptlet support is enabled.
     private let scriptletConfiguration: ScriptletConfiguration?
 
+    /// Tracks whether `loadInstalledExtensions()` is currently running.
+    /// Used to prevent `syncEmbeddedExtensions` from interleaving during load,
+    /// which can cause race conditions where sync deletes files that load still needs.
+    @MainActor
+    public private(set) var isLoadingInstalledExtensions = false
```
ayoy: `public` can be omitted, making this internal to the module, because only the code from the same module is accessing this.

```suggestion
    private(set) var isLoadingInstalledExtensions = false
```

ayoy (approved):
LGTM, extensions are correctly reloaded for all test scenarios. Thanks @miasma13!

PR #4594, opened 2026-04-27
[iOS/macOS] Honor local-network exclusion at OS level + debug toggles

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214310586983071?focus=true

### Description

Local-network features such as AirPlay (and likely CarPlay) stopped working when the VPN was active. This change makes the local-network exclusion setting take full effect when enabled, restoring those features for users who opt in.

### Testing Steps

1. Connect to the VPN with the local-network exclusion setting on. Confirm AirPlay from iPhone to a Mac on the same network now works.
2. Open the VPN debug menu and explore the new toggles to characterise their effect on local-network features.
3. With default settings, confirm normal browsing and kill-switch behaviour is unchanged.

### Impact and Risks

Medium-low. Behaviour change only for users who already had the local-network exclusion setting enabled; for everyone else routing is unchanged. The new toggles ship to the debug menu only.

#### What could go wrong?

The OS-level exclusion was previously suppressed because of past routing concerns. If those resurface, leaving the local-network exclusion off restores prior behaviour.

### Quality Considerations

Defaults are platform-specific where prior effective behaviour differed between iOS and macOS, so users on either platform see no change unless they opt in. OS APIs that require newer system versions are gated appropriately.

### Notes to Reviewer

The product-visible change is small; the rest is debug-only instrumentation for further investigation of local-network compatibility issues.

---
###### Internal references:


samsymons (approved):
Works nicely for me with CarPlay - thanks!

PR #4608, opened 2026-04-28
[executor] Add bookmarklet support for macOS browser

mallexxx:
## Summary

When a user opens a Bookmark that is a bookmarklet (`javascript:` URL), execute it in the current tab's web view context instead of trying to navigate to it. This matches the existing iOS behavior.

## Changes

- **Tab.swift**: Added `executeBookmarklet(url:)` method that evaluates the decoded JavaScript in the page context using `.defaultClient` content world
- **Tab.swift**: Guard in `setContent` to intercept bookmarklet URLs before navigation and execute them inline
- **WindowControllersManager.swift**: Intercept bookmarklet URLs in `open(_:target:with:)` and route to `executeBookmarklet` on the selected tab

## How to test

1. Create a bookmark with URL: `javascript:alert('Hello from bookmarklet!')`
2. Click the bookmark from the bookmarks bar or menu
3. Verify the alert fires in the current tab instead of navigating to a blank/error page

## Notes

Uses existing `URL.isBookmarklet()` / `URL.toDecodedBookmarklet()` helpers from `BookmarkletExtensions.swift` (BSK). iOS already has equivalent logic in `TabViewController.executeBookmarklet(url:)`.

Task: https://app.asana.com/0/0/1211217619381277

### DIFF
--- a/macOS/DuckDuckGo/Tab/Model/Tab.swift
+++ b/macOS/DuckDuckGo/Tab/Model/Tab.swift
@@ -648,6 +648,13 @@ protocol TabDelegate: ContentOverlayUserScriptDelegate {
         return self.setContent(.contentFromURL(url, source: source))
     }
 
+    /// Executes a bookmarklet (javascript: URL) in the current tab's web view context.
+    /// Bookmarklets must not be loaded as regular navigation targets — they run as JavaScript in the page context.
+    func executeBookmarklet(url: URL) {
+        guard let js = url.toDecodedBookmarklet() else { return }
+        webView.evaluateJavaScript(js, in: nil, in: .defaultClient)
+    }
+
     private func handleUrlDidChange() {
         if let url = webView.url {
             let content = TabContent.contentFromURL(url, source: .webViewUpdated)
@@ -1039,6 +1046,12 @@ protocol TabDelegate: ContentOverlayUserScriptDelegate {
         if case .settings = content, case .settings = webView.url.flatMap({ TabContent.contentFromURL($0, source: .ui) }) {
             // replace WebView URL without adding a new history item if switching settings panes
             webView.evaluateJavaScript("location.replace('\(url.absoluteString.escapedJavaScriptString())')", in: nil, in: .defaultClient)
+            return nil
+        }
+
+        // Execute bookmarklets in the current page context rather than navigating to them
+        if url.isBookmarklet() {
+            executeBookmarklet(url: url)
             return nil
         }
 
--- a/macOS/DuckDuckGo/Windows/View/WindowControllersManager.swift
+++ b/macOS/DuckDuckGo/Windows/View/WindowControllersManager.swift
@@ -219,10 +219,20 @@ extension WindowControllersManager {
     }
 
     /// Opens a bookmark in a tab, respecting the current modifier keys when deciding where to open the bookmark's URL.
+    /// Bookmarklets (javascript: URLs) are executed in the current tab's web view rather than loaded as navigation targets.
     func open(_ bookmark: Bookmark, target window: NSWindow? = nil, with event: NSEvent?) {
         guard let url = bookmark.urlObject else { return }
 
-        // Call updated openBookmark
+        // Bookmarklets must be executed in the current tab's context, not navigated to
+        if url.isBookmarklet() {
+            let eventWindowController = mainWindowController(for: window ?? event?.window)
+            let targetWindowController = eventWindowController ?? lastKeyMainWindowController
+            if let currentTab = targetWindowController?.mainViewController.tabCollectionViewModel.selectedTab {
+                currentTab.executeBookmarklet(url: url)
+                return
+            }
+        }
+
         open(url, source: .bookmark(isFavorite: bookmark.isFavorite), target: window, with: event)
     }
 


mallexxx (commented):
**[executor-review]** Danger JS check is failing: "Please, don't forget to add a link to the internal task."

The PR body has `Task: https://app.asana.com/0/0/1211217619381277` at the bottom, but the repo's Danger rule expects the standard template format `Task/Issue URL: <url>` near the top. The PR also doesn't follow [pull-request.mdc](.cursor/rules/pull-request.mdc) template — missing `Task/Issue URL`, `Tech Design URL`, `CC`, `### Impact and Risks`, `#### What could go wrong?`, `### Quality Considerations` sections.

Suggestion: Replace the PR body with the standard template:

```
Task/Issue URL: https://app.asana.com/0/0/1211217619381277
Tech Design URL:
CC:

### Description
- Bookmarklets (`javascript:` URLs) now execute in the current tab's web view context instead of being treated as navigation targets, matching iOS behavior
- `Tab.executeBookmarklet(url:)` decodes and evaluates the JS via `webView.evaluateJavaScript`
- `Tab.setContent` and `WindowControllersManager.open(_:target:with:)` short-circuit on `url.isBookmarklet()`

### Testing Steps
1. Add a bookmark with URL `javascript:alert('Hello from bookmarklet!')`
2. Click the bookmark from the bookmarks bar / menu
3. Verify the alert fires in the current tab — no navigation, no error page

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- Malicious bookmarklet content could exfiltrate page state. Same risk surface as iOS — the bookmark URL must come from a user-saved bookmark, not arbitrary input.
- If no current tab exists, the bookmarklet silently does nothing. Acceptable for now; consider opening a new tab + about:blank target in a follow-up.

### Quality Considerations
- Reuses existing `URL.isBookmarklet()` / `URL.toDecodedBookmarklet()` helpers — no new parsing.
- No new pixels.
- Privacy: bookmarklets execute in the page's JS context; no data leaves the device.
```

mallexxx (commented):
**[executor-review]** `Tab.executeBookmarklet` evaluates JS in the wrong content world.

```swift
// macOS/DuckDuckGo/Tab/Model/Tab.swift
func executeBookmarklet(url: URL) {
    guard let js = url.toDecodedBookmarklet() else { return }
    webView.evaluateJavaScript(js, in: nil, in: .defaultClient)
}
```

`.defaultClient` is an **isolated** content world — the bookmarklet runs in a sandbox separate from the page's own scripts and cannot interact with the page's `window`/`document` state. Most real bookmarklets (e.g. `javascript:document.title='x'`, Pocket/Pinboard "save page" snippets, `window.getSelection()` helpers) only work in the page world.

For comparison, the iOS implementation uses the default content world (`.page`):

```swift
// iOS/DuckDuckGo/TabViewController.swift:1001
public func executeBookmarklet(url: URL) {
    if let js = url.toDecodedBookmarklet() {
        webView.evaluateJavaScript(js)
    }
}
```

Suggestion: Match iOS — drop the explicit content world and call `webView.evaluateJavaScript(js)` (uses `.page` by default). If you want the modern API, use `.page` explicitly: `webView.evaluateJavaScript(js, in: nil, in: .page)`.

mallexxx (commented):
**[executor-review]** No unit tests added for the new bookmarklet logic.

Per [testing.mdc](.cursor/rules/testing.mdc), new behavior should have test coverage. iOS already tests `URL.isBookmarklet()` / `URL.toDecodedBookmarklet()` in `StringExtensionTests.swift`, so the helpers are covered, but the macOS routing logic in this PR is not.

Suggestion: Add at minimum:

1. A test on `Tab` that verifies `setContent(.url(javascriptURL, ...))` short-circuits to bookmarklet execution and does **not** trigger a navigation. Easiest path is to mock `webView.evaluateJavaScript` (or assert no call to `webView.load` happens) for a `javascript:alert(1)` URL.
2. A test on `WindowControllersManager.open(_:target:with:)` that verifies a bookmarklet `Bookmark` routes to `executeBookmarklet` on the selected tab rather than opening a new tab via `open(url, source:target:with:)`.

If mocking `WKWebView` is infeasible in unit tests, consider extracting the `isBookmarklet → executeBookmarklet` decision into a small testable helper.

mallexxx (commented):
**[executor-review]** Edge case in `WindowControllersManager.open(_:target:with:)`: bookmarklet falls through to navigation if no current tab is found.

```swift
if url.isBookmarklet() {
    let eventWindowController = mainWindowController(for: window ?? event?.window)
    let targetWindowController = eventWindowController ?? lastKeyMainWindowController
    if let currentTab = targetWindowController?.mainViewController.tabCollectionViewModel.selectedTab {
        currentTab.executeBookmarklet(url: url)
        return
    }
}
```

If `selectedTab` is nil (e.g. clicked a bookmark from a window with no tabs), the code falls through to `open(url, source:..., target:..., with:)` which will eventually create a tab and try to load `javascript:...`. The new `Tab.setContent` guard catches it, but on a fresh tab with no page loaded `evaluateJavaScript` against an empty `about:blank` is a no-op at best — confusing UX.

Suggestion: Either (a) early-return without doing anything when there's no current tab (current behavior is acceptable but should be explicit), or (b) open a new tab with `about:blank`, wait for it to load, then run the bookmarklet. Option (a) matches iOS semantics. Either way, a comment explaining the chosen behavior would help.

PR #4616, opened 2026-04-28
UTI iOS: Bug - Activating Voice Mode from Expanded UTI does not dismiss keyboard

pikorddg:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211654189969294/task/1213832982626191?focus=true
Tech Design URL:
CC: @aataraxiaa 

#### Issue 1 - original issue from the ticket
Fixes a bug where activating Voice Mode from the expanded UTI (Unified Toggle Input) toolbar would turn voice mode on but leave the keyboard visible. 

**Root cause:** In the expanded UTI, the keyboard's first responder is the UTI input field (`inputBarView` in `UnifiedToggleInputViewController`), not the omnibar. The voice path (`microphoneButtonTapped` → publisher → `unifiedToggleInputDidRequestVoiceSearch()` → `onDuckAIVoiceModeRequested()` → `openAIChatInVoiceMode` → `openAIChatVoiceModeInTab()`) calls `prepareTabForRequest`, which dismisses the omnibar via `viewCoordinator.omniBar.endEditing()`. That call is a no-op when the UTI input owns the first responder, so the keyboard stays up. Voice mode UI loads but the keyboard stays visible together with the UTI input field.

**Fix:** Add a single call to `unifiedToggleInputCoordinator?.dismissOmnibarKeyboard()` in `openAIChatVoiceModeInTab()` before `prepareTabForRequest`. `dismissOmnibarKeyboard()` already guards on `displayState`: it only resigns the input bar's first responder for `.omnibar(.active)` and `.aiTab(.expanded)`, returning early for `.hidden`, `.aiTab(.collapsed)`, and `.omnibar(.inactive)`. This makes the change a precise, scoped fix for the buggy state and a no-op everywhere else. Optional chaining covers the case when the `unifiedToggleInput` flag is OFF (coordinator is `nil`).

#### Issue 2 - spotted during testing
This PR also fixes an issue where sometimes in the NTP → tap on the Duck.ai button in the omni bar → open Duck.ai flow, the button was showing a greyed-out arrow instead of the voice button (assuming the duckAIVoiceShortcut flag is enabled). https://github.com/user-attachments/assets/1ed95f34-1187-4650-884e-ec12c20c50aa

**Root cause (Bug #2):**
The toolbar voice/send button visibility depends on `isToolbarAIVoiceChatActive`, a flag refreshed only via
`UnifiedToggleInputCoordinator.updateToolbarAIVoiceChat()`. While `activateFromOmnibar`, `updateInputMode`, and `syncInputModeFromExternalSource` all call it after changing `inputMode`, `showCollapsed` and `showExpanded` did not — they updated `inputMode` via `setInitialInputMode` without notifying the toolbar. As a result, when UTI transitioned through these two methods (e.g. entering an AI tab fresh), the flag somehow stayed stale and the toolbar rendered a greyed-out submit arrow instead of the active voice mic. The fix adds the missing call to both methods.

### Testing Steps -- Bug #1  (keyboard)
Note: while testing, you may hit rate limit error. If that’s the case, log in to ai.duckduckgo.com

**Repro the original bug (must now pass):**
1. From NTP Open [Duck.ai](http://duck.ai/) (text mode).
2. Tap the input to bring up the keyboard (UTI is now expanded).
3. Tap the voice (mic) button in the UTI toolbar.
4. Expected: voice mode turns on **and** the keyboard is dismissed.

**Regression checks for other voice entry points:**
6. Voice from the omnibar while it is focused — keyboard dismisses and voice mode opens (this path already worked; verify no regression).
7. Add DDG Custom Shortcuts widget on your home screen and later edit it to have a [Duck.ai](http://duck.ai/) Voice shortcut there -> tap -> voice mode opens normally -> no keyboard visible
8. Toggle the `unifiedToggleInput` feature flag OFF (legacy Switch Bar) and trigger voice mode — `editingStateViewController?.dismissAnimated` path still dismisses the keyboard; coordinator is `nil` and the new line is a no-op.

**Testing Steps — Bug #2 (greyed-out submit arrow instead of voice mic in Duck.ai expanded UTI)**

Prerequisites: enable both `unifiedToggleInput` and `duckAIVoiceShortcut` feature flags.
1. Open NTP and dismiss the keyboard. Tap Duck. ai button in the omni bar
2. When toggle is in [Duck.ai](http://duck.ai/), action button (blue one) should show voice button
3. Type any text into the native input → the button switches to the **active arrow up (submit)** icon
4. Clear the text → the button switches back to the **voice mic** icon
5. Toggle to Search mode → native input toolbar hides
6. Toggle back to [Duck.ai](http://duck.ai/) → native input toolbar shows with voice icon
7. Toggle to search, type some text and then toggle to [Duck.ai](http://duck.ai/) -> native input toolbar shows, icon is active arrow up

**Regression checks:**
8. Disable `duckAIVoiceShortcut` (keep `unifiedToggleInput` enabled) and restart the app: the rightmost button should always show a submit arrow (never the voice mic), regardless of input mode. Confirms that the added `updateToolbarAIVoiceChat()` calls correctly compute `false` when the feature is unavailable, producing no UI change from the previous behavior.
9. Disable `unifiedToggleInput` and restart the app: the legacy Switch Bar code path is used, and the new calls in `UnifiedToggleInputCoordinator` are unreachable.

### Impact and Risks

**Low.**
- One-line change in a single, isolated path (`openAIChatVoiceModeInTab`).
- `dismissOmnibarKeyboard()` is internally guarded by `displayState`, so it cannot resign a first responder we don't own.
- Optional chaining prevents any effect when the UTI flag is OFF.
- No public API, configuration, or storage changes.

#### What could go wrong?

- Resigning the UTI input's first responder during a state where another component expects it to remain active. Mitigated by the displayState guard inside `dismissOmnibarKeyboard()` — only the two known-active states (`.omnibar(.active)`, `.aiTab(.expanded)`) trigger the resign.
- Interaction with the existing `keyboardDidHide → collapseExpandedUTIOnKeyboardDismiss → showCollapsed` cascade. Verified manually that this cascade is idempotent: collapsing an already-collapsing UTI has no adverse effect.

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Small, UI-scoped changes limited to keyboard dismissal and toolbar state refresh; no data/auth logic is touched, and new tests cover the updated behavior.
> 
> **Overview**
> Fixes a UI bug where starting Duck.ai Voice Mode from an expanded Unified Toggle Input left the keyboard visible by explicitly calling `unifiedToggleInputCoordinator?.dismissOmnibarKeyboard()` before loading voice mode.
> 
> Also tightens toolbar voice/mic state updates by making `applyToolbarPresentation()` refresh `isToolbarAIVoiceChatActive` (and removing a couple of scattered calls), with new unit tests covering voice-chat state syncing across collapsed/expanded and mode transitions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f969f15e351be4e5cab191cfaddfb1e475b615c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -287,6 +288,7 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
 
         viewController.apply(renderState.viewConfig, animated: false)
         applyToolbarPresentation()
+        updateToolbarAIVoiceChat()
```
jaceklyp: **Major: Bug #2 fix is incomplete on a sibling code path.**

The new `updateToolbarAIVoiceChat()` calls in `showCollapsed`/`showExpanded` are correct — `setInitialInputMode` runs synchronously first (mutates only `inputMode` / `committedInputMode`), then this call reads `inputMode`. `@MainActor` + no async between the two; ordering is sound.

However, `deactivateToOmnibar` (line ~390) has the same drift pattern that Bug #2 fixes here:

```swift
func deactivateToOmnibar(resetView: Bool = true, animateDismiss: Bool = true) {
    guard isOmnibarSession else { return }
    inputMode = committedInputMode   // mutates inputMode directly
    ...
    applyToolbarPresentation()       // refreshes tools but NOT voice/send icon
}
```

It mutates `inputMode` directly (bypassing `updateInputMode`) and calls `applyToolbarPresentation()` but not `updateToolbarAIVoiceChat()`. It's self-correcting because the next state entry (`activateFromOmnibar` / `showCollapsed` / `showExpanded`) re-syncs the flag and `displayState = .hidden` makes the toolbar invisible in the meantime, but the flag still drifts briefly.

**Suggested fix — root cause, not patch:** fold `updateToolbarAIVoiceChat()` into `applyToolbarPresentation()`:

```swift
func applyToolbarPresentation() {
    refreshToolsPresentation()
    updateToolbarAIVoiceChat()
}
```

`updateToolbarAIVoiceChat` is currently called from 7 sites (5 existing + 2 new in this PR). Every site is paired with — or near — an `applyToolbarPresentation()`/state mutation. Folding it in collapses 7 "remember to also call updateToolbarAIVoiceChat" duties into 1 and prevents this exact class of bug. The standalone calls in `updateInputMode`, `updateAIVoiceChatAvailability`, and `syncInputModeFromExternalSource` could either keep the explicit call or pick up the change for free.

If folding is out of scope for this PR, at minimum add `updateToolbarAIVoiceChat()` to `deactivateToOmnibar` for symmetry.

**Suggested tests** in `UnifiedToggleInputCoordinatorTests.swift` — ~20 lines, two cases:

```swift
func test_showCollapsed_whenAIVoiceChatEnabled_setsToolbarVoiceChatActive() {
    // arrange: coordinator with handler.isAIVoiceChatEnabled = true
    sut.showCollapsed()
    XCTAssertTrue(sut.viewController.isToolbarAIVoiceChatActive)
}

func test_showExpanded_inSearchMode_clearsToolbarVoiceChatActive() {
    // arrange: handler.isAIVoiceChatEnabled = true
    sut.showExpanded(inputMode: .search)
    XCTAssertFalse(sut.viewController.isToolbarAIVoiceChatActive)
}
```

This is exactly the state-machine sync category where unit tests pay back — they would have caught Bug #2.

jaceklyp (commented):
Reviewed the 3-line diff in detail. Both bugs' fixes are correct for the paths they target. Two findings below worth addressing before/after merge — one is a Bug #2 completeness concern on a sibling code path, the other is a design-smell follow-up. Plus a small unit-test recommendation for the state-sync logic.

jaceklyp:
tested and it works as described

jaceklyp (approved):
Please take a look at agentic review - I think it makes sense to fix a sibling code path thing.

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
pikorddg: Good catch, thanks! Folded into `applyToolbarPresentation()`. Left explicit calls in `updateAIVoiceChatAvailability` (no `applyToolbarPresentation` call) and `syncInputModeFromExternalSource` (`refreshToolsPresentation` there is conditional)

Tests added as well!

jaceklyp (approved):

iOS/DuckDuckGoTests/UnifiedToggleInput/UnifiedToggleInputCoordinatorTests.swift
```diff
@@ -1360,6 +1360,32 @@ final class UnifiedToggleInputCoordinatorTests: XCTestCase {
         XCTAssertEqual(mockDelegate.committedMode, .aiChat)
     }
 
+    // MARK: - Toolbar Voice Chat State Sync
+
+    func test_showCollapsed_whenAIVoiceChatEnabled_setsToolbarVoiceChatActive() {
+        sut.updateAIVoiceChatAvailability(true)
+        sut.showCollapsed()
+        XCTAssertTrue(sut.viewController.isToolbarAIVoiceChatActive)
+    }
```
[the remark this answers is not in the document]
pikorddg: Will handle in next PR!

PR #4622, opened 2026-04-28
[iOS] Fix AuthV2 event validation

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1214363189973478?focus=true
Tech Design URL:
CC:

### Description

This PR fixes AuthV2 event validation by correcting the list of process names that we support, and adding additional error params.

### Testing Steps

1. Check that the changes look correct, and that the `base_event` change is safe

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Schema-only change that broadens an enum to accept an additional app/process name; risk is limited to telemetry validation/ingestion behavior.
> 
> **Overview**
> Updates the iOS wide-event base schema and the generated `ios-authv2-token-refresh` schemas (`1.0.0` and `1.0.1`) to include `PacketTunnelProvider` as an allowed value for `app.name`.
> 
> This relaxes schema validation so AuthV2 token refresh events sent from the tunnel provider process are accepted instead of rejected due to an unrecognized app/process name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27cbc6c1e8995ca9de6faeca76ad39bedb318d5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/PixelDefinitions/wide_events/definitions/auth-v2.json5
```diff
@@ -61,6 +61,14 @@
                     "underlying_code": {
                         "type": "integer",
                         "description": "The underlying error code"
+                    },
+                    "underlying_domain2": {
```
samsymons: Hmm, it seems like this change doesn't actually pass validation - I thought it would be fine. Any ideas @ladamski?

samsymons: Ah, no of course this won't work since we only handle the top level code and domain [here](https://github.com/duckduckgo/pixel-schema/blob/901331c64caedb6a7831dcc9a6e66d9b2f737ea8/schemas/wide_event_schema.json5#L195).

ladamski: Ah, yes.  Its only matching literal values, not substrings.  Pretty sure we could support a keyPattern here but would also want to update the wide event format since it doesn't actually document this pattern.

Another path here is to support actual arrays (not enums) instead of the bare string.   We're running into a related issue with pixels actually: https://app.asana.com/1/137249556945/task/1214280563850698?focus=true

samsymons: Reverting the code/domain change for now, shipping the PacketTunnelProvider change will squash a ton of validation errors that are coming in and reduce the noise a lot.

ladamski (approved):

PR #4634, opened 2026-04-29
[executor] Hovered link tooltip covers content in the bottom

mallexxx:
## Task
[Hovered link tooltip covers content in the bottom](https://app.asana.com/0/0/1204013224241988)

## Root Cause
The hovered-link URL tooltip was always pinned to the bottom-leading corner *inside* the window. When the cursor was near the bottom edge (e.g. hovering a footer link), the tooltip overlapped the link. The presentation never moved to a child window even when there was space below the parent; in fullscreen / zoomed windows it could not move out of the way at all. Additionally, when the Web Inspector took focus, the in-page hover script's `mouseout` was not delivered, so the tooltip stayed visible after the cursor left the WebView.

## Fix Summary
Introduced `HoveredLinkTooltipPresenter` that owns show/hide of the tooltip and picks one of three placements based on cursor location, fullscreen / zoomed state, and available space:

- **`insideBottomLeft`** — default (existing in-window subview, cursor far from bottom edge)
- **`belowWindow`** — borderless child `NSWindow` floats just under the parent (Chrome-style)
- **`aboveCursor`** — child `NSWindow` above the cursor for fullscreen / zoomed windows where there is no space below

A window-level `mouseMoved` monitor feeds cursor location to the presenter so it can flip presentations and detect when the cursor leaves the WebView's bounds (Inspector takeover) to dismiss the tooltip.

## Testing
- **Build:** ✅ `xcodebuild build -scheme "macOS Browser" -configuration Debug` succeeded
- **Unit tests:** ✅ 13/13 passed in `HoveredLinkTooltipPresenterTests` (covers bottom-edge band, fullscreen, zoomed, no-screen, no-cursor, and cursor-leaves-WebView geometry)
- **UI tests:** Added 4 new cases in `HoveredLinkTooltipUITests` — covered by CI on the macOS UI Tests scheme:
  - `testTooltipMovesOutOfTheWayWhenHoveringBottomLeftLink`
  - `testTooltipUsesInWindowPlacementWhenHoveringMiddleLink`
  - `testTooltipDismissesWhenCursorLeavesWebView`
  - `testTooltipAppearsInFullScreenMode`
  - `testTooltipAppearsInMaximizedWindow`

## Self-review
- [x] Change matches root cause analysis
- [x] Scope limited to affected files (no unrelated changes)
- [x] Regression test added (unit + UI)
- [ ] Privacy/security implications checked — none expected; no data leaves the process

## Notes for review
- The borderless tooltip child window uses `NSWindow.collectionBehavior = [.transient, .ignoresCycle, .fullScreenAuxiliary, .canJoinAllSpaces]` and `level = .floating`, mirroring how Chrome's overlay behaves on macOS. The window is added as a child of the main window so it follows space transitions and is torn down with the parent.
- The "near-bottom" threshold is 96 pt (matches the typical browser status-bar reveal band). Tunable via `HoveredLinkTooltipPresenter.nearBottomEdgeThreshold`.
- The cursor-leaves-WebView dismissal triggers when the cursor moves to a sibling view in the same window (Inspector pane, address bar, etc.). Moving outside the window is still handled by `windowDidResignKey`, which now calls `reset()` so floating tooltips don't linger on a no-longer-key window.

---
Generated by Executor — DRAFT until Alex flips to ready-for-review.

### DIFF
--- a/macOS/DuckDuckGo-macOS.xcodeproj/project.pbxproj
+++ b/macOS/DuckDuckGo-macOS.xcodeproj/project.pbxproj
@@ -1033,6 +1033,7 @@
 		3707C72A294B5D2900682A9F /* URLExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA8EDF2324923E980071C2E8 /* URLExtension.swift */; };
 		3707C72D294B5D4100682A9F /* EmptyAttributionRulesProver.swift in Sources */ = {isa = PBXBuildFile; fileRef = B6AE39F029373AF200C37AA4 /* EmptyAttributionRulesProver.swift */; };
 		3707C72F294B5D4F00682A9F /* WebViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3783F92229432E1800BCA897 /* WebViewTests.swift */; };
+		D14C7C00010419880000B005 /* HoveredLinkTooltipPresenterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D14C7C00010419880000A004 /* HoveredLinkTooltipPresenterTests.swift */; };
 		370A34B12AB24E3700C77F7C /* SyncDebugMenu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 370A34B02AB24E3700C77F7C /* SyncDebugMenu.swift */; };
 		370A34B22AB24E3700C77F7C /* SyncDebugMenu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 370A34B02AB24E3700C77F7C /* SyncDebugMenu.swift */; };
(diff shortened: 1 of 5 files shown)


mallexxx (commented):
Self-review by Executor:
- ✅ Change matches root cause: introduces a presenter that flips between in-window, below-window child window, and above-cursor placements based on cursor position + window state, plus a `cursor-leaves-WebView` dismissal that fixes the Inspector stuck-tooltip bug.
- ✅ Scope limited: 1 modified controller, 1 new presenter, 1 unit test class, 1 UI test class, 1 pbxproj wiring change. No drive-by edits.
- ✅ Test added: 13 unit tests cover the pure-data position computation; 5 UI tests cover the four UX scenarios from the spec.
- ✅ Privacy/security: none — the tooltip text is the same href that was already shown.
- ✅ Matches codebase patterns: uses `addLocalCancellableMonitor`, `ColorView`, and `NSEvent` monitors consistent with neighboring extensions in `BrowserTabViewController`.

PR #4644, opened 2026-04-29
[iOS] Duck.ai multi-section suggestions (chats / top URLs / Search DDG)

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1214043381369789
CC: @aataraxiaa 

### Description

Implements the spec from the linked Asana task: in Duck.ai mode, replaces the URL-fallback page-switching mechanism with one inline three-section suggestions view rendered below the UTI input — **Recent chats** (filtered by query), **Top URLs** (≤3, reusing the same Search-side ranking via `SuggestionLoader`), and an always-visible **Search DuckDuckGo** row that flips the toggle to Search and submits the query (symmetric with Search-side "Ask privately").

Architecture is `DuckAISuggestionsCoordinator` + `DuckAISuggestionsViewController` + `DuckAIURLSuggestionsLoader`, wired through `SwipeContainerManager` and `UnifiedInputContentContainerViewController`. The legacy `OmniBarEditingStateViewController` (non-UTI path) and the iPad floating-panel coordinator are intentionally untouched — both still use `AIChatHistoryManager` directly.

Side fixes captured along the way: ported the URL-suggestion display formatter and URL-only filter to shared files (one copy each across Search-side and Duck.ai-side); added per-fetcher `lastCompletedFetchQuery` tracking so the Dax empty-state doesn't flash during the keystroke-to-result lag (one fetcher's empty result for a new query landing before the other's); added `removeDuplicates()` on the chat fetcher's text subscription to skip redundant re-fetches; discard out-of-order URL fetch callbacks via a `latestDispatchedQuery` check.

### Testing Steps

**Setup:** iPhone simulator/device. Set Duck.ai as the default omnibar mode (Settings → DuckDuckGo AI Chat → Omnibar mode → Duck.ai). Have at least a few recent Duck.ai chats and visited some URLs in history.

1. Navigate to a webpage (any URL). Tap the address bar so the UTI opens with the URL prefilled. Verify the suggestions list shows three sections as applicable: Recent chats matching the URL (or empty), Top URLs (≤3), and the always-visible "Search DuckDuckGo" row. **No empty space, no empty header above empty list, no Dax logo on top of the list.**
2. With Duck.ai-mode UTI open, type a query you've chatted about before. Verify the matching recent chats appear under the chats section.
3. Tap a recent chat row → opens that chat.
4. Tap a Top URL row → loads that URL.
5. Tap the **Search DuckDuckGo** row → toggle flips to Search and the search executes (loads `duckduckgo.com/?q=…`).
6. Type a no-match query (random letters), then backspace all the way to empty. **Verify the Dax logo does not flash on top of the recent chats** during the transition.
7. Dismiss and re-open the UTI several times. Verify state is consistent each time and there's no leak in behavior (e.g. duplicate reload bursts).
8. Switch the omnibar position to top, then bottom. Repeat steps 1–6 in each.
9. Switch to landscape. Verify the suggestions list still renders correctly.
10. **Search-mode regression check:** switch toggle to Search side; type a query; verify autocomplete behaves identically to before this PR.
11. **iPad regression check:** on iPad, open Duck.ai full tab and type a query — verify the floating recent-chats panel still renders (this PR's design changes don't apply on iPad; the existing chat-only panel is preserved).
12. **Fire tab regression check:** in a fire tab, open Duck.ai UTI — verify chat suggestions are not fetched (no chat section), URL suggestions still work normally.

### Impact and Risks

**Medium.** Duck.ai-mode suggestions are a daily-use surface and this PR replaces the rendering machinery wholesale, so visual or interaction regressions would be immediately user-facing.

Notable risk areas:
- The `DuckAISuggestionsCoordinator` settle-tracking gates the Dax empty-state on both fetchers reporting `lastCompletedFetchQuery == currentText`. If a fetcher fails to record completion in some path, Dax will stay permanently suppressed — verified for current paths (success, empty-skip, stale-discarded, cancelled), but worth a careful review.
- Legacy `OmniBarEditingStateViewController` and `IPadTabChatHistoryCoordinator` still use the older `AIChatHistoryManager` install path + URL fallback machinery. The shared chat manager gained a `lastCompletedFetchQuery` field and a `removeDuplicates()` on its text subscription — both are additive and shouldn't change legacy behavior, but flagged.
- `URL+SuggestionDisplay` and `Suggestion+URLFilter` were extracted from fileprivate copies; both call sites (Search-side autocomplete + new Duck.ai loader) now share them. Behavior is identical — pure mechanical extraction.

Mitigations: feature is gated by the existing `aiChatSuggestions` feature flag and `aiChatSettings.isChatSuggestionsEnabled` setting, so it can be turned off without code changes.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes the Duck.ai UTI suggestions surface by replacing the previous URL-fallback/page-switching behavior with a new inline coordinator + table-driven UI, which is user-facing and timing-sensitive. Risk centers on settle-tracking/cancellation and UI update coalescing potentially affecting Dax empty-state visibility and suggestion rendering under fast typing.
> 
> **Overview**
> Duck.ai mode in the UTI now shows an inline **three-section suggestions list** (recent chats, top URL hits, and an always-present *Search DuckDuckGo* row) instead of using the prior URL-fallback mechanism that temporarily surfaced Search suggestions.
> 
> This introduces `DuckAISuggestionsCoordinator`/`DuckAISuggestionsViewController` and `DuckAIURLSuggestionsLoader` (reusing Search-side `SuggestionLoader` ranking, capped to 3 URL-type results, with out-of-order callback guarding), wires them into `SwipeContainerManager` and `UnifiedInputContentContainerViewController`, and updates Dax empty-state gating to wait until both chat and URL fetchers have *settled* for the current query.
> 
> Shared helpers for URL-only filtering and URL display formatting were extracted into `Suggestion+URLFilter` and `URL+SuggestionDisplay`, and `AIChatHistoryManager` now tracks `lastCompletedFetchQuery` and de-dupes text changes to reduce redundant fetches. New unit tests cover settle logic, URL filtering/capping, and suggestions VC section/header behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a824ab16cf2d22ecceb53994ade4acd1dbae73ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/Suggestions/DuckAISuggestionsViewController.swift
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    weak var delegate: DuckAISuggestionsViewControllerDelegate?
+
+    private let chatViewModel: AIChatSuggestionsViewModel
+    private let urlLoader: DuckAIURLSuggestionsLoader
+    private let queryProvider: () -> String
+
+    /// Sized to absorb the gap between chat-fetcher and URL-fetcher settle times so a single reload renders both.
+    private static let reloadCoalesceMilliseconds = 120
```
pikorddg: Can this be related to LP6?

pikorddg: Also, got this when run agent review

DuckAISuggestionsViewController.swift:130-136 — "Search DuckDuckGo" row title lags behind text during fast typing

reload() only fires from (a) the merged 120 ms debounce on $filteredSuggestions + $topURLs and (b) setQueryActive on the empty↔non-empty flip. Cells aren't refreshed per keystroke, so configureSearchCell's queryProvider() read in cellForRowAt shows stale text until fetchers settle (~220 ms after typing pause). Submitted query is correct (read live on tap), but the displayed title visibly lags on the PR's headline feature.

Suggested fix: add currentTextPublisher (with removeDuplicates()) as a third participant in the Publishers.MergeMany, or capture the query at reload time and use that copy in configureSearchCell.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/Suggestions/DuckAISuggestionsCoordinator.swift
```diff
@@ -0,0 +1,157 @@
...
+    private let chatViewModel: AIChatSuggestionsViewModel
+    private let queryProvider: () -> String
+
+    private var viewController: DuckAISuggestionsViewController?
+    private var cancellables = Set<AnyCancellable>()
+
+    var hasCompletedInitialChatFetch: Bool { chatManager.hasCompletedInitialFetch }
+
+    /// True when both fetchers have settled for `query`. Container gates Dax visibility on this to avoid mid-keystroke flashes.
+    func hasSettled(forQuery query: String) -> Bool {
```
pikorddg, line 48: Agentic review: 

DuckAISuggestionsCoordinator.swift:51-54 — hasSettled(forQuery:) (the core of the Dax-flash fix) has no test

The PR description names this as the mechanism that "covers both the initial-load window and the keystroke-to-result lag". The logic (chat == query && url == query) is non-trivial and a regression from dropping either side would slip through silently.

Suggested fix: small coordinator-level test with two mock fetchers covering: "only chat settled" → false; "both settled for current query" → true; "both settled but for a stale query" → false.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/Suggestions/DuckAISuggestionsViewController.swift
```diff
@@ -0,0 +1,452 @@
...
+            tableView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            tableView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            tableView.bottomAnchor.constraint(equalTo: view.keyboardLayoutGuide.topAnchor)
+        ])
+
+        // Reload on fetcher settle (not text change), so the search-row title and section data update together.
+        let chatChanges = chatViewModel.$filteredSuggestions.map { _ in () }.eraseToAnyPublisher()
+        let urlChanges = urlLoader.$topURLs.map { _ in () }.eraseToAnyPublisher()
+        Publishers.MergeMany([chatChanges, urlChanges])
+            .debounce(for: .milliseconds(Self.reloadCoalesceMilliseconds), scheduler: DispatchQueue.main)
```
pikorddg, line 137: **7\.** DuckAISuggestionsViewController.swift:69-70, 134 constant reloadCoalesceMilliseconds = 120 **is coupled to two independent fetcher debounces (100 ms + 150 ms)**
The comment says it's "sized to absorb the gap between chat-fetcher and URL-fetcher settle times", but the three values live in three separate files (AIChatHistoryManager.Constants.debounceMilliseconds = 150, DuckAIURLSuggestionsLoader.debounceMilliseconds = 100, and 120 here). A future change to either debounce won't signal that the coalesce needs updating; if it ends up too short, the user sees two reloads in sequence (URL data first, chat catching up).
**Suggested fix:** colocate the three constants and express the relationship explicitly (coalesce = max(chat, url) + buffer), or at minimum add a comment in reloadCoalesceMilliseconds listing the dependent values and "increase if either changes".

pikorddg (commented):
LP1/ First step fails: despite having toggle position set to [Duck.ai](http://duck.ai/) -> after tap I’m on Search mode

LP2/ Following LP1 (by switching to Duck) -> not sure what you mean by “Empty space”. There is ton of empty space there :D. Looks good, but to be safe, attaching a screenshot

https://github.com/user-attachments/assets/71f5d55c-fdd6-4917-9055-db6cb2b5413b

LP3/ I’m not sure if “Recent chats matching the URL (or empty)” part is working according to the expectations. At this point it should match bandcamp?

https://github.com/user-attachments/assets/03a1a989-1479-4525-8c33-fdf61626344d
https://github.com/user-attachments/assets/ec3c0113-085a-465b-8036-108fc3ef8c50

LP4/ Tap the **Search DuckDuckGo** row → toggle flips to Search and the search executes (loads duckduckgo.com/?q=)
Search is performed correctly but “toggle flips to search” -> that part is not visible -> UTI is dismissed automatically and search is performed. To me this looks correctly, but wanted to be explicit.

LP5/ Step 6 No logo (correctly), but there’s some lag when deleting chat qucikly -> issue in Combine chains? See recording 

https://github.com/user-attachments/assets/b9d0df0c-ebe2-4bcc-8cac-3ca188caa9bc

LP6/ Step 10 I’m afraid something is broken here: because now I’m constantly on Duck.ai instead of search. Suggestions work correctly though.

jaceklyp:
LP1/ Search/Duck.ai default is only for newly created tabs, perhaps this tab was already in duck.ai mode, you would have to do some search so it flips to search for this particular tab
LP2/ LGTM
LP3/ Yes I think so, you put url and it stopped matching recent chats (it's not logic on our side really)
LP4/ LGTM
LP5/ there's no attachment, but please check again, perhaps it's fixed now?
LP6/ I need clarification, I think this might be correct behaviour, if this was search tab (like in LP1)

pikorddg (approved):

PR #4709, opened 2026-05-04
Per-tab state for the unified input field (iOS)

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1214281344740600?focus=true
Tech Design URL:
CC:

### Description
Each iOS tab now remembers its own unified-input draft — text, toggle position, attachments, model id, reasoning mode, and selected tool — across tab switches, with a `UnifiedInputStateStoring` store keyed by `tab.uid` and `LastUsedInputDefaults` seeding new tabs from the most recent deliberate user choice. Submission clears the active tab's entry eagerly; dismiss-time `clearText` only clears the visible input so unsubmitted drafts survive. New tabs are activated even from the no-link/NTP refresh path so the shared coordinator's strip can't carry over the previous tab's attachments, and `applyState` syncs the live model store unconditionally so reasoning/model don't leak between tabs.

### Testing Steps
1. Open two tabs, type different drafts in each, switch back and forth — each tab restores its own text, toggle, and attachments. Open a third tab — it inherits the most recent deliberate model/reasoning, not whatever the active tab was showing.
2. In an AI tab, attach an image, dismiss the omnibar without submitting, switch tabs and back — the attachment and text are still there. Submit a prompt — text and attachments clear in that tab; mode/model/reasoning/tool persist for the next prompt.
3. Set reasoning to `fast` in tab A and `reasoning` in tab B, navigate across several tabs and back to A — A still shows `fast`, not `reasoning`.

### Impact and Risks
Medium

#### What could go wrong?
The shared coordinator/state-store sits on hot UI paths (omnibar focus, tab switch, refresh) — a missed propagation hook or guard could surface as an empty draft, an attachment leaking across tabs, or a reasoning/model selection bleeding between tabs.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes tab persistence (NSCoding) and omnibar/unified-input state handling on hot UI paths, which could cause drafts/attachments or model/tool selections to desync or leak across tabs if any edge case is missed.
> 
> **Overview**
> Each iOS tab now maintains its own unified-input state (draft text, toggle mode, attachments, selected model/reasoning/tool) keyed by `tab.uid`, with a new `UnifiedInputStateStore` that seeds new tabs from *last deliberate user choice* and observes both normal + fire `TabsModel`s for insert/evict.
> 
> Tab persistence is expanded: `Tab` replaces `preferredTextEntryMode` with `unifiedInputState` and encodes/decodes model ID, reasoning mode, and tool so tab restores survive relaunch; `MainViewController`/`UnifiedToggleInputCoordinator` are updated to activate per tab, flush/apply state on tab switches, and eagerly clear the active tab’s stored text/attachments/tool on submission while keeping dismiss-time `clearText()` as a visual-only scrub.
> 
> AI Chat preferences gain `selectedTool` persistence, model selection restoration now writes through nils via `UTIModelStore.applyPersistedSelection`, and `ToggleModeStorage` avoids redundant writes; extensive new unit tests cover per-tab state, legacy decode behavior, and regression cases (tab switching, external submissions, tool/model/reasoning leakage).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6909816e3cda147785ec353906af0da50f363b48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -2086,6 +2087,7 @@ class MainViewController: UIViewController {
             viewCoordinator.omniBar.setDaxEasterEggLogoURL(nil)
             if let tabModel = tabManager.currentTabsModel.currentTab {
                 viewCoordinator.omniBar.setSelectedTextEntryMode(displayedTextEntryMode(for: tabModel))
+                unifiedToggleInputCoordinator?.activateForTab(tabModel.uid)
```
aataraxiaa: Two questions:

1. `activateForTab` is called here directly and also as part of the `refreshUnifiedToggleInput` call below. Possible to only have one of these?

2. Agent flagged below comment, I think relevant as I had issues previoulsy with the UTI due to how many times `refreshOmnibar` is called. Give to agent to assess maybe:

```
activateForTab is called from every refreshOmniBar() — is the same-uid re-apply intentional?                                                                                                                            
  MainViewController.refreshOmniBar is called from a lot of places (focus, blur, navigations, fire-tab toggles, refreshUnifiedToggleInput itself — see grep above). Each call hits coordinator.activateForTab(uid), and when
  previous == uid the code falls through to applyState(stateStore.state(for: uid)) regardless. applyState then:                                                                                                              
                                                                                                                                                                                                                           
  - calls viewController.removeAllAttachments() and re-adds every attachment in state.attachments,                                                                                                                           
  - calls modelStore.applyPersistedSelection(...) which writes through to preferences.selectedModelId/ShortName/selectedReasoningMode,                                                                                     
  - runs handleModelsUpdated, updateReasoningPicker, and refreshes tools presentation.                                                                                                                                       
                                                                                                                                                                                                                             
  The AIChatPreferencesPersistor.selectedReasoningMode setter now early-returns on identical values (good), but the model-id setter and the attachment teardown/reapply don't have that guard. Worth confirming whether you  
  want an early return at the top of activateForTab when previous == uid, or whether you intentionally want this re-sync on every refresh because some other code path mutates the live state out from under the store.      
  (Asking because re-removing/adding attachment chips on every navigation focus event feels like it might cause hard-to-spot UI churn or animation oddities, and I couldn't tell from the tests.) —                          
  UnifiedToggleInputCoordinator.swift:295-340, MainViewController.swift:2090  
```

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputTabState.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+
+import AIChat
+
+/// Per-tab unified address-bar configuration. Persisted with the tab via
+/// NSCoding so reopening the app restores the user's last-used input setup
+/// rather than falling back to the global default.
+public struct UnifiedInputTabState: Equatable {
+    public var preferredTextEntryMode: TextEntryMode
+    public var selectedModelID: String?
```
aataraxiaa, line 27: Agent flag, but I think possibly valid:

```
selectedTool not persisted across app restart — intentional asymmetry?                                                                                                                                                  
  TabInputState carries selectedTool, but UnifiedInputTabState (the one encoded onto Tab via NSCoding) only stores preferredTextEntryMode, selectedModelID, and selectedReasoningMode. So tool selection survives in-process
  tab switches but is lost on app relaunch. Same for the LastUsedInputDefaults seed for new tabs — lastUsed.selectedTool resets to nil because the store's init only reseeds toggleMode, selectedModelID,                    
  selectedReasoningMode from durable storage. If that's the product call (tools feel ephemeral compared to model choice), all good — just want to make sure it's deliberate vs. an oversight. —                            
  UnifiedInputTabState.swift:24-34, UnifiedInputStateStore.swift:34-44        
```

aataraxiaa (approved):
Works great 🚀. Two comments added. Also I think you made some good decisons here re: behavior and so can you document them in a task under ‘Developer Docs’ in the component project. Then flag Android. I did the same here - https://app.asana.com/1/137249556945/project/1214157224317277/task/1214390835017893. Just asked agent to create a spec for another platform so they can implement same behavior.

iOS/DuckDuckGo/MainViewController.swift
Bunn: addressed on https://github.com/duckduckgo/apple-browsers/pull/4709/commits/90cfa297894785e52a407befe75b01ea0bc14fa3

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputTabState.swift
Bunn: Implemented the persistence on tool as well https://github.com/duckduckgo/apple-browsers/pull/4709/commits/41a5a4db8aac62f01eeb98f83c05a423c5c7253f

PR #4743, opened 2026-05-06
Add auto-dismiss capability to Remote Messaging Framework

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1214525802264426
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1214488238400075
CC: N/A

### Description
- Add `displayConditions` wrapper object to RMF message JSON config, nesting `dismissAfterDaysShown` and a new `trigger` field
- Add `MessageTrigger` enum and `DisplayConditions` struct to the domain model; messages with unknown triggers are silently discarded for forward compatibility
- Extend `fetchScheduledRemoteMessage` with an optional `trigger` parameter for filtering messages by their trigger context
- Thread idle-return context through iOS: `HomePageConfiguration.refresh(openedAfterIdle:)` passes `trigger: .afterIdle` when the user returns from idle, and `NewTabPageMessagesModel` forwards this via an `isOpenedAfterIdle` closure
- After-idle messages now appear in both the main NTP and the focussed NTP (suggestion tray), by passing `hasEscapeHatch` to the focussed NTP in `SuggestionTrayViewController`
- Auto-dismiss via `dismissAfterDaysShown` continues to work as before, now nested under `displayConditions`

### Testing Steps

**Test 1 -- Regular message (no trigger, no auto-dismiss):**
1. Apply config from https://pastebin.com/raw/zTS5cAE1 via Debug > Remote Messages > Set source URL, then reset messages and reload
2. Verify the "Regular Message" appears on the NTP on normal launch
3. Dismiss it and verify it stays dismissed

**Test 2 -- After-idle trigger only:**
1. Apply config from https://pastebin.com/raw/MmpU87dY via Debug > Remote Messages > Set source URL, then reset messages and reload
2. Kill and relaunch the app normally -- the message should **not** appear
3. Trigger idle return (background app, wait for threshold, return) -- the "Welcome Back!" message should appear on the main NTP
4. Tap the address bar -- the message should also appear in the focussed NTP
5. Dismiss the escape hatch, open a new tab -- the message should not appear (no idle context)

**Test 3 -- Normal app usage after idle return:**
1. After idle return, verify the message appears
2. Navigate to a URL, then open a new tab -- the message should not appear (no idle context)

**Test 4 -- After-idle + auto-dismiss combo:**
1. Apply config from https://pastebin.com/raw/azMhR7mc
2. Trigger idle return -- "Time-limited idle message" should appear
3. Move the device date forward by 2+ days, trigger idle return again -- the message should no longer appear (auto-dismissed)

**Test 5 -- Unknown trigger (forward compatibility):**
1. Apply config from https://pastebin.com/raw/RhCG3Kaj
2. The message should **never** appear -- the unknown trigger causes it to be discarded during mapping

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- If the escape hatch state is not set before the focussed NTP's `onAppear` fires, the after-idle message may not show in focussed mode. Mitigated by the fact that `setEscapeHatch` is called before the view is installed in the hierarchy
- Messages with unknown triggers are silently dropped, which is the intended forward-compatibility behavior but could cause confusion if a trigger name is misspelled in the server config

### Quality Considerations
- `displayConditions` is fully backward compatible; messages without it behave identically to before
- Unknown triggers are discarded at the mapping layer to avoid showing messages at wrong times
- Unit tests cover: trigger mapping, unknown trigger discard, backward compat, store trigger filtering, auto-dismiss with trigger combo, iOS wiring (HomePageConfiguration and NewTabPageMessagesModel)

### Notes to Reviewer
- The `displayConditions` wrapper replaces the top-level `dismissAfterDaysShown` field in the JSON schema. The domain model groups both fields under `DisplayConditions`
- The focussed NTP fix was a one-line addition in `SuggestionTrayViewController` to forward the escape hatch state

---

> [!NOTE]
> **Medium Risk**
> Adds a Core Data schema migration plus new message filtering logic (trigger matching and time-based auto-dismiss), which can affect which remote messages are shown/dismissed and requires correct persistence/migration behavior.
> 
> **Overview**
> Introduces a new `displayConditions` block in remote message JSON and domain models, adding support for a `trigger` (e.g. `after_idle`) and nesting `dismissAfterDaysShown`; messages with unknown triggers are now dropped during mapping.
> 
> Updates storage and scheduling to support these conditions: Core Data model bumps to v3 with a new `firstShownDate` field, `fetchScheduledRemoteMessage` gains an optional `trigger` filter, and scheduled messages can be automatically dismissed once the `dismissAfterDaysShown` threshold is exceeded (recording `firstShownDate` on first show).
> 
> Threads the idle-return context through iOS NTP code (`HomePageConfiguration`, `NewTabPageMessagesModel`, focussed NTP via `hasEscapeHatch`) so after-idle messages only appear in the appropriate context, with new unit tests covering mapping, trigger filtering, and auto-dismiss behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 739f75026aa3747c0002dcdaf72b97f9764531bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/CoreData/RemoteMessaging.xcdatamodeld/RemoteMessaging 3.xcdatamodel/contents
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/CoreData/RemoteMessaging.xcdatamodeld/RemoteMessaging 3.xcdatamodel/contents
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="23788.4" systemVersion="24G222" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="RemoteMessageManagedObject" representedClassName="RemoteMessageManagedObject" syncable="YES">
+        <attribute name="firstShownDate" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="id" optional="YES" attributeType="String"/>
+        <attribute name="message" optional="YES" attributeType="String"/>
+        <attribute name="shown" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="status" optional="YES" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="surfaces" optional="YES" attributeType="Integer 16" usesScalarValueType="YES"/>
+    </entity>
+    <entity name="RemoteMessagingConfigManagedObject" representedClassName="RemoteMessagingConfigManagedObject" syncable="YES">
+        <attribute name="evaluationTimestamp" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="invalidate" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
+        <attribute name="version" optional="YES" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
+    </entity>
+</model>
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
@@ -126,7 +126,8 @@ struct JsonToRemoteMessageModelMapper {
                 content: content,
                 matchingRules: message.matchingRules ?? [],
                 exclusionRules: message.exclusionRules ?? [],
-                isMetricsEnabled: message.isMetricsEnabled
+                isMetricsEnabled: message.isMetricsEnabled,
+                dismissAfterDaysShown: message.dismissAfterDaysShown
             )
 
             if let translation = getTranslation(from: message.translations, for: Locale.current) {
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
@@ -120,6 +120,21 @@ struct JsonToRemoteMessageModelMapper {
                 return
             }
 
+            let displayConditions: DisplayConditions?
+            if let jsonConditions = message.displayConditions {
+                if let triggerString = jsonConditions.trigger {
+                    guard let trigger = MessageTrigger(rawValue: triggerString) else {
+                        Logger.remoteMessaging.debug("Unknown trigger '\(triggerString, privacy: .public)' for message \(message.id, privacy: .public), discarding")
+                        return
+                    }
+                    displayConditions = DisplayConditions(trigger: trigger, dismissAfterDaysShown: jsonConditions.dismissAfterDaysShown)
+                } else {
+                    displayConditions = DisplayConditions(trigger: nil, dismissAfterDaysShown: jsonConditions.dismissAfterDaysShown)
+                }
+            } else {
+                displayConditions = nil
+            }
+
             var remoteMessage = RemoteMessageModel(
                 id: message.id,
                 surfaces: surfaces,
@@ -127,7 +142,7 @@ struct JsonToRemoteMessageModelMapper {
                 matchingRules: message.matchingRules ?? [],
                 exclusionRules: message.exclusionRules ?? [],
                 isMetricsEnabled: message.isMetricsEnabled,
-                dismissAfterDaysShown: message.dismissAfterDaysShown
+                displayConditions: displayConditions
             )
 
             if let translation = getTranslation(from: message.translations, for: Locale.current) {
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
@@ -121,18 +121,11 @@ struct JsonToRemoteMessageModelMapper {
             }
 
             let displayConditions: DisplayConditions?
-            if let jsonConditions = message.displayConditions {
-                if let triggerString = jsonConditions.trigger {
-                    guard let trigger = MessageTrigger(rawValue: triggerString) else {
-                        Logger.remoteMessaging.debug("Unknown trigger '\(triggerString, privacy: .public)' for message \(message.id, privacy: .public), discarding")
-                        return
-                    }
-                    displayConditions = DisplayConditions(trigger: trigger, dismissAfterDaysShown: jsonConditions.dismissAfterDaysShown)
-                } else {
-                    displayConditions = DisplayConditions(trigger: nil, dismissAfterDaysShown: jsonConditions.dismissAfterDaysShown)
-                }
-            } else {
-                displayConditions = nil
+            do {
+                displayConditions = try mapToDisplayConditions(message: message)
+            } catch {
+                Logger.remoteMessaging.debug("Invalid display conditions for message \(message.id, privacy: .public): \(error.localizedDescription, privacy: .public)")
+                return
             }
 
             var remoteMessage = RemoteMessageModel(
@@ -152,6 +145,15 @@ struct JsonToRemoteMessageModelMapper {
             remoteMessages.append(remoteMessage)
         }
         return remoteMessages
+    }
+
+    static func mapToDisplayConditions(message: RemoteMessageResponse.JsonRemoteMessage) throws -> DisplayConditions? {
+        try message.displayConditions.flatMap { conditions in
+            let validator = MappingValidator(root: conditions)
+            let trigger = try validator.mapEnumIfPresent(\.trigger, to: MessageTrigger.self)
+            let dismissAfterDaysShown = conditions.dismissAfterDaysShown.map { max($0, 1) }
+            return DisplayConditions(trigger: trigger, dismissAfterDaysShown: dismissAfterDaysShown)
+        }
     }
 
     static func mapToSurfaces(jsonSurfaces: [String]?, supportedSurfacesForMessage: RemoteMessageSurfaceType, messageId: String) -> RemoteMessageSurfaceType? {
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/JsonRemoteMessagingConfig.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/JsonRemoteMessagingConfig.swift
@@ -33,7 +33,7 @@ public enum RemoteMessageResponse {
         let translations: [String: JsonContentTranslation]?
         let matchingRules, exclusionRules: [Int]?
         let metrics: JsonMetrics?
-        let dismissAfterDaysShown: Int?
+        let displayConditions: JsonDisplayConditions?
 
         static func == (lhs: JsonRemoteMessage, rhs: JsonRemoteMessage) -> Bool {
             return lhs.id == rhs.id
@@ -42,6 +42,11 @@ public enum RemoteMessageResponse {
         var isMetricsEnabled: Bool {
             metrics?.state.flatMap(JsonMetrics.MetricsState.init) != .disabled
         }
+    }
+
+    struct JsonDisplayConditions: Decodable, Equatable {
+        let trigger: String?
+        let dismissAfterDaysShown: Int?
     }
 
     struct JsonMetrics: Decodable {
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/JsonRemoteMessagingConfig.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/JsonRemoteMessagingConfig.swift
@@ -33,6 +33,7 @@ public enum RemoteMessageResponse {
         let translations: [String: JsonContentTranslation]?
         let matchingRules, exclusionRules: [Int]?
         let metrics: JsonMetrics?
+        let dismissAfterDaysShown: Int?
 
         static func == (lhs: JsonRemoteMessage, rhs: JsonRemoteMessage) -> Bool {
             return lhs.id == rhs.id
--- a/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
@@ -213,7 +213,7 @@ extension RemoteMessagingStore {
 
 extension RemoteMessagingStore {
 
-    public func fetchScheduledRemoteMessage(surfaces: RemoteMessageSurfaceType) -> RemoteMessageModel? {
+    public func fetchScheduledRemoteMessage(surfaces: RemoteMessageSurfaceType, trigger: MessageTrigger? = nil) -> RemoteMessageModel? {
 
         func predicateForSurfaces(_ surfaces: RemoteMessageSurfaceType) -> NSPredicate {
             // Match any message whose surfaces bitmask overlaps with the requested surfaces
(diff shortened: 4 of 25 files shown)


SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/CoreData/RemoteMessaging.xcdatamodeld/RemoteMessaging 3.xcdatamodel/contents
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="23788.4" systemVersion="24G222" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="RemoteMessageManagedObject" representedClassName="RemoteMessageManagedObject" syncable="YES">
+        <attribute name="firstShownDate" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
```
alessandroboron, line 4: The RMF model has now two properties `firstShownDate: Date?` and `shown: Bool` that could be not in sync with each other (e.g. firstShownDate set and shown not set and vice-versa). We could ensure that in code but it’s more code to maintain. How do you think making the impossible states unrepresentable?

What I mean is dropping `shown` from the .xcdatamodel and have it as computed property of the RemoteMessageManagedObject

```
public var shown: Bool {
   firstShownDate != nil
}
```

This requires a couple of code updates:
- 1. Update the predicates that check `shown` boolean condition to use `firstShownDate` != nil
- 2. Add a Mapping model betwen v2 and v3 with a Custom Migration Policy  (`NSEntityMigrationPolicy`) 

In our case should be a simple migration policy as we will be setting a distant paste date if shown is `true`.

```
class RemoteMessageV2ToV3MigrationPolicy: NSEntityMigrationPolicy {                                                                                                 
      override func createDestinationInstances(forSource source: NSManagedObject,                                                                               
                                               in mapping: NSEntityMapping,                                                                                     
                                               manager: NSMigrationManager) throws {                                                                            
          try super.createDestinationInstances(forSource: source, in: mapping, manager: manager)
          guard let destination = manager.destinationInstances(forEntityMappingName: mapping.name, sourceInstances: [source]).first else { return }                    
          if (source.value(forKey: "shown") as? Bool) == true {                                                                                                 
              destination.setValue(Date.distantPast, forKey: "firstShownDate")                                                                                         
          }                                                                                                                                                     
      }                                                                                                                                                         
  }                                      
```

For more info on custom policy migration https://www.polpiella.dev/custom-core-data-migrations/

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/JsonRemoteMessagingConfig.swift
```diff
@@ -33,6 +33,7 @@ public enum RemoteMessageResponse {
         let translations: [String: JsonContentTranslation]?
         let matchingRules, exclusionRules: [Int]?
         let metrics: JsonMetrics?
+        let dismissAfterDaysShown: Int?
```
alessandroboron: Does this need to be updated to reflect latest changes in the [TD](https://app.asana.com/1/137249556945/project/481882893211075/task/1214488238400075?focus=true)?

```
"displayConditions": {
    "trigger": “after_idle”,
    "dismissAfterDaysShown": 5
},
```

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
```diff
@@ -381,6 +392,9 @@ extension RemoteMessagingStore {
                         return
                     }
                     message.shown = shown
+                    if shown && message.firstShownDate == nil {
```
alessandroboron, line 398: Maybe if this is the only place where we need to keep the properties in sync is fine to skip the migration policy.

alessandroboron (commented):
Overall looks good but I think we need to update the JsonRemoteMessage config to reflect the TD changes?

I also left a comment on the Core Data model but I leave it to you to decide if we want to make that change.

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/CoreData/RemoteMessaging.xcdatamodeld/RemoteMessaging 3.xcdatamodel/contents
SabrinaTardio: Given that shown and firstShownDate are only set together in one place (updateRemoteMessage(withID:asShown:)), the sync risk is minimal for now. The custom migration policy adds complexity that doesn't seem warranted yet given the fact this needs to go in this week release. If we find more places where they need to stay in sync, we can revisit and do the migration then.

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/JsonRemoteMessagingConfig.swift
SabrinaTardio: Yes, updated in the latest push. dismissAfterDaysShown is now nested under displayConditions alongside the new trigger field, matching the tech design.

iOS/DuckDuckGo/NewTabPageViewController.swift
```diff
@@ -95,7 +96,8 @@ final class NewTabPageViewController: UIHostingController<NewTabPageView>, NewTa
                                                 messageActionHandler: remoteMessagingActionHandler,
                                                 imageLoader: remoteMessagingImageLoader,
                                                 pixelReporter: remoteMessagingPixelReporter,
-                                                fireModePromotionEligibility: fireModePromotionEligibility)
+                                                fireModePromotionEligibility: fireModePromotionEligibility,
+                                                isOpenedAfterIdle: { hasEscapeHatch })
```
alessandroboron: Question: 
- 1. Could we use `associatedTab.openedAfterIdle` insteaad of passing `hasEscapeHatch` as parameter of `isOpenedAfterIdle`? It seems that flag is set in `MainViewController.attachHomeScreen`.
- 2. `isOpenedAfterIdle` is a closure that evaluates `hasEscapeHatch` when needed. But `hasEscapeHatch` is a constant value passed at init time that doesn’t change over time. I noticed that we have a `setEscapeHatch(_ model: EscapeHatchModel?)` function that gets called at certain point in time, should `isOpenedAfterIdle` read `newTabPageViewModel.escapeHatch != nil` value?

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
```diff
@@ -120,13 +120,29 @@ struct JsonToRemoteMessageModelMapper {
                 return
             }
 
+            let displayConditions: DisplayConditions?
```
alessandroboron, line 123: Questions:
1. Just to confirm we are discarding the message if `MessageTrigger` can’t be parsed?
2. Should we clamp any value <= 0 to 1 for `dismissAfterDaysShown` ?

Suggestions:
This snippet is hard to read due to the nested `if let` / `guard`. I think we can improve the code by:
1. Moving this logic in its own function `static func mapToDisplayConditions(message: RemoteMessageResponse.JsonRemoteMessage) -> DisplayConditions?` like we do for mapToContent, mapToSurfaces etc.
2. When working on What’s New message I added a generic `MappingValidator` to map json fields according to some rules. What it does is also to report the keypath that fails to parse out of the box. If we want to continue re-use that the new function could become:

```
static func mapToDisplayConditions(message: RemoteMessageResponse.JsonRemoteMessage) throws -> DisplayConditions? {
    try message.displayConditions.flatMap { conditions in
        let validator = MappingValidator(root: conditions)
        let trigger = try validator.mapEnumIfPresent(\.trigger, to: MessageTrigger.self)
        return DisplayConditions(trigger: trigger, dismissAfterDaysShown: conditions.dismissAfterDaysShown)
    }
}

// At call site
let displayConditions: DisplayConditions?
    do {
        displayConditions = try mapToDisplayConditions(message: message)
    } catch {
        Logger.remoteMessaging.debug("Unknown trigger for message \(message.id, privacy: .public): trigger: \(error.localizedDescription, privacy: .public)")
         return
}
```

That requires enhancing the `MappingValidator` with
```
func mapEnumIfPresent<T, E: RawRepresentable>(_ keyPath: KeyPath<Root, T?>, to enumType: E.Type) throws(MappingError) -> E? where E.RawValue == T {
    guard let value = root[keyPath: keyPath] else { return nil }
    guard let result = E(rawValue: value) else { throw .invalidValue(keyPath) }
    return result
}
```
Using the `MappingValidator` is just a suggestion but I believe moving the code in a function and try clean up the nested if/let will make the code neater and easy to reason about.

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
```diff
@@ -444,6 +461,14 @@ extension RemoteMessagingStore {
         }
     }
 
+    private func dismissExpiredMessage(withID id: String) {
```
alessandroboron, line 464: Do we need to wait for this task to complete since it’s fire and forget? That way we could also re-use `dismissRemoteMessage(withID id: String) async `:

```
private func dismissExpiredMessage(withID id: String) {
   Task {
      await dismissRemoteMessage(withID: id)
   }
}
```

Only issue may be tests if you’re waiting to read the results as this use unstructured concurrency

alessandroboron (commented):
@SabrinaTardio I left some comments as suggestions and some questions, nothing major.

One thing I noticed: when we switch the toggle from 'Search' to 'Duck.ai', the RMF message disappears momentarily and then reappears when 'Search' is toggled back. I’ll attach a video on the Asana task

iOS/DuckDuckGo/NewTabPageViewController.swift
SabrinaTardio: 1) associatedTab.openedAfterIdle wouldn't work for the focussed NTP, it creates its own temporary Tab that's disconnected from the idle-return flow. The focussed NTP gets its escape hatch state via SuggestionTrayViewController.setEscapeHatch() which sets pendingEscapeHatchModel, so hasEscapeHatch: pendingEscapeHatchModel != nil is the correct signal for both NTPs.

2) Simplified isOpenedAfterIdle from () -> Bool to a plain Bool since the NTP is destroyed and recreated on every state change, making the value fixed for its lifetime.

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
SabrinaTardio: 1) Yes, we discard the entire message if the trigger can't be parsed. If we don't understand the trigger, we can't know when the message should appear, so discarding is the safe choice for forward compatibility with future trigger types.

2) Added clamping, dismissAfterDaysShown values <= 0 are now clamped to 1.

3) Adopted your suggestion: extracted a mapToDisplayConditions function using MappingValidator, and added mapEnumIfPresent to MappingValidator with tests. Much cleaner now.

SabrinaTardio:
> @SabrinaTardio I left some comments as suggestions and some questions, nothing major.
> 
> One thing I noticed: when we switch the toggle from 'Search' to 'Duck.ai', the RMF message disappears momentarily and then reappears when 'Search' is toggled back. I’ll attach a video on the Asana task

Yeah that has always been the case, the duck.ai page is completely separate (it’s not a NTP) so it won’t show the RMF.

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
SabrinaTardio: Yeah I tried both approaches and kept being in two minds about it. performAndWait gives synchronous consistency but introduces nested performAndWait on the write context from within the read-only context's performAndWait. The Task approach avoids that at the cost of the dismiss happening asynchronously. Since the fetch already skips expired messages via continue (so the caller never sees them regardless), I went with the Task approach to maintain the async-only pattern on the write context.

iOS/DuckDuckGoTests/NewTabPageMessagesModelTests.swift
```diff
@@ -290,15 +290,47 @@ final class NewTabPageMessagesModelTests: XCTestCase {
...
-    private func createSUT() -> NewTabPageMessagesModel {
+    // MARK: - openedAfterIdle
+
+    func testWhenOpenedAfterIdleIsTrueThenRefreshPassesOpenedAfterIdleTrue() {
+        let sut = createSUT(isOpenedAfterIdle: true)
+
+        sut.load()
+
+        XCTAssertTrue(messagesConfiguration.didRefresh)
+        XCTAssertEqual(messagesConfiguration.lastRefreshOpenedAfterIdle, true)
```
alessandroboron: nip: `XCTAssertTrue(messagesConfiguration.lastRefreshOpenedAfterIdle)`

iOS/DuckDuckGoTests/NewTabPageMessagesModelTests.swift
```diff
@@ -290,15 +290,47 @@ final class NewTabPageMessagesModelTests: XCTestCase {
...
+    }
+
+    func testWhenOpenedAfterIdleIsFalseThenRefreshPassesOpenedAfterIdleFalse() {
+        let sut = createSUT(isOpenedAfterIdle: false)
+
+        sut.load()
+
+        XCTAssertTrue(messagesConfiguration.didRefresh)
+        XCTAssertEqual(messagesConfiguration.lastRefreshOpenedAfterIdle, false)
```
alessandroboron: nip: `XCTAssertFalse(messagesConfiguration.lastRefreshOpenedAfterIdle)`

iOS/DuckDuckGoTests/NewTabPageMessagesModelTests.swift
```diff
@@ -290,15 +290,47 @@ final class NewTabPageMessagesModelTests: XCTestCase {
...
+    }
+
+    func testWhenDefaultOpenedAfterIdleThenRefreshPassesFalse() {
+        let sut = createSUT()
+
+        sut.load()
+
+        XCTAssertTrue(messagesConfiguration.didRefresh)
+        XCTAssertEqual(messagesConfiguration.lastRefreshOpenedAfterIdle, false)
```
alessandroboron: nip: `XCTAssertFalse(messagesConfiguration.lastRefreshOpenedAfterIdle)`

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
```diff
@@ -251,13 +251,27 @@ extension RemoteMessagingStore {
...
+                   let firstShown = remoteMessageManagedObject.firstShownDate,
+                   let daysSinceFirstShown = Calendar.current.dateComponents([.day], from: firstShown, to: Date()).day,
+                   daysSinceFirstShown >= dismissAfterDays {
+                    self.dismissExpiredMessage(withID: id)
+                    continue
+                }
+
+                if let messageTrigger = remoteMessage.displayConditions?.trigger, messageTrigger != trigger {
+                    continue
+                }
```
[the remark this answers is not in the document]
alessandroboron: @SabrinaTardio this is outlining two possible scenarios:
1. if we disable RMF (very low chance to happen) if the only message scheduled is one with a display condition because we don’t pass the parameter the function returns and deleteScheduledMessages() does not get called. Should we remove ` guard fetchScheduledRemoteMessage(surfaces: RemoteMessageSurfaceType.allCases) != nil else {
      return
   }` from 
 
```
remoteMessagingAvailabilityProvider.isRemoteMessagingAvailablePublisher
    .filter { !$0 }
    .removeDuplicates()
    .sink { [weak self] _ in
       guard fetchScheduledRemoteMessage(surfaces: RemoteMessageSurfaceType.allCases) != nil else {
          return
       }
      Task {
          await self.deleteScheduledMessages()
      }
}
```
2. This may be an issue cause if the message has remote images but we don’t pass the trigger and the message has display condition, the image won’t prefetch?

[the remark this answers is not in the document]
alessandroboron: we synced on MM and decided to address this in a follow up task given the project timeline. Issue 2 would happen if the idle message would use a remote image but at this stage we use a predefined image so no issues.

alessandroboron (approved):
LGTM! 💪

PR #4745, opened 2026-05-06
macOS: Add YouTube ad-blocking Next Steps card

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214572114475566

### Description

Adds a new YouTube Ad Blocking card to the Next Steps cards on macOS, wired into both the legacy "Continue Set Up" widget and the new single-card NTP surface. Also stops showing the Duck Player Next Steps card.

Visibility:
- The YouTube Ad Blocking card shows only when ad-blocking is fully enabled (`AdBlockingAvailabilityProviding.isEnabled` — feature flag on **and** the user has opted in). This is the setup for the public rollout where we will remotely enable the feature for users.
- The Duck Player Next Steps card is removed from the candidate lists on both surfaces. `shouldShowCard(.duckplayer)` also returns `false` as a safety net for any user with a stale `.duckplayer` entry in `orderedCardIDs` from a prior version.

Status:
- Click action for the YouTube Ad Blocking card is currently a placeholder (mirrors the Duck Player action) — final CTA destination is still TBD.

### Testing Steps

1. Open Settings → Ad Blocking and **disable** YouTube Ad Blocking by 
3. Open a New Tab Page and click through the Next Steps cards — neither the **Duck Player** card nor the **YouTube Ad Blocking** card should appear.
4. Open Settings → Ad Blocking and **enable** YouTube Ad Blocking.
5. Reset the Next Steps cards (e.g. via the debug menu) and click through them on the NTP again — the **YouTube Ad Blocking** card should appear and the **Duck Player** card should not.

### Impact and Risks

**Low.**

The new card is gated on the `adBlockingExtension` feature flag (currently off in public), so it cannot render in production until the flag is rolled out.

The Duck Player Next Steps card is removed from the candidate lists on both Next Steps surfaces — a deliberate replacement. Existing users who hadn't dismissed it will no longer see it. The visibility logic also returns `false` defensively, protecting users on advanced ordering from any stale `.duckplayer` entry in their persisted card list.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches macOS New Tab Page/Continue Set Up card ordering, visibility, and persistence (new dismissal key), which could affect what onboarding cards users see; behavior is largely gated behind `AdBlockingAvailabilityProviding.isEnabled`.
> 
> **Overview**
> Adds a new `youtubeAdBlocking` Next Steps card across both the legacy `ContinueSetUpModel` matrix and the newer single-card provider, including mapping in the shared `NewTabPageDataModel.CardID` and a placeholder click action that opens a YouTube URL.
> 
> Updates visibility/dismissal logic to gate the new card on `AdBlockingAvailabilityProviding.isEnabled`, persists its dismissal via a new `home.page.show.youtube.ad.blocking` flag, and defensively stops showing the Duck Player card (removed from default/advanced ordering and `shouldShowCard(.duckplayer)` returns `false`).
> 
> Bumps `content-scope-scripts` to `14.6.0` in SwiftPM resolution files and wires `adBlockingAvailability` through constructors/tests (including a new `MockAdBlockingAvailability`) and adjusts related UI/unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88b2ed72db59741ac8335d511d9792e85dbd3984. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


miasma13:
This is partial PR into `michal/ad-block-ux` feature branch so this will also include C-S-S version update when the [this PR](https://github.com/duckduckgo/content-scope-scripts/pull/2685) is merged.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
```diff
@@ -316,7 +322,10 @@ private extension NewTabPageNextStepsSingleCardProvider {
         case .addAppToDockMac:
             return !isAppStoreBuild && !dockCustomizer.isAddedToDock
         case .duckplayer:
-            return duckPlayerPreferences.duckPlayerModeBool == nil && !duckPlayerPreferences.youtubeOverlayAnyButtonPressed
+            // `.youtubeAdBlocking` takes precedence — hide Duck Player when YT ad-blocking is eligible.
+            return duckPlayerPreferences.duckPlayerModeBool == nil
+                && !duckPlayerPreferences.youtubeOverlayAnyButtonPressed
+                && !adBlockingAvailability.isEnabled
```
rachelmcr: Should the Duck Player card be visible if the YT card has been dismissed and ad-blocking is not enabled?

Currently, dismissing the YT card and then disabling ad-blocking makes the Duck Player card appear in the list. But it seems like they should be mutually exclusive (users should only ever see one or the other).

We could check `!isCardPermanentlyDismissed(.youtubeAdBlocking)` here so the user won't see this card if the other was manually dismissed.

(And any changes here should also be reflected in `HomePageContinueSetUpModel`.)

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
```diff
@@ -325,6 +334,8 @@ private extension NewTabPageNextStepsSingleCardProvider {
             return !appearancePreferences.didChangeAnyNewTabPageCustomizationSetting
         case .sync:
             return syncService?.featureFlags.contains(.all) == true && syncService?.authState == .inactive
+        case .youtubeAdBlocking:
+            return adBlockingAvailability.isEnabled
```
rachelmcr, line 333: Similar to the above, do we want the YT card to show if the user has already interacted with the Duck Player card (either dismissed it or clicked its call to action to enable Duck Player) and then enables ad blocking? It seems like users should only ever interact with one or the other.

Also, are there any states in which the card should hide itself, other than the ad blocking setting (which effectively swaps the YT and Duck Player cards right now)? For example, the Duck Player card is hidden if the user chooses a Duck Player setting, which is part of the card's call to action.

rachelmcr (commented):
This is working as described in the testing steps, but I'm unclear about the expectation around what users should see after they interact with either the YT or Duck Player cards:
- If a user dismisses the YT card, and disables ad blocking, should the Duck Player card appear?
- If a user dismisses the Duck Player card (either before this new card is released or after), and enables ad blocking, should the YT card appear?
- What if a user interacts with the call to action on either card, and then becomes eligible to see the other card?

I'd expect that users will only ever interact with one card or the other: once they either dismiss a card or click on its call to action, they won't see the other one. Does that sound right? In that case, we need to add checks for those states (see review comments).

> Click action is currently a placeholder (mirrors the Duck Player action) — final CTA destination is still TBD.

In addition to the destination itself, it's worth considering if the destination will include some action that should auto-dismiss the Next Steps card. (Will the user always have to manually dismiss the card when they go back to the New Tab Page? Or will the `shouldShowCard(_:)` check include a state that hides the card?)

And a small tip for testing: You can use the build branch from https://github.com/duckduckgo/content-scope-scripts/pull/2685 (`pr-releases/michal/next-step-yt-ad-block`) instead of building the branch locally.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
miasma13: This is a good point, yes you are correct when the `youtubeAdBlocking` card is eligible (dismissed or not) then the `duckplayer` should not be ever presented (you can understand it as the new card overrides it)

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
miasma13: Since we want the logic to effectively use one of the pair of those cards depending on the YouTube Ad blocking rollout state we need to create an inverse condition as well: if `duckplayer` card was already dismissed or suppressed by its condition then we should not suddenly reappear with the `youtubeAdBlocking` card.

Let me review that logic again and probably create some helpers to better capture the conditions for their presentation.

miasma13:
@rachelmcr This is ready for re-review. The logic is now simplified by omitting Duck Player card and instead replacing it by the new YouTube Ad Blocking one.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
```diff
@@ -316,7 +320,7 @@ private extension NewTabPageNextStepsSingleCardProvider {
         case .addAppToDockMac:
             return !isAppStoreBuild && !dockCustomizer.isAddedToDock
         case .duckplayer:
-            return duckPlayerPreferences.duckPlayerModeBool == nil && !duckPlayerPreferences.youtubeOverlayAnyButtonPressed
+            return false
```
rachelmcr, line 323: Replacing `duckplayer` with `youtubeAdBlocking` entirely/unconditionally means that if the `adBlockingExtension` is not enabled, there will be no video-related Next Steps card shown. (This also applies to `HomePage.Models.availableFeatures`.)

What about including both cards in the set of possible cards but making this check include the feature flag?


```suggestion
            return !adBlockingAvailability.isFeatureAvailable && duckPlayerPreferences.duckPlayerModeBool == nil && !duckPlayerPreferences.youtubeOverlayAnyButtonPressed
```

rachelmcr (commented):
The code changes are reasonable but I want to check the new expectations one more time:

1. Should the Duck Player card be removed even when the ad blocking feature flag is disabled? (See code comment below)
2. Since the YouTube Ad Blocking card shows only when ad-blocking is fully enabled, will users be automatically opted in to ad blocking when it rolls out? If not, what is the expectation for new user discovery of the ad blocking feature? (Next Steps cards are only shown to users for 9 days after install, or until they have dismissed all the available cards. If ad blocking is off by default then new users won't see this card unless they enable ad blocking before interacting with all the Next Steps cards.)

That criteria for showing the new card still seems a little odd to me, but if that's the intended behavior it's working. I do think we should handle the scenario where the feature flag is off, though.

macOS/DuckDuckGo/NewTabPage/Features/NextSteps/NewTabPageNextStepsSingleCardProvider.swift
miasma13: Thanks for the questions @rachelmcr. Yes, the Duck Player next step card removal at this stage is deliberate. This follows the removal of Duck Player onboarding step. This is deliberate design and product decision to downplay the Duck Player promotion in favour of YouTube Ad Blocking.

Currently Ad Blocking is during soft launch period (feature is available to all eligible users - web extension support required by os version) but is not enabled by default - so it is left for self discovery in settings.

Next step is planned to gradually roll out the feature as enabled by default for new users in around 3 weeks. We are aware that there might be short period of time that some users might not see the next step card for either of the features, but this is expected.

I hope the above explanation is clear, but if you want to chat about more details we can jump on a call.

rachelmcr (approved):
Thanks for addressing my questions @miasma13. Given the product decisions behind the changes, I have no other concerns. 👍

PR #4764, opened 2026-05-07
Fix crash when back/forward item has a nil URL during commit

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201037661562251/task/1214602405943382?focus=true

### Description

Tentative fix for a rare crash that surfaces when WebKit delivers a page title change during a navigation commit.

The change is purely a defensive nil-guard on a path that previously trapped on `nil`, so functional behavior is unchanged in the happy case.

### Working theory
We noticed in WebKit sources that back/forward item can briefly expose a `nil` URL — despite advertising the property as non-optional (I've checked the WebKit sources and there are a few instances where this happens).

This PR routes the property through an optional path so the transient `nil` is tolerated instead of trapping; if the theory holds, the crash signature should drop.

### Testing Steps

We have no reproduction steps, so we just need to make sure this doesn't introduce issues.

Smoke-test browsing: open and close tabs, navigate forward/back, and confirm tab titles still update correctly.

To verify the new path actually runs, set a breakpoint inside the `if let itemURL = wkBackForwardListItem.safeURL` block in `Tab.updateTitle()` and step through normal browsing — you should hit the block regularly with a non-nil `itemURL`, confirming the KVC-based read returns the expected URL on the happy path.

### Impact and Risks

Low. The change only adds a nil-guard on an existing path; in the no-nil case the behavior is identical.

#### What could go wrong?

If the back/forward URL is nil in some non-pathological case the contract didn't anticipate, the tab title would simply not be persisted onto that back/forward item for that update — minor UX at worst, no crash.

### Quality Considerations

Privacy/security: none. Performance: one extra KVC read per title-change observation, negligible. Monitoring: existing crash reporting will surface a drop in this signature; no new instrumentation added.

### Notes to Reviewer

The justification for reading the URL via KVC instead of trusting the imported non-optional type is documented inline. WebKit's URL bridge explicitly returns nil for invalid or null underlying URLs despite the header's nonnull annotation, so the trap is on Apple's side of the contract — not ours to fix upstream.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Low risk defensive change: replaces direct reads of `WKBackForwardListItem.url` with an optional `safeURL` to avoid a rare nil-bridging crash; behavior is unchanged when the URL is valid.
> 
> **Overview**
> Prevents a rare WebKit nil-bridge crash by adding `WKBackForwardListItem.safeURL` (an optional selector-based URL read) and routing history/title logic through it.
> 
> `HistoryItemIdentity` and macOS `Tab.updateTitle()` now guard on `safeURL` before comparing/persisting titles, so transiently invalid back/forward items no longer trap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01b279e752b292f7cdcca1c6d395582eec11ff17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/Tab/Model/Tab.swift
+++ b/macOS/DuckDuckGo/Tab/Model/Tab.swift
@@ -689,7 +689,8 @@ protocol TabDelegate: ContentOverlayUserScriptDelegate {
         self.title = webView.title?.trimmingWhitespace()
 
         if let wkBackForwardListItem = webView.backForwardList.currentItem,
-           content.urlForWebView == wkBackForwardListItem.url,
+           let itemURL = wkBackForwardListItem.safeURL,
+           content.urlForWebView == itemURL,
            !webView.isLoading,
            title?.isEmpty == false {
             wkBackForwardListItem.tabTitle = title
--- a/SharedPackages/BrowserServicesKit/Sources/Navigation/Extensions/WKBackForwardListItemExtension.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Navigation/Extensions/WKBackForwardListItemExtension.swift
@@ -24,12 +24,12 @@ public extension WKBackForwardListItem {
     /// Safe Optional `url: URL` getter:
     /// `WKBackForwardListItem.url` is imported as non-optional, but WebKit's
     /// implementation can return nil for null or invalid underlying URLs,
-    /// trapping Swift's unconditional `NSURL` → `URL` bridge. Reading via KVC
-    /// keeps the value optional so the conditional `as? URL` cast safely
-    /// returns nil instead of crashing.
+    /// trapping Swift's unconditional `NSURL` → `URL` bridge. Reading via the
+    /// getter selector keeps the value optional so the conditional `as? URL`
+    /// cast safely returns nil instead of crashing.
     /// See: https://app.asana.com/1/137249556945/project/1201037661562251/task/1214602405943382
     var safeURL: URL? {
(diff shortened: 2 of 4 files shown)


macOS/DuckDuckGo/Tab/Model/Tab.swift
```diff
@@ -689,7 +689,8 @@ protocol TabDelegate: ContentOverlayUserScriptDelegate {
         self.title = webView.title?.trimmingWhitespace()
 
         if let wkBackForwardListItem = webView.backForwardList.currentItem,
-           content.urlForWebView == wkBackForwardListItem.url,
+           let itemURL = wkBackForwardListItem.safeURL,
```
diegoreymendez, line 692: If the url is nil, this'll just bail out.

mallexxx (approved):
The change raises no concerns,
However, would it make sense to use the same approach as the one used in the similar `WKNavigationAction.safeSourceFrame` (i.e. `performSelector` instead of KVC) 
It may also make sense to update the`backForwardListItem.url` accessor in `HistoryItemIdentity` (`NavigationType.swift`)

diegoreymendez:
Thanks @mallexxx , good feedback.

I didn't change to using a selector as I wanted the benefit of having a runtime check against the original property.

That said I did extend to apply the same fix for `HistoryItemIdentity` (which meant moving to BSK, but this is fine).

Thanks!

diegoreymendez:
Sorry I misunderstood your selector suggestion, so I've swapped it now.

PR #4774, opened 2026-05-07
macOS: Add native tab and file attachments to the duck.ai omnibar

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1213072873557311?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1214704361806993?focus=true

### Description

Adds tab and PDF/file attachments to the duck.ai **omnibar** mode on macOS, alongside the image attachments that already existed. The attach button opens a native menu with **Add Image or File** and **Add Page Content** (a submenu listing open tabs with multi-select that keeps the menu open while toggling). Selected tabs, images, and files render in a unified horizontally-scrollable carousel, in **insertion order**, regardless of kind.

Two new feature flags gate the rollout independently:
- `aiChatOmnibarAttachMoreTabs` — gates the **Add Page Content** submenu.
- The previous flag `aiChatAttachMoreTabs` was renamed to `aiChatSidebarAttachMoreTabs`

The legacy omnibar-only `AIChatImageAttachmentsContainerView` was retired in a previous milestone — the new carousel is the single attachments surface for image + tab + file kinds, used regardless of feature flag.

#### Setup
1. Open the app. Confirm you're an internal user (Debug → Set as Internal User).
2. Open Debug → Feature Flags. Confirm `aiChatPageContext`, `aiChatSidebarAttachMoreTabs`, `aiChatOmnibarAttachMoreTabs` are listed.
3. Set  https://euw-serp-dev-testing3.duck.ai/ as custom AI chat URL
4. Open 3 real URL tabs (Wikipedia, GitHub, news article).

---

##### A1. Attach menu basics
1. Activate the duck.ai omnibar.
2. Click attach → popover with two items: **Add Image or File** (folder glyph), **Add Page Content** (page glyph), separator between them.
3. Hover **Add Page Content** → submenu with header **Recent Tabs** (or **No open tabs**), one row per URL tab (favicon, title, leading checkmark slot).
5. Click a tab row → black checkmark appears, **menu stays open**, card appears in the carousel.
6. Click another tab row → second checkmark, menu still open, second card joins the carousel.
7. Click a checked tab again → checkmark gone, card removed, menu still open.
8. Click outside to dismiss. Reopen the menu → checkmarks reflect previous state.

##### A2. Add Image or File — images
1. Click **Add Image or File** → file picker opens.
2. Pick a JPEG → image thumbnail card appears.
3. Add 2 more → **Add Image or File** disables; inline "You can only add 3 images at a time" hint.

##### A3. Add Image or File — PDFs
1. With 1 image attached, pick a PDF → file card (square, with PDF pill) appears next to the image.
2. Hover the file card → filename in tooltip.
3. Add 2 more PDFs (3 total) → picker rejects further files; inline "You can only add 3 files per conversation" hint.

##### A4. Insertion order
1. Reset, then add: image → tab → PDF → tab → image, in that order.
2. **Verify** the carousel shows the cards left-to-right in **exactly that order**, regardless of kind.
3. Remove the third card (PDF) → the rest preserve order: image, tab, tab, image.

##### A5. Carousel scrolling and removal
1. Add enough cards to overflow horizontally.
2. **Verify** trackpad / mouse-wheel horizontal scroll works; no scrollbar visible.
3. Click × on any card → card disappears immediately; reopening the menu shows the checkmark/cap counter updated.

##### A6. Per-tab persistence
1. Attach two tabs + one PDF in browser tab A.
2. Switch to tab B → omnibar carousel is empty.
3. Attach one image in tab B.
4. Switch back to tab A → original two tabs + PDF restored. Back to tab B → only the image is there.

##### A7. Submit threading — verify the JSON shape

Open Console.app, filter on `com.duckduckgo.macos.browser`. (The temp verification logger was removed before merge, so instead inspect the prompt via Web Inspector on the duck.ai page that opens — under the JS handler for `getAIChatNativePrompt`.)

Run these submit scenarios and verify the JSON:

- **No tabs attached, only text** → top-level `pageContext` key is **absent** from the JSON. `query.attachedTabIds` does not exist anywhere.
- **One tab attached, NOT the active tab** → top-level `"pageContext": [ { ... } ]` is a JSON **array** with one element. The element has `"tabId": "<uuid>"` (non-null), plus `title`, `url`, `content`, `favicon`, `truncated`, `fullContentLength`.
- **One tab attached, IS the active tab** → top-level `pageContext` is an array with one element. The element has **no `tabId`** (key absent). Per the discriminator, this is the "current page" form.
- **Three tabs attached, mix of active + others** → array of 3 elements, in carousel pick order. The entry matching the active tab has no `tabId`; the other two carry `tabId`.
- **Combined with image + PDF** → JSON has `query.images: [...]`, `query.files: [...]`, AND top-level `pageContext: [...]` simultaneously. Carousel cleared after submit.

Pixel verification (PixelKit debug log): `aichat_addressbar_submit_with_image`, `aichat_addressbar_submit_with_files`, `aichat_addressbar_submit_with_tabs` fire when those attachment kinds are present.

##### A8. Sidebar regression check
1. With `aiChatSidebarAttachMoreTabs` ON, open the **duck.ai sidebar**.
2. Confirm its existing in-web tab picker still works — `getAIChatOpenTabs` / `getAIChatTabContent` are called via the JS-bridge, and attached tabs show up in the chat. The JS-bridge response now stamps `tabId` on each `pageContext`, which the duck.ai web app uses for discrimination but shouldn't change visible behavior.

---

##### B1. Image attachments still work
1. Restart the app with both flags off.
2. Activate omnibar → click attach → **system file picker opens directly** (no menu).
3. Pick JPEG/PNG/WebP → thumbnail card in the carousel, visually identical to pre-PR behavior.
4. Click × → card removed. Add 3 images → picker stops accepting.
5. Submit → duck.ai tab opens with the image attached, carousel cleared.
6. Per-tab persistence: switch tabs → empty; switch back → image card restored.

##### B2. No tab or PDF surfaces
1. **Verify** no **Add Page Content** option anywhere; attach button goes straight to file picker.
2. **Verify** no PDF support — file picker advertises image types only.
3. **Verify** submit JSON has no `pageContext` key.

##### B3. Mixed flag states
1. `aiChatPageContext` ON + `aiChatOmnibarAttachMoreTabs` OFF → still no tab menu.
2. `aiChatPageContext` OFF + `aiChatOmnibarAttachMoreTabs` ON → still no tab menu.
3. Sidebar with `aiChatSidebarAttachMoreTabs` OFF → its in-web picker is gated as before.

### Impact and Risks

**Medium.** This is the headline attachment surface for duck.ai in the omnibar, and it now eagerly fetches per-tab content at submit time.

#### What could go wrong?

- **Regression on image attachments.** The legacy `AIChatImageAttachmentsContainerView` was deleted; everything now flows through the new carousel even with all new flags off. Covered by Test Plan B.
- **Eager per-tab content fetch blocks submit.** Up to 5 s per tab (parallel), so 3 picked tabs add ~5 s worst case to submit time. Tabs that fail extraction are silently dropped — same behavior the sidebar's JS-bridge has today.
- **Prompt JSON grows.** Page content is already truncated upstream (`truncated` flag indicates), but the wire payload is now `~50 KB × N` bigger than the M1 `attachedTabIds` shape.
- **Web app receives an unknown field shape.** The schema is union-typed (`pageContext` may now be an object OR an array). The duck.ai frontend team owns the consumer change for the array form; the field is `encodeIfPresent` so prompts without tabs are wire-identical to pre-PR. Coordinated via the linked tech design.

### Quality Considerations

- **Edge cases covered:** insertion order across kinds, per-tab persistence across tab switches, submit-clears-attachments, model-support filtering, no open tabs ("No open tabs" submenu), menu doesn't shift when carousel grows mid-toggle, active-tab `tabId` stripping per discriminator.
- **Pixels:** `aichat_addressbar_submit_with_tabs`, `aichat_addressbar_file_attached`, `aichat_addressbar_submit_with_files` defined in `aichat_addressbar_pixels.json5`.
- **Privacy:** no new content leaves the device beyond what the existing `images` and `files` fields already carry. Tab attachments embed `AIChatPageContextData` (title, url, content, favicon, truncated, fullContentLength, tabId) — the same shape the sidebar already sends for its current page and via `getAIChatTabContent`.
- **Shared extractor:** `AIChatUserScriptHandler.extractPageContext(from: Tab)` is now the single source of truth for per-tab page-context extraction. The JS-bridge `getAIChatTabContent` consumes it (always stamping `tabId`), and the omnibar's submit path consumes it (stamping per entry, stripping for the active tab).

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the omnibar submit payload/schema (new `pageContext` union, file uploads) and refactors attachment state/rendering to a new unified carousel, which could regress existing image-attachment flows or submission gating.
> 
> **Overview**
> Enables **tab (page content) and file (PDF/etc.) attachments** in the macOS duck.ai omnibar, alongside images, rendered together in a single horizontally scrollable **insertion-ordered** carousel with per-item removal.
> 
> Updates the prompt schema to support multi-context omnibar submits by adding `AIChatPageContextPayload` (object-or-array at `pageContext`) and adding a `tabId` discriminator on `AIChatPageContextData`; macOS omnibar submit now extracts attached tabs’ page context in parallel, includes `files` (base64) and `pageContext`, and clears all attachment kinds after submit.
> 
> Refactors omnibar attachment state to be **per-tab persisted** via `AddressBarSharedTextState.aiChatPanelAttachments` (plus derived image/tab/file lists), replaces the old image-only container, adds new attachment UI components (tab/file cards, unified carousel, keep-open tab picker menu), introduces new pixels for tab/file attach & submit, and splits the privacy flag into `sidebarAttachMoreTabs` vs `omnibarAttachMoreTabs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e14d8e546e17fe8c3e655703a14d4f3d9cbe2307. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
```diff
@@ -423,8 +423,11 @@ public enum AIChatSubfeature: String, Equatable, PrivacySubfeature {
     /// Enables support for adding multiple page contexts to a single chat session
     case multiplePageContexts
 
-    /// Enables attaching content from multiple open tabs to Duck.ai chat
-    case attachMoreTabs
+    /// Enables attaching content from multiple open tabs to the Duck.ai sidebar chat.
+    case sidebarAttachMoreTabs
```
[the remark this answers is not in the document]
tomasstrba: This is intended

SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
```diff
@@ -423,8 +423,11 @@ public enum AIChatSubfeature: String, Equatable, PrivacySubfeature {
     /// Enables support for adding multiple page contexts to a single chat session
     case multiplePageContexts
 
-    /// Enables attaching content from multiple open tabs to Duck.ai chat
-    case attachMoreTabs
+    /// Enables attaching content from multiple open tabs to the Duck.ai sidebar chat.
```
jaceklyp, line 426: **Major:** This rename silently abandons any existing rollout state keyed on `"attachMoreTabs"`.

`PrivacySubfeature` resolves to its `rawValue` string when querying the privacy-config payload, so renaming the case **renames the wire key**. Post-update clients will ask for `"sidebarAttachMoreTabs"` / `"omnibarAttachMoreTabs"`; on a server that hasn't been updated those keys don't exist (= disabled). For users currently rolled into `attachMoreTabs`, this looks like a silent regression on update — the sidebar tab picker disappears until the new key rolls out independently.

Verified there's no fallback string mapping in `BrowserServicesKit`, `FeatureFlag`, or `AIChatSubfeature` extensions.

**Either** keep the original key alive as an alias so the existing rollout flows to the sidebar variant:

```swift
extension AIChatSubfeature {
    static let attachMoreTabs: AIChatSubfeature = .sidebarAttachMoreTabs
}
```

**or** coordinate a server-side migration with the privacy-config owner before this lands, and call it out in the PR description so it's not invisible at review time.

macOS/PixelDefinitions/pixels/definitions/aichat_addressbar_pixels.json5
```diff
@@ -114,6 +114,43 @@
             }
         ]
     },
+    "m_mac_aichat_addressbar_submit_with_tabs": {
```
jaceklyp, line 118: **Major:** Two of the three new pixels are missing the `channel` parameter — live validation on canary/dev builds will reject them.

PixelKit auto-injects `channel` on canary builds (per the fix in commit `c8c85a34ad`). Without `"channel"` declared in `parameters`, validation fails with:
> must NOT have additional properties. Found extra property 'channel'

Sibling `m_mac_aichat_addressbar_file_attached` (this PR, line 137) correctly declares it. The two missing it are this one (`submit_with_tabs`, parameters at line 122) and `submit_with_files` (parameters at line 143).

```json5
"parameters": [
    "appVersion",
    "pixelSource",
    "channel",
    { "key": "tabCount", "type": "integer", ... }
]
```

**Drive-by:** the pre-existing `m_mac_aichat_addressbar_submit_with_image` (line 107) has the same omission — easy to fix in the same pass while you're here.

jaceklyp (commented):
Bundled review for the omnibar attachments PR. **3 majors** (privacy-config rollout migration, missing `channel` on two pixels, submit-path post-await snapshot race), **6 minors**, **4 nits**, plus a single comment on the missing extractor-mock test cluster. Nothing on its own is a blocker; combined, the rollout / race / test items are worth a second pass before merge.

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
```diff
@@ -674,29 +862,152 @@ final class AIChatOmnibarController {
...
+            // JS-bridge (`getAIChatTabContent`) uses, so the page-content + favicon enrichment is
+            // byte-identical across both flows. Each entry carries `tabId` so the duck.ai web app
+            // sees the discriminator (presence = tab-picker context), except for the entry whose
+            // tab UUID matches the active tab: that one becomes the no-`tabId` form, i.e. "the
+            // page you're chatting about", per the tech design.
+            //
+            // When there are no attached tabs we skip the `await` entirely — keeping the
+            // submit Task linear in the common case (and matching the pre-M8 number of
+            // suspension points, which a number of tests rely on).
+            let tabAttachments = self.activeTabAttachments
```
jaceklyp: **Major:** The whole payload is read **after** `await waitForAttachmentsReady`. A tab switch during that await corrupts the prompt — the attachments shipped will reflect the new active tab, not the tab whose prompt the user submitted.

`sharedTextState` is rebound on tab change, so every post-await read (`self.activeImageAttachments`, `self.activeTabAttachments`, `self.activeFileAttachments`, and `self.tabCollectionViewModel.selectedTabViewModel?.tab.uuid` at line 886) reflects whatever tab is active when the await resumes — not when submit was clicked. The discriminator strip is also computed against the post-await active tab, so an entry the user picked as picker context can have its `tabId` stripped, or vice versa.

`waitForAttachmentsReady` can take seconds for large images, so the race is exploitable in practice (user clicks submit, switches tabs, the prompt for tab A goes out with tab B's payload). Pre-PR this race existed for images alone with limited impact; tab + file + the discriminator widen the corruption surface a lot.

**Proposed:** snapshot before the await.

```swift
let snapshotTabAttachments   = self.activeTabAttachments
let snapshotImageAttachments = canSendImages ? self.activeImageAttachments : []
let snapshotFileAttachments  = self.selectedModelSupportsFileUpload ? self.activeFileAttachments : []
let snapshotActiveTabUUID    = self.tabCollectionViewModel.selectedTabViewModel?.tab.uuid

Task { @MainActor [snapshotTabAttachments, snapshotImageAttachments,
                   snapshotFileAttachments, snapshotActiveTabUUID] in
    await waitForAttachmentsReady?()
    let images = Self.nativePromptImages(from: snapshotImageAttachments, supportedFormats: …)
    let pageContextPayload = snapshotTabAttachments.isEmpty ? nil
        : await self.extractPageContextsForOmnibarSubmit(
            tabAttachments: snapshotTabAttachments,
            activeTabUUID: snapshotActiveTabUUID)
    …
}
```

Bonus: snapshot-before-await also makes the submit deterministic for tests — they currently lean on `for _ in 0..<5 { await Task.yield() }` to hope the right reads land. See the test-coverage comment for the cluster of tests this unlocks.

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatScriptUserValues.swift
```diff
@@ -171,6 +171,40 @@ public struct AIChatNativeConfigValues: Codable {
...
+/// tab-picker contexts (non-nil) from the current sidebar page (nil).
+public enum AIChatPageContextPayload: Codable, Equatable {
+    case single(AIChatPageContextData)
+    case multiple([AIChatPageContextData])
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        // Decode array first — a JSON object would fail array decoding cleanly, and we want
+        // to disambiguate without relying on Swift's behavior of partially decoding objects.
+        if let array = try? container.decode([AIChatPageContextData].self) {
```
jaceklyp, line 189: **Minor:** Using `try?` on the array-decode branch fires the fallback on the wrong condition — a structurally-correct array with **one malformed element** swallows the array-element error and falls into the single-decode path (which then also throws, with a single-shape error that misleads future investigators).

Path: `[ {ok}, {malformed} ]` → `try? container.decode([AIChatPageContextData].self)` throws on the malformed element → returns `nil` → falls into `try container.decode(AIChatPageContextData.self)` → throws a single-decode error. The caller sees "couldn't decode object" when the actual problem was element 2 of the array.

Doesn't matter today (only macOS native emits arrays), but the contract is now union-typed and worth making the disambiguation crisp.

**Proposed:** decide branch on shape first, then decode each element throwing.

```swift
public init(from decoder: Decoder) throws {
    let container = try decoder.singleValueContainer()
    if var unkeyed = try? decoder.unkeyedContainer() {
        var items: [AIChatPageContextData] = []
        while !unkeyed.isAtEnd {
            items.append(try unkeyed.decode(AIChatPageContextData.self))
        }
        self = .multiple(items)
        return
    }
    self = .single(try container.decode(AIChatPageContextData.self))
}
```

Verified the encode side is symmetric (no `encodeIfPresent` slip); only the decoder fallback is fragile.

macOS/DuckDuckGo/AIChat/AIChatPanelAttachment.swift
```diff
@@ -0,0 +1,65 @@
...
+        case .tab(let attachment):
+            return "tab:\(attachment.id)"
+        case .file(let attachment):
+            return "file:\(attachment.id.uuidString)"
+        }
+    }
+
+    static func == (lhs: AIChatPanelAttachment, rhs: AIChatPanelAttachment) -> Bool {
+        switch (lhs, rhs) {
+        case let (.image(a), .image(b)):
```
jaceklyp: **Minor:** The "id + identity" equality rule for image attachments is duplicated across three sites — here, `AIChatTabAttachment.==` at line 40 of this file, and `AddressBarSharedTextState.setAIChatAttachments` line 130 (change-detection inside `zip`). A future tweak (e.g. include `fileName` in equality) has to touch all three.

**Proposed:** fold the rule into `AIChatImageAttachment: Equatable` (its fields are equatable; `NSImage` uses `===` identity) so `AIChatPanelAttachment: Equatable` becomes synthesizable.

```swift
extension AIChatImageAttachment: Equatable {
    public static func == (lhs: Self, rhs: Self) -> Bool {
        lhs.id == rhs.id && lhs.image === rhs.image
    }
}
```

Removes the whole class of bug rather than patching this one.

Side note: the doc comment block above this `case` says `setAIChatAttachments` does an "id+identity" check — it actually uses `image === image` only as a change-detection check inside a `zip`, not as the overall equality. Worth tightening or just dropping the cross-reference if the rule lands in `AIChatImageAttachment` itself.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -237,17 +259,28 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
             height += Constants.toolButtonSize + Constants.toolButtonBottomInset
 
-            // Add attachments area when there are visible attachments
-            if !attachmentsContainerView.isHidden && !attachmentsContainerView.attachments.isEmpty {
-                height += Constants.attachmentsRowHeight + Constants.attachmentsBottomSpacing
-                if attachmentsContainerView.hasExcessAttachments {
-                    height += Constants.attachmentsErrorHeight
-                }
+            if !tabAttachmentsCarouselView.attachments.isEmpty {
+                height += Constants.tabAttachmentsTotalPanelReservation
```
jaceklyp: **Minor:** `totalPassthroughHeight` and `additionalContentHeight` (line 237 just above) compute the same "carousel reservation" but disagree about gating.

`additionalContentHeight` reserves carousel space unconditionally; `totalPassthroughHeight` only reserves it when `omnibarController.isOmnibarToolsEnabled || !imageUploadButton.isHidden`. Not currently reachable (the carousel is populated only via tools UI) but the parallel computation will drift over time — and the project convention elsewhere in this file is "single source of truth", which two sites independently computing this contradicts.

**Proposed:** extract one helper and call it from both.

```swift
private var attachmentRowReservation: CGFloat {
    var height: CGFloat = 0
    if !tabAttachmentsCarouselView.attachments.isEmpty {
        height += Constants.tabAttachmentsTotalPanelReservation
    }
    if hasVisibleImageExcess || hasVisibleFileExcess {
        height += Constants.attachmentsErrorHeight
    }
    return height
}
```

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
```diff
@@ -136,6 +140,13 @@ final class AIChatOmnibarController {
         featureFlagger.isFeatureOn(.aiChatOmnibarVoiceChatAccess)
     }
 
+    /// Whether the omnibar's tab picker (Attach Page Content) is available.
+    /// Requires both `aiChatPageContext` (the underlying extraction pipeline) and
+    /// `aiChatOmnibarAttachMoreTabs` (the omnibar surface gate).
+    var isOmnibarTabPickerEnabled: Bool {
```
jaceklyp, line 147: **Minor:** This flag is read at button-display time but **not** re-checked at submit. If the privacy-config flag flips off remotely while the user has tabs picked, the in-flight `pageContext` payload still goes out.

Sidebar isn't affected because `supportsTabPicker` gates the JS-bridge surface, and the FE doesn't have a long-lived picked-tabs store. Omnibar persists tab attachments to shared state, so a remote disable doesn't clear the picked tabs — they survive into the next submit.

**Either** gate `tabAttachments` reads in `submit()` (and silently drop the entries when the flag is off), **or** add a comment justifying why in-flight attachments should still ship after a remote disable. Probably the latter — it's a corner case — but worth being explicit.

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
```diff
@@ -524,6 +565,137 @@ final class AIChatOmnibarController {
...
+        guard current.contains(where: { $0.id == id }) else { return }
+        current.removeAll { $0.id == id }
+        persistFileAttachmentsToActiveTab(current)
+    }
+
+    /// Returns the open browser tabs (pinned + regular) in this controller's window as candidate
+    /// attachments, with native `NSImage` favicons resolved from the favicon manager. Used by the
+    /// omnibar attach menu to populate the "Attach Page Content" submenu. Non-URL tabs (settings,
+    /// new-tab page, etc.) are filtered out.
+    func openTabsForOmnibarPicker() -> [AIChatTabAttachment] {
```
jaceklyp, line 710: **Minor:** This filter (URL-tabs only) doesn't match what the sidebar's tab picker uses. After commit `87c7f0c8ae` ("Move AIChat tab-picker URL filter to a reusable shared helper") the sidebar skips `about:blank`, duck.ai itself, and the DDG homepage. The omnibar picker will offer them, which is a visible UX divergence between the two surfaces.

**Proposed:** route through the same shared helper instead of duplicating the rule.

```swift
func openTabsForOmnibarPicker() -> [AIChatTabAttachment] {
    AIChatTabPickerCandidates                       // (or whatever the shared helper is named)
        .filter(from: tabCollectionViewModel)
        .map { tab in
            let url = tab.url
            let title = tab.title ?? url.host ?? ""
            let favicon = NSApp.delegateTyped.faviconManager
                .getCachedFavicon(for: url, sizeCategory: .small)?.image
            return AIChatTabAttachment(id: tab.uuid, title: title, url: url, favicon: favicon)
        }
}
```

Stops divergence at the source, not at the symptom.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1020,36 +1119,177 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
-        panel.allowedContentTypes = allowedContentTypes(for: omnibarController.selectedModelImageFormats)
+        panel.allowedContentTypes = pickerAllowedContentTypes()
 
         guard let window = view.window else { return }
         panel.beginSheetModal(for: window) { [weak self] response in
             guard let self, response == .OK else { return }
-            let remaining = Constants.attachmentsDisplayCap - self.attachmentsContainerView.attachments.count
-            for url in panel.urls.prefix(max(remaining, 0)) {
-                self.addImageAttachment(from: url)
+            // Track per-kind slot consumption so we never overshoot either cap. Both kinds use
```
jaceklyp: **Minor:** Two issues on the cap path.

**(a) Duplicated per-kind cap math.** This site keeps local `imagesAdded` / `filesAdded` counters with `imageCap = Constants.imageAttachmentsDisplayCap`, while `omnibarController.addImageAttachmentToActiveTab` (`AIChatOmnibarController.swift:618`) already caps images via `< Self.imageAttachmentsDisplayCap`. Two cap-check sites for the same rule.

**(b) The file path has no overall cap at all.** `addFileAttachmentToActiveTab` (`AIChatOmnibarController.swift:660`) caps only by id-uniqueness — there's no `< 3` guard. The picker happens to be the only entry point today, but any future path (drag-drop, paste, debug menu, deep link) bypasses the local counter and overshoots silently.

**Proposed:** centralize in the controller's `add*` methods so this site shrinks to:

```swift
for url in panel.urls {
    let utType = UTType(filenameExtension: url.pathExtension.lowercased())
    if let utType, utType.conforms(to: .image) {
        self.addImageAttachment(from: url)
    } else {
        self.addFileAttachment(from: url)
    }
}
```

…and `addFileAttachmentToActiveTab` grows the same `< fileAttachmentsDisplayCap` guard the image side has.

Aside (not blocking): consider phrasing this block's lead comment as "Hold" / "Record" instead of "Track" — see the umbrella comment at line 114 for the suggestion.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -96,20 +113,23 @@ final class AIChatOmnibarContainerViewController: NSViewController {
 
     /// Tracks ongoing resize tasks by attachment ID. Used to ensure resizes complete before submission.
```
jaceklyp: Nit: consider rewording "Tracks" → "Holds" / "Records" / "Captures" in this context — same word recurs on the new comments at line 938 (`// Track the real height even when suppressed`) and line 1151 (`// Track per-kind slot consumption…`). For property/state-storage comments "tracks" reads a little oddly; the alternatives feel more natural. Optional, but worth a one-pass sweep through all three new uses if you agree.

macOS/DuckDuckGo/AIChat/AIChatTabAttachmentsCarouselView.swift
```diff
@@ -0,0 +1,225 @@
...
+            viewsByAttachmentId[id] = view
+            stackView.insertArrangedSubview(view, at: index)
+        }
+
+        onAttachmentsChanged?()
+    }
+
+    /// Wipes the carousel without firing `onAttachmentsChanged` — used by panel teardown where
+    /// the upstream shared state must not be persisted to.
+    func clearWithoutNotification() {
```
jaceklyp: Nit: this function looks like dead code. No callers anywhere in `iOS/` or `macOS/` (verified via grep). Can drop the whole method (and `viewsByAttachmentId.removeAll()` if nothing else needs it).

macOS/DuckDuckGo/AIChat/AIChatTabAttachmentsCarouselView.swift
```diff
@@ -0,0 +1,225 @@
...
+/// image thumbnails and tab cards in a single ordered stack — the order matches the user's
+/// insertion order (driven by `AIChatPanelAttachment` from `AddressBarSharedTextState`), so a
+/// sequence of toggles like `tab, tab, image, tab` renders exactly that way.
+///
+/// The view is intentionally agnostic about how attachments come and go: it just renders the
+/// list it's handed. Removal is delegated upward via the `onImageAttachmentRemoveRequested` /
+/// `onTabAttachmentRemoveRequested` callbacks; the container VC routes those into the data
+/// model (shared state), the publisher fires, and the carousel re-renders. Single source of
+/// truth, no in-carousel mutation.
+final class AIChatTabAttachmentsCarouselView: NSView {
```
jaceklyp: Nit: the class name (and its uses in `AIChatOmnibarContainerViewController` as `tabAttachmentsCarouselView` / `tabAttachmentsHeightConstraint` / `tabAttachmentsRowHeight` etc.) no longer reflect what's in the view — it holds image + file + tab attachments now, as the file's own doc comment at line 26 acknowledges ("Holds both image thumbnails and tab cards"). Consider renaming to `AIChatAttachmentsCarouselView` (and the property uses along with it) so the next reader looking for "the file-attachment carousel" actually finds it. Pure rename, low risk, can be a follow-up.

macOS/DuckDuckGo/Common/Localizables/UserText.swift
```diff
@@ -683,7 +687,8 @@ struct UserText {
...
     static let aiChatImageGenDeactivateTooltip = NSLocalizedString("aichat.image-gen-button.deactivate-tooltip", value: "Deactivate image generation", comment: "Tooltip for the image generation button when image generation mode is active")
-    static let aiChatAttachmentsLimitError = NSLocalizedString("aichat.attachments.limit-error", value: "You can only attach 3 images at a time", comment: "Error message shown when the user tries to attach more than 3 images in AI chat")
+    static let aiChatAttachmentsLimitError = NSLocalizedString("aichat.attachments.limit-error", value: "You can only add 3 images at a time", comment: "Error message shown when the user tries to add more than 3 images in AI chat")
+    static let aiChatFileAttachmentsLimitError = NSLocalizedString("aichat.file-attachments.limit-error", value: "You can only add 3 files per conversation", comment: "Error message shown when the user tries to add more than 3 files (PDFs etc.) in AI chat")
```
jaceklyp, line 692: Nit: two near-identical user-facing strings with slightly different phrasing — "You can only add 3 **images at a time**" (line 690) vs "You can only add 3 **files per conversation**" (line 691). Both caps are `3` from `maxImageAttachments` / `maxFileAttachments`. Worth flagging to copy review whether the wording difference is intentional or accidental.

Also: the `"3"` is hard-coded in both strings rather than parameterized from the constants. If either cap ever moves to 4, the string is wrong until copy is refreshed and there's no compile-time link between them. Not blocking, just flagged.

macOS/UnitTests/AIChat/AIChatOmnibarControllerTests.swift
```diff
@@ -506,28 +511,255 @@ final class AIChatOmnibarControllerTests: XCTestCase {
...
+        // submit flow via the existing image-side path.
+        let sharedState = tabCollectionViewModel.selectedTabViewModel?.addressBarSharedTextState
+        XCTAssertTrue(sharedState?.aiChatTabAttachments.isEmpty ?? false,
+                      "Tab attachments are cleared from shared state after a successful submit")
+    }
+
+    func testWhenSubmitWithTabAttachments_ThenSubmitProceedsAndPixelFires() async {
+        // Given — two tab attachments. Note: in this unit test, the `TabCollectionViewModel`
+        // doesn't contain real `Tab` instances for "tab-A" / "tab-B", so the page-context
+        // extractor returns nil for both and the prompt's `pageContext` ends up nil. We
```
jaceklyp, line 620: The honest comment here calls out what I think is the biggest single test gap in this PR: there's no mock for the per-tab page-context extractor, so the **entire M8 wire-format contract** (the active-tab discriminator strip, mixed-tab ordering, silent-drop on timeout, silent-drop on closed-tab, cancel mid-fetch) is untested at the controller level.

Landing a `MockAIChatPageContextExtracting` (or making the extraction injectable into `AIChatOmnibarController`) unlocks a cluster of high-value tests:

```swift
// 1. End-to-end discriminator: active tab → tabId stripped
func testWhenSubmitWithActiveTabPicked_ThenPageContextEntryHasNilTabId() async { … }

// 2. Mixed array preserves pick order + strips active
func testWhenSubmitWithMixedActiveAndPickerTabs_ThenPageContextOrdersWithActiveStripped() async { … }

// 3. Silent-drop on hung extractor (5s timeout)
func testWhenOneTabFetchHangs_ThenOnlyThatEntryIsDropped() async { … }

// 4. Combined kinds
func testWhenSubmitWithImageFileAndTab_ThenPromptCarriesAllThree() async { … }

// 5. Closed-tab silent drop
func testWhenPickedTabIsClosedBeforeSubmit_ThenItIsDropped() async { … }

// 6. Cancel mid-fetch
func testWhenOmnibarDismissedMidSubmit_ThenPendingFetchesAreCancelled() async { … }
```

Other smaller gaps worth picking up while you're in this file:

- The test on line 616, `testWhenSubmitWithTabAttachments_ThenSubmitProceedsAndPixelFires`, doesn't actually assert a pixel — the name lies. Either rename to drop `_ThenSubmitProceedsAndPixelFires` or add the `PixelKit` mock assertion (`aichat_addressbar_submit_with_tabs` fired exactly once).
- 3-image / 3-file cap enforcement isn't covered. If caps live at the menu/UI layer that's defensible, but `addFileAttachmentToActiveTab` has no cap guard at the controller level (see `AIChatOmnibarController.swift:660`) — worth at least one negative test for "4th file rejected".
- Pervasive `Task.sleep(50_000_000)` (line 1387) + `for _ in 0..<5 { await Task.yield() }` (lines 606, 630, 687, 714, 727, 820, 836, 851). Hand-tuned waits are the classic timing smell. If the snapshot-before-await refactor on `AIChatOmnibarController.swift:883` lands, the submit path becomes deterministic and these can collapse into a single `await submitTask.value`-style hook.

Sister-file gaps (`AIChatNativePromptTests.swift`):

- Decoder symmetry for the **absent** case is missing. Encoder side asserts the field is omitted, but no decoder test confirms `pageContext` decodes as `nil` when the key is absent. One-liner addition:
  ```swift
  func decodingQueryWithoutPageContextKey_ThenPageContextIsNil() throws { … }
  ```
- Decoder side for `tabId: nil` on the active entry: `encodingQueryWithMultiplePageContexts` only asserts the encoder. There's no test that the decoder reads back `tabId: nil` for the active entry — and this is **the** discriminator contract with the duck.ai web app. Worth pinning.
- `encodingQueryWithMultiplePageContexts` accepts the active entry's `tabId` value as `nil` *or* `NSNull` — should tighten to "key is absent from the JSON", matching the wire-format spec exactly.

jaceklyp:
UX: when attaching pages, shouldn’t the current tab be pinned to the top? I think most use cases involve attaching the page you’re currently viewing, and once you have multiple tabs open like I do, it becomes extremely difficult to find it.

jaceklyp:
Question: Should the visible pages only be from the given window?

jaceklyp:
1. I think we should display tooltips for image attachments as well, for consistency. It would also improve usability: image thumbnails are small, and when several images look similar, it’s hard to tell which one was attached without opening or reattaching it again.
2. When I attach a page, enter a prompt, and press Enter on the keyboard, there’s no visual feedback indicating that anything is happening. The text in the omnibar simply disappears, and only after a delay does Duck.ai start loading.
3. For some reason also that attachment is not being carried to the duck.ai.
See video for 2 and 3:

https://github.com/user-attachments/assets/30f72653-0187-41bb-8409-6c18411892ce

update: it works for site like filmweb.pl, it doesn't for github.com

tomasstrba:
Thanks for the review, @jaceklyp. I'll be looking at all your comments and Cursor comments throughout today. Will post an update when it's ready for another look

SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
tomasstrba: This is intentional. The key wasn't set in the privacy configuration yet. And we also don't plan to set it after this (even internal release is delayed)

Agree with the side effect. If someone is already opted in, they have to opt in again. The advantage on the other hand is we will follow convention with other feature flags.

tomasstrba:
All code comments are addressed except following:
- https://github.com/duckduckgo/apple-browsers/pull/4774#discussion_r3232037711
- https://github.com/duckduckgo/apple-browsers/pull/4774#discussion_r3232037723
- https://github.com/duckduckgo/apple-browsers/pull/4774#discussion_r3232037750
- https://github.com/duckduckgo/apple-browsers/pull/4774#discussion_r3232037751
The reasons are added as comments

I've fixed UX issues you raised about current tab not being on top (good catch!) and also building the tab list only from the current window.

**Can't reproduce** 

The github.com tab attachment works for me well. Tab I attached is visualized on duck.ai in following way:
Screenshot 2026-05-13 at 23 29 06
No one in the ship review mentioned problems with attachments either.

**Out of the scope for now**

> I think we should display tooltips for image attachments as well, for consistency. It would also improve usability: image thumbnails are small, and when several images look similar, it’s hard to tell which one was attached without opening or reattaching it again.

> When I attach a page, enter a prompt, and press Enter on the keyboard, there’s no visual feedback indicating that anything is happening. The text in the omnibar simply disappears, and only after a delay does Duck.ai start loading.

Likely caused by the custom AI Chat URL

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatScriptUserValues.swift
tomasstrba: I'd lean towards failing entirely if one item is malformed

macOS/DuckDuckGo/Common/Localizables/UserText.swift
tomasstrba: This copy is intentional. See duck.ai or sidebar

macOS/UnitTests/AIChat/AIChatOmnibarControllerTests.swift
tomasstrba: I'd prefer to cover functionality with UI tests, rather than make the view controller implementation a lot more complex for unit testing. I'll implement UI tests as a follow-up once the ship review is resolved.

jaceklyp (approved):
nice work :)

PR #4780, opened 2026-05-08
Add a CI failure diagnosis skill

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1214627103230178?focus=true
Tech Design URL:
CC:

### Description

This PR adds a skill for diagnosing CI failures.

### Testing Steps

1. Find a recent CI failure in https://github.com/duckduckgo/apple-browsers/actions?query=is%3Afailure and run the skill against it, check whether the results are useful & accurate - also check whether the LLM is prompting you with the right questions and making the right decisions (for example, not forcing you to use the Xcode MCP if you use another tool for building)

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Adds a new documentation-only Claude skill file; no production code paths are modified, so risk is limited to internal workflow/usage guidance being incorrect.
> 
> **Overview**
> Adds a new Claude skill, `ddg-diagnose-ci-failure`, that provides a step-by-step procedure to resolve a GitHub Actions run (from run/PR/branch/Asana inputs), collect logs and JUnit artifacts, classify failure types, and guide follow-up analysis for test failures.
> 
> The skill includes explicit safeguards around requiring authenticated `gh` usage, preserving run `attempt`/`job` context, and *only* querying the Apple CI Failing Tests Asana project as an escalation due to MCP tool-locking side effects, plus guidance to wait for user confirmation before proposing fixes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e76e0d34eaef7d4ba714b27ba2f55635a4d793ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/skills/ddg-diagnose-ci-failure/SKILL.md
+++ b/.claude/skills/ddg-diagnose-ci-failure/SKILL.md
@@ -154,14 +154,14 @@ If the test or its production target changed in the last few commits, that's alm
 When the failing PR's own diff doesn't explain the failure, check whether another PR changed shared test infrastructure (helpers, fixtures, mocks under `iOS/SharedTestUtils/`, `macOS/SharedTestUtils/`, etc.) between this PR's last green run and main HEAD:
 
 ```bash
-# Find this PR's last green run, capture the SHA
-gh run list --branch=<pr_branch> --status=success --limit=1 --json headSha,createdAt
+# Find this PR's last green run for the failed workflow, capture the SHA
+gh run list --branch=<pr_branch> --workflow="<workflow_name>" --status=success --limit=1 --json headSha,createdAt
 
 # What landed in main between that SHA and HEAD?
 git log --oneline <last_green_sha>..HEAD -- <suspected_shared_paths>
 ```
 
-A diff in a helper or mock used by the failing test is a strong signal: the PR's CI passed in isolation, but a parallel-merged PR shifted the contract underneath it. The failing test is then often a real test-side mismatch, not a production bug.
+Use the workflow name identified in Step 1. A diff in a helper or mock used by the failing test is a strong signal: the PR's CI passed in isolation, but a parallel-merged PR shifted the contract underneath it. The failing test is then often a real test-side mismatch, not a production bug.
 
 **Workflow history (flaky / environmental failures):**
 
@@ -169,13 +169,13 @@ When local git shows no relevant changes, or the failure looks intermittent:
 
 ```bash
 # Last green runs of the same workflow on main
-gh run list --workflow={workflow_name} --branch=main --status=success --limit=5 --json headSha,createdAt,displayTitle
+gh run list --workflow="{workflow_name}" --branch=main --status=success --limit=5 --json headSha,createdAt,displayTitle
 
 # Recent runs to see the fail/pass pattern
-gh run list --workflow={workflow_name} --branch=main --limit=20 --json conclusion,headSha,createdAt
+gh run list --workflow="{workflow_name}" --branch=main --limit=20 --json conclusion,headSha,createdAt
 
 # Runs on this PR branch - did the test fail across all pushes (deterministic) or only some (flaky)?
-gh run list --branch={pr_branch} --workflow={workflow_name} --limit=10
+gh run list --branch={pr_branch} --workflow="{workflow_name}" --limit=10
 ```
 
 The diff between the last green SHA and the failing SHA brackets the regression - useful when git log on individual files doesn't surface the cause (env shifts, indirect dependencies, build-config tweaks).
--- a/.claude/skills/ddg-diagnose-ci-failure/SKILL.md
+++ b/.claude/skills/ddg-diagnose-ci-failure/SKILL.md
@@ -9,13 +9,13 @@ Investigate a GitHub Actions workflow failure, classify it, trace test failures 
 
 ## Inputs
 
-The skill resolves any of the following to a single GitHub Actions run to investigate. The repo is always `duckduckgo/apple-browsers`. Resolve the input first, then start at Step 1 with the resulting `run_id`.
+The skill resolves any of the following to a single GitHub Actions run to investigate. The repo is always `duckduckgo/apple-browsers`. Resolve the input first, then start at Step 1 with the resulting `run_id` and any optional `job_id` or `attempt_number`.
 
 ### Actions run URL
 
 Example: `https://github.com/duckduckgo/apple-browsers/actions/runs/25519433222` (optionally with `/job/<id>` or `/attempts/N`).
 
-Extract `run_id` (and `job_id` if present). Use directly.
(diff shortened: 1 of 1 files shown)


.claude/skills/ddg-diagnose-ci-failure/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+  --jq '{run_started_at, updated_at}'
+```
+
+Download by artifact ID - this disambiguates duplicate names and avoids relying on `gh run download`'s name-matching behaviour:
+
+```bash
+mkdir -p /tmp/ci-artifacts/{name}
+gh api repos/duckduckgo/apple-browsers/actions/artifacts/{artifact_id}/zip > /tmp/artifact.zip
+unzip -o /tmp/artifact.zip -d /tmp/ci-artifacts/{name}/
+```
```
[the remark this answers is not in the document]
samsymons: This doesn't seem to be true based on local testing, the skill is working fine. I asked Codex for a second opinion:

```
Not legitimate, based on empirical test against a real artifact on the current branch. I downloaded artifact 6867751496 from a real run using exactly the command in the skill:

gh api repos/duckduckgo/apple-browsers/actions/artifacts/6867751496/zip > artifact.zip

Result on gh 2.92.0:
- 92,173-byte file, exit 0, no stderr
- file reports Zip archive data, at least v2.0 to extract, compression method=deflate
- Magic bytes are 50 4b 03 04 (proper PK\x03\x04 zip header)
- unzip -l artifact.zip lists unittests.xml (817,485 bytes uncompressed) cleanly

So the binary passes through gh api to the file without corruption, the 302 to Azure blob storage is followed correctly, and unzip is happy.

The reviewer is probably remembering cli/cli#5853 (https://github.com/cli/cli/issues/5853)-era behaviour from ~2022, where gh api did add a trailing byte/newline to non-JSON output. That was fixed long ago. There's no current "gh api is documented as producing corrupted files when redirecting binary output" — that's not in gh api --help (just checked) and the live behaviour contradicts it.

Two extra reasons not to take the suggestion:

1. gh run download doesn't accept artifact ID. It only filters by --name or --pattern. Switching to it would re-introduce the exact name-collision-across-attempts ambiguity that commit d1353453fc was solving. That's a regression of the prior review fix.
2. curl --output would need a separate call to get the presigned URL first (the artifact endpoint returns a 302 to Azure storage), which is more complexity for no benefit since gh api already handles the redirect.

I'd push back on the comment: the skill's approach is correct, and the alternatives the reviewer proposes either don't solve the problem the previous review fix was solving, or add complexity without benefit. If you want, I can paste this empirical evidence into a reply.
```

.claude/skills/ddg-diagnose-ci-failure/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+Example: `https://github.com/duckduckgo/apple-browsers/pull/4752`.
+
+Extract the PR number, then resolve as below.
+
+### Bare PR number
+
+Example: `4752` or `#4752`.
+
+```bash
+gh pr checks <pr_number> --repo duckduckgo/apple-browsers --json name,state,link,workflow
```
samsymons: There's not really a better way to get the info we need than the `gh` tool, which we also recommend as part of release workflows when triggering them locally. The skill could probably benefit from a single line up front to mention bailing out early if `gh` is not installed? **(EDIT: updated the skill to mention this)**

.claude/skills/ddg-diagnose-ci-failure/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+description: Invoke ONLY when the user explicitly runs /ddg-diagnose-ci-failure or names this skill by name (e.g. "use ddg-diagnose-ci-failure on PR 4780"). Do NOT auto-invoke from symptom/intent matching - the skill may query the Apple CI Failing Tests Asana project (Step 3e), which locks other MCPs for the rest of the session under the lethal-trifecta policy, so it must be user-initiated. If the user asks about CI failures, GitHub Actions runs, or flaky tests without naming this skill, answer directly instead. Accepts one optional argument: a GitHub Actions run URL, a PR URL or bare PR number, or an Apple CI Failing Tests Asana task URL - or no argument to use the current branch's open PR.
+---
+
+# Diagnose CI Failure
+
+Investigate a GitHub Actions workflow failure, classify it, trace test failures back to the local code, and propose a fix.
+
+## Inputs
+
+The skill resolves any of the following to a single GitHub Actions run to investigate. The repo is always `duckduckgo/apple-browsers`. Resolve the input first, then start at Step 1 with the resulting `run_id` and any optional `job_id` or `attempt_number`.
```
samsymons: >The repo is always `duckduckgo/apple-browsers`

Can probably remove this. I had this in the skill originally when it was in my user-level Claude directory, since Claude would sometimes get confused about what repo I was referring to when I asked it to diagnose a failure. Now that it's part of the Apple repo directly this _shouldn't_ be necessary, and it would be nice to be able to point it at TrackerRadarKit, Fastlane plugin repo, etc. **(EDIT: removed explicit monorepo references so that the skill can be used more generically)**

aataraxiaa (approved):
Nice @samsymons. Ran against linked CI and results seem useful:

```
 1. What the tests do — they exercise three entry points into NewTabPageViewController (viewDidAppear, onboardingCompleted, showNextDaxDialog) and assert that the new-style Dax dialog provider was invoked (nextHomeScreenMessageNewCalled == true).                                                                                   
  2. What failed — every test's XCTAssertTrue(specProvider.nextHomeScreenMessageNewCalled) is false: the new-style dialog provider was never called.                                                                                                                                                                                    
  3. Why — PR #4785 adds an early-return guard at the top of presentNextDaxDialog() in iOS/DuckDuckGo/NewTabPageViewController.swift:286:                                                                                                                                                                                                 
  guard DefaultTutorialSettings().hasSeenOnboarding else { return }                                                                                                                                                                                                                                                                       
  3. DefaultTutorialSettings() is a global-backed (UserDefaults) settings object. In the test target it has no test fixture and hasSeenOnboarding is false, so the guard short-circuits before showNextDaxDialogNew(...) runs — meaning specProvider.nextHomeScreenMessageNewCalled is never set to true. The new production guard is     
  correct behaviour; the tests just don't satisfy its precondition.                                                                                                                                                                                                                                                                       
  4. Test or production? — Test side. The production change is intentional ("don't surface a Dax dialog until onboarding is done"). The tests were not updated to set hasSeenOnboarding = true in setUpWithError, and the SUT pulls DefaultTutorialSettings directly rather than through an injected dependency, so the tests have no seam
   to bypass the new guard. Fix is in the test (set hasSeenOnboarding = true in setUp, restore in tearDown) — or, as a sturdier follow-up, refactor NewTabPageViewController to take a TutorialSettings protocol so a mock can be injected. Step 3e (Asana lookup) was skipped — diagnosis is unambiguous and would lock the session's    
  other MCPs.    
```

Bunn (approved):

PR #4791, opened 2026-05-08
iOS: add one-tap OTA install for ad-hoc builds

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214200115953388/task/1214633075025736?focus=true
Tech Design URL:
CC:

### Description
Adds an over-the-air install page to the iOS ad-hoc build workflow so testers with a UDID-registered iPhone can install the build by tapping a single Safari link, instead of going through AirDrop / Apple Configurator. The workflow now uploads a manifest plist and a small install HTML alongside the IPA, and the success Asana comment + GitHub step summary surface the install URL.

### Testing Steps
1. Run the **iOS Ad-Hoc Build** workflow from Actions on this branch (any build type — Alpha is fine — and pass an Asana task URL).
2. When the run finishes, open the workflow summary and confirm there's an "Install on iPhone" link, and that the Asana task got a comment with both the install URL and the IPA URL.
3. On a UDID-registered iPhone, open the install URL in Safari and tap **Install on iPhone** — the iOS install sheet should appear and the build should install. Open the same URL in Chrome / Slack / an in-app webview and confirm the install button is hidden and the "open in Safari" warning is shown instead.

### Impact and Risks
Low — release infra only, no app code changes.

#### What could go wrong?
Most plausible regression: the per-build-type bundle ID in the workflow drifts from the IPA's actual bundle ID, which would surface on device as a generic "Unable to install". The mapping is set explicitly in the new workflow step (Alpha / Release / Experimental) and matches the existing IPA naming.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches release CI by generating and publicly uploading OTA install artifacts to S3; failures or mismatched per-build bundle IDs could prevent installs or expose incorrect links, though no app code is changed.
> 
> **Overview**
> Adds a new composite action (`generate-ios-adhoc-ota-install`) that renders an OTA `manifest.plist` and Safari-friendly `install.html` from templates, uploads them next to the IPA in S3, and outputs an `install-url`.
> 
> Updates the `ios_adhoc` workflow to run this step after uploading the IPA (with `continue-on-error`), include the install link (or a warning) in the GitHub step summary, and switch Asana reporting to a `template_name:"ios-adhoc-build-available"` comment.
> 
> Bumps `fastlane-plugin-ddg_apple_automation` from `4.1.1` to `4.1.3` for both iOS and macOS fastlane configs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2696716d12cfef78663e5c6ddd53aaaafdfe6404. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/ios_adhoc.yml
```diff
@@ -191,23 +191,115 @@ jobs:
           aws s3 cp $IPA_LOCAL_PATH $ipa_s3_path --acl public-read
           echo "ipa-s3-path=${ipa_s3_path}" >> $GITHUB_OUTPUT
 
+      - name: Generate and upload OTA install manifest and page
```
alessandroboron, line 194: If this step fails (AWS error, render.py issue, anything in the pipline), the whole job fails and the Report success step gets skipped but the IPA and dSYMs have already been uploaded successfully at that point, so the primary deliverable is fine

Would it make sense to make this step non-fatal and have Report success posting the Asana comment with just the IPA / dSYM / workflow links, omitting the Install link? Something like:

```
- name: Generate and upload OTA install manifest and page                                                                                                                                               
    id: generate-ota
    continue-on-error: true
```

and then in Report success, gate the Install-link bullet on `$INSTALL_URL` being non-empty.

The step still shows as ⚠️ in the UI so the OTA path being broken will be noticeable, but we don't lose access to the IPA if OTA generation fails. 

Let me know what you think :)

.github/workflows/ios_adhoc.yml
```diff
@@ -191,23 +191,115 @@ jobs:
...
+            s|.*https://app\.asana\.com/1/[0-9]+(?:/[0-9a-z/]*)?/task/([0-9]+).*|\1|;
+          ' <<< "$ASANA_TASK_URL")
+          asana_task_id="${asana_task_id//[^0-9]/}"
+
+          if [[ -z "$asana_task_id" ]]; then
+            echo "Could not parse Asana task GID from: ${ASANA_TASK_URL}" >&2
+            exit 1
+          fi
+
+          html_body="<body><strong>${TITLE} ${app_version} (${build_number})</strong><ul><li><a href=\"${INSTALL_URL}\">Install on iPhone</a> — open in Safari</li><li><a href=\"${IPA_URL}\">Download IPA</a></li><li><a href=\"${WORKFLOW_URL}\">Workflow run</a></li></ul></body>"
```
alessandroboron: @ayoy a question for you:

The `asana_add_comment` [action](https://github.com/duckduckgo/fastlane-plugin-ddg_apple_automation/blob/main/spec/asana_add_comment_action_spec.rb) in `fastlane-plugin-ddg_apple_automation` supports rich text via the `template_name: parameter`

Should we define the template in the automation plugin?

.github/workflows/ios_adhoc.yml
pballart: Hi @alessandroboron thanks for the review. I like your suggestion to be more defensive. I think the chance of failure is very low given that render.py is a small script with small scope and AWS errors would affect both the ipa and the manifest (both are uploaded to the same folder using the same method). So while the risk I think is low, your suggestion is worth it and I’ll implement it right now 👍

.github/workflows/ios_adhoc.yml
pballart: Yeah, agreed. This would be cleaner as an `asana_add_comment` template instead of doing task parsing + direct `html_text` posting in the workflow.

I can move it into `fastlane-plugin-ddg_apple_automation` and then update this workflow to use `template_name`. It’s a bit of cross-repo scope, but I’m happy to do it if that’s the direction we want.

ayoy: Yes, this sounds good, thanks for spotting this @alessandroboron and thanks for the plugin patch @pballart! 🙌

.github/workflows/ios_adhoc.yml
```diff
@@ -191,23 +191,103 @@ jobs:
...
+
+          export MANIFEST_URL="${ipa_dir_url}/${manifest_filename}"
+          install_url="${ipa_dir_url}/${install_filename}"
+
+          # URL-encode the manifest URL for embedding in the itms-services link.
+          export MANIFEST_URL_ENCODED="$(python3 -c 'import sys, urllib.parse; print(urllib.parse.quote(sys.argv[1], safe=""))' "${MANIFEST_URL}")"
+
+          template_dir="${{ github.workspace }}/.github/templates"
+          python3 "${template_dir}/render.py" "${template_dir}/ios_adhoc_manifest.plist" "${manifest_filename}"
+          python3 "${template_dir}/render.py" "${template_dir}/ios_adhoc_install.html"   "${install_filename}"
```
ayoy: If I can be picky, calling "free-floating" scripts from a workflow doesn't sound great to me. [0]

`.github/templates` directory is not standard, and in current shape it's common to all workflows, so putting a generic-sounding `render.py` inside could make future maintainers scratch their heads over this.

I'd suggest rewriting this step into a standalone action. You could ask an agent to do it - they usually do it great. And with a standalone action you'd get a directory that would be home to your templates and render.py script, plus you could put all the code here in a `.sh` file too.

---

[0] - we do it in a couple of places, but it's mostly legacy, and the scripts are still put either in scripts directory or its subdirectory that clearly states the purpose of the script.

ayoy (commented):
@pballart please see my comment about moving this code to a standalone GitHub action. Let me know in case of questions. Thanks!

.github/workflows/ios_adhoc.yml
pballart: Great suggestion! Makes a lot of sense, I’ll do it like that 👍

.github/actions/generate-ios-adhoc-ota-install/generate_ota.sh
```diff
@@ -0,0 +1,64 @@
...
+    exit 1
+    ;;
+esac
+
+export BUILD_TYPE
+export IPA_URL
+export BUNDLE_VERSION
+export BUILD_NUMBER
+export COMMIT_SHORT_SHA="$(git rev-parse --short HEAD)"
+export BUILD_TIMESTAMP="$(date -u +'%Y-%m-%d %H:%M UTC')"
```
ayoy: Do you really need all these export statements? It doesn't seem that any of these variables are used by child processes, especially `render.py` so these could be just declarations (this applies to all export statements in this file, also outside of the mentioned diff).

```suggestion
case "${BUILD_TYPE}" in
  Alpha)
    BUNDLE_ID="com.duckduckgo.mobile.ios.alpha"
    TITLE="DuckDuckGo Alpha"
    ;;
  Release)
    BUNDLE_ID="com.duckduckgo.mobile.ios"
    TITLE="DuckDuckGo"
    ;;
  Experimental)
    BUNDLE_ID="com.duckduckgo.mobile.ios.experimental"
    TITLE="DuckDuckGo Experimental"
    ;;
  *)
    echo "Unknown build type: ${BUILD_TYPE}" >&2
    exit 1
    ;;
esac

COMMIT_SHORT_SHA="$(git rev-parse --short HEAD)"
BUILD_TIMESTAMP="$(date -u +'%Y-%m-%d %H:%M UTC')"
```

ayoy (approved):
LGTM, tested with [a sample task](https://app.asana.com/1/137249556945/task/1214926569640634/comment/1214926886210051?focus=true) 💅 great job, @pballart! Leaving a comment about a possible simplifying of the bash script, but it's minor overall. Thanks!

.github/actions/generate-ios-adhoc-ota-install/generate_ota.sh
pballart: Good catch, it’s probably a leftover from before when it was not a standalone action. I’ll clean that up.
Also side note on the asana task reporting only works if the task is public to the organization, if it’s just a private task it will fail but I think this is the expected behavior from the current asana comment plugin and we should be good.

ayoy: Thanks!

And yes, it's expected that this will fail if Dax doesn't have access to the task. It's something we have to live with :)

pballart: I pushed a commit with the fixes. I changed the approach instead of exporting vars to just pass what is needed as arguments to be more explicit. Let me know what you think.

ayoy: This is looking great, thanks @pballart! Good to merge :)

.github/workflows/ios_adhoc.yml
```diff
@@ -191,23 +191,50 @@ jobs:
...
           IPA_URL: ${{ steps.upload-ipa-to-s3.outputs.ipa-url }}
           IPA_S3_PATH: ${{ steps.upload-ipa-to-s3.outputs.ipa-s3-path }}
           DSYM_URL: ${{ steps.upload-dsyms-to-s3.outputs.dsym-url }}
           DSYM_S3_PATH: ${{ steps.upload-dsyms-to-s3.outputs.dsym-s3-path }}
+          INSTALL_URL: ${{ steps.generate-ota.outputs.install-url }}
+          TITLE: ${{ steps.generate-ota.outputs.title }}
           WORKFLOW_URL: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
           ASANA_ACCESS_TOKEN: ${{ secrets.ASANA_ACCESS_TOKEN }}
         run: |
+          set -euo pipefail
```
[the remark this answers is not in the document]
pballart: Thanks, but this is intentional. OTA generation is allowed to fail gracefully because the IPA/dSYM artifacts are still the primary deliverable. Posting the final Asana success comment is part of the workflow’s success contract, though, so if `asana_add_comment` fails we do want the job to fail rather than silently report success.

PR #4825, opened 2026-05-11
Update PIR timer handling

quanganhdo:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1214591852011937?focus=true
Tech Design URL:
CC:

### Description

Fixes an issue with PIR timers causing crashes.

### Testing Steps

1. Smoke test PIR by setting the timeouts in BrokerJobExecutionConfig to shorter internal (e.g. 30s) then monitor the logs to confirm actions time out properly. 

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches timeout/cancellation timer lifecycle for PIR actions; mistakes could cause missed timeouts or duplicate error reporting, but scope is limited and covered by updated tests.
> 
> **Overview**
> Fixes PIR timer-related crashes by moving timer creation/invalidation onto the main actor and main run loop. `DataBrokerProtectionFeature` now uses `Timer(timeInterval:)` + `RunLoop.main.add` (instead of `scheduledTimer`) and ensures timers are invalidated safely in `deinit` via `DispatchQueue.main.asyncOrNow`.
> 
> Timeout cancellation is now explicitly awaited from async message handlers (`onActionCompleted`/`onActionError`), and the test suite updates timeouts and adds coverage to ensure timeouts are cancelled even when completion/error handlers run off-main.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7de99987f5ca15702a2222386dc4ad70e49dbfe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/CCF/DataBrokerProtectionFeature.swift
```diff
@@ -168,25 +175,42 @@ public class DataBrokerProtectionFeature: Subfeature {
         installActionTimer(for: (params as? Params)?.state.action)
     }
 
+    @MainActor
     private func installActionTimer(for action: Action?) {
+        actionResponseTimer?.invalidate()
+        actionResponseTimer = nil
```
samsymons, line 181: Previously we didn't do anything before checking that `action` was non-nil - is it okay to do this before that check? Do we ever expect `action` to be nil in this case?

samsymons (approved):
Smoke test of PIR worked as expected and the change seems sensible - just one comment about a behavior change.

SharedPackages/DataBrokerProtectionCore/Sources/DataBrokerProtectionCore/CCF/DataBrokerProtectionFeature.swift
quanganhdo: Good catch — I had reordered those lines as part of unrelated cleanup, but as you noted it changed behavior in the `nil`-action case: the old code left any existing timer untouched when no new action was being installed, but my version was invalidating it unconditionally. Reverting locally to match the original ordering.

On whether `action` is ever nil here: yes, in principle. The only call site is `installActionTimer(for: (params as? Params)?.state.action)` (in `pushAction`), where both the `as?` cast and the optional `state.action` can produce nil. Keeping the original "leave the existing timer alone when no new action is being installed" behavior avoids a semantics change in this PR.

PR #4828, opened 2026-05-11
[iOS] Fix SIGTRAP crash from concurrent mutation during TabsModel archiving

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211767540372501/task/1214486604108348

### Description

- Fix SIGTRAP crash in `Link.encode` caused by concurrent mutation of `TabsModel` during `NSKeyedArchiver` serialization
- Add `NSCopying` conformance to `Link`, `Tab`, and `TabsModel` to enable deep-copying the object graph before archiving
- `TabsModelPersistence.save` now archives a frozen snapshot instead of the live mutable model
- Add `@MainActor` to `TabManager.addHomeTab` and `TabManager.save()` to ensure mutations and saves run on the same thread

### Testing Steps

1. Open the app and create several tabs (mix of home tabs and web pages)
2. Rapidly tap "+" to add new tabs while the tab switcher animation is running
3. Switch between tabs quickly while new tabs are being created
4. Verify no crashes occur during rapid tab creation and switching
5. Close and reopen the app — verify all tabs are restored correctly
6. Repeat steps 2-5 with Fire Mode enabled

### Impact and Risks

**Impact Level: Low**

This is a targeted fix for a crash with minimal behavioral change — the only difference is that archiving now operates on a snapshot copy rather than the live object graph. Tab persistence behavior is otherwise unchanged.

#### What could go wrong?

- The deep copy adds a small memory/CPU overhead per save. Given the typical tab count (<100 tabs with lightweight properties), this should be negligible.
- If a `Tab` property is added in the future but not included in `copy(with:)`, it could be silently dropped during persistence. Mitigated by the copy using the main `init`, so any new required parameter will produce a compile error if not provided.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes tab persistence to archive deep-copied snapshots and adds MainActor annotations around save/mutations; risk is moderate because it touches tab storage/threading and could affect what state is persisted.
> 
> **Overview**
> Fixes a SIGTRAP during tab persistence by **archiving frozen snapshots** of the `TabsModel` object graph instead of serializing live, mutable instances.
> 
> Adds `NSCopying` support for `Link` and introduces `archivalSnapshot()` deep-copy helpers on `Tab` and `TabsModel`, then updates `TabsModelPersistence.save` to archive the snapshot. Also tightens threading by marking `TabManager.save`, `TabManager.addHomeTab`, and bulk-generation `finished()` as `@MainActor`, and adds tests to verify snapshot correctness and isolation from subsequent mutations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35ce628a9366fe471f82b90a724b234cef279e58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/Tab.swift
+++ b/iOS/DuckDuckGo/Tab.swift
@@ -32,7 +32,7 @@ protocol TabObserver: AnyObject {
     
 }
 
-public class Tab: NSObject, NSCoding, NSCopying {
+public class Tab: NSObject, NSCoding {
 
     struct WeaklyHeldTabObserver {
         weak var observer: TabObserver?
@@ -236,7 +236,8 @@ public class Tab: NSObject, NSCoding, NSCopying {
         // Note: type is not encoded as it's now a computed property based on the link URL
     }
 
-    public func copy(with zone: NSZone? = nil) -> Any {
+    /// Returns a frozen deep copy containing only the fields that are persisted via NSCoding.
+    func archivalSnapshot() -> Tab {
         Tab(uid: uid,
             link: link?.copy() as? Link,
             viewed: viewed,
--- a/iOS/DuckDuckGo/Tab.swift
+++ b/iOS/DuckDuckGo/Tab.swift
@@ -32,7 +32,7 @@ protocol TabObserver: AnyObject {
     
 }
 
-public class Tab: NSObject, NSCoding {
+public class Tab: NSObject, NSCoding, NSCopying {
 
     struct WeaklyHeldTabObserver {
         weak var observer: TabObserver?
@@ -234,6 +234,19 @@ public class Tab: NSObject, NSCoding {
         coder.encode(unifiedInputState.selectedTool?.rawValue, forKey: NSCodingKeys.selectedTool)
         // Note: isExternalLaunch and shouldSuppressTrackerAnimationOnFirstLoad are not encoded as they are transient flags
         // Note: type is not encoded as it's now a computed property based on the link URL
+    }
+
+    public func copy(with zone: NSZone? = nil) -> Any {
+        Tab(uid: uid,
+            link: link?.copy() as? Link,
+            viewed: viewed,
+            desktop: isDesktop,
+            lastViewedDate: lastViewedDate,
+            daxEasterEggLogoURL: daxEasterEggLogoURL,
+            contextualChatURL: contextualChatURL,
+            supportsTabHistory: supportsTabHistory,
+            fireTab: fireTab,
+            unifiedInputState: unifiedInputState)
     }
 
     public override func isEqual(_ other: Any?) -> Bool {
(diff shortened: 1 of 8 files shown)


iOS/DuckDuckGo/TabsModel.swift
```diff
@@ -117,6 +117,11 @@ public class TabsModel: NSObject, NSCoding, TabsModelManaging {
         coder.encode(mode.rawValue, forKey: NSCodingKeys.mode)
     }
 
+    public func copy(with zone: NSZone? = nil) -> Any {
+        let tabsCopy = tabs.compactMap { $0.copy() as? Tab }
+        return TabsModel(tabs: tabsCopy, currentIndex: _currentIndex, desktop: false, mode: mode)
+    }
```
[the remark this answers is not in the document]
hassaanelgarem: Addressed in b821dd0c923348b1bcc7b1d592adfc94cd8026c0

iOS/DuckDuckGo/Tab.swift
```diff
@@ -236,6 +236,19 @@ public class Tab: NSObject, NSCoding {
         // Note: type is not encoded as it's now a computed property based on the link URL
     }
 
+    public func copy(with zone: NSZone? = nil) -> Any {
```
pballart: `Tab.copy()` doesn't forward `isExternalLaunch`, `shouldSuppressTrackerAnimationOnFirstLoad`, or `aichatDebugSettings` — `init` defaults kick in. Fine for the archival path (the first two aren't encoded; the
  third is debug-only), but the PR description's *"any new required parameter will produce a compile error"* only holds for params **without** defaults. Every `Tab.init` param except `fireTab` has one, so future fields might silently drop from the snapshot.

Just wanted to raise awareness of this, we might be good with it as it is or we might want to either forward every field in `copy(with:)` explicitly, or rename the entry point to something like `archivalSnapshot()` to signal it's intentionally lossy (`NSCopying` implies "functionally identical,” which this isn't).

pballart (commented):
I had a look at the existing coverage — `TabsModelPersistenceTests` round-trips through `save()` (which calls `copy()` internally now), so in theory a regression in field forwarding would already fail one of
  those tests. But the assertions only check `count`, `currentIndex`, `mode`, and a couple of tabs built via the helper, so a new `Tab` field with a default could quietly drop and the suite would stay green.

If you fancy it, two small additions in `TabsModelTests` would close that gap nicely:

  1. **Direct copy equivalence** — assert every `Tab` field round-trips through `copy()`. That's the one that would immediately catch a silently-dropped field next time someone adds one.
  2. **Independence** — mutate the source after copying, verify the snapshot is unaffected. Nothing currently exercises this and it's the actual property the fix relies on.

Totally optional — the manual testing steps are solid, just thought CI coverage might be worth it.

iOS/DuckDuckGo/Tab.swift
hassaanelgarem: Addressed in 354d7926d650ccf590ff96ae3de36f67c1e7b29c 👍

hassaanelgarem:
@pballart thanks for the thorough review! I addressed both of your points. 👍

pballart (approved):

PR #10, opened 2025-02-13
Release workflows

kshann:
Task/Issue URL: https://app.asana.com/0/1205419239275955/1209302248763523
Tech Design URL:
CC:

**Description**:
Updated release workflows suitable for monorepo.

**Optional E2E tests**:
- [ ] Run macOS PIR E2E tests
	Check this to run the Personal Information Removal end to end tests. If updating CCF, or any PIR related code, tick this.

**Steps to test this PR**:
1. Testing was done in test monorepo repository. 

**Definition of Done**:

* [x] Does this PR satisfy our [Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f)?

---
###### Internal references:
[Pull Request Review Checklist](https://app.asana.com/0/1202500774821704/1203764234894239/f)
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)
[Pull Request Documentation](https://app.asana.com/0/1202500774821704/1204012835277482/f)

PR #4838, opened 2026-05-12
macOS: @-mention tab picker in the duck.ai omnibar

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1213072873557311?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1214704361806993?focus=true

### Description

Typing `@` in the duck.ai omnibar input opens a native picker panel that lists the user's open tabs, filtered live by whatever follows the `@`. The user can:

- Keep typing to filter (substring match on title + URL; spaces stay inside the active query, so `@apple plus` is one multi-word filter).
- Use ↑ / ↓ to move the keyboard highlight (wraps from first ↔ last).
- Mouse-hover to highlight the row under the cursor.
- Press Enter on a highlighted row, or click it, to splice the `@token` out of the input and toggle that tab's attachment on the carousel.
- Press Esc to dismiss the picker only (the omnibar keeps focus so the user can keep typing).

The picker is a borderless, non-activating `NSPanel` (not an `NSMenu`) anchored just below the `@` character via `NSTextView.firstRect(forCharacterRange:)`. `NSMenu`'s modal tracking loop is incompatible with the "keep typing in the input while the menu filters live" requirement; the panel-based design keeps the text view first responder throughout.

Existing surfaces are unchanged: the "Add Page Content" submenu under the attach button stays as-is; the `@`-picker is an additional path to the same `AIChatOmnibarController.toggleTabAttachment(...)` data flow. Gated behind the same `aiChatOmnibarAttachMoreTabs` feature flag as the parent PR (#4774, already merged), so the `@` trigger is a no-op when the flag is off.

### Testing Steps

> Setup: enable `aiChatPageContext` + `aiChatOmnibarAttachMoreTabs` in Debug → Feature Flags. Open several URL tabs.

#### A. Token detection

1. Activate the duck.ai omnibar.
2. Type `@` → "Recent Tabs" picker drops down just under the `@` character.
3. Type `foo@` (text + a non-whitespace char before `@`) → picker does **not** open (email-style `@` is not a trigger).
4. Type `@@goo` → picker opens (closest valid `@` wins).

#### B. Filtering

1. Type `@wiki` → picker shows only tabs whose title or URL contains "wiki", case-insensitive.
2. Type `@apple plus` (with a space) → picker filters by the full `apple plus` phrase. Spaces stay inside the active query.
3. Type `@xyzabc` (no matches) → picker shows a single "No matching tabs" disabled row; the "Recent Tabs" header is hidden in this state.

#### C. Selection + keyboard

1. After typing `@` → first row is highlighted by default.
2. ↓ a few times → highlight moves down; past the last row, wraps to the first.
3. ↑ from the first row → wraps to the last.
4. Move the mouse over a different row → highlight follows the cursor.
5. Press Enter on a highlighted row → the `@token` substring is removed from the input, the tab is attached to the carousel, the picker dismisses.
6. Press Enter on a row that's **already attached** (checkmark visible) → the `@token` is removed and the existing attachment is toggled off (matches the existing "Add Page Content" submenu's toggle behavior).
7. Click a row directly with the mouse → same as Enter.

#### D. Dismissal

1. With the picker open, press Esc → picker dismisses, omnibar **keeps focus** (typed `@token` stays in the input, caret is preserved, user can keep editing).
2. With the picker open, click outside the omnibar → picker dismisses.
3. With the picker open, switch apps with Cmd-Tab → picker dismisses.
4. With the picker open, press backspace until the `@` is gone → picker dismisses automatically (the detector returns nil once the trigger is removed).
5. Submit the prompt with picker still showing the empty-state row → Enter falls through to normal submit; the picker dismisses as the omnibar clears.

#### E. Activation hook

1. Click into the address bar in Search mode and type `@`.
2. Toggle into Duck.ai mode. The `@` text carries over and the picker opens automatically on activation (the detection re-evaluates after focus + caret restoration).

#### F. Feature-flag gate

1. Turn `aiChatOmnibarAttachMoreTabs` **off**. Type `@` in the duck.ai omnibar → picker does **not** open.
2. Turn the flag on while a picker is open (via debug overrides that apply live) → picker dismisses on the next text edit / selection change.

#### G. Pixels (Debug → Pixels panel)

1. Open the "Add Page Content" submenu → `attach_tabs_picker_shown` fires once. Close without toggling → `attach_picker_canceled` fires.
2. Open submenu, toggle a tab on → `attach_tab_chosen`. Toggle off → `attach_tab_removed`. Close → no cancel.
3. Click × on a tab card in the carousel → `attach_tab_removed` (regardless of how the tab was added).
4. Type `@` → `mention_picker_shown` fires once. Keep typing within the same token → no re-fire. Esc → `mention_picker_canceled`.
5. Type `@`, click an unattached row → `mention_tab_chosen`. Type `@` again, click that same now-attached row → `mention_tab_removed`.
6. Attach a PDF → existing `file_attached`. × on the file card → new `file_removed`.

#### H. Sidebar regression check

1. With the `aiChatSidebarAttachMoreTabs` flag also on, open the duck.ai sidebar and verify its existing in-web tab picker continues to work — the `@`-picker is omnibar-only.

### Impact and Risks

**Low.** Pure addition on top of the merged #4774; no schema or shared-state changes. The picker is gated behind the existing `aiChatOmnibarAttachMoreTabs` flag (off by default), and it shares data with the existing "Add Page Content" submenu, so the attachment flow on submit is identical.

#### What could go wrong?

- **Feature-flag bypass.** Earlier iteration of `updateMentionTokenDetection` didn't check the omnibar tab-picker flag, so the picker would open even when the rollout flag was off. Now gated at the top of detection with an early-return that also dismisses any already-presented picker if the flag flips off remotely while the panel is on screen.
- **Esc handling.** The address bar's existing `escapeKeyDown()` resigns the omnibar's first-responder, which would orphan the picker (and force the user to refocus). Mitigated by a small callback hook (`AddressBarViewController.aiChatOmnibarHandledEscape`) that lets the picker swallow Esc *before* the focus-resign logic runs. With picker closed, Esc behavior is unchanged.
- **Scroll smoothness.** The picker uses an `NSScrollView` + custom row views. Risk areas: per-row tracking areas (we centralized hover on one tracking area on the row stack), layout passes during scroll (cached accessory widths in `init`, layer-backed scroll view + row stack so AppKit composites instead of CG-redrawing). Tested on long lists.
- **Filter result that lands on the same index.** Earlier regression where filtering "many rows, row 0 highlighted" → "one row, also row 0" left the new row unhighlighted because of an equality short-circuit in `setHighlightedIndex`. Fixed by resetting `highlightedIndex = nil` inside `setTabs` before re-applying. Covered by a dedicated test.

### Quality Considerations

- **Unit tests:** `AIChatMentionTokenDetectorTests` covers the `@`-trigger rule across edge cases (start-of-input, after whitespace, mid-token caret, multi-word, two `@`-tokens on a line, newline boundary, out-of-bounds caret). `AIChatMentionPickerFilterTests` covers scoring (title beats URL), stable tie-breaking, no-match, empty-query passthrough. `AIChatMentionPickerViewControllerTests` covers default highlight, arrow-wrap, empty-state mode, the filter-to-same-index regression, and `fittingContentSize` clamping.
- **Strings:** one new key — `aichat.mention-picker.no-matches` ("No matching tabs"). Two existing keys also had their English values updated per Ship Review:
  - `aichat.attach-menu.no-open-tabs`: "No open tabs" → "No page content available"
  - `aichat.attach-menu.image-or-file`: "Add Image or File" → "Add Images or PDFs"
  - Re-translation needed for the two updated values.
- **Pixels:** nine new daily+count pixels added to `aichat_addressbar_pixels.json5` for address-bar tab/file picker engagement, per the [analytics plan](https://app.asana.com/1/137249556945/project/72649045549333/task/1214707945870680?focus=true):
  - Submenu: `attach_tabs_picker_shown`, `attach_tab_chosen`, `attach_tab_removed`, `attach_picker_canceled`
  - `@`-picker: `mention_picker_shown`, `mention_tab_chosen`, `mention_tab_removed`, `mention_picker_canceled`
  - File: `file_removed` (mirrors existing `image_removed`)
  - `attach_tab_removed` fires on the submenu re-toggle path AND the × on a tab card (the carousel can't source-attribute, so it's a uniform tab-removal signal); `mention_tab_removed` stays as the clean signal for the `@`-picker deselect-via-re-toggle path.
  - Existing `aichat_addressbar_submit_with_tabs` is unchanged (raw `tabCount: Int` retained per project decision).
- **Accessibility:** rows use `NSTextField(labelWithString:)` for labels (VoiceOver-readable); tooltip on each row is the tab's URL. The empty-state row is non-interactive.
- **SwiftLint:** clean — including `for_where` / `legacy_nsgeometry_functions` warnings on the mouse-hit loop in `AIChatMentionPickerViewController`, fixed by combining `for … where row.frame.contains(point)` into one expression.

### Notes to Reviewer

- The picker reuses the data layer (`openTabsForOmnibarPicker()`, `toggleTabAttachment(_:)`, `activeTabAttachments`) of the existing "Add Page Content" submenu — there's no parallel state.
- The biggest file is `AIChatMentionPickerCoordinator.swift`: it owns the panel/VC lifecycle, anchoring via `NSTextView.firstRect(forCharacterRange:)`, repositioning on window move/resize, dismissal on resignKey, the splice-on-accept that uses `shouldChangeText` / `didChangeText` so the splice is undoable, and the `mention_*` pixel firings tied to panel-visibility transitions.
- The "Add Page Content" submenu's `attach_tabs_picker_shown` / `attach_picker_canceled` pixels are fired via a small private `AttachTabsSubmenuTracker: NSMenuDelegate` retained on the VC (NSMenu holds its delegate weakly). Toggle-row callbacks set the tracker's "did mutate" flag so cancel only fires when nothing was toggled during the session.
- Chrome material switched from a static `controlBackgroundColor.cgColor` to an `NSVisualEffectView` with `.menu` material so the picker matches the native NSMenu in both light and dark mode — the previous hardcoded `cgColor` snapshotted appearance at `loadView` time and didn't refresh on appearance change.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Medium risk: introduces new AppKit panel lifecycle + keyboard/Esc interception in the address bar path and adds new pixel events, which could affect focus/command handling and attachment UX if edge cases regress.
> 
> **Overview**
> Typing `@` in the Duck.ai omnibar now opens a **non-activating `NSPanel` mention picker** that lists open tabs, live-filters by the token query, supports hover + ↑/↓ navigation, and accepts with Enter/click to toggle the tab attachment while splicing the `@token` out of the prompt (feature-flag gated).
> 
> The omnibar text container and address bar were updated to manage picker presentation/dismissal (caret moves, focus loss, window move/resize, Esc handling) and to keep attachments UX smooth by **scrolling the most recently added carousel item into view** after layout settles.
> 
> Adds new localization strings and **new PixelKit events** for attach-menu sessions, mention-picker engagement, and file removal; the existing “Add Page Content” submenu now fires chosen/removed/canceled pixels and the carousel’s remove actions emit corresponding removal pixels. Includes new unit tests for token detection, filtering/scoring, and picker view-controller behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3e3947c772d9e076b6c2306645b435481244279. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jaceklyp (commented):
Solid feature on top of merged #4774 — coordinator / VC / filter / detector split is clean, pixel set matches the agreed analytics plan, and the Ship Review's open items (AV3/AV4/AV5 + JL1/JL2) are addressed. Two Major findings below: the `track`/`tracking` naming (project privacy convention — multiple sites) and a `DispatchQueue.main.async` timing workaround in the carousel auto-scroll. The rest are Minor / Nit cleanups.

**Cross-cutting note, not anchored anywhere because it's everywhere — the doc-comments are too long.** Multiple `///` blocks across the new files run 3–8 lines and explain context, history, and rationale that belongs in the PR description or a follow-up design doc, not on every declaration. Default to **1–2 lines max**; rely on well-named identifiers for the WHAT and reserve comments for the non-obvious WHY (a hidden constraint, a specific bug worked around). Examples worth tightening on a follow-up pass: the multi-paragraph class doc on `AIChatMentionPickerCoordinator`, the responsibility-list block on `AIChatMentionPickerViewController`, the 12-line scheduling rationale above `scrollLatestAddedAttachmentIntoView`, the rules-list on `AIChatMentionTokenDetector`. Not a merge blocker — fold into a cleanup pass.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1225,6 +1233,13 @@ final class AIChatOmnibarContainerViewController: NSViewController {
         return menu
     }
 
+    /// Tracker installed as the `NSMenuDelegate` of the "Add Page Content" submenu so we can
+    /// fire the picker-shown / picker-canceled pixels exactly once per open/close cycle. The
+    /// row's `onToggle` callback below flips the tracker's `didMutateDuringSession` flag so
+    /// the canceled pixel is only fired when nothing was toggled during the session.
+    /// Retained on the VC because `NSMenu.delegate` is weak.
+    private var attachTabsSubmenuTracker: AttachTabsSubmenuTracker?
```
jaceklyp: **Major:** Let's rename `track` / `tracking` / `tracker` in code (variable, type, function names, and comments).`AttachTabsSubmenuTracker` - Suggested:

```suggestion
    private var attachTabsSubmenuObserver: AttachTabsSubmenuObserver?
```

Also rename the type declaration at line 1807, the alloc at line 1255, the doc-comment at line 1238 ("tracker's `didMutateDuringSession` flag"), and the internal references at lines 1814 / 1823 / 1834.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerViewController.swift
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+}
+
+/// `NSStackView` subclass that owns one `NSTrackingArea` covering its entire bounds and
+/// forwards `mouseMoved:` / `mouseExited:` to the view controller. Centralizing the
+/// tracking here (rather than per-row) eliminates the multi-row-highlighted artifact that
+/// happens when fast trackpad scroll fires `mouseEntered:` on several rows in succession
+/// without the matching `mouseExited:` deliveries.
+private final class HoverTrackingStackView: NSStackView {
```
jaceklyp: **Major:** Same project rule — rename to drop the `Tracking` substring.

```suggestion
private final class HoverObservingStackView: NSStackView {
```

Also update the `rowStack` property's declared type at line 64 to match (`private let rowStack = HoverObservingStackView()`). Class doc-comment above this line can stay; it just talks about `NSTrackingArea` (Apple API — fine).

macOS/DuckDuckGo/AIChat/AIChatMentionPickerCoordinator.swift
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    private func removeWindowObservers() {
+        let center = NotificationCenter.default
+        windowObservers.forEach { center.removeObserver($0) }
+        windowObservers.removeAll()
+    }
+
+    @MainActor
+    private func repositionIfStillTracking(textView: NSTextView) {
```
jaceklyp: **Major:** Same project rule — rename.

```suggestion
    private func repositionIfStillAnchored(textView: NSTextView) {
```

Two call sites to update at lines 328 and 334.

macOS/DuckDuckGo/AIChat/AIChatAttachmentsCarouselView.swift
```diff
@@ -180,9 +185,38 @@ final class AIChatAttachmentsCarouselView: NSView {
...
+    /// card's 62pt height against the still-zero carousel height — producing an
+    /// "unable to simultaneously satisfy" warning and, more practically, a `scrollToVisible`
+    /// that reads stale frames (lands on the next-to-last card).
+    private func scrollLatestAddedAttachmentIntoView(previousIds: Set<String>, currentList: [AIChatPanelAttachment]) {
+        let addedIds = Set(currentList.map(\.attachmentId)).subtracting(previousIds)
+        guard !addedIds.isEmpty,
+              let lastAdded = currentList.last(where: { addedIds.contains($0.attachmentId) })
+        else { return }
+        let lastAddedId = lastAdded.attachmentId
+        DispatchQueue.main.async { [weak self] in
```
jaceklyp: **Major:** This is a timing workaround dressed as an async hop — the docstring above explains it as "the carousel's own height constraint hasn't updated yet at the moment we're inside `setAttachments`." That's a deterministic ordering problem, not an async one.

The container VC owns the reactive sink that flips the carousel between `height == 0` and `height == expandedHeight`. Root-cause fix: expose an explicit `scrollLastAddedAttachmentIntoView()` on the carousel and have the container VC call it **after** it has applied the new height constraint and laid out, rather than the carousel self-scheduling on the next runloop tick.

Minimum acceptable: wrap in `NSAnimationContext.runAnimationGroup` with the scroll in its `completionHandler`, or `superview?.layoutSubtreeIfNeeded()` after the height constraint is known applied — both deterministically wait for layout rather than guessing one runloop tick.

If you keep the `DispatchQueue.main.async`, please document the invariant explicitly: "the container VC must apply the new height constraint within one main-runloop tick of `setAttachments`." Today that's implicit and fragile.

macOS/DuckDuckGo/AIChat/AIChatOmnibarTextContainerViewController.swift
```diff
@@ -289,6 +306,62 @@ final class AIChatOmnibarTextContainerViewController: NSViewController, ThemeUpd
...
+    /// click-outside, Cmd-Tab to another app, etc. The picker should not stay floating
+    /// above an unfocused input.
+    ///
+    /// Esc is intercepted *before* it gets here (see `dismissMentionPickerIfPresented` —
+    /// the address bar's `escapeKeyDown` consults that hook and short-circuits when the
+    /// picker is open so the omnibar's focus is preserved).
+    func textDidEndEditing(_ notification: Notification) {
+        // `?.` — skip the lazy allocation when the user never typed `@`. The coordinator
+        // would no-op anyway, but avoiding the alloc keeps the omnibar's idle path cheap.
+        mentionPickerCoordinator?.dismiss()
```
jaceklyp: **Major:** Two paths can dismiss the picker independently: this `textDidEndEditing` hook, and the coordinator's own `NSWindow.didResignKeyNotification` observer (`AIChatMentionPickerCoordinator.swift:318`). On a cross-window de-focus event (Cmd-Tab away), **both** fire. They're idempotent today — the second `dismiss()` no-ops via `panel.isVisible == false` and the canceled-pixel guard handles it — but the interleaving is order-dependent reasoning future maintainers will have to re-derive.

They also cover subtly different surfaces: this hook catches within-window focus moves (textView → sibling control); the window-observer catches window-level de-focus. Both real cases, but the design is implicit.

Fold into one source of truth. Either (a) move the within-window signal into the coordinator (e.g. observe `NSTextDidEndEditingNotification` for the textView there) and drop this hook, or (b) make `dismiss(reason:)` explicit and have both paths pass their reason — then the canceled-pixel guard reads off the reason instead of inferring from `wasPresented` / `didAcceptDuringPresentation`.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerCoordinator.swift
```diff
@@ -0,0 +1,355 @@
...
+        // user's action (chosen vs removed) — the toggle flips it, so post-toggle the
+        // signal would be inverted.
+        let wasAttached = omnibarController.activeTabAttachments.contains(where: { $0.id == attachment.id })
+        // Set the accept flag BEFORE `spliceTokenFromTextView()` — the splice's
+        // `didChangeText` posts `NSText.didChangeNotification` synchronously, which
+        // cascades into the omnibar's `textDidChange → updateMentionTokenDetection →
+        // dismiss()` path. If the flag isn't already set when that re-entrant `dismiss`
+        // runs, it sees `didAcceptDuringPresentation == false` and erroneously fires the
+        // `mention_picker_canceled` pixel alongside the chosen/removed pixel below.
+        didAcceptDuringPresentation = true
```
jaceklyp: **Minor:** Commit `aee4ce2605` correctly fixed the canceled-pixel double-fire by setting the flag *before* the splice's re-entrant dismiss. Verified the ordering — this is fine as written. But the **class of bug** ("did I flip the flag early enough?") will keep recurring as long as `accept` triggers a re-entrant `dismiss` via `didChangeText` → `updateMentionTokenDetection` → `dismiss`.

Eliminate the re-entrance instead of guarding against it. Add `isSplicing` on the coordinator, set/clear it around `spliceTokenFromTextView`, and early-return from `updateMentionTokenDetection` / `dismiss` while it's true. Then `accept` becomes:

```swift
@MainActor
private func accept(attachment: AIChatTabAttachment) {
    let wasAttached = omnibarController.activeTabAttachments.contains { $0.id == attachment.id }
    spliceTokenFromTextView()                  // sets/clears isSplicing internally
    omnibarController.toggleTabAttachment(attachment)
    PixelKit.fire(wasAttached ? .aiChatAddressBarMentionTabRemoved
                              : .aiChatAddressBarMentionTabChosen,
                  frequency: .dailyAndCount, includeAppVersionParameter: true)
    dismiss()                                   // single dismiss, no pre-flag
}
```

Removes the load-bearing 6-line comment above and the whole "where exactly must the flag be set" class of bugs.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerCoordinator.swift
```diff
@@ -0,0 +1,355 @@
...
+        let attachedIds = Set(omnibarController.activeTabAttachments.map(\.id))
+        let currentTabId = omnibarController.currentTabUUID
+        let openTabs = omnibarController.openTabsForOmnibarPicker()
+        let filtered = AIChatMentionPickerFilter.filter(openTabs, query: token.query, currentTabId: currentTabId)
+        viewController.setTabs(
+            filtered,
+            currentTabId: currentTabId,
+            attachedTabIds: attachedIds
+        )
+        viewController.onAccept = { [weak self] attachment in
```
jaceklyp: **Minor:** Reassigned on every `presentIfNeeded` call (one per keystroke while the user types the query). Set once at construction:

```swift
private func ensureViewController() -> AIChatMentionPickerViewController {
    if let viewController { return viewController }
    let vc = AIChatMentionPickerViewController()
    vc.onAccept = { [weak self] attachment in self?.accept(attachment: attachment) }
    _ = vc.view
    self.viewController = vc
    return vc
}
```

Then drop lines 116–118 from `presentIfNeeded`.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerCoordinator.swift
```diff
@@ -0,0 +1,355 @@
...
+        if let parent = attachedToWindow {
+            parent.removeChildWindow(panel)
+        }
+        attachedToWindow = nil
+        removeWindowObservers()
+        panel.orderOut(nil)
+        lastTokenRange = nil
+        anchoredTextView = nil
+
+        if shouldFireCanceled {
```
jaceklyp, line 192: **Minor:** Pixel-attribution corner case worth confirming intentional. Sequence:

1. User types `@nope` — no matches, empty-state row visible.
2. User presses Enter.
3. `acceptHighlighted()` returns `false` (empty state); `doCommandBy` falls through and the omnibar's `submit()` runs.
4. `submit()` likely resigns first responder → `textDidEndEditing` → `dismiss()` → `didAcceptDuringPresentation == false` → `mention_picker_canceled` fires here.

So a user who actually **submitted** a prompt while the picker was on screen counts as a "cancel" in pixel terms. Two reasonable readings:
- Cancel = "didn't pick a tab" → today's behaviour is correct; document it in the JSON5 description so analytics readers don't mis-interpret.
- Cancel = "explicitly dismissed without acting" → need a distinct "fell-through to submit" signal, and `acceptHighlighted` falling through should mark `didAcceptDuringPresentation = true` (or a similar bit) before the dismiss runs.

Pick one; today the semantics aren't documented.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerFilter.swift
```diff
@@ -0,0 +1,89 @@
...
+        // *after* scoring so the current tab only appears when it actually matches the query
+        // (e.g. user types `@goog` on a non-Google current tab → "(Current Tab)" stays
+        // hidden because it doesn't match). When the current tab matches, the user expects
+        // it to be the obvious first choice regardless of score order.
+        guard let currentTabId,
+              let currentIndex = result.firstIndex(where: { $0.id == currentTabId }),
+              currentIndex != 0 else {
+            return result
+        }
+        var reordered = result
```
jaceklyp: **Minor:** `remove(at:)` + `insert(at: 0)` rebuilds the array with a shift. Single-allocation equivalent:

```swift
return [result[currentIndex]] + result.prefix(currentIndex) + result.suffix(from: currentIndex + 1)
```

(Drops the next three lines too.) Functionally identical; matters only at large tab counts.

macOS/DuckDuckGo/AIChat/AIChatMentionTokenDetector.swift
```diff
@@ -0,0 +1,124 @@
...
+        // Within that line range, scan backwards from the caret looking for an `@` whose
+        // preceding character is start-of-input OR whitespace. Whichever we find first is
+        // the active token's trigger.
+        var index = caret - 1
+        while index >= lineStart {
+            if nsString.character(at: index) == Self.atSign {
+                let isAtLineStart = (index == 0)
+                let isPrecededByWhitespace = index > 0 && Self.isWhitespace(nsString.character(at: index - 1))
+                if isAtLineStart || isPrecededByWhitespace {
+                    let queryRange = NSRange(location: index + 1, length: caret - index - 1)
```
jaceklyp, line 79: **Minor:** `caret` is passed in as `selectedRange.location` (selection start), not `selectedRange.upperBound`. So with selection `@foo|bar|` (where `|...|` marks the user's selection over `bar`), the detector returns `query = "foo"` and the splice removes just `@foo` — leaving `bar` selected. Inconsistent with Enter's normal "collapse selection" semantics.

Either fix at the call site (have the container VC pass `selectedRange.length > 0 ? selectedRange.upperBound : selectedRange.location`), or document the trade-off here so the next reader doesn't change it accidentally.

macOS/DuckDuckGo/AIChat/AIChatOmnibarTextContainerViewController.swift
```diff
@@ -319,6 +392,40 @@ final class AIChatOmnibarTextContainerViewController: NSViewController, ThemeUpd
...
+            case #selector(insertNewline(_:)),
+                 #selector(insertNewlineIgnoringFieldEditor(_:)):
+                if coordinator.acceptHighlighted() {
+                    return true
+                }
+                // No real selection — fall through to the normal Enter path below.
+            default:
+                break
+            }
+        } else if let coordinator = mentionPickerCoordinator,
```
jaceklyp: **Minor:** This `else if` duplicates the `cancelOperation` case from the main branch above (line 408). Easier to read as one branch keyed off `isPresented`, with `canHandleKeyCommands` gating only the keys that need a real selection:

```swift
if let coordinator = mentionPickerCoordinator, coordinator.isPresented {
    switch commandSelector {
    case #selector(NSResponder.cancelOperation(_:)):
        coordinator.dismiss()
        return true
    case #selector(NSResponder.moveDown(_:)) where coordinator.canHandleKeyCommands:
        coordinator.moveHighlightDown(); return true
    case #selector(NSResponder.moveUp(_:)) where coordinator.canHandleKeyCommands:
        coordinator.moveHighlightUp(); return true
    case #selector(insertNewline(_:)), #selector(insertNewlineIgnoringFieldEditor(_:)):
        if coordinator.canHandleKeyCommands, coordinator.acceptHighlighted() { return true }
    default:
        break
    }
}
```

macOS/DuckDuckGo/AIChat/AIChatMentionPickerViewController.swift
```diff
@@ -0,0 +1,425 @@
...
+    private var currentTabIdForWidthMeasurement: String?
+
+    // MARK: - Hover (centralized)
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        rowStack.onMouseMoved = { [weak self] pointInStack in
+            self?.updateHighlightFromMousePosition(pointInStack)
+        }
+        rowStack.onMouseExited = { [weak self] in
```
jaceklyp: Nit: This closure binds `self` weakly only to discard the binding (`_ = self`). The intent ("don't clear highlight on exit — keyboard nav may want to keep it") is fine, but a no-op closure reads more confusingly than just not assigning the callback — `HoverTrackingStackView` already no-ops on `nil`.

```swift
// Don't set onMouseExited — keyboard nav may want to keep the current highlight visible
// even after the cursor leaves the panel. Highlight only resets on setTabs(...).
```

Drop the assignment block; the absence of the call site is the contract.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerViewController.swift
```diff
@@ -0,0 +1,425 @@
...
+    /// If `tabs` is empty, renders the "No matching tabs" empty-state row instead — the
+    /// picker stays on screen so the user can keep typing (the next keystroke that drops
+    /// the filter or matches something will repopulate the rows).
+    func setTabs(_ tabs: [AIChatTabAttachment], currentTabId: String?, attachedTabIds: Set<String>) {
+        currentTabIdForWidthMeasurement = currentTabId
+
+        // Rebuild rows. For now this is a simple "tear down + rebuild" — measured to be fine
+        // for typical N (≈ open tabs in a window). If we ever see jank with hundreds of tabs
+        // we can move to NSTableView diffing.
+        rowViews.forEach { $0.removeFromSuperview() }
```
jaceklyp, line 237: Nit: Full row tear-down + rebuild on every keystroke. The comment above already calls this out and it's fine for typical N. Flagging as a follow-up only — if profiling ever shows jank with 50+ tabs, switching to an `NSTableView` (or a manual diff) avoids re-allocating two `NSTextField`s + an `NSImageView` per row per keystroke.

macOS/DuckDuckGo/AIChat/AIChatMentionPickerCoordinator.swift
```diff
@@ -0,0 +1,355 @@
...
+        guard isPresented, let vc = viewController else { return false }
+        return !vc.isShowingEmptyState
+    }
+
+    init(omnibarController: AIChatOmnibarController) {
+        self.omnibarController = omnibarController
+    }
+
+    deinit {
+        // Drop the notification-center registrations synchronously — they don't touch UI.
```
jaceklyp: Nit: The comment isn't quite right — the observer blocks **do** touch UI (`Task @MainActor { self?.dismiss() }`). What's actually safe is that after `removeObserver`, the blocks won't *fire* again, so the MainActor hop isn't needed during deinit. Tighten:

```swift
// After removeObserver, the observer blocks won't fire again, so no MainActor dismiss hop
// is needed here. The panel itself is owned by its parent window (via addChildWindow);
// when the parent closes, AppKit tears the child down.
```

macOS/DuckDuckGo/AIChat/AIChatMentionPickerViewController.swift
```diff
@@ -0,0 +1,425 @@
...
+        let chromeHeight = headerHeight + rowsHeight + Layout.outerInset
+        let totalHeight = chromeHeight + (Layout.outerInset * 2)
+        return NSSize(width: width, height: totalHeight)
+    }
+
+    /// Convenience: derives the content size from the current row count and a width that
+    /// adapts to the longest title currently shown. The empty-state row uses `minWidth`.
+    var fittingContentSize: NSSize {
+        let isEmpty = rowViews.isEmpty
+        let effectiveCount = isEmpty && isShowingEmptyState ? 1 : rowViews.count
```
jaceklyp: Nit: Reads two state fields where one suffices. `isEmpty` derives from `rowViews.isEmpty` and `isShowingEmptyState` derives from `emptyRowView != nil`; in practice `setTabs` keeps them in sync, so the combined check is redundant.

```suggestion
        let effectiveCount = isShowingEmptyState ? 1 : rowViews.count
```

Keep `isEmpty` on line 332 — it's still used by the `contentWidth` branch below.

macOS/DuckDuckGo/AIChat/AIChatMentionTokenDetector.swift
```diff
@@ -0,0 +1,124 @@
...
+            let previous = nsString.character(at: lineStart - 1)
+            if Self.isLineBoundary(previous) { break }
+            lineStart -= 1
+        }
+
+        // Within that line range, scan backwards from the caret looking for an `@` whose
+        // preceding character is start-of-input OR whitespace. Whichever we find first is
+        // the active token's trigger.
+        var index = caret - 1
+        while index >= lineStart {
```
jaceklyp, line 74: Nit: Backward scan from caret is O(line length). Irrelevant for prompt-sized input; a future reader pasting a multi-thousand-character prompt with `@x` at the end would walk the whole line per keystroke. Either cap the scan distance:

```swift
let scanFloor = max(lineStart, caret - 1024)
while index >= scanFloor { ... }
```

or add a one-line comment noting the bound so the next reader doesn't reach for the cap reflexively.

jaceklyp (approved):

tomasstrba:
Pushed [`116c949d98`](https://github.com/duckduckgo/apple-browsers/commit/116c949d98) addressing the review comments. Summary of what changed vs. what was intentionally deferred:

### Applied

**Major**
- **`track`/`tracking` renames** (JL #7-#9): `AttachTabsSubmenuTracker` → `AttachTabsSubmenuObserver`, `HoverTrackingStackView` → `HoverObservingStackView`, `repositionIfStillTracking` → `repositionIfStillAnchored`. Apple-API mentions (`NSTrackingArea`) left as-is.
- **Carousel root-cause fix** (JL #10): dropped `DispatchQueue.main.async` from `scrollLatestAddedAttachmentIntoView`. `setAttachments` now stashes `lastAddedAttachmentId`; the container VC calls `superview?.layoutSubtreeIfNeeded()` then `scrollLastAddedAttachmentIntoView()` after the height constraint flips to `expandedHeight`. Also hooked into `menuDidClose` so the deferred-layout (submenu) path scrolls too.
- **Explicit `dismiss(reason:)`** (JL #11): added a `DismissReason` enum (`accept`, `userEscape`, `textEndedEditing`, `windowResignKey`, `tokenGone`, `featureFlagOff`); the canceled-pixel guard now reads `reason != .accept`. Removed the `didAcceptDuringPresentation` flag and its 6-line ordering comment along the way.

**Minor + Nit**
- **JL #13** — `onAccept` is now assigned once in `ensureViewController()` instead of on every `presentIfNeeded` keystroke.
- **JL #14** — documented the empty-state Enter → cancel-pixel semantics in the `mention_picker_canceled` JSON5 description (kept current behavior).
- **JL #15** — single-allocation rewrite of the current-tab pin in `AIChatMentionPickerFilter`.
- **JL #16** — `caret` now uses `selection.upperBound` when the selection has length, so the splice sweeps the full `@token`. Fixed at the two call sites (`updateMentionTokenDetection` + `spliceTokenFromTextView`); the detector contract is unchanged.
- **JL #17** — folded the duplicated `cancelOperation` branch in `doCommandBy` into a single `isPresented` switch with `where coordinator.canHandleKeyCommands` clauses on the arrow / Enter cases.
- **JL #18** — dropped the no-op `onMouseExited = { _ = self }` block; replaced with a one-line comment.
- **JL #20** — rewrote the deinit comment per your suggested wording.
- **JL #21** — `effectiveCount` in `fittingContentSize` no longer reads both `isEmpty` and `isShowingEmptyState`.
- **JL #22** — added a one-line note that the backward scan is O(line length), bounded by `lineStart`.

### Cursor Bugbot — already resolved on this branch (no action)

The six earlier Bugbot comments are all addressed by commits already on the branch: feature-flag gate (`eab9c6f2af`), orphan xcstrings (never existed on HEAD), unused `backgroundView` (it's wired up), unused `rowCount` (`d95770ab62`), unnecessary lazy alloc on dismiss (`e88fafba4f`), spurious canceled-pixel double-fire (`aee4ce2605`).

### Intentionally deferred

- **JL #12** (`isSplicing` flag refactor) — the dismiss(reason:) refactor in #11 already eliminates the load-bearing 6-line comment by removing the flag dance entirely; `accept(attachment:)` now calls `dismiss(reason: .accept)` first so any re-entrant `dismiss(reason: .tokenGone)` from the splice is a panel-already-gone no-op.
- **JL #19** (NSTableView migration for very long tab lists) — flagged as follow-up only; not addressed.
- **Cross-cutting doc-comment shortening** — the 12-line `scrollLatestAddedAttachmentIntoView` block went away with the root-cause fix above. The other long blocks (`AIChatMentionPickerCoordinator` class doc, `AIChatMentionPickerViewController` responsibility list, `AIChatMentionTokenDetector` rules) are left for the follow-up cleanup pass you mentioned.

### Verification

- `xcodebuild -scheme "macOS Browser" -configuration Debug` ✅
- `xcodebuild -scheme "macOS Unit Tests" -only-testing:AIChatMentionTokenDetectorTests / -only-testing:AIChatMentionPickerFilterTests / -only-testing:AIChatMentionPickerViewControllerTests` → 42 tests, 0 failures ✅

macOS/DuckDuckGo/AIChat/AIChatMentionPickerCoordinator.swift
```diff
@@ -0,0 +1,366 @@
...
+    /// dismisses the picker.
+    @MainActor
+    private func accept(attachment: AIChatTabAttachment) {
+        // The splice's `didChangeText` posts `NSText.didChangeNotification` synchronously,
+        // which cascades into `updateMentionTokenDetection → dismiss(reason: .tokenGone)`.
+        // Dismiss the picker explicitly with `.accept` first so that re-entrant dismiss
+        // becomes a no-op (panel already torn down) instead of firing the canceled pixel.
+        let wasAttached = omnibarController.activeTabAttachments.contains(where: { $0.id == attachment.id })
+        dismiss(reason: .accept)
+        spliceTokenFromTextView()
```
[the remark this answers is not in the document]
tomasstrba: Good catch — this is a real bug introduced by my dismiss-before-splice reordering. Fixed in [`b3e3947c77`](https://github.com/duckduckgo/apple-browsers/commit/b3e3947c77): restored splice-first ordering and added the `isSplicing` flag (the JL #12 recommendation I'd originally deferred), so the splice's re-entrant `dismiss(reason: .tokenGone)` is a no-op while splicing. The final explicit `dismiss(reason: .accept)` then runs after the splice + toggle + pixel.

tomasstrba:
Addressed the comments and executed the testing steps again. This is not going to be enabled by default, only for people who opt-in. Moving forward with the merge

PR #4857, opened 2026-05-13
Restore last-used chat model in UTI

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1214718692771943?focus=true
Tech Design URL:
CC:

### Description
Adds a `DuckAiLastUsedModelProviding` protocol + implementation that reads the persisted chat's `model` field via the existing `DuckAiChat.decode` parser and applies it to the iPad UTI's model store. The coordinator restores the last-used model on tab bind for existing chats, and re-applies it whenever the JS bridge writes the chat via `putChat`, so the toolbar (model chip, reasoning, tools) reflects the model used mid-conversation. Parse and storage failures surface a new `duckAiNativeStorageLastUsedModelParseError` pixel and never crash.

### Testing Steps
1. Open an existing duck.ai chat from history that was last used with a non-default model (e.g. `meta-llama/Llama-4-Scout-17B-16E-Instruct`) — the UTI's model chip should match that model.
2. Within the same chat, switch the model from the JS UI and confirm the toolbar updates without re-binding the tab.
3. Open a chat with no assistant reply yet, or a brand-new tab — the UTI should fall back to the persisted preference (no error pixel).

### Impact and Risks
Low

#### What could go wrong?
Wrong model surfaced on the UTI for a tab if `duckAIChatID` extraction misreads a non-standard URL, in which case the existing per-tab state fallback still applies.

---

> [!NOTE]
> **Medium Risk**
> Changes UTI model-selection behavior based on persisted chat data and adds a new cross-layer event/pixel path; bugs could surface the wrong model or cause extra storage reads, but failures are handled non-fatally.
> 
> **Overview**
> **UTI now restores the per-chat last-used model from native storage** when binding an existing Duck.ai chat tab, and re-applies it when the chat is updated via the native-storage JS bridge so the model chip/tools stay aligned with FE model changes.
> 
> Adds `DuckAiLastUsedModelProviding` + `DuckAiLastUsedModelProvider` to read the persisted chat’s top-level `model` field, makes `DuckAiChat.decode` `throws` (callers updated), and introduces a `chatUpdatesPublisher` on `DuckAiNativeStorageUserScript` plus a small in-memory cache in `UnifiedToggleInputCoordinator`.
> 
> Introduces a new parse-error telemetry path (`lastUsedModelParseError`) wired into iOS pixels (and defined in `aichat_native_storage_pixels.json5`), with macOS currently ignoring this event.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64bbcc0dd8191bd22d7ea255b70a0a4c9cad169b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
```diff
@@ -510,6 +510,12 @@ private extension MainViewController {
     func bindAITabIfPossible(tab: TabViewController, coordinator: UnifiedToggleInputCoordinator, hasExistingChat: Bool) {
         if let userScript = tab.userScripts?.aiChatUserScript {
             coordinator.bindToTab(userScript, hasExistingChat: hasExistingChat)
+            if hasExistingChat, let chatID = tab.webView.url?.duckAIChatID {
+                coordinator.restoreLastUsedModel(forChatID: chatID)
```
aataraxiaa, line 514: Agent, non-blocking -> This does a synchronous GRDB read. Fine when blob is small but might impact tab swiping etc. Could add a last-result cache for re-binds.

aataraxiaa (approved):
Tested and LGTM. One non-blocking comment.

iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
Bunn: Fixed on 65ca108137a4fba68d7aa71e943c15c70df5b34b

PR #4860, opened 2026-05-13
[macOS] Improve voice chat permission handling for Duck.ai

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1214808879504912?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1214511463798723?focus=true
CC:

### Description

Move Duck.ai voice chat microphone permission handling from the web layer to native on macOS. Instead of relying on the Duck.ai front-end to render a consent modal and prompt the OS, the native browser now:

- **Read-path override** (`DuckAiVoiceChatPermissionOverride`): forces `duck.ai` microphone to `.allow` at read time via a new `PermissionDecisionOverriding` slot on `PermissionManager`. Nothing is written to storage — turning the flag off restores the user's real persisted decision (`.deny`/`.ask`/none) with no migration footprint.
- **Legacy consent cleanup** (`DuckAiVoiceChatLegacyConsentCleanup`): one-shot at app init that clears the FE-side `hasVoiceModeConsent` for users who previously persisted `.deny`. Avoids the case where the FE believes consent was given for a capability the user had blocked. Skipped on subsequent launches via a `UserDefaults` flag.
- **OS-denial remediation** (`DuckAiVoiceChatFailureHandler` → `SystemDisabledPermissionInfoView`): when the FE posts `voiceChatStartFailed` with `NotAllowedError` AND the OS microphone is denied, native surfaces the standalone "System microphone disabled. Turn it on in System Settings → Privacy" popover anchored to the address-bar shield. The shield stays visible for as long as the OS-denied state holds (`isDuckAiVoiceChatSystemMicDenied`), so the user can re-open the popover after dismissing it.
- **Permission Center scoping**: the mic row is stripped from the Permission Center on `duck.ai` while the flag is on, since any edit would be silently re-masked by the override. The OS-denied remediation surface is the standalone popover, not a row.
- **Address-bar shield suppression**: the shield is hidden on `duck.ai` when only mic is in play, so users don't see a stale shield for the auto-granted mic. The OS-denied derived check overrides the suppression when remediation is needed. A pending `permissionAuthorizationQuery` also overrides it (covers the flag-off case where the standard permission popover needs an anchor).
- **Debug URL honoring**: `URL.duckAi` honors the AI Chat Debug menu's custom URL on internal builds, so every host check (override scoping, suppression, cleanup) follows staging/test environments.

Everything is gated behind the new `aiChatNativeVoicePermissionFlow` feature flag (subfeature `AIChatSubfeature.nativeVoicePermissionFlow`). The capability is advertised to the front-end via the existing `getAIChatNativeConfigValues` payload (new `supportsNativeVoicePermissionHandler` field — `true` on macOS when the flag is on, `false` on iOS). When the flag is off, the override returns nil, the standalone popover is never opened by this flow, and the old web-driven flow is unchanged.

New pixel:
- `m_mac_aichat_voice_chat_start_failed` (param `reason`: `mic_os_denied` / `other`) — fired when Duck.ai reports a voice-chat startup failure, segmented by whether we surfaced the OS-denied popover or ignored it.

Screenshot 2026-05-15 at 3 54 00 PM

### Testing Steps

**Helpful commands to reset state between cases:**
- Reset OS microphone authorization for the debug build: `tccutil reset Microphone com.duckduckgo.macos.browser.debug`
- Clear per-site permission state and FE consent: `macOS/clean.sh debug`

1. **Legacy consent cleanup for previously-denied user** (exercises the cleanup path):
   1. Launch debug build with `aiChatNativeVoicePermissionFlow` **disabled**.
   2. Open `duck.ai`, start a voice chat, accept the FE consent modal (this sets `hasVoiceModeConsent = true`).
   3. In Permission Center for `duck.ai`, set **Microphone → Never Allow**.
   4. Enable `aiChatNativeVoicePermissionFlow`. Quit and relaunch.
   5. Start a new voice chat from any entry point. **Verify**:
      - FE consent modal re-appears (cleanup cleared `hasVoiceModeConsent`).
      - After accepting, voice chat starts immediately with no browser-level mic prompt (the override masks the persisted `.deny` to `.allow`).
      - No `m_mac_aichat_voice_chat_*` pixel is fired in the success path.
2. **OS allowed path**: System Settings → Privacy → Microphone → enable DuckDuckGo. Voice chat works, no shield/permission-center button on `duck.ai`.
3. **OS denied path**: System Settings → Privacy → Microphone → disable DuckDuckGo. Start voice chat. Verify the standalone "Duck.ai voice chat requires microphone access. / System microphone disabled. Turn it on in System Settings → Privacy" popover anchored to the shield. Tap the link → opens System Settings → Privacy → Microphone. Pixel `aichat_voice_chat_start_failed` fires with `reason=mic_os_denied`. Dismiss the popover — shield should stay; tap shield to re-open the same popover.
4. **Override scoping**: visit a non-`duck.ai` mic site (e.g. `https://microphone-test.com/`). Verify the mic row in Permission Center is editable, OS-denied surfaces use the regular inline-yellow row, and the override does not mask anything.
5. **Feature flag off**: disable `aiChatNativeVoicePermissionFlow`. FE falls back to its own modal flow (`supportsNativeVoicePermissionHandler=false`). Persisted mic state surfaces normally in Permission Center, no shield suppression on `duck.ai`. If a pending getUserMedia happens, the standard authorization popover anchors to the shield (the shield is force-shown for that duration via the `permissionAuthorizationQuery` predicate).

### Impact and Risks

**Medium** — changes touch the per-site permission read path (override masks duck.ai/mic), the Permission Center UI, and the address-bar button visibility logic. All changes are gated behind a feature flag and scoped to the `duck.ai` host (or the Debug-overridden host).

#### What could go wrong?
- **Override leak onto non-Duck.ai hosts** — could in principle force `.allow` for mic on the wrong host. Mitigated by `DuckAiVoiceChatPermissionOverride.decision(forDomain:permissionType:)` early-returning unless `permissionType == .microphone`, `domain == aiChatURL.host`, AND the flag is on. Unit test coverage in `DuckAiVoiceChatPermissionOverrideTests`.
- **Permission Center row leaks** — the OS-denied warning could surface on non-Duck.ai sites under the new flag. Mitigated by the duck.ai-scoped filter in `PermissionCenterViewModel.collectPermissions` (matches `domain == URL.duckAi.host` AND flag on) gating the mic-strip; non-Duck.ai sites keep their regular row + inline-yellow warning. Unit test coverage in `PermissionCenterViewModelTests`.
- **Address-bar shield disappears on Duck.ai when another permission is needed** — could leave the user with no way to manage a non-mic permission. Mitigated by `shouldSuppressShieldOnDuckAi` only suppressing when `persistedPermissionTypes` and `usedPermissions` contain *no* permission other than `.microphone`.
- **Custom URL override leaks to production** — `URL.duckAi`'s debug override is gated on `UserDefaults.appConfiguration.isInternalUser`. Non-internal builds always resolve to `https://duck.ai`.
- **Popover anchored to a hidden view** — historically crashed at `NSPopoverExtension.swift:85`. Fixed by `isDuckAiVoiceChatSystemMicDenied` keeping the shield visible whenever the OS-denied state holds, plus `layoutSubtreeIfNeeded()` before anchoring in `handleAIChatVoiceChatPermissionRequested`.
- **Front-end sends `voiceChatStartFailed` before flag is rolled out** — `AIChatUserScriptHandling.voiceChatStartFailed` early-returns when the flag is off. Safe to roll out FE first.

### Quality Considerations

- **Edge cases**:
  - Mic permission already `.allow`: override is a no-op (it still returns `.allow`, same as the storage would).
  - Legacy consent cleanup with no native storage handler: skipped silently — the override still masks the persisted `.deny` regardless.
  - OS state `.systemDisabled` (audio service off entirely): no remediation surface — system-disabled isn't actionable from System Settings → Privacy → Microphone.
  - Failure reason other than `NotAllowedError` (e.g. `AbortError`, no device): we record the `other` pixel and do not surface a UI — front-end keeps existing handling.
  - Multiple rapid `voiceChatStartFailed` messages: handler dedupes by checking the popover isn't already shown.
- **Performance**: override is an O(1) string/enum compare on every permission read for `.microphone` on `duck.ai`; nothing else is touched. Legacy cleanup runs once at launch and short-circuits via UserDefaults flag.
- **Monitoring**: `aichat_voice_chat_start_failed` lets us detect spikes in `reason=other` (potential regressions in the FE/native contract) and confirm the `mic_os_denied` remediation path is reached.
- **Privacy/security**: no new data is collected. The legacy cleanup writes only to local `PermissionManager` and `UserDefaults`. No PII in pixels (only `reason` enum).
- **Documentation**: tech design is at `macOS-voice-chat-permission-tech-design.md` (local, not in repo).

### Notes to Reviewer

- `AddressBarButtonsViewController.updatePermissionCenterButton` has three branches: (1) `isDuckAiVoiceChatSystemMicDenied(forDomain:)` → force-show (anchors `SystemDisabledPermissionInfoView`); (2) `shouldSuppressShieldOnDuckAi` → hide when only mic is in play on duck.ai; (3) normal `shouldShowPermissionCenterButton` path, with a `hasPendingAuthorizationQuery || !isAIChatPanelActive` predicate so the standard auth popover still has an anchor when the native flag is off.
- `URL.duckAi` honoring the Debug menu's custom URL is a small but load-bearing change — every host check in this PR (override scoping, suppression logic, cleanup) derives from `URL.duckAi.host`, so the debug override is what makes staging testing actually work.
- The `supportsNativeVoicePermissionHandler` rename (was `supportsNativeVoiceChatPermissionPrompt`) was Moon's request to match the FE-side naming in the existing `getAIChatNativeConfigValues` contract.
- New files (need confirmation that pbxproj entries are correct):
  - `macOS/DuckDuckGo/AIChat/DuckAiVoiceChatPermissionOverride.swift`
  - `macOS/DuckDuckGo/AIChat/DuckAiVoiceChatLegacyConsentCleanup.swift`
  - `macOS/DuckDuckGo/AIChat/DuckAiVoiceChatFailureHandler.swift`
  - `macOS/UnitTests/AIChat/DuckAiVoiceChatPermissionOverrideTests.swift`
  - `macOS/UnitTests/AIChat/DuckAiVoiceChatLegacyConsentCleanupTests.swift`
  - `macOS/UnitTests/AIChat/DuckAiVoiceChatFailureHandlerTests.swift`

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Modifies the per-site permission read path and address-bar permission UI behavior on macOS (feature-flagged and duck.ai-scoped), so regressions could affect microphone/permission center handling if the override or scoping is wrong.
> 
> **Overview**
> **Moves Duck.ai voice-chat microphone permission handling on macOS from FE-driven UI to native, gated by the new `nativeVoicePermissionFlow`/`aiChatNativeVoicePermissionFlow` flag.**
> 
> Native now (a) intercepts `PermissionManager.permission(...)` via a new `PermissionDecisionOverriding` hook to force `duck.ai` microphone to `.allow` without writing to storage, (b) performs a one-time cleanup of stale FE consent in native storage for users who previously persisted mic `.deny`, and (c) handles FE-reported `getUserMedia` failures via a new `voiceChatStartFailed` JS-bridge message, showing a system-microphone-disabled popover anchored to the address-bar shield when OS mic access is denied.
> 
> This updates address-bar shield visibility/click-routing logic specifically for duck.ai (hide shield when only mic is in play; force-show when OS-mic denied; keep an anchor for pending auth queries), expands system-permission querying to include microphone, adds a new pixel `aichat_voice_chat_start_failed` with reason segmentation, and introduces new localized strings for microphone permission messaging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eae3741c3f67c1b09d3e2de083563592293c3bd7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/AIChat/Sources/AIChat/Shared/AIChatScriptUserValues.swift
```diff
@@ -127,7 +133,8 @@ public struct AIChatNativeConfigValues: Codable {
                                         supportsOpenAIChatLink: true,
                                         supportsAIChatSync: false,
                                         supportsMultipleContexts: false,
-                                        supportsNativeStorage: false)
+                                        supportsNativeStorage: false,
+                                        supportsNativeVoicePermissionHandler: true)
```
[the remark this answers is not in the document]
jotaemepereira: This is not correct.

macOS/DuckDuckGo/AIChat/DuckAiVoiceChatFailureHandler.swift
```diff
@@ -0,0 +1,140 @@
...
+            pixelFiring?.fire(
+                AIChatPixel.aiChatVoiceChatStartFailed(reason: .other),
+                frequency: .dailyAndCount
+            )
+            return
+        }
+
+        // Dedupe: skip if the popover is already on screen for this webView (or any, since
+        // the popover is window-scoped). Avoids stacking when the user mashes the voice button.
+        guard !permissionCenterPresenter.isPermissionCenterPresented(for: sourceWebView) else { return }
```
tomasstrba, line 86: The handler's dedupe is wired with `isPresentedProvider: { _ in false }` in production (`AIChatUserScriptHandling.swift:153`), so this `guard` only ever fires in tests. The real dedupe happens at the address-bar observer — which means rapid failures fire multiple `mic_os_denied` pixels even though only one popover is shown. Worth either letting the presenter query the address bar, or moving the pixel firing behind the actual "did present" callback so prod and tests agree.

macOS/DuckDuckGo/AIChat/DuckAiVoiceChatPermissionMigration.swift
```diff
@@ -0,0 +1,82 @@
...
+        let isPersisted = permissionManager.hasPermissionPersisted(forDomain: host, permissionType: .microphone)
+
+        switch permissionManager.permission(forDomain: host, permissionType: .microphone) {
+        case .deny:
+            // `aiChatNativeStorage` is at 100% rollout in production, so `storageHandler`
+            // is effectively always non-nil here. In dev builds with that flag off the
+            // FE's `hasVoiceModeConsent` lives in localStorage out of native's reach,
+            // and the consent-clear becomes a no-op — accepted since it can't reach prod.
+            try? storageHandler?.deleteEntry(key: Self.voiceModeConsentKey)
+            permissionManager.setPermission(.allow, forDomain: host, permissionType: .microphone)
```
tomasstrba: Minor ordering nit: if the app dies between `deleteEntry` and `setPermission`, the user has a cleared FE consent flag with a still-`.deny` per-site permission. Re-running on next launch recovers (worst case: extra consent modal shown once), so not a bug — but flipping the order (permission first, then consent clear) would make the recovery state identical to "never ran".

macOS/DuckDuckGo/Permissions/ViewModel/PermissionCenterViewModel.swift
```diff
@@ -515,15 +524,30 @@ final class PermissionCenterViewModel: ObservableObject {
         }
     }
 
-    private func buildPermissionItem(for permissionType: PermissionType) -> PermissionCenterItem {
+    private func buildPermissionItem(for permissionType: PermissionType) -> PermissionCenterItem? {
+        let isLocked = isDuckAiVoiceMicrophoneRow(domain: domain, permissionType: permissionType)
```
tomasstrba: The comment justifies `.authorized` and `.notDetermined` but not `.systemDisabled`, which is also in the case. The PR description explains it ("avoid a confusing 'Off in System Settings' link that won't help") — worth pulling that one-liner into the source comment.

tomasstrba:
TS1/ **Suggestion: replace the `.allow` migration with a code-level override**

Right now we're writing `.allow` into `PermissionManager` on the user's behalf, which makes storage represent capability fiat rather than user intent. A lot of this PR exists to compensate for that:

- the **lock UI** in Permission Center exists only because we stored something we don't want the user to edit
- the **`auto_granted` pixel** exists only because we did a migration
- the **row-hiding** in `buildPermissionItem` exists only because the row has nothing to say
- every fresh user gets a phantom `duck.ai`/mic entry after first launch
- turning the flag off later leaves `.allow` entries behind in storage

A read-path override — either special-cased in `PermissionManager.permission(forDomain:)` or injected via a `PermissionDecisionOverriding` protocol — would collapse all of this. No migration, no lock, no row-hiding, no auto-grant pixel, and flag rollback is a one-line revert. Legacy `.deny`/`.ask` entries become harmless shadows.

The only thing we'd still need is a one-shot "forget legacy voice-mode consent for duck.ai" cleanup (smaller and more honestly named than a permission migration), and the OS-denied warning row would become a synthetic item built by the view model (it's already special-cased there anyway).

Worth considering before this lands — happy to be wrong if there's a constraint I'm missing.

tomasstrba:
TS2/ I cannot get the build to present the site permission dialog on duck.ai when the feature flag is turned off. I have to run the main branch.
Screenshot 2026-05-15 at 12 26 26
Can you see this too?

jotaemepereira:
TS1/ Yes, I believe you are right. I locked myself into that architectural change (using the migration class), and some of the changes are a consequence of it. We could simplify this.

TS2/ I was able to do it, but I will double-check now that I’m probably refactoring the code.

jotaemepereira:
@tomasstrba I’ve made changes to the code, now instead of migrating and persisting the permission, we just do it in the model layer. Let me know your thoughts, I’ve tested the different scenarios and they work as expected.

macOS/DuckDuckGo/NavigationBar/View/AddressBarButtonsViewController.swift
```diff
@@ -1800,6 +1900,19 @@ final class AddressBarButtonsViewController: NSViewController {
         let url = tabViewModel.tab.content.urlForWebView ?? .empty
         let domain = (url.isFileURL ? .localhost : (url.host ?? "")).droppingWwwPrefix()
 
+        // On duck.ai with OS mic denied, the shield only exists as an anchor for the
+        // system-disabled remediation surface — route the click to that popover and toggle it
+        // if it's already shown.
+        if isDuckAiVoiceChatSystemMicDenied(forDomain: url.host ?? "") {
```
tomasstrba: **Click routing regression on duck.ai when OS denies mic + another permission is in play**

When the user is on duck.ai with OS mic denied **and** has another permission active (e.g. a previously granted notifications permission), the shield is visible via the `isDuckAiVoiceChatSystemMicDenied` branch in `updatePermissionCenterButton`. But because `shouldSuppressShieldOnDuckAi` returns false (non-mic permissions present), the shield is the *real* Permission Center entry point — yet this short-circuit always routes the click to the OS-denied mic popover and `return`s. The user can't reach Permission Center to manage their notification permission.

Suggest gating this short-circuit on "no other permissions to manage" — roughly the inverse of `shouldSuppressShieldOnDuckAi`'s body — so the OS-denied popover only takes over the click when there's nothing else for Permission Center to show.

tomasstrba (approved):
LGTM! Just a small cosmetic issue above that would be good to address in case we add other type of permissions in future (notifications)

Also, I still cannot reproduce the original authorization dialog when the feature flag is disabled, but since the duck.ai voice mode works as expected in this case, it's not a blocker.

Nice work 👏

PR #4864, opened 2026-05-13
[macOS] Suppresses the first-time quit survey when termination wasn't initiated by the user

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1213985848142870?focus=true
Tech Design URL:
CC:

### Description

Suppresses the first-time quit survey when termination wasn't initiated by the user explicitly quitting the app — specifically:

- **Sparkle update relaunch** — when the app is restarting itself to install an update, detected via `StateRestorationManager.isRelaunchingAutomatically`.
- **System logout / restart / shutdown** — detected by inspecting the `kAEQuitReason` attribute on the current Apple Event (`kAELogOut`, `kAEReallyLogOut`, `kAEShowRestartDialog`, `kAERestart`, `kAEShowShutdownDialog`, `kAEShutDown`, `kAEQuitAll`).

In both cases the user didn't choose to quit, so prompting them with "why are you quitting?" is wrong and pollutes the survey data.

Wired through two new closures on `QuitSurveyDecider` / `QuitSurveyAppTerminationDecider` (`isAppRelaunchingForUpdate`, `isSystemQuitting`), plus a `resetRelaunchFlagOnCancel` callback used to clear the persisted Sparkle relaunch flag when the termination chain is cancelled by a later decider — without it, a stale flag could silently suppress the survey on the next legitimate quit in the same session.

Gated behind a new macOS feature flag `firstTimeQuitSurveySkipNonUserQuit` (remote-releasable, default `.enabled`), so the bypass can be remotely disabled if it misbehaves.

### Testing Steps

> Prereq: be a "new user" (install date within 0–3 days) on a build where `firstTimeQuitSurvey` is enabled. Easiest path: run the `clean-app.sh debug` inside the macOS folder.

1. **Baseline — user-initiated quit**: with the feature flag on, ⌘Q the app as a new user → first-time quit survey appears. (sanity check)
2. **System restart bypass**: with the app running, choose Apple menu → Restart… → confirm. The app should terminate without showing the survey. (Equivalent: Log Out / Shut Down.)
3. **Update relaunch bypass**: trigger a Sparkle update install (or simulate via the updater debug menu so the app relaunches itself) → survey should NOT appear.
4. **Feature flag off**: disable `firstTimeQuitSurveySkipNonUserQuit` via the internal feature-flag debug menu (or set it to disabled in privacy config), then repeat step 2 or 3 → survey SHOULD appear (proves the gating works).
5. **Cancelled-termination reset**: with an active download running and Sparkle in the middle of an update relaunch, the downloads dialog should appear → click Cancel. Then quit normally with no download. Survey should still appear on this second quit (proves `resetRelaunchFlagOnCancel` clears the stale relaunch flag).

### Impact and Risks

**Low.** This only changes whether a survey dialog is shown in two non-user-initiated termination paths. Worst case if the detection misfires:

- **False negative** (we wrongly think it's a non-user quit): a legitimate first-time quitter doesn't see the survey — we lose one data point, no user-facing harm.
- **False positive** (we wrongly think it's a user quit): survey shows during shutdown/restart — user sees a brief dialog they can ignore as the system tears down. Same behavior as before this PR.

No data loss, no privacy regression (no new data collected), no change to core browsing or quit behavior.

#### What could go wrong?

- **Apple Event reason detection misses an edge case** — e.g., a future macOS adds a new system-quit reason we don't enum-match. **Mitigation**: falls back to the existing behavior (survey shows); also the feature flag lets us disable the bypass entirely.
- **Sparkle `isRelaunchingAutomatically` not set in time** — if Sparkle's relaunch path doesn't set the flag before `applicationShouldTerminate` runs, the bypass won't trigger. **Mitigation**: this is the same flag tab-restoration already relies on, so any regression there would surface in restoration tests too.
- **Stale relaunch flag suppressing a later quit** — the cancellation-reset closure is explicitly there to prevent this. The `QuitSurveyAppTerminationDecider` sits at index 0 of the decider chain, so its `deciderSequenceCompleted` always fires even when a later decider cancels; this is tested.
- **Carbon `OpenScripting` deprecation** — `kAEQuitReason`, `kAELogOut`, etc. live in Carbon, which Apple has long signalled they'd like to deprecate. **Mitigation**: still fully supported in current macOS SDKs; if removed in a future SDK, the build break is loud and obvious.

### Quality Considerations

- **Edge cases**: cancelled termination chains (handled), repeated quits in the same session (handled by the reset), feature flag off (handled by gating).
- **Performance**: zero hot-path impact — the new checks only run inside `applicationShouldTerminate`.
- **Monitoring**: none added. The existing first-time-quit-survey pixels naturally measure the change — if the bypass misfires we'll see survey-shown counts move on macOS-restart days; if the bypass works, we'll see fewer surveys per install (the desired effect).
- **Privacy/security**: no new data collected. Apple Event inspection is local and synchronous.

### Notes to Reviewer

- The `isQuittingForSystemEvent` extension on `NSAppleEventManager` is only meaningful inside `applicationShouldTerminate(_:)` (when AppKit is dispatching the `kAEQuitApplication` Apple Event). I noted this in the doc comment.
- Step 5 (cancelled-termination reset) is the subtlest scenario — please pay attention to the `deciderSequenceCompleted` reset and the unit tests for it.
- The feature flag is wired remote-releasable with `defaultValue: .enabled`. If you'd prefer `.internalOnly` for staged rollout, easy switch in `FeatureFlag.swift`.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only the conditions for showing the first-time quit survey during app termination, gated behind a new feature flag, with added unit coverage; no browsing, security, or data-handling logic is modified.
> 
> **Overview**
> Prevents the **first-time quit survey** from showing when the app is terminating due to a *non-user-initiated quit* (Sparkle update relaunch or system logout/restart/shutdown), gated behind a new `firstTimeQuitSurveySkipNonUserQuit` flag (remote-releasable, default enabled).
> 
> Wires new termination-context signals into `QuitSurveyDecider`/`QuitSurveyAppTerminationDecider` from `AppDelegate`, adds AppleEvent-based system-quit detection via `NSAppleEventManager.isQuittingForSystemEvent`, and resets the update-relaunch flag when the termination decider chain is cancelled to avoid suppressing a later legitimate quit.
> 
> Updates privacy/feature flag enums to include the new flag and adds unit tests covering the bypass and cancellation-reset behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2430e85587c017e39088bb913f2935978351060f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
@@ -151,6 +151,10 @@ public enum MacOSBrowserConfigSubfeature: String, PrivacySubfeature {
     /// https://app.asana.com/1/137249556945/inbox/1203972458584425/item/1212200919350194/story/1212483080081687
     case firstTimeQuitSurvey
 
+    /// Suppresses the first-time quit survey when termination wasn't initiated by the user
+    /// (Sparkle update relaunch, or system logout/restart/shutdown).
+    case firstTimeQuitSurveySkipNonUserQuit
+
     /// Web Notifications API polyfill - allows websites to show notifications via native macOS Notification Center
     /// https://app.asana.com/1/137249556945/project/414235014887631/task/1211395954816928?focus=true
     case webNotifications
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -1728,6 +1728,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 installDate: AppDelegate.firstLaunchDate,
                 persistor: persistor,
                 reinstallUserDetection: DefaultReinstallUserDetection(keyValueStore: keyValueStore),
+                isAppRelaunchingForUpdate: { [weak self] in
+                    self?.stateRestorationManager?.isRelaunchingAutomatically ?? false
+                },
+                isSystemQuitting: {
+                    NSAppleEventManager.shared().isQuittingForSystemEvent
+                },
+                resetRelaunchFlagOnCancel: { [weak self] in
+                    guard let stateRestorationManager = self?.stateRestorationManager,
+                          stateRestorationManager.isRelaunchingAutomatically else { return }
+                    stateRestorationManager.resetRelaunchFlag()
+                },
                 showQuitSurvey: { [weak self] in
                     guard let self else { return }
                     let presenter = QuitSurveyPresenter(windowControllersManager: self.windowControllersManager, persistor: persistor, featureFlagger: self.featureFlagger, historyCoordinating: self.historyCoordinator, faviconManaging: self.faviconManager)
--- a/macOS/DuckDuckGo/QuitSurvey/QuitSurveyDecider.swift
+++ b/macOS/DuckDuckGo/QuitSurvey/QuitSurveyDecider.swift
@@ -16,6 +16,7 @@
 //  limitations under the License.
 //
 
+import Carbon.OpenScripting
 import Common
 import FeatureFlags
 import Foundation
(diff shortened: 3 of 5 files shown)


jleandroperez (approved):
All test scenarios verified. Quality of life improvement @jotaemepereira !!

:shipit:

PR #4865, opened 2026-05-13
[macOS | iOS] Extract VPN rekey logic into KeyRotator

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203108348835387/task/1214783836303686?focus=true

### Description

Rekey is critical to VPN session health — a regression here silently stops key rotation across all sessions — but the flow is untestable in place. It lives inline on `PacketTunnelProvider`, which extends `NEPacketTunnelProvider`. That base class can't be instantiated outside an extension process, so code attached to it has no realistic unit-test path.

This PR makes rekey testable by lifting the lifecycle out of the provider into a dedicated class that can be instantiated and exercised directly in tests.

### Testing Steps

1. Enable VPN in the app and confirm it's connected.
2. Open **Console.app**, select your device, and filter by subsystem `Network protection` (or by category `Key Management`).
3. Trigger a rekey:
   - **macOS**: from the menu bar, **Debug → VPN → Registration Key → Expire Now**.
   - **iOS**: in-app Debug menu → **Network Protection** → **Expire Registration Key Now**.
4. Validate in Console.app:
   - Look for `Querying the current key pair` entries appearing immediately after the menu action, with a new `publicKey` value and a future `expirationDate` — confirms a fresh keypair was generated.
   - The VPN stays connected throughout (the rekey is non-reasserting; no visible reconnection).
5. Repeat on the other platform.

### Impact and Risks

**Medium** — touches scheduled key rotation. A regression would impact users silently (sessions wouldn't rotate keys).

#### What could go wrong?

- Rekey doesn't fire when it should → key never rotates. *Mitigated by:* unit tests on the new class covering the settings gate and the success/failure paths.
- Access-revoked errors stop being routed → subscription state stays stale. *Mitigated by:* routing goes through the same handler the two other tunnel-op call sites already use, so they all agree.
- Provider deinit mid-rekey reports false success. *Mitigated by:* the closure surfaces this as a failure so telemetry isn't fabricated.

### Quality Considerations

- New unit tests on the rotator and on the leak-check decision helper. Full VPN test suite passes on macOS.
- The new class has a doc comment spelling out what is and isn't in scope.

### Notes to Reviewer

- The new class is concrete, not a protocol — no caller needs to mock it.
- The rekey op is passed as a closure rather than added to `TunnelLifecycleManaging`, so the rotator stays ignorant of how a rekey is performed and the protocol doesn't grow a method that only one caller would use.
- `KeyExpirationTester` is unchanged.

---
###### Internal references:


jleandroperez (approved):
✅ Verified on both iOS and macOS that the Key gets regenerated

:shipit:

PR #4900, opened 2026-05-15
Fix RMF trigger filtering and idle return message display

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204186595873227/task/1214833717753157?focus=true
Tech Design URL: N/A
CC: N/A

### Description

- Introduce `TriggerFilter` enum (`.any`, `.noTrigger`, `.specific`) to replace the ambiguous optional `MessageTrigger?` parameter in `fetchScheduledRemoteMessage`, fixing a regression where cleanup and prefetch paths skipped triggered messages
- Fix after-idle remote messages not appearing in the unified toggle input focused mode by making `NewTabPageMessagesModel.isOpenedAfterIdle` a closure that reads from the NTP view model's escape hatch state, ensuring the messages model always reflects the current idle return state even when the NTP is created before the escape hatch is set

### Testing Steps

**Prerequisites:**
- Enable `showNTPAfterIdleReturn` feature flag
- Set After Inactivity option to "New Tab" in Settings > General
- Set idle threshold low enough to trigger easily (e.g. 30s)

**Test 1 — After-idle message only ([JSON config](https://pastebin.com/raw/JT3T7GRD)):**
1. Point RMF config to the JSON URL above
2. Launch the app and open a new tab
3. Verify: **no** remote message appears on the standard NTP
4. Navigate to any website, background the app
5. Wait for idle threshold, then return to the app
6. Verify: the "Idle Return Prefetch Test" message appears on the NTP with an orange square image (not the megaphone placeholder — confirms prefetch worked)
7. If unified input FF is ON: tap the input field and verify the message also appears in focused mode

**Test 2 — Standard + after-idle messages (queued by JSON order)**
1. Point RMF config to the [JSON](https://pastebin.com/raw/5XYMNyjj).
2. Launch app, open a new tab.
3. Verify “Standard NTP Message” appears (first in config priority).
4. Dismiss “Standard NTP Message”, then refresh RMF config (debug menu or relaunch).
5. Open a new tab (normal, not after idle).
6. Verify no remote message (idle is next in queue but has after_idle, so it’s hidden on normal NTP).
7. Navigate to a site, background, wait for idle threshold, return (after-inactivity → New Tab).
8. Verify “Idle Return Message” (orange square).
9. If unified input FF is on: tap input and verify the same idle message in focused mode.

**Test 3 — No triggered messages (standard RMF behavior):**
1. Use any standard RMF config without `displayConditions`
2. Verify messages display normally on NTP as before (no regression)

### Impact and Risks
**Impact Level: Medium**

#### What could go wrong?
- Remote messages with triggers could fail to display after idle return if the trigger filter logic doesn't match correctly — mitigated by unit tests covering all filter combinations
- The shared `HomePageConfiguration` instance is refreshed by both focused and unfocused NTP paths; the closure approach ensures the correct `openedAfterIdle` value is always read at refresh time rather than relying on stale cached state

### Quality Considerations
- Edge case: messages without `displayConditions` are treated as having `nil` trigger, correctly matching `.noTrigger` and `.any` but not `.specific`
- Edge case: prefetch and cleanup paths use `.any` to find all messages regardless of trigger
- Privacy: no new data collection or network calls; trigger filtering is local-only

### Notes to Reviewer
- The `TriggerFilter` enum replaces the previous `MessageTrigger?` parameter which was ambiguous (nil could mean "no filter" or "only non-triggered messages" depending on the call site)
- The closure change in `NewTabPageMessagesModel` follows the same pattern as `isStillOnboarding` and `onTryFireModeRequested` already used in the codebase

---

> [!NOTE]
> **Medium Risk**
> Changes message selection logic for triggered vs non-triggered remote messages and updates new-tab idle-return detection; mistakes could hide messages or cause incorrect message prefetch/cleanup behavior.
> 
> **Overview**
> Fixes remote message trigger filtering by replacing the ambiguous optional `MessageTrigger?` parameter with an explicit `TriggerFilter` (`.any`, `.noTrigger`, `.specific`) and updating store/protocol/callers so prefetch and feature-flag cleanup paths can include triggered messages when desired.
> 
> Updates NTP message selection to correctly prefer `after_idle` messages only when returning from idle (and otherwise show only non-triggered messages), and makes `NewTabPageMessagesModel` read the idle-return state via a closure tied to `escapeHatch` so focused/unfocused refreshes don’t use stale state. Tests are updated/expanded to cover the new filter semantics and idle fallback behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a302885bd7c84b1b1ece67f997047eca5138ada8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/RemoteMessageModel.swift
```diff
@@ -22,6 +22,12 @@ public enum MessageTrigger: String, Codable {
     case afterIdle = "after_idle"
 }
 
+public enum TriggerFilter: Equatable {
```
alessandroboron, line 26: Suggestion: can we add documentation for the TriggerFilter cases? In particular what happens if noTrigger is passed and message has display condition

SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/RemoteMessagingStore.swift
```diff
@@ -260,8 +260,13 @@ extension RemoteMessagingStore {
                     continue
                 }
 
-                if let messageTrigger = remoteMessage.displayConditions?.trigger, messageTrigger != trigger {
-                    continue
+                switch triggerFilter {
```
alessandroboron: Probably it would be good to add comments here to explain the logic for other devs and for us to remember in the future.

[suggestion] We could add the below function to TriggerFilter

```
public enum TriggerFilter: Equatable {
   …
   func matches(_ trigger: MessageTrigger?) -> Bool {
          switch self {                                                                                                                                                                                   
          case .any: return true                                                                                                                                                                        
          case .noTrigger: return trigger == nil
          case .specific(let required): return trigger == required                                                                                                                                        
          }
   }
```

Then at call site we just have

`guard triggerFilter.matches(remoteMessage.displayConditions?.trigger) else { continue }`
That way the matching rule lives within the type that defines it and the call site is cleaner.

iOS/DuckDuckGo/NewTabPageViewController.swift
```diff
@@ -95,13 +95,14 @@ final class NewTabPageViewController: UIHostingController<NewTabPageView>, NewTa
                                             favoriteDataSource: FavoritesListInteractingAdapter(favoritesListInteracting: interactionModel),
                                             faviconLoader: faviconLoader,
                                             faviconsCache: faviconsCache)
+        let viewModel = newTabPageViewModel
```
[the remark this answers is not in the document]
alessandroboron: @SabrinaTardio we can remove `hasEscapeHatch` since it’s not used anymore

alessandroboron (approved):
Tested and works as expected. I left some minor suggestions. I’m approving in the meantime.

PR #4902, opened 2026-05-15
iOS: Snapshot unifiedToggleInput flag at launch

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214820742244139?focus=true
Tech Design URL:
CC:

### Description
Re-evaluating `featureFlagger.isFeatureOn(.unifiedToggleInput)` on every read let mid-session debug-menu flips leave the old toggle UI half-wired — `SwitchBarHandler`'s container choice and the fade listener's guard would disagree with what was actually mounted, producing no-op taps and swapped Search/Duck.ai mappings. This adopts the existing `FireModeCapability.resolve(using:)` pattern: capture the flag into `UserDefaults` once in `MainCoordinator.init`, then route `SwitchBarHandler` and the AIChat user-script / contextual sheet / contextual web view through `UnifiedToggleInputFeature.isFeatureFlagEnabled`. Toggling the flag from the debug menu now requires a relaunch, which is the correct behavior given how much UI wiring binds to this flag at startup.

### Testing Steps
1. Launch the app with `unifiedToggleInput` disabled; confirm the old Search/Duck.ai toggle picker works correctly (tap Search → search page, tap Duck.ai → chat page).
2. Without relaunching, enable `unifiedToggleInput` from the debug menu, then re-open the omnibar — the old toggle UI must continue to behave correctly (no swapped buttons, no dead taps).
3. Kill and relaunch the app — the new Unified Toggle Input UI should now take over.

### Impact and Risks
Low

#### What could go wrong?
A consumer that genuinely needs runtime flag changes (rather than session-stable) would now require a relaunch — none of the six call sites updated here have that requirement.

---

> [!NOTE]
> **Medium Risk**
> Changes feature-flag evaluation semantics for `unifiedToggleInput` to a session-stable snapshot stored in `UserDefaults`, so any missed/late `resolve()` call could lead to inconsistent UI behavior until relaunch.
> 
> **Overview**
> Makes `unifiedToggleInput` a **launch-resolved, session-stable** flag by adding `UnifiedToggleInputFeature.resolve(using:)` which snapshots the flag into `UserDefaults` and exposing `isFeatureFlagEnabled` for consumers.
> 
> Routes AI Chat contextual components and `SwitchBarHandler` to consult `UnifiedToggleInputFeature.isFeatureFlagEnabled` (instead of re-checking `featureFlagger.isFeatureOn`), and calls `UnifiedToggleInputFeature.resolve(using:)` during `MainCoordinator` init to lock the value for the session; tests are updated/added to verify the snapshot does not change with mid-session flag mutations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63d9cef1d3bacfac84bf7b29d028a2beea56edb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
+++ b/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
@@ -73,6 +73,7 @@ final class AIChatContextualSheetCoordinator {
     private let contentBlockingAssetsPublisher: AnyPublisher<ContentBlockingUpdating.NewContent, Never>
     private let featureDiscovery: FeatureDiscovery
     private let featureFlagger: FeatureFlagger
+    private let unifiedToggleInputFeature: UnifiedToggleInputFeatureProviding
     private let duckAiNativeStorageHandler: DuckAiNativeStorageHandling?
     private let duckAiFireModeStorageHandler: DuckAiNativeStorageHandling?
     private let debugSettings: AIChatDebugSettingsHandling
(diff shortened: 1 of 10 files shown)


samsymons (commented):
I won't formally approve it since I can't speak confidently about the code changes, but this did fix the issue for me - enabling the flag kept the old toggle appearance (which worked correctly), and the unified toggle only appeared after a relaunch. Same when disabling the flag, the unified toggle remained active until relaunch.

aataraxiaa (approved):
Nice one @Bunn 🚀

PR #4903, opened 2026-05-15
iOS: Use Reasoning Level from Native Storage on Chat Continuation

pikorddg:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1210947754188321/task/1214769493606669?focus=true
Tech Design URL:
CC:

### Description
When a user reopened a previously saved Duck .ai chat from native storage, the reasoning picker in the UI did not reflect the mode the chat had been conducted in — it always showed the global default. This PR reads the new `reasoningMode` field from the stored chat payload (per `chatID`) and applies it to the live preferences and reasoning picker, mirroring the existing `restoreLastUsedModel` pattern.

There is no DB migration needed: storage is an opaque JSON blob and the decoder now accepts an optional `reasoningMode` field, which keeps it backward- and forward-compatible. The global default (UserDefaults key `aichat.omnibar.selected-reasoning-mode`) is not touched directly — only the per-tab restoration path is updated. A new pixel `m_duck-ai_native-storage_last-used-reasoning-mode-parse_error` is fired when the stored JSON cannot be decoded, matching the existing `lastUsedModelParseError` pattern.

### Testing Steps
Prerequisite: `unifiedToggleInput` feature flag ON. Use a model that supports multiple reasoning modes (e.g. `gpt-5.2`, `claude-sonnet-4-5`).

1. Firstly, create some chats using production AI chat URL. These chats will be handy to test backward compatibility. Ideally name them descriptively, like `Haiku, Reasoning, Prod`, `GPT Fast prod`
2. Confirm production is not storing reasoning mode per chat
   1. open chat # 1 in Fast mode
   2. open chat # 2 in Extended Reasoning mode
   3. come back to chat # 1, the default is Extended Reasoning
3. Set custom AI chat url to: https://euw-serp-dev-testing21.duck.ai/
4. Apply temporary code workaround so that methods like `bindAITabIfPossible` can work correctly
   1. In `SharedPackages/AIChat/Sources/AIChat/Shared/URL+Extension.swift`, replace the first host check inside `isDuckAIURL` with: `if host == DuckDuckGo.aiHost || host?.hasSuffix(".\(DuckDuckGo.aiHost)") == true { return true }`
5. From NTP/opened website or directly from [duck.ai](http://duck.ai/), create Chat # 1: set the reasoning mode to *Fast* in the picker, send a prompt, wait for the response. Hint: you can name the chat like `GPT Fast`.
6. Create new chat, Chat # 2: set the reasoning mode to *Reasoning* (or *Extended Reasoning*), send a prompt.
7. Switch back to Chat # 1 → the picker should display *Fast*.
8. Switch back to Chat # 2 → the picker should display *Reasoning* / *Extended Reasoning*
9. Open a brand-new chat → the picker should show the global default (the most recently used mode), unaffected by the payloads of the existing chats.
10. Backward compatibility: now we’ll use chats from step 1: reopen these chats. The app must not crash, and the picker should remain on the global default because the payload has no `reasoningMode` field.
11. Optional — stale-mode safety net: switch to a model that does not support a given reasoning mode (e.g. `gpt-4o`) and reopen a chat that was saved with another model using `extendedReasoning`. After `fetchModels` resolves, the picker should clear the mode (via the existing `clearStaleReasoningModeIfNeeded`).

### Impact and Risks

**Low.**
- The entire UnifiedToggleInput module is gated behind the `unifiedToggleInput` feature flag, so the changes ship behind that flag.
- The change to the shared `DuckAiChat` is the addition of an optional field: backward-compatible with legacy storage payloads and forward-compatible (`JSONDecoder` ignores unknown keys).
- No DB migration: the `duck_ai_chats` table schema is `chatId` + opaque `data BLOB`.
- The implementation is a 1:1 mirror of `restoreLastUsedModel` and relies on the same safety net (`clearStaleReasoningModeIfNeeded` inside `fetchModels`).
- No new network or disk operations are introduced — only an additional decode pass over the existing JSON blob using the existing storage handler.

#### What could go wrong?

- **Cold-start race:** if `bindAITabIfPossible` runs before `fetchModels` resolves, `applyChatPersistedReasoningMode` writes preferences without the model-supports validity check. This is intentional and mirrors `applyPersistedSelection`. The existing `clearStaleReasoningModeIfNeeded` corrects the state once the resolved model lacks support for the mode.
- **Legacy chats / dropped support:** chats saved before this change have no `reasoningMode` field — the provider returns `nil` and the restore method is a no-op (picker keeps its previous state). Chats whose model later drops support for a given mode are handled by the existing stale-clear path.
- **Unknown future modes:** if a new mode is added on the FE before iOS knows about it, `AIChatReasoningMode(rawValue:)` returns `nil` and the restore method becomes a no-op — no crash, no incorrect picker state.

### Quality Considerations

- **Backward compatibility verified:** older chats without `reasoningMode` in the payload → provider returns `nil` → restore method is a no-op.
- **Unknown values:** `AIChatReasoningMode(rawValue:)` returning `nil` is handled explicitly with a debug log and a no-op.
- **Per-chat vs. global:** the PR updates the per-tab live `preferences.selectedReasoningMode` and refreshes the picker. The global default is mirrored on every tab change via the pre-existing pattern — this PR does not change that.
- **Privacy:** no new data is collected beyond what already exists in storage. The new pixel `m_duck-ai_native-storage_last-used-reasoning-mode-parse_error` only carries `appVersion`, `errorDomain`, and `errorCode` — no user content.
- **macOS:** `DuckAiNativeStoragePixelAdapter.fire` on macOS has a `break` for the new event, mirroring how `.lastUsedModelParseError` is handled today. macOS does not currently emit these pixels; if it ever needs to, a case in the adapter plus a definition in macOS PixelDefinitions would be enough.
- **Test coverage:** 13 new unit tests in `DuckAiLastUsedReasoningModeProviderTests` cover the three known modes, missing field, unknown value, `null` value, missing `chatId`, decode error, missing record, storage throwing, and full FE-contract payload.

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Low risk: adds an optional `reasoningMode` field decode and uses it to restore UI state, with no schema migration and graceful no-ops for missing/unknown values; main risk is minor state mismatches if restoration timing conflicts with model loading.
> 
> **Overview**
> When continuing an existing Duck.ai chat, the Unified Toggle Input now restores the chat’s persisted `reasoningMode` from the native-storage JSON payload (in addition to the existing model restore) and refreshes the reasoning picker accordingly.
> 
> This adds optional `reasoningMode` decoding to `DuckAiChat`, introduces `DuckAiLastUsedReasoningModeProvider` to read the value per `chatId`, and wires a new parse-error pixel (`m_duck-ai_native-storage_last-used-reasoning-mode-parse_error`) through iOS/macOS adapters. `UTIModelStore` gains `applyChatPersistedReasoningMode` to apply trusted persisted state before models finish loading, and unit tests cover provider behavior for missing/null/unknown values and error cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38e27522dc6926fbcbc631e92d6ff12ad3070d44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/AIChat/Sources/AIChat/Shared/NativeStorage/DuckAiChat.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/Shared/NativeStorage/DuckAiChat.swift
@@ -38,13 +38,17 @@ public struct DuckAiChat: Equatable {
     /// UUIDs of files referenced by this chat, stored in the native file store.
     public let fileRefs: [String]
 
+    /// Raw FE-supplied reasoning-mode string for this chat.
+    public let reasoningMode: String?
+
     public init(
         chatId: String,
         title: String,
         model: String,
         lastEdit: String,
         pinned: Bool,
-        fileRefs: [String] = []
+        fileRefs: [String] = [],
+        reasoningMode: String? = nil
     ) {
         self.chatId = chatId
         self.title = title
@@ -52,6 +56,7 @@ public struct DuckAiChat: Equatable {
         self.lastEdit = lastEdit
         self.pinned = pinned
         self.fileRefs = fileRefs
+        self.reasoningMode = reasoningMode
     }
 }
 
@@ -71,7 +76,8 @@ extension DuckAiChat {
             model: blob.model ?? "",
             lastEdit: blob.lastEdit ?? "",
             pinned: blob.pinned ?? false,
-            fileRefs: blob.fileRefs ?? []
+            fileRefs: blob.fileRefs ?? [],
+            reasoningMode: blob.reasoningMode
         )
 
         let firstUserMessage = blob.messages?
@@ -92,6 +98,7 @@ private struct ChatBlob: Decodable {
     let pinned: Bool?
     let fileRefs: [String]?
     let messages: [MessageBlob]?
+    let reasoningMode: String?
 }
 
 private struct MessageBlob: Decodable {
--- a/SharedPackages/AIChat/Sources/AIChat/Shared/NativeStorage/DuckAiChatReasoningModeProviding.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/Shared/NativeStorage/DuckAiChatReasoningModeProviding.swift
@@ -1,0 +1,52 @@
+//
+//  DuckAiChatReasoningModeProviding.swift
+//
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import DuckAiDataStore
+import Foundation
+
+/// Reads the last-used `reasoningMode` raw string for a stored Duck.ai chat.
+public protocol DuckAiChatReasoningModeProviding {
+    /// Returns the raw `reasoningMode` string persisted with `chatId`, or `nil` when:
+    /// - the chat is not in storage,
+    /// - the payload has no `reasoningMode` field, or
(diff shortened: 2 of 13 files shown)


aataraxiaa (approved):
LGTM 👍

PR #4906, opened 2026-05-15
iOS: Move Duck.ai native storage out of app group container

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1214837562649183?focus=true
Tech Design URL:
CC:

### Description
The Duck.ai SQLCipher DB lives in the shared app-group container, whose default `NSFileProtectionComplete` makes it inaccessible while the device is locked — producing 0xdead10cc kills on the background auto-clear path. This PR relocates the default and fire-mode containers to `Library/Application Support/`, migrating existing user data once with up-to-3 retries across launches, and marks both directories as excluded from iCloud / Finder backups. New container-migration pixels (`default` / `fire-mode`) report success, no-data, orphan-cleaned, attempt-failed, and gave-up outcomes.

### Testing Steps
Feature gated behind FF: duckAINativeStoragePathMigration

1. Install a build with the old (app-group) storage, send a few Duck.ai messages, then upgrade to this build and confirm chat history is still present after launch.
2. Reboot the device, lock it, and trigger the auto-clear path to verify no 0xdead10cc termination.
3. Delete the app, install a clean version, check if duck.ai storage is creating the folder in the correct path and data is being saved there (There are logs printing the path, and for validating data you can check on the AI Chat storage debugger. Debug -> AI Chat -> Start Server)

### Impact and Risks
High

#### What could go wrong?
Migration fails for a user with non-trivial chat data and the give-up path discards their history after 3 attempts; pixels will surface the rate if it happens.

---

> [!NOTE]
> **High Risk**
> One-time filesystem migration of encrypted chat DBs at launch can block native storage or strand history if move/conflict handling fails; mitigated by preserving source data on give-up and deferring when protected data or KV reads are unavailable.
> 
> **Overview**
> Adds a **remote-gated** (`duckAINativeStoragePathMigration` / `nativeStoragePathMigration`, internal-only default) path that relocates Duck.ai **default** and **fire-mode** native SQLCipher containers from the shared **App Group** to **`Library/Application Support`**, with the legacy App Group layout unchanged when the flag is off.
> 
> Introduces **`DuckAiNativeStorageContainerMigration`**: one-time directory move with KV-backed state, protected-data deferral, up to three move retries per launch (budget resets across launches), destination-conflict handling that preserves the old copy, async recursive **`NSFileProtectionCompleteUntilFirstUserAuthentication`**, and **`isExcludedFromBackup`**. **`Launching.makeNativeStorageHandler`** and **`FireModeNativeStorageController`** run migration before opening storage and return **`nil`** on **`.skip`** so an empty DB is not created mid-failure.
> 
> Adds labeled **container-migration** pixels (default / fire-mode) and extensive unit tests for retry, recovery, and launch short-circuit behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c56643b11c40a0cdf1539b9f3637ee0f53995010. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
+++ b/iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
@@ -23,77 +23,6 @@ import Persistence
 import UIKit
 import os.log
 
-/// Identifies the container being migrated. Raw value is suffixed onto pixel names.
-enum DuckAiNativeStorageContainerMigrationLabel: String {
-    case `default`
-    case fireMode = "fire-mode"
-}
-
-/// Outcomes for the container relocation. Distinct from the JS→native chat-data
-/// migration pixels (`duckAiNativeStorageMigration*`).
-enum DuckAiNativeStorageContainerMigrationEvent {
-    case notNeeded(label: DuckAiNativeStorageContainerMigrationLabel)
-    case success(label: DuckAiNativeStorageContainerMigrationLabel)
-    case attemptFailed(label: DuckAiNativeStorageContainerMigrationLabel, error: Error)
-    case gaveUp(label: DuckAiNativeStorageContainerMigrationLabel, error: Error?)
-    /// Move succeeded but `applyDefaultFileProtection` failed — the DB kept its
-    /// inherited `NSFileProtectionComplete` class (the 0xdead10cc condition).
-    case protectionFailed(label: DuckAiNativeStorageContainerMigrationLabel, error: Error)
-    /// Destination has data but the migrated flag isn't set — could be a partial
-    /// move or a DB created while the App Group entitlement was unavailable. The
-    /// destination is claimed going forward and the source is preserved.
-    case destinationConflict(label: DuckAiNativeStorageContainerMigrationLabel)
-    /// `isExcludedFromBackup` or the destination directory creation failed.
-    /// The encrypted DB may end up in iCloud backups.
-    case excludeFromBackupFailed(label: DuckAiNativeStorageContainerMigrationLabel, error: Error)
-    /// Deferred because protected data is unavailable (pre-first-unlock
-    /// background launch). No state is modified.
-    case protectedDataUnavailable(label: DuckAiNativeStorageContainerMigrationLabel)
-}
-
-enum DuckAiNativeStorageContainerMigrationOutcome {
-    case proceed
-    case skip
-}
-
-protocol DuckAiNativeStorageContainerMigrationPixelFiring {
-    func fire(_ event: DuckAiNativeStorageContainerMigrationEvent)
-}
-
-struct NullDuckAiNativeStorageContainerMigrationPixelFiring: DuckAiNativeStorageContainerMigrationPixelFiring {
-    func fire(_ event: DuckAiNativeStorageContainerMigrationEvent) {}
-}
-
-struct DuckAiNativeStorageContainerMigrationPixelAdapter: DuckAiNativeStorageContainerMigrationPixelFiring {
-    func fire(_ event: DuckAiNativeStorageContainerMigrationEvent) {
-        switch event {
-        case .notNeeded(let label):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationNotNeeded(label: label.rawValue))
-        case .success(let label):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationSuccess(label: label.rawValue))
-        case .attemptFailed(let label, let error):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationAttemptFailed(label: label.rawValue), error: error)
-        case .gaveUp(let label, let error):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationGaveUp(label: label.rawValue), error: error)
-        case .protectionFailed(let label, let error):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationProtectionFailed(label: label.rawValue), error: error)
-        case .destinationConflict(let label):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationDestinationConflict(label: label.rawValue))
-        case .excludeFromBackupFailed(let label, let error):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationExcludeFromBackupFailed(label: label.rawValue), error: error)
-        case .protectedDataUnavailable(let label):
-            Pixel.fire(pixel: .duckAiNativeStorageContainerMigrationProtectedDataUnavailable(label: label.rawValue))
-        }
-    }
-}
-
-/// Runs the one-time move of a Duck.ai container from the shared App Group into
-/// the app's Application Support directory.
-protocol DuckAiNativeStorageContainerMigrating {
-    @discardableResult
-    func run() -> DuckAiNativeStorageContainerMigrationOutcome
-}
-
 /// One-time move of a Duck.ai container from the shared App Group into the
 /// app's Application Support directory.
 ///
@@ -103,16 +32,7 @@ protocol DuckAiNativeStorageContainerMigrating {
 /// `NSFileProtectionCompleteUntilFirstUserAuthentication`, which stays readable
 /// after first unlock.
 ///
-/// State (in `keyValueStore`):
-/// - `<migrationKey>.migrated` (Bool) — set when we own the destination going
-///   forward (success / notNeeded / destinationConflict). Once set, `run()`
-///   is a no-op apart from `ensureProtection`.
-/// - `<migrationKey>.attempts` (Int) — failed move attempts. Cleared when
-///   `migrated` is set. Left at the cap on `.gaveUp` so the next launch
-///   detects the exhausted budget and resets for a fresh retry.
-/// - `<migrationKey>.protectionAttempts` (Int) — retry budget for
-///   `applyDefaultFileProtection`. `maxAttempts` is the "done" sentinel
-///   (set by success or by exhausting retries).
+/// Persistent state is owned by `MigrationStateStore`.
 struct DuckAiNativeStorageContainerMigration: DuckAiNativeStorageContainerMigrating {
 
     static let defaultMaxAttempts = 3
@@ -122,13 +42,13 @@ struct DuckAiNativeStorageContainerMigration: DuckAiNativeStorageContainerMigrat
 
     let oldURL: URL
     let newURL: URL
-    let migrationKey: String
     let label: DuckAiNativeStorageContainerMigrationLabel
-    let keyValueStore: ThrowingKeyValueStoring
     let fileManager: FileManager
     let pixelFiring: DuckAiNativeStorageContainerMigrationPixelFiring
     let isProtectedDataAvailable: () -> Bool
     let maxAttempts: Int
+
+    private let stateStore: MigrationStateStore
 
     init(oldURL: URL,
          newURL: URL,
@@ -141,43 +61,54 @@ struct DuckAiNativeStorageContainerMigration: DuckAiNativeStorageContainerMigrat
          maxAttempts: Int = DuckAiNativeStorageContainerMigration.defaultMaxAttempts) {
         self.oldURL = oldURL
         self.newURL = newURL
-        self.migrationKey = migrationKey
         self.label = label
-        self.keyValueStore = keyValueStore
         self.fileManager = fileManager
         self.pixelFiring = pixelFiring
         self.isProtectedDataAvailable = isProtectedDataAvailable
         // 0 / negative would give up on the first failure.
         self.maxAttempts = max(1, maxAttempts)
+        self.stateStore = MigrationStateStore(keyValueStore: keyValueStore, migrationKey: migrationKey)
     }
 
     // MARK: - Entry point
 
     @discardableResult
     func run() -> DuckAiNativeStorageContainerMigrationOutcome {
-        // Background launches before first unlock can't read
-        // `NSFileProtectionComplete` source bytes. Defer rather than burn
-        // retries on a transient condition.
         guard isProtectedDataAvailable() else {
             Self.logger.info("[NativeStorage] [\(label.rawValue, privacy: .public)] protected data unavailable; deferring")
             pixelFiring.fire(.protectedDataUnavailable(label: label))
             return .skip
         }
 
-        resetExhaustedAttemptsIfNeeded()
+        do {
+            return try performMigration()
+        } catch let kvError as MigrationStateStore.ReadError {
+            Self.logger.error("[NativeStorage] [\(label.rawValue, privacy: .public)] key-value store read failed; deferring: \(kvError.underlying.localizedDescription, privacy: .public)")
+            pixelFiring.fire(.keyValueStoreReadFailed(label: label, error: kvError.underlying))
+            return .skip
+        } catch {
+            assertionFailure("Unexpected throw from performMigration: \(error)")
+            Self.logger.error("[NativeStorage] [\(label.rawValue, privacy: .public)] unexpected migration error; deferring: \(error.localizedDescription, privacy: .public)")
+            return .skip
+        }
+    }
 
-        if isMigrated {
-            ensureProtection()
+    private func performMigration() throws -> DuckAiNativeStorageContainerMigrationOutcome {
+        var state = try stateStore.load()
+        resetExhaustedAttemptsIfNeeded(&state)
+
+        if state.isMigrated {
+            ensureProtection(priorAttempts: state.protectionAttempts)
             return .proceed
         }
 
-        Self.logger.info("[NativeStorage] [\(label.rawValue, privacy: .public)] starting (prior attempts: \(attempts)); old=\(oldURL.path, privacy: .public) new=\(newURL.path, privacy: .public)")
+        Self.logger.info("[NativeStorage] [\(label.rawValue, privacy: .public)] starting (prior attempts: \(state.attempts)); old=\(oldURL.path, privacy: .public) new=\(newURL.path, privacy: .public)")
 
         guard fileManager.fileExists(atPath: oldURL.path) else {
             Self.logger.info("[NativeStorage] [\(label.rawValue, privacy: .public)] no old directory; marking done")
-            markMigrated()
+            stateStore.markMigrated()
             pixelFiring.fire(.notNeeded(label: label))
-            ensureProtection()
+            ensureProtection(priorAttempts: state.protectionAttempts)
             return .proceed
         }
 
@@ -188,16 +119,16 @@ struct DuckAiNativeStorageContainerMigration: DuckAiNativeStorageContainerMigrat
             // `excludeFromBackup` on a skipped launch — remove and proceed.
             if destinationHasData {
                 Self.logger.error("[NativeStorage] [\(label.rawValue, privacy: .public)] destination has data without migrated flag; claiming new and preserving old for recovery")
-                markMigrated()
+                stateStore.markMigrated()
                 pixelFiring.fire(.destinationConflict(label: label))
-                ensureProtection()
+                ensureProtection(priorAttempts: state.protectionAttempts)
                 return .proceed
             }
             Self.logger.info("[NativeStorage] [\(label.rawValue, privacy: .public)] empty destination directory present; removing to clear path for move")
             do {
                 try fileManager.removeItem(at: newURL)
             } catch {
-                return handleMoveFailure(error, operation: .removingEmptyDestination)
+                return handleMoveFailure(error, operation: .removingEmptyDestination, priorAttempts: state.attempts)
             }
         }
 
@@ -206,12 +137,12 @@ struct DuckAiNativeStorageContainerMigration: DuckAiNativeStorageContainerMigrat
                                             withIntermediateDirectories: true)
             try fileManager.moveItem(at: oldURL, to: newURL)
             Self.logger.info("[NativeStorage] [\(label.rawValue, privacy: .public)] moved successfully")
-            markMigrated()
+            stateStore.markMigrated()
             pixelFiring.fire(.success(label: label))
-            ensureProtection()
+            ensureProtection(priorAttempts: state.protectionAttempts)
             return .proceed
         } catch {
-            return handleMoveFailure(error, operation: .move)
+            return handleMoveFailure(error, operation: .move, priorAttempts: state.attempts)
         }
     }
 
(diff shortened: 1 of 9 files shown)


iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
```diff
@@ -0,0 +1,233 @@
...
+    /// this step a locked device can still trip 0xdead10cc on SQLite reads. Setting
+    /// the directory itself also fixes the default class new sidecars inherit.
+    static func applyDefaultFileProtection(at url: URL, fileManager: FileManager = .default) {
+        let attributes: [FileAttributeKey: Any] = [.protectionKey: FileProtectionType.completeUntilFirstUserAuthentication]
+
+        func apply(to path: String) {
+            do {
+                try fileManager.setAttributes(attributes, ofItemAtPath: path)
+            } catch {
+                Self.logger.error("[NativeStorage] failed to set file protection on \(path, privacy: .public): \(error.localizedDescription, privacy: .public)")
```
aataraxiaa, line 198: Question on this path: if `setAttributes` fails on the directory itself or on `chats.db` / `chats.db-wal`, the migration still reports `success` and the caller opens the DB with whatever protection class came across from the app-group move. That's exactly the 0xdead10cc condition this PR is fixing — but silently, with no telemetry.

Worth considering: treating a failure on the directory-level `setAttributes` (or on a `chats.db*` path) as a migration failure so the next launch retries, or at minimum firing a dedicated pixel from the catch block. Right now we'd only see it as a rate of repeat crashes post-rollout.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
```diff
@@ -0,0 +1,233 @@
...
+                pixelFiring.fire(.orphanRemoved(label: label))
+                return .proceed
+            } catch {
+                // newURL is intact and usable, but the old container still holds
+                // sensitive data. Don't claim success — keep retrying deletion
+                // across launches until it works or we hit the attempt cap.
+                let attempt = priorAttempts + 1
+                userDefaults.set(attempt, forKey: attemptsKey)
+                if attempt >= maxAttempts {
+                    Self.logger.error("[NativeStorage] [\(label.rawValue, privacy: .public)] orphan removal failed (attempt \(attempt)/\(maxAttempts)); giving up — old data may remain in app group: \(error.localizedDescription, privacy: .public)")
```
aataraxiaa: Once `gaveUp` fires here we set `done = true` and never look at the old container again, so user chat content stays in the app-group container indefinitely under `NSFileProtectionComplete`. Not a crash risk, but it's the same sensitive content the rest of the PR is moving for safety reasons.

Two cheap options:
- Add a `try? fileManager.removeItem(at: oldURL)` best-effort sweep just before setting `done` on this branch.
- Add a pixel parameter or separate event for "abandoned data still present" so we have visibility into how often this lands.

Either is fine; the current behavior just feels worth at least one more shot before we lose track of it.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
aataraxiaa, line 198: A `setAttributes` failure on chats.db doesn't reach the outer retry block - we log moved successfully, set `done = true`, fires the success pixel, and never retries. The file ends up at the new path with its original NSFileProtectionComplete class still applied.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
aataraxiaa: Maybe some cleanup needed here. `try? fileManager.removeItem(at: oldURL)` immediately before setting done on the give-up branch.

aataraxiaa (commented):
Tested and mostly looks good - logs show the migration and creation in new location. 2 inline comments.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    private func handleMoveFailure(_ error: Error, operation: FailingOperation) -> DuckAiNativeStorageContainerMigrationOutcome {
+        let attempt = attempts + 1
+        setAttempts(attempt)
+        if attempt >= maxAttempts {
+            Self.logger.error("[NativeStorage] [\(label.rawValue, privacy: .public)] \(operation.rawValue, privacy: .public) failed (attempt \(attempt)/\(maxAttempts)); giving up: \(error.localizedDescription, privacy: .public)")
+            pixelFiring.fire(.gaveUp(label: label, error: error))
+            ensureProtection()
+            return .proceed
```
jaceklyp: **Major — data-loss orphan path on give-up.**

After 3 failed `moveItem` attempts, this returns `.proceed` with `migrated=false` and `attempts` pinned at the cap. The caller in `Launching.swift:399-425` (and the matching block in `FireModeNativeStorageController.swift:94-101`) branches on `.skip` only — `.proceed` falls through to `excludeFromBackup(containerURL)` (creates the destination directory) and `DuckAiNativeStorageHandler(.disk(path: containerURL, ...))` (creates `chats.db` + `settings.plist` + `files/`).

The next launch: `resetExhaustedAttemptsIfNeeded` clears `attempts` (still `!isMigrated`), control reaches the `fileExists(newURL.path)` branch at line 188, `destinationHasData` is `true` (the empty DB + sidecars created by the caller after gave-up), the `.destinationConflict` branch at line 195 fires, calls `markMigrated()`, and **commits the empty new container as the source of truth forever**. The user's chats at `oldURL` stay on disk but no code path ever reads them again — silently orphaned.

The pixel description for `destination-conflict_*` even promises "the old App Group container is preserved so the user's pre-upgrade data remains recoverable" — but there is no recovery code path anywhere in the PR. The orphan rate ≈ gave-up rate × P(next-launch-migration-runs), and the latter is ~100% because of the budget reset.

**Suggested fix:** return `.skip` instead of `.proceed` on give-up, and drop the `ensureProtection()` call here since there's nothing at the destination to protect. This matches the contract already verified by `testGivenFailedMoveWhenFactoryRunsThenDestinationStaysAbsent`: caller returns `nil` handler, destination is not created, source stays untouched. User loses Duck.ai for that launch, but no orphaning. Next launch's `resetExhaustedAttemptsIfNeeded` retries cleanly.

Current:
```swift
if attempt >= maxAttempts {
    Self.logger.error(...)
    pixelFiring.fire(.gaveUp(label: label, error: error))
    ensureProtection()
    return .proceed
}
```

Proposed:
```swift
if attempt >= maxAttempts {
    Self.logger.error(...)
    pixelFiring.fire(.gaveUp(label: label, error: error))
    return .skip
}
```

Also resolves the misleading `gave-up_*` pixel description ("migration will not be retried and old data is abandoned") — once gave-up returns `.skip`, the budget reset still gives transient errors another chance, but no shadow-empty-DB ever lands at `newURL`.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
```diff
@@ -0,0 +1,370 @@
...
+            // it as a possibly-partial copy: claim the destination but keep the
+            // source for recovery. If empty, it's scaffolding from a prior
+            // `excludeFromBackup` on a skipped launch — remove and proceed.
+            if destinationHasData {
+                Self.logger.error("[NativeStorage] [\(label.rawValue, privacy: .public)] destination has data without migrated flag; claiming new and preserving old for recovery")
+                markMigrated()
+                pixelFiring.fire(.destinationConflict(label: label))
+                ensureProtection()
+                return .proceed
+            }
```
jaceklyp, line 132: Tied to the give-up finding on line 359: this branch is the silent-orphan trigger. If give-up returns `.skip` (per the suggested fix), this branch becomes nearly unreachable from the same-app gave-up flow and only fires for genuinely-weird states (sibling process / external sync touched the destination) — which is the intended use.

If you'd rather keep `.proceed` here for some reason, then `destinationHasData` needs to be smarter than `!children.isEmpty` (line 313) — currently it can't distinguish "the SQLCipher-empty-DB the caller created post-give-up" from "a real partial-copy with user data". A per-file check (does `chats.db` exist *and* is non-empty *and* lacks the in-progress marker the caller would have to drop) would work, but it's load-bearing only because `.proceed` on give-up is. Easier to fix at the root.

iOS/PixelDefinitions/pixels/definitions/aichat_native_storage_pixels.json5
```diff
@@ -87,6 +87,118 @@
...
+    },
+    "m_duck-ai_native-storage_container-migration_attempt-failed_fire-mode": {
+        "description": "Fires when a fire-mode container move attempt fails and will be retried on a subsequent launch (up to 3 attempts total)",
+        "owners": ["Bunn"],
+        "triggers": ["other"],
+        "suffixes": ["platform", "form_factor"],
+        "parameters": ["appVersion", "errorDomain", "errorCode"]
+    },
+    "m_duck-ai_native-storage_container-migration_gave-up_default": {
+        "description": "Fires once after the default-container move has failed the maximum number of attempts; migration will not be retried and old data is abandoned",
```
jaceklyp: **Minor — description disagrees with code reality.**

"Migration will not be retried and old data is abandoned" — but `DuckAiNativeStorageContainerMigration.resetExhaustedAttemptsIfNeeded` (migration file lines 329-334) clears the attempts counter on every launch where `!isMigrated && attempts >= maxAttempts`, so the migration **does** retry on the next launch with a fresh budget. And "old data is abandoned" is only true *de facto* because of the orphan path flagged on `DuckAiNativeStorageContainerMigration.swift:359` — it's not the intended design (the corresponding test `testWhenMoveFailsRepeatedlyThenGivesUpAtMaxAttemptsWithoutSweepingSource` asserts source is preserved).

Suggested rewording, assuming the `.skip`-on-give-up fix lands:

> Fires once after the default-container move has failed the maximum number of attempts in a single launch. The migration is retried on subsequent launches with a fresh attempt budget; the user's pre-migration container at the old App Group path is preserved on disk in the meantime.

Same applies to `gave-up_fire-mode` on line 140.

iOS/PixelDefinitions/pixels/definitions/aichat_native_storage_pixels.json5
```diff
@@ -87,6 +87,118 @@
...
+    },
+    "m_duck-ai_native-storage_container-migration_protection-failed_fire-mode": {
+        "description": "Fires alongside container-migration_success_fire-mode when at least one setAttributes call failed while applying NSFileProtectionCompleteUntilFirstUserAuthentication after the fire-mode container move — destination data may have retained the original NSFileProtectionComplete class",
+        "owners": ["Bunn"],
+        "triggers": ["other"],
+        "suffixes": ["platform", "form_factor"],
+        "parameters": ["appVersion", "errorDomain", "errorCode"]
+    },
+    "m_duck-ai_native-storage_container-migration_destination-conflict_default": {
+        "description": "Fires when the default Duck.ai destination directory already contains a chats.db at migration time but no migration marker is present — could indicate a partial moveItem copy that crashed, or a previous launch that created data while the App Group entitlement was unavailable; the new location is claimed going forward and the old App Group container is preserved so the user's pre-upgrade data remains recoverable",
```
jaceklyp: **Minor — description references a marker that no longer exists; recovery claim is non-actionable.**

Two issues:

1. "No migration marker is present" — after the `49e0f4764f` refactor the marker is a `keyValueStore` Bool, not a file. "Marker" is technically an abstract term but reading the description today suggests a sentinel file is involved. Tighten to e.g. "no migration flag is set in the key-value store".
2. "The old App Group container is preserved so the user's pre-upgrade data remains recoverable" — no recovery code exists in this PR. The data stays on disk but nothing in the app ever reads from it again, so "recoverable" oversells the contract. Either drop that clause or qualify it: e.g. "the old App Group container is preserved on disk but not surfaced to the user; manual recovery would require operating-system-level access".

Same applies to `destination-conflict_fire-mode` on line 168.

iOS/DuckDuckGoTests/AIChat/DuckAiNativeStorageContainerMigrationTests.swift
```diff
@@ -0,0 +1,675 @@
...
+        migrate(from: oldURL, to: newURL)
+
+        // Migrated state must NOT delete oldURL. After the destination
+        // conflict path runs, oldURL can be the only complete copy.
+        XCTAssertTrue(FileManager.default.fileExists(atPath: oldURL.appendingPathComponent("chats.db").path),
+                      "oldURL data must be preserved once migrated flag is set")
+        XCTAssertTrue(pixelSpy.firedEventNames.isEmpty)
+    }
+
+    func testWhenGaveUpLeftPartialDestinationAndNextLaunchRunsThenDestinationConflictFires() throws {
```
jaceklyp: **Major — integration test gap for the orphan path.**

This test manually stages the post-give-up state by writing `"partial"` to `newURL` and setting `attempts = 3` directly. It proves *if* the gave-up + caller-creates-file state exists, `destinationConflict` fires correctly — but it does **not** prove the gave-up flow actually produces that state through the caller. The dangerous sequence `run() → .proceed → caller (Launching.makeNativeStorageHandler) creates DB at newURL → next-launch run() → destinationConflict → markMigrated → orphan` is the most important user-data-loss path in the PR and is not end-to-end verified.

Proposed test (sketch — don't expect it to compile cleanly, but the shape is what matters):

```swift
func testWhenGaveUpThenCallerCreatesDestinationThenNextLaunchOrphansOldData() throws {
    let oldURL = sandbox.appendingPathComponent("old/DuckAi")
    let newURL = sandbox.appendingPathComponent("new/DuckAi")
    try FileManager.default.createDirectory(at: oldURL, withIntermediateDirectories: true)
    try Data("chats".utf8).write(to: oldURL.appendingPathComponent("chats.db"))

    // Launch 1: 3 failed moves through the factory; on the gave-up .proceed,
    // the caller creates newURL/chats.db just like Launching.makeNativeStorageHandler.
    for _ in 0..<3 {
        _ = runFactory(
            oldURL: oldURL, newURL: newURL,
            fileManager: FailingMoveFileManager(),
            createHandler: {
                DuckAiNativeStorageContainerMigration.excludeFromBackup(
                    newURL, label: .default, pixelFiring: self.pixelSpy)
                try? Data("caller-created".utf8).write(
                    to: newURL.appendingPathComponent("chats.db"))
            })
    }

    // Launch 2: real FS, migrator re-runs.
    _ = migrate(from: oldURL, to: newURL)

    // Document the orphan as a regression test.
    XCTAssertEqual(pixelSpy.firedEventNames.last, "destinationConflict")
    XCTAssertTrue(keyValueStore.bool(forKey: migratedKey))
    XCTAssertTrue(FileManager.default.fileExists(atPath: oldURL.appendingPathComponent("chats.db").path),
                  "user's pre-upgrade data must remain on disk for recovery")
}
```

Value: if `.skip`-on-gave-up lands, this test changes — `runFactory`'s `createHandler` closure must never run, so the second launch's migration succeeds normally. Either way the test pins the contract.

While you're in here, two adjacent gaps worth a couple of lines each:

1. **Fire-mode UUID shape.** Every test uses the flat `chats.db` layout. The `.fireMode` label is never exercised against a `<UUID>/chats.db` subdirectory destination, even though `destinationHasData` is documented (line 305-306) as handling that shape.
2. **Recursive `applyDefaultFileProtection` with child failure.** Only root-fail and nil-enumerator are tested. The interesting case is root succeeds, child fails — the `errorHandler` closure on `enumerator(at:)` (migration file lines 271-275). A regression there would silently re-introduce `NSFileProtectionComplete` on the very SQLite sidecar files this PR exists to fix.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
```diff
@@ -0,0 +1,370 @@
...
+    private var protectionAttemptsKey: String { migrationKey + ".protectionAttempts" }
+
+    private var isMigrated: Bool {
+        (try? keyValueStore.object(forKey: migratedKey) as? Bool) ?? false
+    }
+
+    private var attempts: Int {
+        (try? keyValueStore.object(forKey: attemptsKey) as? Int) ?? 0
+    }
+
```
jaceklyp: Nit — silent error swallowing on KV reads.

`(try? keyValueStore.object(forKey: ...) as? Bool) ?? false` masks a genuinely broken store as "not migrated". The migrator then re-attempts every launch; if `oldURL` is already gone, fires `.notNeeded` forever; if not, ends up in `destinationConflict` once the destination has data. No data loss, but observability noise + a missed signal that the store itself is unhealthy. Consider firing a one-shot `keyValueStoreReadFailed` pixel or letting the catch surface to skip the migration entirely so we don't bury the real fault. Same applies to `attempts` (line 303) and `protectionAttempts` (line 308).

iOS/DuckDuckGo/AppLifecycle/AppStates/Launching.swift
```diff
@@ -372,12 +376,47 @@ struct Launching: LaunchingHandling {
...
+
+        let containerURL: URL
+        if featureFlagger.isFeatureOn(.duckAINativeStoragePathMigration) {
+            guard let appSupportURL = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first else {
+                return nil
+            }
+            containerURL = appSupportURL.appendingPathComponent(DuckAiNativeStorageHandler.defaultDirectoryName)
+
+            if let groupContainer = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: Global.appConfigurationGroupName) {
+                let outcome = DuckAiNativeStorageContainerMigration(
```
jaceklyp, line 391: Nit — main-thread perf risk on large containers.

`Launching` is `@MainActor` and this migration `.run()` is synchronous. The recursive `applyDefaultFileProtection` walks every file in `newURL` and issues a `setAttributes` syscall per entry (migration file lines 257-283). For a worst-case user (years of chats, large `files/` tree), the cumulative syscall cost could push past the iOS launch watchdog — turning a 0xdead10cc fix into a different 0xdead10cc cause.

Probably fine in practice (Duck.ai data is typically small), but the cost is unbounded by user data size. Two cheap mitigations to consider:

1. Defer `ensureProtection` to a background dispatch after `Launching.init` returns. Correctness is preserved — files don't need re-protection until the next lock event, which is post-launch.
2. Or at minimum, fire a timing pixel around the migration block so you can detect launch-budget regressions in production data.

jaceklyp (commented):
Reviewed the migrator end-to-end with three parallel agent passes (data-loss safety, test coverage, manual testing). Overall the design is sound and the test suite is thorough for the state machine. Internal-only flag default contains the rollout blast radius nicely.

One **major data-loss path** to address before ramping the flag wider — flagged inline at the `handleMoveFailure` give-up branch. Two **pixel descriptions** disagree with the code (gave-up's "will not be retried / abandoned" wording, destination-conflict's "old container is preserved so data remains recoverable" — no recovery code exists). One **test integration gap** for the orphan path. Plus a couple of minor observability/perf items inline.

Not flagged (verified safe so you don't have to re-trace):
- Keychain access class — `DuckAiKeyStoreProvider` uses `kSecAttrAccessibleAfterFirstUnlock`, the keychain analogue of `completeUntilFirstUserAuthentication`. No locked-device decrypt failure post-migration.
- `moveItem` cross-volume — both paths are on the iOS data partition, so `rename(2)` and atomic.
- `completeUntilFirstUserAuthentication` vs `complete` downgrade — SQLCipher independently encrypts at rest with a per-install key, so file-level protection is defense-in-depth, not the only line.
- `ensureProtection` retry cap being permanent post-success — new SQLite WAL/SHM files inherit the parent directory's protection class, and the directory's class was set during the initial walk. Verified safe.

aataraxiaa (approved):
Approving based on test scenarios passing (migration worked, did not reproduce crash). Agree with @jaceklyp’s comments. Look like small fixes and should be good to go once they are made.

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
Bunn: fixed on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/ce67886b3deb99973658893e78fc5f2d99928513

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
Bunn: fixed on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/ce67886b3deb99973658893e78fc5f2d99928513

iOS/PixelDefinitions/pixels/definitions/aichat_native_storage_pixels.json5
Bunn: fixed on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/22697f107433c7e08d1f54bf60cd510fe98931e7

iOS/PixelDefinitions/pixels/definitions/aichat_native_storage_pixels.json5
Bunn: fixed on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/22697f107433c7e08d1f54bf60cd510fe98931e7

iOS/DuckDuckGoTests/AIChat/DuckAiNativeStorageContainerMigrationTests.swift
Bunn: fixed on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/f32486a0910ca9f5d55f9bff8a9b675c5163e5e2

iOS/DuckDuckGo/AIChat/DuckAiNativeStorageContainerMigration.swift
Bunn: Fixed on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/d72711efeb3ed1af00ae6cda67ebf9920c00e75d, also did a small refactor to better encapsulate the kvs calls

jaceklyp (approved):
nice work :)

iOS/DuckDuckGo/AppLifecycle/AppStates/Launching.swift
Bunn: done on https://github.com/duckduckgo/apple-browsers/pull/4906/commits/bfcb0fe679ae5083ea2e4bb286d64f28c98201cc

PR #4977, opened 2026-05-20
Fix duck.ai suggestion truncation and View all chats double selection

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1214450331857403
https://app.asana.com/1/137249556945/project/1204912272578138/task/1212943926426596
https://app.asana.com/1/137249556945/project/1148564399326804/task/1213734134708912?focus=true

### Description

Seven small fixes to address bar suggestions, all related to the new duck.ai surface:

1. **"Ask privately" wrapped to two lines on long input.** The user-text attributed string built in `SuggestionTableCellView.display(userText:style:icon:isBurner:)` did not include a paragraph style, so the `byTruncatingTail` line break mode set in the XIB was overridden by the attributed string default and the text wrapped. The regular suggestion path goes through `SuggestionViewModel.tableCellViewAttributedString`, which already applies `SuggestionViewModel.paragraphStyle` — use the same paragraph style for the search / aiChat / visit cells so very long input stays on one line and truncates with "…".

2. **"View all chats" stayed highlighted after switching to keyboard navigation.** `AIChatSuggestionsView.updateSelection(_:isKeyboardNavigating:)` already clears `isHovered` on every regular row when keyboard nav starts, but the same clear was missing for the "View all chats" footer row, so it kept its hover highlight in addition to the keyboard-selected row. Clear the footer row's hover too.

3. **Suggestion window didn't reopen on edit when backend returned nothing.** `AddressBarTextField.subscribeToSuggestionResult` opened the window only when `result.count > 0`. For long input that returns no backend suggestions, refocusing and typing another character left the window hidden even though the synthetic "Ask privately" / "Search the web" rows would otherwise display. Gate on `numberOfRows > 0` instead, mirroring the close-side check in `SuggestionViewController.displayNewSuggestions`.

4. **Address-bar suffix stayed " – Search DuckDuckGo" when the Ask-privately row was keyboard-selected.** The synthetic aiChat row has no real `SuggestionViewModel`, so selection always fell through to the default search suffix. Synthesize a `.askAIChat` view model in `updateSelectedSuggestionViewModel()` when that row is selected, and route a new `Suffix.aiChat` case to " – Duck.ai" — matching the dropdown row's own suffix.

5. **"⌃ ⏎" key caps used the wrong color in the highlighted Ask-privately chip.** The caps hardcoded `selectedSuggestionTint` (pure white) while the chip label and arrow use `accentContentSecondary`. In dark mode this made the caps read as foreign elements. Switch the caps to the same design-system token.

6. **Duck.ai row highlight rendered the light-mode accent on dark-mode windows.** The CALayer-based row was assigning `.cgColor` of the dynamic accent color from a property `didSet`, where `NSAppearance.current` defaults to `.aqua` and resolves the wrong variant. Run the highlight setup inside `effectiveAppearance.performAsCurrentDrawingAppearance` and refresh on `viewDidChangeEffectiveAppearance`.

7. **Keyboard cycling trapped the user on the only selectable row.** With the search and Ask-privately synthetic rows visible but no backend suggestions, arrowing past the ends wrapped back to the first/last row (gated by the `aiChatOmnibarToggle` flag) — never landing on the "no row selected → just typed input" state. Drop the wrap branch in `wrapAroundOrClearSelection`, matching what `AIChatSuggestionsViewModel` already does for the duck.ai chat list.

### Testing Steps

1. Type a very long search query in the address bar. The "Ask privately" row stays on one line and truncates with "…".
2. Open duck.ai mode with enough chats to show "View all chats". Hover that row, then press up; only one row highlighted. Reverse direction; still only one row highlighted.
3. Type a very long input that returns no backend suggestions, unfocus the address bar, click back in, and type another character. The window reopens with the "Ask privately" / "Search the web" rows.
4. With any input in search mode, arrow down past the search row onto the Duck.ai row. Address-bar suffix shifts from " – Search DuckDuckGo" to " – Duck.ai" and back as you cycle.
5. In dark mode, hover or arrow-select the Duck.ai row. The "⌃" / "⏎" key caps read in the same near-black tone as "Ask privately" and the arrow.
6. In dark mode, compare the row highlight in search-mode suggestions and duck.ai chat suggestions — the blue should match.
7. With only the two synthetic rows visible, arrow down repeatedly. Cycle goes: search row → Ask-privately row → no row selected (just typed input) → back to search row. Arrow up does the mirror.

### Impact and Risks

Low: visual / behavioural fixes to the suggestions dropdown. No new model state — only view-layer wiring and the existing selection/value plumbing.

#### What could go wrong?

Fix 4 makes the address-bar value transition into `.suggestion(synthetic)` whenever the Ask-privately row is selected. The synthetic view model produces the same `string` as the user's typed input, so no visible text change. Fix 7 changes keyboard cycling for users with the `aiChatOmnibarToggle` feature flag on (currently the dev/internal default) — they lose the wrap-around and gain an escape-to-typed-input state. Fix 6's `viewDidChangeEffectiveAppearance` override may run on every appearance toggle, but `updateAppearance` is cheap (a CATransaction with disabled animations).

### DIFF
--- a/macOS/DuckDuckGo/AIChat/Suggestions/AIChatSuggestionRowView.swift
+++ b/macOS/DuckDuckGo/AIChat/Suggestions/AIChatSuggestionRowView.swift
@@ -34,19 +34,11 @@ protocol SuggestionRowThemeProviding {
 /// Default implementation that uses the app's theme manager.
 struct DefaultSuggestionRowThemeProvider: SuggestionRowThemeProviding {
     var accentPrimaryColor: NSColor {
-        var color: NSColor = .controlAccentColor
-        NSAppearance.withAppAppearance {
-            color = NSApp.delegateTyped.themeManager.theme.palette.accentPrimary
-        }
-        return color
+        NSApp.delegateTyped.themeManager.theme.palette.accentPrimary
     }
 
     var selectedTintColor: NSColor {
-        var color: NSColor = NSColor(designSystemColor: .accentContentPrimary)
-        NSAppearance.withAppAppearance {
-            color = NSApp.delegateTyped.themeManager.theme.palette.accentContentPrimary
-        }
-        return color
+        NSApp.delegateTyped.themeManager.theme.palette.accentContentPrimary
     }
 
     var suggestionHighlightCornerRadius: CGFloat {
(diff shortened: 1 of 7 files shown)


jotaemepereira (approved):
This looks great @tomasstrba

PR #4982, opened 2026-05-20
Fix IME return handling in address bar

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199178362774117/task/1214960575971803?focus=true

### Description
Fixes the native macOS address bar Return handling while IME marked text is active:
- lets IME composition-confirm Enter continue through AppKit instead of being handled by the window key monitor
- adds defensive field delegate handling for any newline command received while marked text is still active
- covers both paths in AddressBar integration tests

### Testing Steps
1. In System Settings > Keyboard > Text Input, enable Japanese - Romaji and switch to Hiragana.
2. Run DuckDuckGo macOS.
3. Focus the address bar and type `tesuto`.
4. Press Enter once.
5. Confirm the address bar contains `てすと`, remains focused, and no search/navigation has occurred.
6. Press Enter again and confirm search/navigation submits.
7. Disable `addressBarIMEConfirmFix`, repeat steps 3-4, and confirm the first Enter submits immediately, matching the pre-fix rollback behavior.
8. Optional: repeat while suggestions are visible and with Shift/Option/Command+Enter during composition to confirm they do not submit or trigger alternate actions before composition confirmation.

### Impact and Risks
Low

#### What could go wrong?
Return shortcuts in the address bar could fail to submit or trigger AI Chat if marked-text detection is incorrect. The guard only applies while the active field editor has marked text, and tests cover the key monitor and delegate fallback paths.

### Quality Considerations
- Added integration coverage for the local key monitor path used by `MainViewController.customKeyDown`.
- Added defensive coverage for the field delegate newline path.
- Manual QA confirmed Japanese-Romaji/Hiragana Enter first confirms composition and the second Enter submits.
- `git diff --check` passed.
- XCTest suite not run locally because repo instructions require explicit permission before running tests.

### Notes to Reviewer
- The MainViewController guard is the primary fix because it runs before AppKit/IME via the local event monitor.
- The AddressBarTextField guard is defensive for input paths that reach the delegate while marked text is still active.
- The local Xcode scheme language/region change used for manual Japanese testing was intentionally left out of this PR.

---

> [!NOTE]
> **Low Risk**
> Low risk: behavior is gated behind a remote flag and only skips Return handling while marked composition text is active; mis-detection could affect submit shortcuts but tests cover the main paths.
> 
> **Overview**
> Adds a remotely toggled **`addressBarIMEConfirmFix`** flag and changes macOS address bar **Return** handling so the first Enter during IME composition confirms text instead of navigating or searching.
> 
> **`MainViewController`** no longer consumes Return in the local key monitor when the address bar field editor has marked text (primary path). **`AddressBarTextField`** applies the same check in the newline delegate path as a fallback. Integration tests cover both paths with the flag on and off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a624c6e412c73e1d38b3c32d809af6e0cd9dbe2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ayoy (approved):
Working great now, thanks @jotaemepereira!

PR #4983, opened 2026-05-20
macOS: Fix wrong drop index moving a bookmark into the overflow menu

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1214230995537229?focus=true
Tech Design URL:
CC:

### Description

Dragging a bookmark from the bookmarks bar onto the overflow popover (`»` clipped-items menu) and dropping it inside the popover put the bookmark in the wrong place — frequently back inside the bar. Moving a bookmark into a *sub-folder* of the overflow worked correctly; only moves that stayed at the same (top) level were broken.

Root cause is in `BookmarkOutlineViewDataSource.outlineView(_:acceptDrop:item:childIndex:)`. The clipped-items menu only shows the overflow subset of top-level entities, so the data source translates the visible outline-row index into a real top-level index by looking at the entity preceding the drop point. The cast used in that translation was `as? BookmarkFolder`, a leftover from when the code only handled folders-only mode (Bookmark Manager). When the preceding outline row was a plain `Bookmark`, the cast failed, the translation was skipped, and the raw outline-child index — which only counts the clipped subset — was sent to `BookmarkManager.move`, landing the bookmark back inside the bar.

The fix widens the cast to `as? BaseBookmarkEntity` so the translation works regardless of whether the preceding overflow row is a folder or a bookmark. `BaseBookmarkEntity` is `Equatable`, so the subsequent `siblings.firstIndex(of:)` lookup still compiles. Folders-only mode is unaffected (folders are still `BaseBookmarkEntity`), and trailing `SpacerNode` / `MenuItemNode` rows still correctly fall through (they aren't `BaseBookmarkEntity`).

### Testing Steps
1. Have a window narrow enough that the bookmarks bar overflows — at least one bookmark should be clipped behind the `»` button.
2. Drag a bookmark from the bookmarks bar onto the `»` button and wait for the overflow popover to open.
3. Drop the bookmark between two bookmarks already in the popover.
4. Expected: the bookmark stays at the dropped position inside the popover.
5. Also try with the dragged source coming from inside the popover itself (reorder within overflow) and dropping into a sub-folder of the overflow (which previously already worked) to confirm no regression.

### Impact and Risks

Low — small, localized change in a single drop-index branch that only widens an `as?` cast. The code path affected is exclusively the bookmarks-bar overflow / folders-only Bookmark Manager tree view; ordinary in-window bookmark reorders use a different branch.

#### What could go wrong?
- Folders-only Bookmark Manager tree view (the original consumer of this branch) could regress. — Mitigated: folders are also `BaseBookmarkEntity`, so behavior is unchanged when every outline row is a folder.
- A trailing `SpacerNode` / "Open all in new tabs" row at the end of the overflow could now be picked up as the nearest neighbor. — Mitigated: neither type is a `BaseBookmarkEntity`, so the cast still fails and the existing fallback path is used.

### Quality Considerations
- New regression test `testWhenDroppingBookmarkIntoClippedItemsMenu_AndPrecedingRowIsBookmark_ThenMoveIndexIsRealTopLevelIndex` covers the exact scenario from the report (preceding overflow row is a `Bookmark`, not a folder). Verified to fail without the fix and pass with it; the surrounding 12 tests in the file still pass.
- `BookmarkStoreMock.move(...)` now captures `toIndex` (previously only `objectUUIDs` and `parentFolderType` were captured) so future drop-index regressions can be asserted directly.
- No privacy/security or performance impact.

### Notes to Reviewer
The widened cast preserves the existing `siblings.firstIndex(of:)` lookup — `BaseBookmarkEntity` is `Equatable` and both `Bookmark` and `BookmarkFolder` inherit from it, so the index math is identical, only the gate on entering the branch changes.

---

> [!NOTE]
> **Low Risk**
> Low risk, localized change to drag-and-drop index calculation for bookmarks overflow/clipped-items menus plus a regression test; minimal impact outside this specific UI path.
> 
> **Overview**
> Fixes a drag-and-drop bug where dropping a bookmark into the bookmarks bar overflow (clipped-items) menu could compute the wrong destination index and place the item back in the visible bar.
> 
> The drop-index translation in `BookmarkOutlineViewDataSource.outlineView(_:acceptDrop:item:childIndex:)` now treats the nearest preceding row as any `BaseBookmarkEntity` (not just `BookmarkFolder`), so visible outline indices are correctly mapped to real sibling indices even when the preceding item is a bookmark.
> 
> Adds a unit test covering the overflow-menu reorder scenario and extends `BookmarkStoreMock` to capture `toIndex` so tests can assert the computed move target.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 032cacbc794e2327a2b778b5711b4b35bf8f1cd4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/Bookmarks/Model/BookmarkOutlineViewDataSource.swift
+++ b/macOS/DuckDuckGo/Bookmarks/Model/BookmarkOutlineViewDataSource.swift
@@ -325,11 +325,12 @@ final class BookmarkOutlineViewDataSource: NSObject, BookmarksOutlineViewDataSou
         }
 
         let index = {
-            // for folders-only calculate new real index based on the nearest folder index
+            // for folders-only and clipped-items menus the outline doesn‘t show every sibling,
+            // so translate the visible outline index into a real index in the parent folder.
             if contentMode == .foldersOnly || representedObject is PseudoFolder,
                index > -1,
-               // get folder before the insertion point (or the first one)
-               let nearestObject = (outlineView.child(max(0, index - 1), ofItem: item) as? BookmarkNode)?.representedObject as? BookmarkFolder,
+               // get the entity before the insertion point (or the first one)
+               let nearestObject = (outlineView.child(max(0, index - 1), ofItem: item) as? BookmarkNode)?.representedObject as? BaseBookmarkEntity,
                // get all the children of a new parent folder (take actual bookmark list for the root)
                let siblings = ((representedObject is PseudoFolder ? nil : representedObject) as? BookmarkFolder)?.children ?? bookmarkManager.list?.topLevelEntities {
 
--- a/macOS/DuckDuckGo/Bookmarks/Services/BookmarkStoreMock.swift
+++ b/macOS/DuckDuckGo/Bookmarks/Services/BookmarkStoreMock.swift
@@ -238,9 +238,11 @@ final class BookmarkStoreMock: BookmarkStore, CustomDebugStringConvertible {
 
     var moveObjectUUIDCalled = false
     var capturedObjectUUIDs: [String]?
+    var capturedToIndex: Int?
     func move(objectUUIDs: [String], toIndex: Int?, withinParentFolder: ParentFolderType, completion: @escaping (Error?) -> Void) {
         moveObjectUUIDCalled = true
         capturedObjectUUIDs = objectUUIDs
+        capturedToIndex = toIndex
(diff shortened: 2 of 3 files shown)


mallexxx (approved):
LGTM! 🔥

PR #4882, opened 2026-05-14
iOS: Fix SwiftUI Previews on iOS browser scheme

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214604323343316?focus=true

### Description
Fixes SwiftUI Previews for iOS and macOS project.
- iOS Browser previews work in the new (Xcode 26 default) execution mode.
- macOS Browser previews work with Legacy Previews Execution enabled (Editor → Canvas → Use Legacy Previews Execution).
- Modules marked `??` below were not individually tested (applied 80/20 rule). The only known failure is the Onboarding package, investigated in the follow-up below.

Module                       | # Previews | Fixed
-----------------------------|-----------:|:----:
iOS Browser                  |         71 |  ✅
macOS Browser                |         70 |  ✅ (legacy mode)
Onboarding                   |         10 |  ❌
UIComponents                 |          3 |  ✅
AutofillCredentialProvider   |          3 |  ??
DataBrokerProtection-macOS   |          2 |  ??
NetworkProtectionUI          |          2 |  ??
SetDefaultBrowserUI          |          2 |  ??
SubscriptionUI               |          1 |  ??
SwiftUIExtensions            |          1 |  ??
SyncUI-iOS                   |          1 |  ??
iOS Widgets Extension        |          1 |  ??

### Onboarding Module Follow-Up
The current PR fixes SwiftUI Previews for the main iOS Browser target, but not previews opened directly from the Onboarding Swift package.
Investigation showed that Onboarding package previews crash when the preview agent loads the transitive PixelKit -> Common -> URLPredictor dependency graph. Removing direct Common usage helped, but previews still crashed while PixelKit remained in the Onboarding target. Removing PixelKit from Onboarding made the package previews work again.

### How
Extends the existing `ENABLE_PREVIEWS` guard in our build scripts to also recognize `ENABLE_XOJIT_PREVIEWS` (Xcode JIT previews), and adds matching skips to the PacketTunnelProvider embed phase, the ContentScopeScripts strip script, and the loc-update script — these either mutate the `.app` bundle previews are using or just spam the canvas with warnings.

### Testing Steps
- [x] Open `macOS/DuckDuckGo-macOS.xcodeproj`, toggle **Editor → Canvas → Use Legacy Previews Execution** on, open any SwiftUI view in the macOS Browser target (e.g. `QuitSurveyView.swift`), confirm preview
  renders without the `CouldNotFindLibrary "libswiftWebKit.dylib"` error.
[...]

PR #5021, opened 2026-05-22
iOS: Fix SwiftUI previews by making loc update idempotent

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214604323343316?focus=true
Tech Design URL: N/A
CC:

### Description
Follow up of https://github.com/duckduckgo/apple-browsers/pull/4882 that was reverted due to bugs introduced.
Fixes iOS SwiftUI Previews without using `ENABLE_XOJIT_PREVIEWS` to skip build phases.

The previous approach made previews render by skipping build scripts when `ENABLE_XOJIT_PREVIEWS=YES`, but that variable is also set for normal new-preview-capable Debug builds. As a result, normal app launches could skip build phases that still need to run.

This PR keeps the `Update Localizable.strings` build phase enabled and makes `loc_update.sh` safe to run during previews. The script now extracts localization strings into a temporary directory, converts the generated file to UTF-8 there, compares it with the existing checked-in `Localizable.strings`, and only replaces the checked-in file when the content actually changed. This avoids touching source-tree localization files on every build, which was causing Xcode automatic preview builds to restart repeatedly.

Preview mode expectations:

- iOS should use the new preview execution system. Legacy iOS previews fail because the legacy preview thunk cannot link private/new SwiftUI pieces such as `SwiftUICore` and local Swift package dependencies used by the app target.
- macOS should use Legacy Previews Execution. Xcode 26's new macOS preview analyzer fails while resolving Swift WebKit overlay libraries, specifically the `libswiftWebKit.dylib` path behind the SDK/cryptex layout for our deployment target. Legacy mode avoids that analyzer failure.

### Testing Steps

1. Open `iOS/DuckDuckGo-iOS.xcodeproj`.
2. Leave **Editor -> Canvas -> Use Legacy Previews Execution** disabled for iOS.
3. Open a SwiftUI preview in the iOS Browser target.
4. Confirm the preview renders while the `Update Localizable.strings` build phase remains enabled.
5. Confirm Xcode no longer repeatedly restarts automatic preview builds after `loc_update.sh` runs.
6. For macOS, open `macOS/DuckDuckGo-macOS.xcodeproj`, enable **Editor -> Canvas -> Use Legacy Previews Execution**, and confirm a Browser preview renders in legacy mode.

### Impact and Risks

Low. This changes build tooling only. Runtime app behavior should not change.

#### What could go wrong?

The localization update script now writes through a temporary directory and compares generated output before replacing files. If that comparison logic is wrong, localization files might not update when source strings change, or they might still be touched unnecessarily. The script still replaces `Localizable.strings` when generated content differs, so normal localization updates should continue to work.

### Quality Considerations

This avoids using `ENABLE_XOJIT_PREVIEWS` as a proxy for "currently rendering a preview." In Xcode's new preview system, Debug build products are preview-capable, so build scripts must be safe to run in normal Debug builds and preview builds alike.

The fix specifically addresses source-tree churn from `loc_update.sh`: unchanged generated localization content no longer updates file modification times, which prevents Xcode from invalidating and restarting previews in a loop.

### Notes to Reviewer

The key behavior to review is that `loc_update.sh` only updates checked-in `Localizable.strings` files when the generated UTF-8 content differs.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Low risk build-tooling change, but it alters how `Localizable.strings` files are generated and may cause missed updates or continued churn if the compare/convert logic is incorrect.
> 
> **Overview**
> Keeps the `Update Localizable.strings` build phase always running (removes the `ENABLE_PREVIEWS` early-exit) so Debug builds and SwiftUI previews don’t accidentally skip required scripts.
> 
> Refactors `loc_update.sh` to be *idempotent*: extracts strings into a temp directory, converts to UTF-8, compares against the existing `Localizable.strings`, and only replaces the checked-in file when content changes (with added error handling and cleanup traps).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f3417cf918927609a6b82d9edea5fb5cd72b66e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
+++ b/iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
@@ -12203,7 +12203,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [ $ENABLE_PREVIEWS == \"YES\" ]\nthen\n  exit 0\nelse\n  \"$SOURCE_ROOT/scripts/loc_update.sh\"\nfi\n";
+			shellScript = "\"$SOURCE_ROOT/scripts/loc_update.sh\"\n";
 		};
 		B6409DC62BC7D9BF00D66F9E /* SwiftLint */ = {
 			isa = PBXShellScriptBuildPhase;
--- a/iOS/scripts/loc_update.sh
+++ b/iOS/scripts/loc_update.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 # Get the directory where the script is stored
 script_dir=$(dirname "$(readlink -f "$0")")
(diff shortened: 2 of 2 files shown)


miasma13 (approved):
Worked as expected 👍

PR #5026, opened 2026-05-22
iOS: Reset UTI state when frontend signals a new chat

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1215056187001453?focus=true

### Description

On a duck.ai chat tab with the Unified Input feature flag enabled, tapping **Start new chat** in the duck.ai webview sidebar left the host's unified input in stale state (previous input text and attachments still visible, collapsed/expanded state wrong). The sidebar entry point is frontend-only and never reached the native host.

This adds a dedicated `newChatStarted` FE→native user-script message and routes it to a single UTI reset, used by **both** the sidebar and the native chat-header new-chat button. The native header button no longer resets state up front — it submits to the FE and the FE round-trip drives the reset, so there's one code path with one source of truth.

### Testing Steps

**Prerequisites:**
- Unified Input feature flag enabled.
- Point Duck.ai at the dev domain that ships the `newChatStarted` message: `https://euw-serp-dev-testing20.duck.ai/`.

**Steps:**
1. Open a duck.ai chat tab on the dev domain and submit a prompt so there's chat content.
2. Tap the top-left button in the chat header to open the duck.ai sidebar.
3. Tap **New chat** in the sidebar.
4. Expected: a new chat is started, the UTI is expanded, and its state is reset (no stale text or attachments).

### Impact and Risks

**Low** — change is scoped to the duck.ai chat tab + Unified Input flag path.

#### What could go wrong?
- The native header **new chat** button used to reset UTI state synchronously before pushing to the FE. After this change it relies on the FE round-trip. If the FE ever stops emitting `newChatStarted` (e.g. version downgrade, regression), the native button will appear to do nothing — but that is also the current behaviour for the sidebar path, so this is a net improvement, not a regression.
- The delegate hook dispatches to the main queue because the user-script handler is invoked off the main actor. Verified on simulator: calling `startNewChat()` synchronously from the async path otherwise trips a UIKit autolayout assertion and crashes. Hopping to main fixes it for both the real message path and the eventual swift-concurrency main-actor migration.

### Quality Considerations
- Added a unit test (`testDidReceiveNewChatStartedNotifiesDelegate`) verifying the message routes to the new delegate method.
- macOS handler falls through `default:` for unknown cases, so the new enum case is a no-op on macOS until/unless macOS wires it up.

### Notes to Reviewer
- The `newChatStarted` message is intentionally fire-and-forget with no params — it's an event, not a request. Same shape as `voiceSessionStarted`.
- The default no-op in the `AIChatContentHandlingDelegate` extension means `AIChatContextualSheetViewController` (and the test mock) don't need to opt in.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Scoped to duck.ai full-tab + Unified Input; native new-chat now depends on the FE emitting `newChatStarted`, same as the sidebar path.
> 
> **Overview**
> Adds a **`newChatStarted`** duck.ai web→native bridge message so iOS can reset **Unified Toggle Input** (clear text/attachments, expand AI chat mode) when the user starts a new chat from the **in-webview sidebar**, not only from native chrome.
> 
> **`AIChatContentHandler`** forwards the message to a new delegate callback; **`MainViewController`** runs `startNewChat()` and `showExpanded(inputMode: .aiChat)` on the **main queue**. The native header **New chat** control no longer resets UTI locally—it only calls **`submitStartChatAction()`** and waits for the FE event so sidebar and header share one path.
> 
> Unit test and mock delegate updated; macOS ignores the new enum case until wired.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f557b15db735313de0feb3d81ca509b4cd323ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pikorddg (approved):
The fix looks correct (though I wasn't able to reproduce it on main), but the code looks good — approving!

PR #4544, opened 2026-04-23
Implement Duck.ai chat-path onboarding flow with tracker-blocking demo and subscription upsell

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1213551050320249
Tech Design URL: N/A
CC: N/A

### Description
- Add chat-first onboarding path for the Duck.ai query experiment: after fire step, show "try visiting a site" NTP prompt, tracker-blocked browsing dialog, and chat-specific "You've got this!" EOJ dialog
- Subscription promotion "Try it free" button opens duckduckgo.com/pro (via `SubscriptionURL.StaticURLs.proPageURL`) instead of in-app purchase flow when coming from the chat path
- Fix fire button pulse lifecycle: pulse starts with the fire education Dax dialog, is suppressed if the fire confirmation sheet is already on screen, stops permanently on fire completion
- Fix premature `disableContextualDaxDialogs()` when subscription promo is still pending
- Fix `isChatPathSubscriptionPromo` evaluating to false by the time "Try it free" is tapped (capture value at dialog creation, before `onFirstAppear` sets `subscriptionPromotionDialogSeen`)
- Harden control-group isolation: all chat-path branches gate on `chatPathVisitSiteSeen` (only set in experiment path)
- Add `SubscriptionURL.StaticURLs.proPageURL` constant to BSK instead of hardcoding the URL

#### Setup (required before each test run)
1. Enable the `onboardingDuckAIQueryTrackersDemoExperiment` feature flag in Settings → Debug → Feature Flags
2. Reset to a fresh-install state **without reinstalling**:
   - Go to Settings → Debug → Onboarding
   - Set **User Type** picker → **"New User"**
   - Tap **"Reset All Onboarding"** and confirm
   - Kill and relaunch the app — onboarding will restart from the beginning

#### Happy path
3. Complete linear onboarding via the Duck.ai path
4. On Duck.ai, submit a query and wait for response — verify fire button blinks and "That's Duck.ai" dialog appears
5. Tap fire → verify blink stops, confirmation sheet appears; cancel → verify blink restarts; tap fire early (before dialog) → verify no blink while confirmation is shown
6. Confirm fire → verify NTP opens and "Try visiting a site" prompt appears
7. Navigate to any site — verify tracker-blocked browsing dialog appears
8. Dismiss it → verify "You've got this!" (chat variant, no pointer/tail) appears on NTP with address bar active
9. Tap "Got it" → verify subscription promo appears
[...]

PR #5043, opened 2026-05-25
Switch conflict-watch cross-ref filter to three-dot diff

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203301625297703/task/1215097277192572?focus=true

### Description

The daily conflict-watch script was flooding Asana with tasks naming the same files: 4 OmniBar files on 05-21 (~12 tasks), `UnifiedToggleInput` files on 05-24 (~6 tasks), `AIChatContentHandler.swift` on 05-24 (~25 tasks). Root cause is in `_branch_files_vs_main`: it used two-dot `git diff main..branch` to decide whether a branch "actually touches" a conflicting file. Two-dot is symmetric — it also includes files where **main has moved ahead** of the branch's base, so any branch that predates a recently-landed multi-file PR looks like it touched the PR's files. The cross-reference filter then stops dropping the inherited files, and any pair of such branches gets a task naming them.

Example: `tom/fix-ntp-autocomplete-suggestions-setting` (macOS-only, 12 files in its 3-dot diff, no OmniBar) was paired with `sabrina/fix-pro-url-interception` (3 files in 3-dot diff, also no OmniBar) and the task named 4 OmniBar files. Neither branch's 3-dot diff against main includes OmniBar — both just predate #4544 (Duck.ai chat-path onboarding), which landed on main on 05-19.

Fix is one operator: `..` → `...` in `_branch_files_vs_main`. Three-dot is `diff $(merge-base main branch)..branch` — only what the branch contributed on top of its base, which is what the docstring already described as the intent. `_is_merged_into_default` keeps two-dot tip-tree semantics (its docstring explains why) and is unchanged. Docstring rewritten to spell out the contrast. Self-test reproducer added that fails with two-dot and passes with three-dot.

### Testing Steps

1. `ASANA_PAT=stub python3 scripts/conflict-watch/conflict-watch.py --self-test` — passes, including the new three-dot case.
2. Local `--once` probes against yesterday's false-positive pairs all return `has_conflicts: false`:
   - `pikor/uti/new-pixels..tom/fix-ntp-autocomplete-suggestions-setting`
   - `jacek/voice-mode-keep-tab-open..diego/route-handlers-through-tab-index-helper`
   - `jacek/voice-mode-keep-tab-open..tom/fix-ntp-autocomplete-suggestions-setting`
3. Control pair `pikor/uti/new-pixels..jacek/voice-mode-keep-tab-open` (both branches genuinely edit a UTI file) still surfaces — correctly as a soft conflict since their line ranges don't overlap.

### Impact and Risks

Low — internal tooling. No user-facing code path.

#### What could go wrong?

`_is_merged_into_default` deliberately uses two-dot tip-tree semantics; not touched. Stale false-positive tasks won't all auto-close on the next scheduled run — `sweep_existing_tasks` only closes pairs whose hard-conflict-line count drops below `CW_MIN_CONFLICT_LINES` (20). Most should fall below since the named files are the bulk of those totals, but a manual bulk-close may be wanted.

### DIFF
--- a/scripts/conflict-watch/conflict-watch.py
+++ b/scripts/conflict-watch/conflict-watch.py
@@ -860,21 +860,28 @@ _branch_files_vs_main_cache: dict[str, set[str]] = {}
 
 
 def _branch_files_vs_main(branch: str) -> set[str]:
-    """Set of files where the branch's tip differs from main's tip.
+    """Set of files the branch added on top of its merge-base with main.
 
-    Used by probe_pair to drop "false-positive via merge-from-main"
-    conflicts: a branch that merged main in inherits main's recent
-    edits, which from an old pair-merge-base look like the branch's
-    own changes. If diff(main..branch) doesn't include a file, the
-    branch's owner isn't actually contributing changes to it — so
-    pinging them about a conflict on that file is misleading.
+    Used by probe_pair to drop "false-positive via inheritance" conflicts:
+    a branch whose apparent change to a file came from somewhere other
+    than its own commits (either an explicit merge-from-main or — far
+    more commonly — main moving ahead of the branch's base) shouldn't be
+    pinged when that file conflicts with another branch's real edit.
+
+    Uses three-dot (``main...branch``, == ``diff $(merge-base main
+    branch)..branch``) deliberately. Two-dot would also include files
+    where main moved past the branch's base — those show up in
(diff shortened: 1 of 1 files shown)


samsymons (approved):
Near as I can tell, this is a great change - thanks!

PR #5052, opened 2026-05-25
[iPad perf] Debounce tabManager.save and move write off main

dus7:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206226850447395/task/1214542891270915
Tech Design URL:
CC:

### Description

`tabManager.save()` is called on every tab loading-state change. On iPad at 200 tabs the underlying `NSKeyedArchiver` encode and disk write cost ~30 ms avg / 47 ms max on the main thread, contributing visible hangs during heavy switching workloads.

This change splits the save into a debounce step on the call path and an off-main disk write on a serial persist queue. Bursts coalesce, and a 1 s max-wait ceiling guarantees that the first save in a burst always lands within 1 s so sustained activity cannot starve persistence.

`TabsModelPersistence.save` uses `TabsModel.archivalSnapshot()` (race fix from #4828) so the encode is safe regardless of which thread the persist queue picks up.

`flushPendingSave()` cancels the pending debounce and persists synchronously via `saveSynchronously`. Wired into `willResignActive`, `willTerminate`, `didReceiveMemoryWarning`, and the data-clearing path so we don't lose state at lifecycle boundaries.

The new logic is behind the `tabsSaveOptimization` feature flag (enabled by default).

### Testing Steps

1. Open the app on an iPad with 10+ tabs across heavy media pages (YouTube, Netflix, news).
2. Switch rapidly between tabs and open a few background tabs from links. Confirm the app feels at least as responsive as main, with no visible regression in tab switching.
3. Trigger backgrounding: tap the home gesture or press Cmd+H in the simulator. Re-open the app, switch through tabs and confirm the state persists (URLs, current selection, tabs bar order).
4. Force-quit from app switcher mid-load while a few tabs are still loading. Re-launch and confirm the latest committed URLs are restored.
5. Trigger a memory warning in the simulator (Hardware > Memory > Memory Warning) and confirm the app does not crash and tabs survive a relaunch.
6. Clear data via fire button or settings. Confirm the data-clearing flow returns the real `Result` (no regression in clear-on-failure paths).
7. Disable the `tabsSaveOptimization` flag in the debug menu (or via remote config). Repeat steps 2-6 and confirm the synchronous fallback path behaves like main.

### Impact and Risks

Medium. Touches the persistence path used by every tab open, close, switch, and reorder. Lifecycle hooks and data-clearing path were rewired.

#### What could go wrong?

- The app is suspended between scheduling and writing the debounced save. Mitigated by force-flushing on `willResignActive`, `willTerminate`, `didReceiveMemoryWarning`, and on the data-clearing path before the `Result` is returned.
- Sustained activity could starve the trailing-edge debounce. Mitigated by the 1 s max-wait ceiling.
- `save()` now returns `Void`. No caller read the previous `Result`. The data-clearing path is the only consumer that needs the real result and it explicitly uses `flushPendingSave()` -> `saveSynchronously`. Encode/write errors continue to fire `tabsStoreSaveError` daily pixel. cc @hassaanelgarem 
- New path could regress in the wild. Mitigated by the `tabsSaveOptimization` failsafe switch.

### Quality Considerations

Measured A/B on iPad A16 (iOS 26.4), 12-tab heavy-media workload, both runs ~42-45 s:

| metric                              | main      | this branch | delta   |
|-------------------------------------|----------:|------------:|--------:|
| `TabManager.save` count / s         | 2.81      | 2.44        | similar |
| `TabManager.save` avg duration      | 7.28 ms   | 0.06 ms     | -99.2%  |
| `TabManager.save` cumulative / s    | 20.50 ms  | 0.15 ms     | -99.3%  |
| Big-hang rate (>=500 ms) / s        | 0.095     | 0.089       | similar |

Sustained-activity run (33.5 s, signposts split queue-wait vs write):

| signpost                              | n   | p50      | p95      | max      |
|---------------------------------------|----:|---------:|---------:|---------:|
| `TabsModelPersistence.save.enqueued`  | 84  | 1.55 ms  | 9.02 ms  | 13.3 ms  |
| `TabsModelPersistence.save.write`     | 84  | 2.07 ms  | 6.50 ms  | 11.9 ms  |
| inter-fire gap (write-start to next)  | 83  | 12 ms    | 1009 ms  | 1531 ms  |

84 fires over 33.5 s comfortably above the 25-33 floor; inter-fire p95 ~1 s confirms the 1 s max-wait ceiling triggers periodic flushes under load. Queue-wait and write both stay sub-15 ms in the worst case.

### Notes to Reviewer

- The `dispatchPrecondition(condition: .onQueue(persistQueue))` inside `write()` exists because `DispatchQueue.sync` on a serial queue runs the block inline on the calling thread: "running inside the persistQueue serial section" regardless of which thread runs it.
- `flushPendingSave()` and the 4 lifecycle `@objc` selectors (`onApplication{BecameActive,WillResignActive,WillTerminate}`, `onMemoryWarning`) are `@MainActor`, so the main-thread requirement is enforced at compile time rather than via a runtime `dispatchPrecondition`. `save()` is already `@MainActor`.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes the global tab persistence path used on every tab switch/load; mitigations include lifecycle flushes, max-wait debouncing, and a remote kill switch, but async saves can still race with suspend if flush hooks miss an edge case.
> 
> **Overview**
> Introduces **`tabsSaveOptimization`** (remote kill switch, on by default) and reworks how iOS persists open tabs when `tabManager.save()` runs often during tab activity.
> 
> When the flag is on, **`TabManager.save()`** no longer blocks on archive + disk I/O: it **debounces** bursts (300 ms quiet, **1 s max-wait** per burst) and **`flushPendingSave()`** cancels debounce and persists synchronously with a real **`Result`**. The flag off path uses **`flushPendingSave()`** immediately (prior synchronous behavior). Lifecycle hooks **`willResignActive`**, **`willTerminate`**, and **memory warning** force flush so debounced work is not lost; **data clearing** uses **`flushPendingSave()`** instead of **`save()`**.
> 
> **`TabsModelPersistence`** archives on the caller thread, enqueues file writes on a **serial `userInitiated` queue**, and adds **`flush()`** / **`saveSynchronously()`**. Call sites drop the unused **`save()`** return value. Tests cover debounce coalescing, max-wait under sustained saves, and immediate flush.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 768d18e2237d3cd4e85d242e2e2164d444613a47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


brindy (approved):
LGTM - note the unit test failures and the AI comment though :)

PR #4885, opened 2026-05-14
iOS: Voice mode opens in new tab; X button closes Duck.ai tab

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214810245123542

### Description

Two voice-mode behavior changes on Duck.ai:

1. **Close tab on X tap.** Listen for the new `voiceSessionUserEnded` FE message (added by Moon in [duckduckgo/ddg#45732](https://dub.duckduckgo.com/duckduckgo/ddg/pull/45732)) and close the originating tab. We can't reuse `voiceSessionEnded` because it also fires on idle timeout / network drop — closing the tab in those cases would be hostile. macOS ignores the new message. No iOS flag added — the FE only sends this message when `supportsNativeChatInput` is reported true (which already requires `unifiedToggleInput`), so the path is transitively gated.

2. **Open voice mode in a new tab when current tab has content** (gated by `unifiedToggleInput`). Previously, tapping a voice entry point on a website replaced the page; now voice opens in a new tab. Empty/NTP tabs are still reused. Deeplink path is preserved with its prior new-tab behavior regardless of the flag.

### Testing Steps

**Setup:** Enable the `unifiedToggleInput` feature flag in Settings → Debug → Feature Flags. In Settings → Debug → AI Chat, set the custom AI chat URL to `https://euw-serp-dev-testing20.duck.ai` (the FE dev instance for the new `voiceSessionUserEnded` message).

**Close tab on X tap:**
1. Open a new Duck.ai tab, enter voice mode, tap the X (close) button → the tab closes.
2. Open a new Duck.ai tab, enter voice mode, let it idle out (don't tap X) → the tab stays open.

**Voice mode opens in new tab (with `unifiedToggleInput` ON):**
3. On the NTP, tap the voice button (toolbar / address bar / NTP shortcut) → voice loads in the current tab.
4. On any website (e.g. duckduckgo.com), tap the voice button → a new Duck.ai voice tab opens; the website tab stays as it was.
5. On an existing Duck.ai chat tab, tap the voice button → a new Duck.ai voice tab opens; the prior chat is preserved.
6. Deeplink into voice mode (when current tab has content) → still opens in a new tab.

**With `unifiedToggleInput` OFF (sanity):**
7. On a website, tap the voice button → voice loads **in the current tab** (legacy behavior).
8. Deeplink with content in current tab → still opens in new tab (unchanged).

### Impact and Risks

[...]

PR #5053, opened 2026-05-25
iOS: Voice mode keep Duck.ai tab open on X tap; align UTI with FE

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214934861426454
Tech Design URL:
CC:

### Description

Partial revert of #4885 plus an alignment of UTI visibility with the Web/Android model, following the spec change [from the Asana thread](https://app.asana.com/1/137249556945/task/1214934861426454) (Aitor's TL;DR after syncing with Pete): *"voice mode can open a new tab, but ending voice mode should not close that tab."*

- Stop closing the originating tab on `voiceSessionUserEnded`; FE now renders the post-voice view in the same tab. The "voice opens in a new tab when current tab has content" half of #4885 is preserved.
- Drop the defensive native show-on-voice-end UTI override (both current-tab and background-tab paths). FE drives `hideChatInput`/`showChatInput` exclusively — same model Android uses.
- Replace the URL-based pre-emptive UTI hide with an intent flag (`voiceMode`) threaded through `loadUrlInNewTab` / `addTab`, so the bottom chrome doesn't flash white when opening voice from a non-Duck.ai tab.
- Remove the now-dead `voiceSessionUserEnded` enum case from the shared `AIChat` package (zero consumers across iOS and macOS).

### Testing Steps

**Setup**
1. Settings → Debug → Feature Flags → enable `unifiedToggleInput`.
2. Settings → Debug → AI Chat → set the custom AI chat URL to `https://euw-serp-dev-testing20.duck.ai` (the FE dev instance that emits the new voice-session messages used here).
3. Reopen the app so the new URL takes effect.

**Tap X in voice mode no longer closes the tab**
4. Open a new Duck.ai tab. Enter voice mode. Tap X. **Expected:** the Duck.ai tab stays open and the FE renders the post-voice view (transcript / new chat). Previously the tab closed.
5. Open a new Duck.ai tab. Enter voice mode. Let it idle out (don't tap X). **Expected:** tab stays open; UTI behavior follows whatever the FE sends.

**No bottom-chrome flash when opening voice from a non-Duck.ai tab**
6. Open a regular web page (e.g. `duckduckgo.com`).
7. Tap the voice shortcut (omnibar voice button, or swap the fire button in tab-bar customisation for the voice button and tap it). **Expected:** new Duck.ai voice tab opens edge-to-edge; no thin white bar in the bottom safe-area.
8. From an existing Duck.ai chat tab, tap the voice button. **Expected:** new tab opens cleanly, prior chat preserved (unchanged from #4885).
9. Deep-link into voice mode while the current tab has content. **Expected:** still opens in a new tab (unchanged from #4885).

**UTI visibility is FE-driven**
10. On a regular Duck.ai chat → UTI shows. Enter voice mode → UTI hides (`hideChatInput` from FE). End voice → UTI follows whatever the FE sends; no native restore.
11. Tap the conversation icon in the Duck.ai header, switch to a non-voice chat. **Expected:** UTI behavior follows FE messages, with no native override.

**Regression sanity**
12. Open a regular Duck.ai chat from the NTP / omnibar. UTI appears as before.
13. Tap "new chat" from the Duck.ai header. UTI appears (locally set to `.visible` in `startNewChat()` — kept for the local entry path).
14. Switch between AI and non-AI tabs. Toolbar/chrome reconciliation is unchanged.

### Impact and Risks

**Low–Medium.** Scoped to Duck.ai voice mode entry/exit and the UTI visibility contract.

#### What could go wrong?

- **FE doesn't send `showChatInput` for a chat view we expected** — UTI would stay hidden. Mitigated by keeping `aiChatInputBoxVisibility = .visible` in `startNewChat()` for the local new-chat entry point, and keeping the in-memory per-tab visibility store so prior FE-set state survives tab switches.
- **Voice cold-start tab restore** — if a tab was last on a voice URL and the app is killed/restored, the new `voiceMode` intent flag isn't persisted, so the synchronous chrome hide doesn't apply. UTI may flash briefly until FE sends `hideChatInput`. Not chasing in this PR (rare path); the FE generally navigates off `?mode=voice` once voice ends.
- **macOS** — untouched. `voiceSessionUserEnded` was already a no-op on macOS per #4885's description.

### Quality Considerations

- **Edge case — opening voice from omnibar shortcut vs. tab-bar swap vs. deep link**: all three now route through `loadUrlInNewTab(..., voiceMode: true)`, so chrome timing is consistent.
- **Recent chats bug**: Aitor reported voice chats not appearing in recent chats. Worked theory was tab destruction killing the WKWebView before the FE could persist; keeping the tab open should let the FE save naturally. Worth verifying alongside this PR — if still missing, that's an FE-side investigation, not iOS.
- **Test coverage**: removed two tests (`testDidReceiveVoiceSessionUserEndedNotifiesDelegate`, `testDidReceiveVoiceSessionEndedDoesNotNotifyVoiceSessionUserEndedDelegate`) since the delegate path they covered is gone. No new tests added; the surface that's left is exercised by existing `AIChatContentHandlerTests` (30 passing) and behaviour is best verified manually per Testing Steps.

### Notes to Reviewer

- The intent-flag approach (vs. URL inspection) is a deliberate response to the constraint that `tab.url` lags during a freshly-opened tab. The flag is set in `addTab` before `attachTab` triggers `refreshUnifiedToggleInput`, so the chrome refresh sees the right state on first read.
- `voiceSessionUserEnded` removal from the shared package is intentional. Verified zero references across `iOS/`, `macOS/`, `SharedPackages/`. If FE re-emits this message in future for an iOS-specific reaction, we can add it back behind a fresh delegate method.
- One existing URL inspection (`tabURL?.isDuckAIVoiceMode` at the `refreshAction` site) is preserved — that decides whether to auto-expand the UTI on a fresh tab. Different concern from the flicker fix and works correctly with persisted-on-disk voice URLs at cold start.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches tab lifecycle, unified input chrome, and the native/FE contract for voice; regressions would show as stuck-hidden UTI or brief chrome flash on voice entry.
> 
> **Overview**
> **iOS Duck.ai voice** no longer closes the tab when the user taps X to end voice—the shared `voiceSessionUserEnded` bridge message and the delegate path that called `closeCurrentTab()` are removed so the same tab can show the post-voice FE view.
> 
> **Unified toggle input (UTI)** visibility after voice ends is no longer forced visible in native code (`applyVoiceSessionTransition` / `TabInputState.applyingVoiceSessionTransition`); only the FE `hideChatInput` / `showChatInput` messages drive the bar, matching Android.
> 
> Opening voice in a **new tab** now threads a `voiceMode` intent through `loadUrlInNewTab` / `addTab` into `tab.isVoiceModeRequested`, which **one-shot** hides the bottom chrome on first AI-tab refresh (replacing URL-based pre-hide) so the safe area does not flash before the voice URL commits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cba58c0341010f87af0682aa44ff69888d1d0398. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aataraxiaa (approved):
Code looks good. Adding one comment in Asana re: edge case.

PR #5070, opened 2026-05-25
iOS: Add Navigation Bar toggle to Manage Duck.ai Shortcuts settings

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215119474036174
Tech Design URL:
CC:

### Description

> **Note on base branch:** this PR is stacked on top of #5068 (the feature flag declaration). Once #5068 lands the base will be changed to `main` automatically.

Adds storage and the UI row for the new **"Navigation Bar"** toggle in **Settings → AI Features → Manage Duck.ai Shortcuts**. The toggle is the user-facing control for showing/hiding the Duck.ai chrome shortcut in the iPad tabs bar.

**Storage layer**
- New `isAIChatNavigationBarUserSettingsEnabled` getter and `enableAIChatNavigationBarUserSettings(enable:)` setter on `AIChatSettingsProvider`.
- Concrete impl in `AIChatSettings` persists `aichat.settings.showAIChatNavigationBar` to the shared key-value store (default `true`), is gated on the global `isAIChatEnabled`, and broadcasts via the existing `.aiChatSettingsChanged` notification.
- `MockAIChatSettingsProvider` updated to match.

**Settings UI**
- New `SettingsCellView` row with the subtitle *"Show in the browser navigation bar or address bar, depending on window size."*
- Row is gated on **iPad** AND the `aiChatChromeShortcutIPad` feature flag, so it's invisible on iPhone and in production builds while the flag is off.
- Visibility logic is extracted to `DuckAIChromeShortcutVisibility.isSettingsRowVisible(isIPad:featureFlagger:)` so the platform check is injectable for tests.

**Tests** — 8 new tests in `AIChatSettingsTests`:
- Storage: default value, persistence, global-Duck.ai-disable interaction, notification broadcast.
- Visibility helper: iPad × flag matrix (4 cases).

### Testing Steps

1. Run `UnitTests/AIChatSettingsTests` — all pass.
2. Build the **iOS Browser** scheme on an iPad simulator.
3. Enable the `aiChatChromeShortcutIPad` feature flag via Debug → Feature Flags.
4. In the app, open Settings → AI Features → Manage Duck.ai Shortcuts → confirm the **Navigation Bar** toggle appears between "Address Bar" and "Voice Search" with the subtitle copy above.
5. Toggle it on/off — the value is persisted; no observer wired in this PR yet, so no visible browser-chrome change is expected here (that lands in the visibility-wiring PR).
[...]

PR #5075, opened 2026-05-25
iOS: Add Duck.ai chrome button + visibility wiring on iPad

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215105825212340
Tech Design URL:
CC:

### Description

> **Note on base branch:** stacked on top of #5070 (Navigation Bar storage + Settings row), which is itself stacked on #5068 (the feature flag). Base will roll up to \`main\` automatically as each lands.

Adds the **Duck.ai pill** to the iPad tabs bar — the text-only "open Duck.ai in a new tab" entry point — with full visibility gating. This is the chrome surface that the rest of the [iPad] – 1-click Duck.ai access project decorates (split-button icon, long-press menu, pixels, Maestro tests) in subsequent PRs.

### Testing Steps

1. Run \`UnitTests/AIChatSettingsTests\` — all 16 pass.
2. Build the **iOS Browser** scheme on an iPad simulator (iOS 18.5 known-good).
3. Enable internal-user mode in your sim (Settings → Debug, or your usual mechanism) so the \`.internalOnly\` feature flag is on.
4. Launch the app on iPad → confirm the **Duck.ai** pill appears in the tabs bar between the **+** button and the **fire** button.
5. Tap the pill → confirm Duck.ai opens in a new tab.
6. Open Settings → AI Features → Manage Duck.ai Shortcuts → flip **Navigation Bar** off → return to browser → confirm the pill disappears live (no app restart needed). Flip back on → confirm it reappears.
7. Disable Duck.ai globally (Settings → AI Features → main toggle off) → confirm the pill disappears even with Navigation Bar still on. Re-enable Duck.ai → it reappears.
8. On iPhone → confirm nothing changes (tabs bar is iPad-only).

### Impact and Risks

**Low** — protected by three independent gates:
- Feature flag \`aiChatChromeShortcutIPad\` (default \`.internalOnly\` → off for external users).
- Per-shortcut Navigation Bar preference (default on, but only visible in Settings when the flag is on).
- Global \`isAIChatEnabled\` short-circuits the per-shortcut getter.

`TabsBarViewController` is iPad-only (gated by `guard isPad else { return }` in `MainViewController.loadTabsBarIfNeeded`), so the button cannot leak to iPhone.

#### What could go wrong?

- If the `.aiChatSettingsChanged` notification stops firing for some reason (e.g. a future refactor breaks the broadcast), the chrome state would go stale until the next `viewWillAppear`. Acceptable — the Settings UI flips immediately because it re-reads on every binding access, and `viewWillAppear` fires on every return from a presented screen.
- The button is added to the stack even when hidden — it doesn't consume layout space (`isHidden = true` excludes it from `UIStackView` arrangement). No visual side effect.

### Quality Considerations

- Notification observer uses `addObserver(self:selector:)` with automatic teardown — no manual cleanup needed, no Combine import added.
- Mirrors the iOS pattern for chrome buttons (lazy property, action-target wiring through the delegate, intrinsic content size).
- No new strings — reuses `UserText.actionOpenAIChat` ("Duck.ai"), already localized.

### Notes to Reviewer

- Stacked on #5070 → review that first.
- The visibility logic deliberately lives next to the Settings-row helper in \`DuckAIChromeShortcutVisibility\` so a future refactor can see both rules in one place. If a third surface needs visibility logic later, it goes here too.

---

> [!NOTE]
> **Low Risk**
> iPad-only UI behind internal-only feature flag and user preference, with snapshot overlay explicitly keeping the pill hidden; no auth or data-path changes.
> 
> **Overview**
> Adds a **Duck.ai** text pill to the iPad tabs bar (between **+** and fire) that opens Duck.ai via a new delegate path—new tab when the current tab has content, otherwise in-place—matching macOS behavior.
> 
> Visibility is centralized in `DuckAIChromeShortcutVisibility.isChromeButtonVisible` (feature flag `aiChatChromeShortcutIPad` **and** the Navigation Bar shortcut preference). `TabsBarViewController` wires `featureFlagger`, starts hidden to avoid flicker, and refreshes on `viewWillAppear` and `.aiChatSettingsChanged` (Combine). `MainViewController` injects `featureFlagger` when loading the tabs bar; the blank snapshot overlay intentionally omits flag/settings so the pill stays off that chrome.
> 
> Unit tests cover the new visibility helper; iPhone is unchanged (tabs bar remains iPad-only).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02e8b12d641712231e4ce4c22c80530e71bce228. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Bunn (approved):
LGTM. One issue, tapping on the duck.ai button while on duck.ai is refreshing the same tab instead of opening a new one like macOS. Fine to merge as is and fix on a new PR.
Thanks

PR #5076, opened 2026-05-25
Open Duck.ai prompt in new tab when Duck.ai is the current tab

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1214486828098117?focus=true

### Description

When Duck.ai was already loaded in the current tab, submitting a prompt via the omnibar called `currentTab.reload()` so the loaded page's user script consumed the freshly stashed prompt and appended it to the existing conversation. That implicit "submit to the current chat" behavior surprised users and was inconsistent with how the omnibar works elsewhere.

This change routes the `.currentTab + isDuckAIURL + hasPrompt` branch in `WindowControllersManager.openAIChat` through `.newTab(selected: true)`, so every omnibar prompt submission starts a fresh chat. The original Duck.ai tab is left untouched.

Unchanged:
- `.currentTab + isDuckAIURL + chatID` (explicit navigation to a saved chat).
- `.currentTab` when the current tab is not Duck.ai (e.g. NTP omnibar still navigates the current tab to Duck.ai with the prompt).
- `.newTab` / `.newWindow` callers, including the Cmd+Enter path.

### Testing Steps

1. Open Duck.ai, send a couple of messages so the conversation has visible state.
2. With that tab active, type a new prompt in the address bar and hit **Enter** (no modifiers). A new tab opens, becomes selected, loads Duck.ai with a fresh chat, and auto-submits the prompt. Switching back to the original tab shows the conversation untouched.
3. Repeat with **Shift+Enter** (focused Duck.ai omnibar) — same behavior.
4. Regression checks:
   - **Cmd+Enter** on Duck.ai with a prompt — still opens a new tab.
   - From a non-Duck.ai tab, submit a Duck.ai prompt via the AI-chat suggestion — current tab navigates to Duck.ai with the prompt.
   - From the NTP, submit a prompt via the NTP omnibar — current tab loads Duck.ai with the prompt.
   - On Duck.ai, open a specific saved chat via a \`chatID\` URL — current tab navigates to that chat.

### Impact and Risks

Low — UX routing change with no schema, persistence, or networking impact. The new-chat path is already exercised by \`.newTab\` callers today; the prompt-handler consumption mechanism is unchanged.

#### What could go wrong?

- Anyone who relied on the legacy "append to the current chat" behavior will see a new tab open instead. That's the intended behavior change — submitting into the loaded chat was the source of the confusion.
- If a future caller invokes \`openAIChat(..., hasPrompt: true)\` while on Duck.ai expecting an in-place reload, they'll get a new tab instead. No such caller exists today and the doc comment on \`openAIChat\` now describes the new contract.

---

> [!NOTE]
> **Low Risk**
> Small UX routing change in `WindowControllersManager.openAIChat` with no persistence or networking impact; only callers expecting in-place reload on Duck.ai with a prompt would see different behavior.
> 
> **Overview**
> When Duck.ai is already the active tab and the user submits a prompt with **current-tab** open behavior (typical omnibar **Enter**), the browser no longer **reloads** that tab to inject the stashed prompt into the open conversation. It now opens Duck.ai in a **new selected tab** so the existing chat stays as-is and the prompt starts a fresh session.
> 
> `chatID` navigation on the current Duck.ai tab and current-tab loads from non–Duck.ai tabs are unchanged. `openAIChat`’s `hasPrompt` documentation was updated to describe this contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f7f865454aa552239f94d3a616887bc533a12c67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
@@ -226,8 +226,9 @@ extension WindowControllersManager: AIChatTabManaging {
     /// - Parameters:
     ///   - url: The AI chat URL to open.
     ///   - linkOpenBehavior: Specifies where to open the URL. Defaults to `.currentTab`.
-    ///   - hasPrompt: If `true` and the current tab is an AI chat, reloads the tab. Ignored if `target` is `.newTabSelected`
-    ///                or `.newTabUnselected`.
+    ///   - hasPrompt: With `.currentTab`, if the current tab is already an AI chat and a prompt was supplied,
+    ///                opens a fresh chat in a new selected tab so the loaded conversation is left untouched.
+    ///                Ignored for `.newTab` / `.newWindow`, which always open a new tab/window.
     func openAIChat(_ url: URL, with linkOpenBehavior: LinkOpenBehavior = .currentTab, hasPrompt: Bool) {
 
         let tabCollectionViewModel = mainWindowController?.mainViewController.tabCollectionViewModel
@@ -236,7 +237,10 @@ extension WindowControllersManager: AIChatTabManaging {
         case .currentTab:
             if let currentTab = tabCollectionViewModel?.selectedTab, currentTab.url?.isDuckAIURL == true {
                 if hasPrompt {
-                    currentTab.reload()
+                    // Omnibar submission while Duck.ai is already loaded: open a fresh chat in a new
+                    // selected tab rather than injecting the prompt into the loaded conversation,
+                    // which users found confusing.
+                    open(url, with: .newTab(selected: true), source: .ui, target: nil)
                 } else if url.getParameter(named: "chatID") != nil {
                     // Navigate to a specific existing chat — must load even if already on duck.ai
                     show(url: url, source: .ui, newTab: false)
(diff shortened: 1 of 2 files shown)


jotaemepereira (approved):
This looks good to me 👍🏼. Is it possible to add unit tests to the `AIChatTabOpener`? Or a UI test could work here.

tomasstrba:
UI test added. I've [ran the check](https://github.com/duckduckgo/apple-browsers/actions/runs/26578382671) to make sure it passes on CI

EDIT: I'm still polishing the test 🥴

jotaemepereira (approved):
This looks good to merge after test passes 👍🏼

tomasstrba:
@jotaemepereira, the UI test is starting to be a pain for this particular case. The reason is that it requires an existing Duck.ai chat loaded in the browser. Given that it has already wasted some of my time, I would recommend moving on without it and rely just on manual testing. What do you think?

jotaemepereira:
> @jotaemepereira, the UI test is starting to be a pain for this particular case. The reason is that it requires an existing Duck.ai chat loaded in the browser. Given that it has already wasted some of my time, I would recommend moving on without it and rely just on manual testing. What do you think?

@tomasstrba Yes, sure. Let’s remove the UI test then and merge

PR #5085, opened 2026-05-26
iOS: Persist after-idle escape hatch across bg/fg and tab switcher

SabrinaTardio:
Task/Issue URL:  https://app.asana.com/1/137249556945/project/414235014887631/task/1215083465183342?focus=true

Also fixes: https://app.asana.com/1/137249556945/project/1204912272578138/task/1215106279592504

### Description

Makes the after-idle escape hatch persist across background → foreground transitions, across tab-switcher round trips, and across explicit close/burn actions on the hatch itself.

Two changes:

1. **`MainCoordinator.showNewTabPageAfterIdleReturn`** — added an early-return when the current tab is any NTP (`link == nil`). Above-threshold foregrounds used to tear down and rebuild the NTP, which removed the existing escape hatch state and bounced the omnibar/keyboard. With the guard in place, the rebuild is skipped whenever the user is already on the NTP — preserving the existing hatch (or absence of one), avoiding the bounce, and avoiding surfacing a stale hatch on a consumed/fresh NTP. The basic after-idle feature is unaffected because it fires when the user was on a URL tab (`link != nil`), which doesn't match the guard.

2. **`MainViewController.showTabSwitcher`** — removed the unconditional `openedAfterIdle = false` line. The after-idle session is now only "consumed" when the user actually transitions to a different tab (already handled by `transitionTo`). Opening the switcher and dismissing back to the same NTP no longer ends the session, which keeps the `*FromNTP(afterIdle:)` pixels accurate for actions taken after a switcher peek.

### Testing Steps

Shorten the idle threshold via `IdleReturnDebugStorageKeys.idleReturnThresholdSecondsDebugOverride` in `UserDefaults.app` (e.g. 3 seconds) so each cycle completes in seconds. Run every scenario with **UTI on and UTI off**.

#### Basic after-idle feature (regression)

1. Cold launch on a URL tab (e.g. bbc.com). Background past the idle threshold. Foreground. **Expected:** NTP shown with the escape hatch pointing back at bbc.com.

#### Hatch persists across bg/fg 

2. From scenario 1, background within the idle threshold, foreground. **Expected:** NTP and hatch unchanged, no flicker.
3. From scenario 1, background past the idle threshold, foreground. **Expected:** NTP and hatch unchanged, no NTP rebuild, no omnibar/keyboard bounce.

#### Tab switcher peek doesn't end the session

4. From scenario 1, open the tab switcher, then dismiss it (cancel or tap "Done") back to the same NTP. Background past the idle threshold. Foreground. **Expected:** hatch still showing, no bounce.

#### Consumed moment, no stale hatch and no bouncing

5. From scenario 1, open the tab switcher, select a different (URL) tab. Then go back to the NTP via the switcher again. Background past the idle threshold. Foreground. **Expected:** NTP shown *without* a hatch, no omnibar/keyboard bounce.

#### Swipe-close preserves the tabs pill 

6. With the Duck.ai toggle enabled and "NTP after idle" set to *always*, open a few tabs, navigate to bbc.com. Background past the threshold. Foreground. **Expected:** "Return to" card visible next to the compact tabs pill.
7. From scenario 6, swipe-close the "Return to" card. **Expected:** tabs pill expands to fill the space. Background past the idle threshold. Foreground. **Expected:** tabs pill still expanded, no bounce.

#### Explicit close-tab action on the hatch

8. From scenario 1, use the close-tab action on the hatch (the swipe action, or the close item from any actions menu if `escapeHatchActions` is enabled) to close the target tab. **Expected:** target tab is removed and the tabs pill expands in place of the "Return to" card. Background past the threshold. Foreground. **Expected:** pill still expanded, no bounce.

#### Consumed moment, then fresh idle on URL tab (regression)

9. From scenario 1, tap the escape hatch to switch to the target URL tab. (You're now on bbc.com; the NTP tab was closed.) Background past the threshold. Foreground. **Expected:** fresh idle return fires; NTP shown with a new hatch pointing at bbc.com.

#### Voice session edge case (regression)

10. While a voice session is active, idle-return foreground should not trigger an NTP rebuild that interferes with voice playback. **Expected:** existing voice-session guard in `showNewTabPageAfterIdleReturn` short-circuits before the NTP-guard.

### Impact and Risks

Low. Scoped to the after-idle NTP escape-hatch flow. No data loss possible; worst case is the hatch shows/doesn't show when the user expects otherwise.

#### What could go wrong?

1. **Pixel drift.** Six `*FromNTP(afterIdle:)` pixels (`appBackgroundedFromNTP`, `barUsedFromNTP`, `backButtonUsedFromNTP`, `toggleUsedFromNTP`, `tabSwitcherSelectedFromNTP`, `returnToPageTapped`) now report `afterIdle: true` more often — specifically after the user opens the tab switcher and returns to the same NTP, which is now considered part of the same session. This is the more accurate label under the new semantics, but it is a behavior shift in the data. Pixel owners should be aware.
2. **Long-running NTP session.** A user who keeps the after-idle NTP open across many bg/fg cycles will keep seeing the same hatch. The target-tab-closed case is already covered: the existing `clearEscapeHatch()` call in `onSwitchToTab` fires when the target tab no longer exists at tap time.
3. **`postIdleSessionInstrumentation.sessionEnded(reason: .tabSwitcherSelected)`** still fires unconditionally on switcher open even when the session technically continues. Pre-existing behavior; left as-is intentionally to minimize blast radius.

### Quality Considerations

1. `Tab.openedAfterIdle` is *not* in `NSCoding` encode/decode and not in `Tab.init`. Always defaults to `false` on app launch, so the "sticky" semantics never survive a kill/relaunch.
2. Audited all read sites of `tab.openedAfterIdle`: six pixels (above) plus the `shouldSaveTabs` change-detection in `attachHomeScreen`. No timer-based or notification-based readers. No persistence layer reads.
3. Audited write sites: only `attachHomeScreen` (sets to param value), `prepareTabForRequest` (false on navigation), and `transitionTo` (false on previous tab on actual transition).
4. The NTP-guard in `showNewTabPageAfterIdleReturn` matches any tab with `link == nil`, including home tabs in fire mode. This is intentional: the existing eligibility logic in `buildEscapeHatch` already returns nil for fire-mode current tabs, so skipping the rebuild on a fire NTP can't lose a hatch.
5. Added `LastActiveTabStoreTests.swift` as incidental coverage of a previously-untested module (round-trip, latest-write-wins, cross-instance persistence, empty default). Not driven by the fix itself.

### Notes to Reviewer

1. `LastActiveTabStoreTests.swift` is on disk but **not yet added to the `DuckDuckGoTests` Xcode target** — the `.pbxproj` is `objectVersion = 60` (no auto-sync). Please drag it into the test group in Xcode before merging so CI picks it up.
2. `MainCoordinator` and `MainViewController` lack a unit-test harness in this repo (the only `MainViewController*Tests.swift` file in `iOS/DuckDuckGoTests` is empty). The two production-code changes don't have direct unit coverage as a result; tested manually with the scenarios above. Worth a follow-up to add a thin coordinator-level testability seam.
3. There is a known code-quality wart that this PR intentionally does *not* touch: `LastActiveTabStore.recordActiveTab` is called from `attachHomeScreen` even though the store's variable name is `lastActiveNonEmptyTabUID`. The home-tab UID gets written into the "non-empty" store, which is technically a misuse — but this fix relies on that misuse to suppress the hatch in the "consumed moment, back to NTP" scenario. Fixing the misuse properly would require pairing the rename/cleanup with an explicit "is the after-idle moment consumed?" suppression check; out of scope here.
4. See Asana threads for design context: Beah and Aitor confirmed they want the hatch to persist across bg/fg, and the no-clear-on-tab-switcher change aligns with Beah's stated expectation that opening the switcher and coming back should not end the moment.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Scoped to after-idle NTP/hatch UX and `openedAfterIdle` semantics; main side effect is more frequent `afterIdle: true` analytics after switcher peek-and-return.
> 
> **Overview**
> Keeps the **after-idle “return to” escape hatch** on the new tab page across background/foreground cycles and tab-switcher peeks, without rebuilding the NTP or clearing the session when the user only opens the switcher.
> 
> **`MainCoordinator.showNewTabPageAfterIdleReturn`** now returns early when the current tab is already an NTP (`link == nil`), so idle foreground no longer tears down and recreates the home screen (preserving hatch state and avoiding omnibar/keyboard bounce). URL tabs still go through the existing idle NTP flow.
> 
> **`MainViewController.showTabSwitcher`** no longer sets `openedAfterIdle = false` on open; the after-idle session ends only on an actual tab change (e.g. **`transitionTo`**), so dismissing the switcher back to the same NTP keeps analytics `*FromNTP(afterIdle:)` aligned with user actions.
> 
> Adds **`LastActiveTabStoreTests`** for UserDefaults round-trip, latest-write-wins, and persistence across store instances (incidental coverage; may need adding to the Xcode test target per PR notes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc78b47af50e6bb54279b28d343e144cc5627d62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jleandroperez (approved):
✅ All test scenarios are green
✅ Code looks good!

:shipit:

PR #4454, opened 2026-04-14
Redesign macOS contextual SearchDone dialog for rebranding

afterxleep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204099484721401/task/1214258622759101?focus=true
Tech Design URL:
CC:

### Description

Redesigns the macOS contextual onboarding flow as part of the 2026 rebranding:

- Introduces a layered dialog system (separate bubble and background NSHostingControllers) for the rebranded flow
- Adds rebranded dialog screens: TrySearch, SearchDone, TrySite, Trackers, FireButton, EndOfJourney
- Removes all background illustration animations (previously slid in from below) — background is now static
- Adds bubble transition animations: fade-out on exit, slide-up + fade-in on entrance, opacity cross-fade for inline transitions (SearchDone→TrySite, Trackers→Fire)
- Fixes background illustration positioning to use overlay instead of ZStack alignment, eliminating UIKit animation context absorption that caused left-to-right sweep artefacts
- Removes the rebranding debug badge
- Fixes missing `optionsListHoverColor`/`optionsListPressedColor` parameters in iOS rebranding theme

### Testing Steps

1. Enable the `onboardingRebranding` feature flag
2. Reset contextual onboarding state via Debug menu
3. Open a new tab — verify TrySearch dialog appears with no background animation
4. Perform a search — verify SearchDone dialog transitions in with bubble slide-up animation and background fade
5. Tap "Got it" — verify inline transition to TrySite content with opacity cross-fade
6. Visit a site with trackers — verify Trackers dialog appears with bubble animation
7. Tap "Got it" — verify inline transition to Fire dialog content
8. Use the Fire button — verify EndOfJourney dialog appears
9. Verify no background illustration animation on any screen (background should appear immediately)
10. Verify light and dark mode for all screens

### Impact and Risks

**Low** — scoped entirely to the rebranded onboarding flow behind a feature flag. The legacy flow is untouched.

#### What could go wrong?

- Bubble slide animation may feel off on slower machines — duration is `screenTransitionPhaseDuration` (0.2s)
- Inline transitions (SearchDone→TrySite, Trackers→Fire) now use `.transition(.opacity)` where `.identity` was used before; if any screen has unusual sizing this could cause a layout jump during the cross-fade

### Quality Considerations

- All animation changes are confined to the rebranded factory and `BrowserTabViewController` layered dialog methods
[...]

PR #5091, opened 2026-05-26
iOS: Fix rebranded contextual onboarding bugs

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1215001148269878?focus=true

### Description

Bug fixes for the rebranded contextual onboarding (`AppRebrand` FF on):

- **Typing animation restarted on rotation.** A process-wide cache keyed by `(title, message)` lets the view-rebuild triggered by an orientation change skip the typing animation and render the static text immediately.
- **Pre-existing text-disappearing glitch.** The same cache fix also resolves a separate pre-existing issue where the dialog text vanished briefly during rebuild.
- **Landscape background image alignment.** The illustration was misaligned in landscape and left a gap above the web view / gradient. Background is now anchored to the bottom edge in landscape.
- **Gray safe-area bands in iPhone landscape.** The modifier's frame respected horizontal safe area (notch / Dynamic Island) and the home-indicator bottom inset, leaving gray bands. The modifier now ignores the horizontal container safe area always, plus the bottom edge on iPhone landscape (`vSizeClass == .compact`). Portrait and iPad layouts are unchanged.

### Testing Steps

1. Be sure to start with no app on the device
2. Enable the `AppRebrand` feature flag.
3. Walk through the contextual onboarding dialogs and confirm:
   - Typing animation plays exactly once per dialog. Rotating the device does NOT restart it.
   - Dialog text remains visible across rotation; no flash of empty bubble.
   - Background illustration sits flush at the bottom of the panel in both portrait and landscape, with no gap above the web view / gradient.
   - In iPhone landscape there are no gray bands left/right (notch / Dynamic Island) or at the bottom (home indicator).
   - Portrait iPhone and iPad layouts look unchanged from before this PR.

### Impact and Risks

Low. All changes are scoped to the rebranded onboarding code path, which is FF-gated.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to FF-gated rebranded onboarding UI (typing state and background safe-area/frame), with no auth, data, or network impact.
> 
> **Overview**
> Fixes rebranded contextual onboarding so **rotation and view rebuilds** do not replay the Dax typing sequence or briefly hide dialog copy.
> 
> **Dialog typing:** Adds parent-level `alreadyRevealed` state, passed into `TypingTitleMessageStack` and `AnimatableTypingText` via `skipAnimation`. `onAppear` skips starting typing when already revealed; completion sets the flag and on iOS initializes static message visibility from it so title/message show immediately after the first reveal.
> 
> **Background layout (iOS):** Anchors the illustration with **bottom alignment** inside its max-height frame. Ignores **container** safe area on horizontal edges always, and on the **bottom** edge when `verticalSizeClass` is compact (iPhone landscape), so the panel background fills edge-to-edge without gray bands; portrait and iPad bottom safe area behavior stay as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb91eeee9a93cb01ef6072eefb6ee6d2ee752467. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/Onboarding/Sources/Onboarding/Rebranding/Shared/RebrandedContextualDaxDialogContent.swift
```diff
@@ -39,9 +58,9 @@ extension OnboardingRebranding {
         private let titleBodyVerticalSpacingOverride: CGFloat?
         private let content: Content
 
-        @State private var startTypingTitle = false
-        @State private var startTypingMessage = false
-        @State private var shouldShowContent = false
+        @State private var startTypingTitle: Bool
+        @State private var startTypingMessage: Bool
+        @State private var shouldShowContent: Bool
```
alessandroboron, line 47: @federicocappelli there's a swiftier alternative here that I think can replace the cache entirely.

The @State variables declared here survive the branching VStack vs HStack. That means we can swap `RebrandedContextualDaxDialogCache` for a single `@State alreadyRevealed: Bool`  on `ContextualDaxDialogContent`.

We still need:
-`@State private var showStaticMessage: Bool` changes you made on `TypingTitleMessageStack` plus `_showStaticMessage = State(initialValue: initiallyRevealed)` in init as this is what actually fixes the disappearing-text bug
- Pass `skipAnimation: $initiallyRevealed` to `AnimatableTypingText.

That way we can just set `alreadyRevealed = true` in `animateContentIn()` (or in `revealStaticMessageAndFinish` if needs to be set immediately)

SharedPackages/Onboarding/Sources/Onboarding/Rebranding/Shared/RebrandedContextualDaxDialogContent.swift
```diff
@@ -249,19 +307,16 @@ private extension OnboardingRebranding {
...
             .fixedSize(horizontal: false, vertical: true)
         }
 
         #if os(iOS)
-        // iOS: only the title types; message is static and fades in once typing completes.
-        // No-title path: the parent flips `startTypingMessage` directly — handled in
-        // `messageTypingView` below.
         @ViewBuilder
         private func titleTypingView(_ title: NSAttributedString, alignment: TextAlignment) -> some View {
             AnimatableTypingText(
```
alessandroboron, line 288: I thought that the new `TypingText `was a replacement for `AnimatableTypingText`. but I noticed that linear onboading uses  `TypingText` and contextual `AnimatableTypingText`. What prevents us to just use `TypingText`?

SharedPackages/Onboarding/Sources/Onboarding/Rebranding/Shared/Styles/RebrandedOnboardingStyles+Background.swift
```diff
@@ -133,7 +133,7 @@ extension OnboardingRebranding.OnboardingStyles {
                             backgroundType.image
                                 .resizable()
                                 .scaledToFit()
-                                .frame(maxHeight: maxHeightMetrics)
+                                .frame(maxHeight: maxHeightMetrics, alignment: .bottom)
```
alessandroboron, line 137: The positioning looks fixed, but some assets appear stretched in landscape (still better than being mispositioned). The positioning was addressed during the onboarding rebranding project.

I noticed that there were [quite few changes to this class](https://github.com/duckduckgo/apple-browsers/pull/4454/changes#diff-8e7fdfd2364a06b90ba2e6ec5e14dd03a97417a73a407eb6a251b13b9f1f9343) during the macOS project, but actually, macOS does not use this class. 🤷 

I’d recommend showing it to Denis before merging it. Also you can refere to testflight build 7.216.0 to check original asset positioning in portait and landscape. I’ve tried to upload a video but no luck

SharedPackages/Onboarding/Sources/Onboarding/Rebranding/Shared/RebrandedContextualDaxDialogContent.swift
federicocappelli: TLDR: The refactoring is not straight forward and the risk of introducing regressions is sizeable.

Two things prevent a straight swap:
- Platform: TypingText is #if os(iOS) only. AnimatableTypingText is shared across iOS and macOS, and RebrandedContextualDaxDialogContent builds for both, so the macOS branch needs it.
- Per-instance skip: the contextual dialog passes skipAnimation: .constant(initiallyRevealed) per instance (so a re-rendered dialog from the cache appears already typed). TypingText only exposes skip via @Environment(\.typingAnimationSkip), which is subtree-wide; flipping it would skip every TypingText in scope, not just one.
The iOS title could migrate by short-circuiting (render plain Text when initiallyRevealed, mount TypingText otherwise), but macOS still needs AnimatableTypingText, so it doesn't eliminate it.

SharedPackages/Onboarding/Sources/Onboarding/Rebranding/Shared/Styles/RebrandedOnboardingStyles+Background.swift
federicocappelli: Yes, that's the real issue, the images are not made for landscape, and I haven't found a solution.
Proposal: let's merge this to fix the most prominent bug, and then I'll ask Denis for a full review.

SharedPackages/Onboarding/Sources/Onboarding/Rebranding/Shared/RebrandedContextualDaxDialogContent.swift
federicocappelli: Good point, implemented and tested

alessandroboron (approved):

PR #5096, opened 2026-05-26
iOS: Fix UTI auto-collapse on mid-load redirects

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215043849443418
CC:

### Description

Fixes the JL1 bug where the UTI auto-collapses (losing the user's typed text) when the user reopens it mid-load and the page then completes a chained redirect (e.g. `cnn.com` → `https://cnn.com` → `www.cnn.com` → `edition.cnn.com`).

Each redirect reassigns `tab.url`, which fires `tabLoadingStateDidChange` while `tab.isLoading == true`. The guard in `refreshAction(for:coordinator:)` required `!tab.isLoading` to take the `.preserveOmnibarSession` branch, so the in-flight redirect routed through `.refreshNonAITab` and tore down the coordinator. The fix drops the `!isLoading` qualifier; the user-driven submit path explicitly deactivates the coordinator before nav starts, so the auto-collapse on real submissions is unaffected (covered by `test_nonAITab_coordinatorActive_notOmnibarSession_refreshesNonAITab`).

The decision is also extracted into a pure static `MainViewController.decideRefreshAction(for: UnifiedToggleInputRefreshActionInputs)` so it's unit-testable without spinning up UIKit.

### Testing Steps

Requires `unifiedToggleInputFeature` ON.

1. From the omnibar/NTP, type `cnn.com` and press Enter.
2. Before the page settles, tap the omnibar to reopen the UTI and start typing (e.g. `youtube.com`).
3. Wait for CNN to fully resolve through its redirect chain (http → https → www → edition).
4. **Expected:** UTI stays open with your typed text intact.
5. Regression: submit a URL from the UTI on any page — the UTI should still collapse normally on submission (the submit path deactivates the coordinator before nav starts).
6. Flag OFF: production behaviour unchanged.

### Impact and Risks

Low. Scope is one boolean removed from a single guard in `refreshAction`; behavior change is limited to the non-AI-tab branch when an omnibar session is open. The submit path is already deactivated by the time the decision runs, so the auto-collapse on real submissions is preserved.

#### What could go wrong?

A user-opened UTI session could now persist across a navigation event that historically tore it down. Mitigated by:
- Submit path deactivates the coordinator before the new nav reaches the decision (verified in logs and by `test_nonAITab_coordinatorActive_notOmnibarSession_refreshesNonAITab`).
- Tab-switch path explicitly calls `resetUnifiedToggleInputForTabTransition` before refreshing for the new tab, so `isOmnibarSession` is false by then.

### Notes to Reviewer

- The refactor adds `UnifiedToggleInputRefreshActionInputs` + a static `decideRefreshAction(for:)` to make the decision unit-testable; the instance `refreshAction(for:coordinator:)` is now a thin adapter. The inputs struct deliberately omits `isLoading` so the regression is also locked in structurally.
- Tests in `MainViewControllerRefreshActionTests.swift`; the regression case was verified to fail on the pre-fix code (`refreshNonAITab` instead of `preserveOmnibarSession`) then pass with the fix.

### DIFF
--- a/iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
+++ b/iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
@@ -2094,6 +2094,7 @@
 		CB1143DE2AF6D4B600C1CCD3 /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = CB1143DC2AF6D4B600C1CCD3 /* InfoPlist.strings */; };
 		CB14D5832E269A7F002E897E /* UserAgentConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = CB14D5822E269A73002E897E /* UserAgentConfiguration.swift */; };
 		CB18A3662E2790C700BE9D8D /* UserAgentConfigurationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CB18A3652E2790C700BE9D8D /* UserAgentConfigurationTests.swift */; };
+		CB22D7A22FC603EB00128979 /* MainViewControllerRefreshActionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CB22D7A12FC603EB00128979 /* MainViewControllerRefreshActionTests.swift */; };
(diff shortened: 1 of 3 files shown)


aataraxiaa (approved):
Tested, LGTM

PR #5102, opened 2026-05-26
iOS: Only commit UTI toggle position on submit

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1215138338466450?focus=true

### Description

`UnifiedInputStateStore.recordUserChoice` was persisting the toggle mode to global `toggleModeStorage` (and updating `trackedLastUsed.toggleMode`) on every in-flight toggle change inside the UTI, not only on submit. So toggling to the other mode inside the UTI and then inline-dismissing without submitting would leak the in-flight choice into the global last-used storage — the next UTI activation on the same tab opened in the in-flight mode instead of the tab's actual committed mode.

This change decouples in-flight toggle changes from global commit:

- `recordUserChoice` no longer writes `toggleMode` to `toggleModeStorage` or to `trackedLastUsed.toggleMode`. The other deliberate-choice fields (model, reasoning, tool) continue to write through as before.
- New `UnifiedInputStateStoring.commitToggleMode(_:)` is the single chokepoint for promoting a toggle change to global. Called from `UnifiedToggleInputCoordinator.commitCurrentToggleState()` on submission, replacing the direct `toggleModeStorage.save(...)` write that previously lived there.

Net effect:

- In-flight toggle inside UTI → no global write → next omnibar tap on the same tab re-derives the same mode as before. ✓
- Submit → global write happens through `commitToggleMode` → new tabs inherit the submitted mode. ✓
- Symmetric across both directions (Search → Duck.ai dismiss, and Duck.ai → Search dismiss).

### Testing Steps

Repro of the bug (failure on `main`, fixed on this branch). Setup: UTI on, toggle enabled.

1. Open a new tab — observe whichever mode the UTI/omnibar opens in (Search or Duck.ai, per the current global last-used).
2. Tap the omnibar → UTI opens in the tab's default mode.
3. Toggle to the OTHER mode inside the UTI. Do NOT submit.
4. Tap the inline back chevron to dismiss.
5. Tap the omnibar again.

Expected: UTI opens in the original tab default (no commit happened, so default unchanged).
Pre-fix: UTI opened in the in-flight (toggled) mode, as if the user had committed.

Verify the commit path is still intact:

6. Toggle to the OTHER mode and submit a query.
7. Open a new tab. UTI should open in the just-committed mode (it is now the new global default).

### Impact and Risks

Low. Restores the pre-`c7810aac60` behavior for the in-flight toggle case while preserving the deliberate-commit-on-submit contract.

#### What could go wrong?

- Per-tab `states[uid].toggleMode` still updates on every in-flight change. That's intentional for UI continuity in tab-switch flows, but if a future change starts reading per-tab toggle for cross-tab decisions, the in-flight leak could resurface in a different shape. The new dedicated `commitToggleMode` makes the intentional commit path explicit and easy to audit.

### Quality Considerations

- Updated two `UnifiedInputStateStoreTests` cases that pinned the old (buggy) write-through behavior to assert the new contract:
  - `test_recordUserChoice_doesNotWriteToggleModeToStorage` (was `..._writesThroughToggleModeToStorage`)
  - `test_recordUserChoice_updatesLastUsedExceptToggleMode` (was `..._updatesLastUsed`)
- Added three new tests for `commitToggleMode`: write-through to storage, `lastUsed.toggleMode` update, and preservation of other `lastUsed` fields.
- Existing `UnifiedToggleInputCoordinatorTests` submit / commit / toggle / external-submission / inline-dismiss assertions continue to pass with no changes.

### Notes to Reviewer

`recordUserChoice` still updates the per-tab `states[uid].toggleMode` but not the global — by design. The per-tab cache is the source of truth for tab-switch UI continuity; global `toggleModeStorage` is the seed for new tabs and should only reflect deliberate submissions. The protocol's doc comment was updated to reflect this split.

---

> [!NOTE]
> **Low Risk**
> Localized UTI state-store behavior with explicit submit path and test coverage; per-tab toggle still updates in-flight for UI continuity.
> 
> **Overview**
> **Search vs Duck.ai toggle mode is no longer saved globally while the user is only previewing a switch inside the unified toggle input (UTI).** Previously, `recordUserChoice` pushed in-flight toggle changes into `toggleModeStorage` and `lastUsed`, so dismissing without submitting could reopen the omnibar in the wrong mode.
> 
> `recordUserChoice` still updates per-tab state and other deliberate choices (model, reasoning, tool) but **leaves global toggle mode unchanged**. A new **`commitToggleMode`** on `UnifiedInputStateStoring` / `UnifiedInputStateStore` is the only path that updates `trackedLastUsed.toggleMode` and persists to storage; **`UnifiedToggleInputCoordinator.commitCurrentToggleState()`** calls it on submit instead of writing `toggleModeStorage` directly. The coordinator drops its redundant `toggleModeStorage` property in favor of the state store.
> 
> Tests and fakes were updated to assert the new contract (including coordinator tests that in-flight `updateInputMode` must not mutate `lastUsed.toggleMode`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6f41b789626714e16b9062d12f33e561c676a54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aataraxiaa (approved):
Tested and LGTM. Agent flagged stale comment in related code:

```
Stale doc comment on unifiedToggleInputDidCommitMode

func unifiedToggleInputDidCommitMode(_ mode: TextEntryMode) {
    // No per-tab persistence: existing tabs read from URL; new tabs read from setting + app-wide last-used.
    // The app-wide last-used is written through `UnifiedInputStateStore.recordUserChoice` on submission.   // ← stale
}

After this PR, recordUserChoice no longer writes the toggle to the app-wide last-used. The new chokepoint is commitToggleMode, called from commitCurrentToggleState() — which is precisely what fires this delegate call. The comment now contradicts the updated protocol doc comment in UnifiedInputStateStoring.swift:46-48 and will mislead the next reader.

Replace the second line with: "The app-wide last-used is written through UnifiedInputStateStore.commitToggleMode on submit, which fires this delegate.”
```

PR #2726, opened 2025-12-01
CPM-specific breakage detection

muodov:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1163321984198618/task/1212148047186714?focus=true
Tech Design URLs:
- https://app.asana.com/1/137249556945/project/72649045549333/task/1212150080905623?focus=true (reload loop detection)
- https://app.asana.com/1/137249556945/project/1163321984198618/task/1212238528964166?focus=true (performance pixels)
CC: @noisysocks @miasma13 @SabrinaTardio 

### Description
This PR adds:
- CPM reload loop detection and prevention for macOS
- corresponding CPM pixel parameter: `autoconsent_error_reload-loop`
- `consentRule` and `consentReloadLoop` parameters on the breakage report

### Testing Steps

1. Set remote config URL to https://duckduckgo.github.io/privacy-configuration/pr-4173/v4/macos-config.json
2. Test that a reload loop is prevented on https://www.sofiebadet.dk/book-tid after two reloads
3. Test that CPM still works on other sites (e.g. https://doodle.com, https://etsy.com, https://www.castrol.com/)
4. Try to send a breakage report for https://www.sofiebadet.dk/book-tid. The breakage pixel should contain `consentRule=Complianz+optin` and `consentReloadLoop=1`

### Notes to Reviewer

- Note that the new UI test will fail until https://github.com/duckduckgo/privacy-test-pages/pull/360 is merged

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> Add macOS autoconsent reload-loop detection/prevention with a new pixel, extend breakage reports with `consentRule`/`consentReloadLoop`, and update iOS/macOS code and tests accordingly.
> 
> - **macOS – Autoconsent**:
>   - Implement reload-loop detection/prevention in `AutoconsentUserScript`:
>     - Track `lastHandledCMPName`/`reloadLoopDetected`, clear state on main-frame navigation, disable `autoAction` when loop detected, and record CMPs.
>     - Report state via `refreshDashboardState(... consentReloadLoop, consentRule ...)` and fire new pixel on detection.
> - **Pixels**:
>   - Add `AutoconsentPixel.errorReloadLoop` and include in `summaryPixels`, `standardParameters`, and `name` mapping.
[...]

PR #5175, opened 2026-05-29
[macOS] NTP omnibar: attach open tabs and files to Duck.ai

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1215138415338926?focus=true
Tech Design URL:
CC:

### Description

Adds native macOS support for attaching content from open tabs (and PDF files) to a Duck.ai chat started from the NTP omnibar. The web layer (content-scope-scripts [#2726](https://github.com/duckduckgo/content-scope-scripts/pull/2726)) owns the picker UI; native implements the bridge contract behind a new `aiChatNtpAttachMoreTabs` flag (remote-releasable, off by default):

- **`omnibar_getOpenTabs`** — returns open-tab metadata (excludes the requesting NTP tab, applies the same picker-exclusion rule as the sidebar).
- **`omnibar_getTabContent`** — extracts page content for a tab id, or `null` on failure.
- **`omnibar_submitChat`** — now carries the echoed `pageContext[]` and `files[]`; native maps them to `AIChatPageContextData` (`.multiple`) and `NativePromptFile` and forwards them through the existing `queryPrompt`.
- **`omnibar_getConfig` / `omnibar_onConfigUpdate`** — add `enableAttachTabs` (reactive to the flag) and `supportedFileTypes` per model.

A new `NewTabPageOmnibarTabsProvider` reuses the Duck.ai sidebar's tab enumeration and `PageContextUserScript` extraction (`AIChatUserScriptHandler.extractPageContext`, `shouldExcludeFromTabPicker`) so both flows share one implementation.

> [!WARNING]
> Draft: the second commit temporarily pins content-scope-scripts to the `pr-releases/alavrenchuk/ntp-attach-tabs` branch for end-to-end testing. It must be reverted to a released version (the merged bump of #2726) before this leaves draft. Depends on #2726 merging first.

### Testing Steps
1. Enable `aiChatNtpAttachMoreTabs` (Debug → Feature Flags).
2. Open a few tabs (include a pinned tab, a `duck://`/settings page, and normal pages).
3. New tab → switch omnibar to Duck.ai mode → the paperclip / `@` picker appears.
4. Open the picker: the list shows normal tabs only, excludes the current NTP tab and restricted pages; `@`-typeahead filters client-side.
5. Pick a tab → a chip renders; a closed/restricted tab yields no chip.
6. Attach a PDF on a model whose `supportedFileTypes` includes `application/pdf`.
7. Submit with N tab chips + a file → the opened Duck.ai chat receives the prompt with the N page contexts and the file.
8. Disable the flag → affordances disappear without a reload; plain submit still works.

### Impact and Risks
Low. Gated by `aiChatNtpAttachMoreTabs` (off by default); when absent, `enableAttachTabs` is false and existing omnibar flows are unchanged.

#### What could go wrong?
- Page extraction can be slow or fail for background/suspended tabs — handled by returning `pageContext: null` so the web drops the chip.
- Page content is echoed back through JS on submit (per the agreed contract); native trusts content it originally produced from first-party NTP.

### Notes to Reviewer
- New messages live in the existing `NewTabPageOmnibarClient`; tab/content work is delegated via the new `NewTabPageOmnibarTabsProviding` so the NewTabPage package stays free of an AIChat dependency.
- Covered by unit tests (`NewTabPageOmnibarClientTests`) and an integration test (`NewTabPageOmnibarActionsHandlerTests`) asserting the attachments reach the native prompt.

---

> [!NOTE]
> **Medium Risk**
> Feature-flagged, but changes shared tab enumeration and page-context extraction (including waking suspended tabs across windows), which affects omnibar and sidebar attach flows—not only NTP.
> 
> **Overview**
> Adds **multi-tab and PDF attachment** for Duck.ai chats started from the **New Tab Page omnibar**, behind the remote **`aiChatNtpAttachMoreTabs`** flag (`enableAttachTabs` on `omnibar_getConfig` / `omnibar_onConfigUpdate`).
> 
> Introduces **`AIChatTabPickerSource`** as the shared tab list for address-bar omnibar, Duck.ai sidebar JS bridge, and NTP. Regular windows aggregate tabs across non–Fire windows; Fire Windows stay isolated. **`extractPageContext(forTabId:…)`** can **materialize suspended tabs**, reload, and wait for navigation so background tabs aren’t dropped at submit. The address-bar omnibar **prewarms** tabs on attach and uses the shared source for picker + submit extraction.
> 
> NTP gains **`NewTabPageOmnibarTabsProvider`**, **`omnibar_getOpenTabs` / `omnibar_getTabContent`**, and extended **`omnibar_submitChat`** with echoed **`pageContext`** and **`files`**, mapped into the native **`queryPrompt`**. Model config adds **`supportedFileTypes`**; the NTP file picker allows **PDF**.
> 
> Sidebar **`getAIChatOpenTabs` / `getAIChatTabContent`** now use the same origin-scoped picker instead of only the key main window’s loaded tabs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6fbd0c955c389fbc19b8748c80b1e90bb29401ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jotaemepereira (commented):
@tomasstrba Overall, it looks good, but I want to discuss one approach. Right now, the UX for suspended tabs isn't ideal. If you are a user with many tabs and some of them get suspended, you can select them, and when submitting the prompt, they are dropped without any warning or error.

We should go one of the following routes:

1.  Hide suspended tabs from the picker
2.  Wake the tab on selection, then extract
3. Keep offering all tabs, but make failure explicit and early

Thoughts?

jotaemepereira:
@tomasstrba Another bug that I found that we should bring to Alex’s attention, is that if you use the `@` to attach tabs and the tabs are larger than the picker can show, if you move with the keyobard it doesn’t work as expected.


https://github.com/user-attachments/assets/4de50503-7182-43fd-a41c-e345d19b530d

tomasstrba:
@jotaemepereira, that was a great catch! I added the logic to wake up suspended tabs. Regarding the frontend issue with scrolling, I've scoped [a task for Alex](https://app.asana.com/1/137249556945/project/1148564399326804/task/1215750276717137?focus=true), and he'll tackle it when he's back on CSS layer. (The feature won't be enabled for now)

Ready for another look

macOS/DuckDuckGo/AIChat/UserScript/AIChatTabPickerSource.swift
```diff
@@ -0,0 +1,119 @@
...
+import AIChat
+import AppKit
+import WebKit
+
+/// Single source of truth for which open tabs a Duck.ai "attach tabs" picker may offer, shared by
+/// all three surfaces (address bar omnibar, Duck.ai sidebar, New Tab Page omnibar).
+///
+/// Scope depends on the window where the picker was opened:
+/// - **Regular window** → tabs from every regular (non–Fire) window; Fire Window tabs are excluded.
+/// - **Fire Window** → only that same Fire Window's tabs (each Fire Window is an isolated session).
```
jotaemepereira, line 28: @tomasstrba ❓ Just to be sure, this is not true for NTP in a Fire Window, right? Because we do not show NTP on a Fire Window.

jotaemepereira (approved):
Looks great!

macOS/DuckDuckGo/AIChat/UserScript/AIChatTabPickerSource.swift
tomasstrba: @jotaemepereira, I've modified the codebase a little and created a shared tab picking logic for address bar, sidebar and NTP. This way, we will ensure all three entry points have same tabs in the list all the time. That's why you see the comment there.

PR #5195, opened 2026-05-31
Sync Exchange V2

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215621770470726?focus=true
Tech Design URL:
CC:

### Description
Implementation of "Exchange V2 Unified Algorithm" https://app.asana.com/1/137249556945/project/72649045549333/task/1214739740392701?focus=true  

**Setup:**
In tab go to https://staging.duck.ai/setup-aichat, click on ‘production’ to set up as an internal chat user

On iOS, ensure you are seeing the new Sync UI (check via Debug screen > Feature Flags > simplifiedSyncSetupExperiment > treatment ) as the experiment was successful, so this will be set as the default for all users in the next day or so. 

Controlled by feature flags:
`syncScopedAccessCredentials` - gating for underlying Sync foundation account changes
`syncCanUseV2ConnectFlow` - will be on by default, supports **scanning** V1 & V2 codes
`syncCanShowV2ConnectCode` - this will be the last to be enabled once all platforms are fully rolled out, enables **presenting** V2 code  

**Notes:**
- V2 functionality implemented.
- Note the 3p devices don’t appear in the device list on native after enabling Sync (intentional scope cut) and vice-versa on 3p browsers. Sync connection can be tested by validating that chats sync between platforms.

**Helpful Links:**
- All Asana specifications are under  https://app.asana.com/1/137249556945/project/72649045549333/task/1214702966683638?focus=true
- Ship review with guide on how to set up for testing: https://app.asana.com/1/137249556945/project/1142021229838617/task/1215282097325722?focus=true
- Python client ref implementation https://dub.duckduckgo.com/duckduckgo/sync-pairing-v2-demo
- Swagger for endpoint updates: https://sync-staging.duckduckgo.com/sync/docs/?secret=dax

### Testing Steps
1. Work through https://dub.duckduckgo.com/pages/duckduckgo/chat-sync-exchange-v2/algorithm.html matrix of test cases 😅 

#### What could go wrong?

---
###### Internal references:

---

> [!NOTE]
> **High Risk**
> Large changes to sync pairing, authentication, and cryptographic account upgrade paths; mis-handling could block pairing or affect account credentials across native and third-party clients.
> 
> **Overview**
> Adds **Pairing V2 (Exchange V2)** for DDG Sync: when rollout flags allow, connect/exchange presentation uses relay-based QR pairing instead of legacy exchange/connect polling, while scanning still supports legacy URLs and V2 `code2` payloads.
> 
> **`SyncConnectionController`** is extended with a `PairingV2Coordinator` (state machine, encrypted relay messages on `sync/v2/exchange`, host/joiner confirmation hooks) and richer delegate callbacks (`updateRequired`, third-party recovery errors, “already connected”, optional wait after sending a recovery key). Legacy flows call `prepareForNewFlow()` before starting and decode recovery via **`defaultCredentialRecoveryKey()`** (V2 default `ddg` only; third-party codes are rejected on the recovery path).
> 
> **Recovery codes & accounts:** `SyncAccount` exposes **`legacyRecoveryCodeV1`** and **`recoveryCodeV2`**; `DDGSync.recoveryCode` picks V2 when both pairing V2 flags are on. **`prepareThirdPartyRecoveryCode`** no longer hard-fails when scoped credentials are disabled; scoped passwords are cached in a background task. New **`upgradeThirdPartyAccountToDefaultCredential`** runs **`ThirdPartyAccountUpgradeCoordinator`** (temp 3party login → create native credential / rewrap keys → final native login).
> 
> **Infrastructure:** `isScopedAccessCredentialsEnabled` init callback is replaced by **`SyncFeatureFlagProviding`** (scoped access + pairing V2 scan/present gates). JWE codec gains **RSA-OAEP-256** wrapping; shared **RSA/DER** helpers; login tolerates nullable device `type`; scoped-access API payloads use `[ProtectedKey]` directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6be6bcf4f6dcf423133040bfadb5e7d084186088. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AppServices/SyncService.swift
```diff
@@ -84,9 +84,17 @@ final class SyncService {
             privacyConfigurationManager: privacyConfigurationManager,
             keyValueStore: keyValueStore,
             environment: environment,
-            isScopedAccessCredentialsEnabled: {
-                featureFlagger.isFeatureOn(for: FeatureFlag.syncScopedAccessCredentials)
-            },
+            syncFeatureFlags: SyncFeatureFlagProvider(
```
ayoy, line 87: ❤️

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2/PairingV2MessageExchanger.swift
```diff
@@ -0,0 +1,194 @@
...
+        channelURL(channelID).appendingPathComponent("messages")
+    }
+
+    private func executeRequest(_ request: HTTPRequesting, validatingStatusWith validateStatusCode: (Int) throws -> Void) async throws -> HTTPResult {
+        do {
+            let result = try await request.execute()
+            try validateStatusCode(result.response.statusCode)
+            return result
+        } catch SyncError.unexpectedStatusCode(let statusCode) {
+            try validateStatusCode(statusCode)
```
ayoy, line 120: Why is `validateStatusCode` called again in the catch clause? Is it to cover the case when `request.execute()` throws? If so, it sounds like a good candidate for a code comment :D

(The same applies to `executeRequestIgnoringResponse` below)

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2MessageExchanger.swift
```diff
@@ -0,0 +1,194 @@
...
+final class PairingV2MessageExchanger: PairingV2MessageExchanging {
+
+    let endpoints: Endpoints
+    let api: RemoteAPIRequestCreating
+    private let firstMessagePost404RetryDelays: [UInt64]
+    private var channelsWithCompletedFirstMessagePost: Set<String> = []
+
+    init(endpoints: Endpoints,
+         api: RemoteAPIRequestCreating,
+         firstMessagePost404RetryDelays: [UInt64] = [200_000_000, 500_000_000]) {
```
ayoy: I have two notes here:
* The name includes "404" which is an implementation detail, and the error that represents 404 is called `relayChannelUnavailable`. Shall we rename this parameter e.g. as `firstMessagePostChannelUnavailableRetryDelays`?
* These must be nanoseconds, so perhaps let's add a comment to the property explaining that the values are in nanoseconds.

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2/PairingV2Coordinator.swift
```diff
@@ -0,0 +1,452 @@
...
+        lastProcessedSequence = 0
+        hasClosedLocalChannel = false
+
+        let commands = stateMachine.handle(
+            .scannedCode(.v2Linking(peerChannelID: qrPayload.channelId, localChannelID: keyPair.channelID), localClient: localClient(isPresenter: false), flags: flags)
+        )
+        try await execute(commands)
+    }
+
+    func pollOnce() async throws {
```
ayoy, line 114: I believe this function is only called by `pollUntilFinished` and could be private if not for unit tests. I'd consider adding a comment mentioning that.

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2/PairingV2Coordinator.swift
```diff
@@ -0,0 +1,452 @@
...
+        }
+    }
+
+    func pollUntilFinished(timeout: TimeInterval = PairingV2PollingDefaults.sessionTimeout,
+                           pollInterval: UInt64 = PairingV2PollingDefaults.pollIntervalNanoseconds,
+                           onStateUpdate: ((PairingV2State) async -> Void)? = nil) async throws -> PairingV2State.Completion {
+        let timeoutDate = Date().addingTimeInterval(timeout)
+
+        while true {
+            if let completion = try completedPairingOrThrowFailure() {
```
ayoy: (Minor)
This function checks if the state machine reached one of the final states: completion or failure.

How about calling it `checkPairingCompletion` or something similar? I'd stick to a verb here (instead `completed`), and the "ThrowFailure" part is not necessary because the method is throwing and called using `try`.

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2/PairingV2Coordinator.swift
```diff
@@ -0,0 +1,452 @@
...
+                return
+            }
+            try await handle(message.encryptedMessage)
+            lastProcessedSequence = max(lastProcessedSequence, message.seq)
+        }
+    }
+
+    func pollUntilFinished(timeout: TimeInterval = PairingV2PollingDefaults.sessionTimeout,
+                           pollInterval: UInt64 = PairingV2PollingDefaults.pollIntervalNanoseconds,
+                           onStateUpdate: ((PairingV2State) async -> Void)? = nil) async throws -> PairingV2State.Completion {
```
ayoy: Minor: `onStateUpdate` may be misleading as the callback would be triggered also when the state didn't update. Perhaps `onDidPoll` would fit better?

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2/PairingV2Coordinator.swift
```diff
@@ -0,0 +1,452 @@
...
+            try await Task.sleep(nanoseconds: pollInterval)
+        }
+    }
+
+    func cancel() async {
+        _ = stateMachine.handle(.failed(.cancelled))
+        await closeLocalChannel()
+    }
+
+    func closeLocalChannel() async {
```
ayoy: It looks like this function could be marked `private`.

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/ScopedAccess/ThirdPartyAccountUpgradeCoordinator.swift
```diff
@@ -0,0 +1,440 @@
...
+    private static let finalNativeLoginRetryDelays: [UInt64] = [500_000_000, 1_000_000_000]
+
+    private(set) var retrySleep: @Sendable (UInt64) async throws -> Void = {
+        try await Task.sleep(nanoseconds: $0)
+    }
+
+    private let jweCompactCodec = JWECompactCodec()
+    private let scopedAccessCredentialEnvelope = ScopedAccessCredentialEnvelope()
+
+    func upgradeThirdPartyAccountToDefaultCredential(_ recoveryCode: String, deviceName: String, deviceType: String) async throws -> UpgradedThirdPartyAccount {
```
ayoy, line 94: We should either link the Asana documentation here or (imo better, but I'm sure it's personal) add inline comments in the function body indicating the stages of 3rd party account upgrade. Whatever you feel more comfortable with :)

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/PairingV2/PairingV2StateMachine.swift
```diff
@@ -0,0 +1,583 @@
...
+    let localChannelID: String?
+    let peerStatus: PairingV2PeerStatus?
+    let purpose: String
+    let hasReceivedHello: Bool
+
+    init(localClient: PairingV2LocalClient,
+         peerChannelID: String?,
+         localChannelID: String? = nil,
+         peerStatus: PairingV2PeerStatus? = nil,
+         purpose: String = "ai_chats",
```
ayoy, line 79: Maybe it's worth extracting `ai_chats` into a constant, but then again I can see that we're only using it in 2 places (and 2 different contexts) in DDGSync module. Just flagging it in case you had a good candidate for where to put it, but also fine to keep it as-is for now 👍

ayoy (commented):
@amddg44 I'm dropping a few comments, mostly related to naming. Otherwise the code is looking great. I had some test failing, either permanently or occasionally, so please have a look at it – the fix for BSK tests is in main so you could merge it into your branch and verify in CI.

I'm done reading through BSK changes and will only have to go through client apps tomorrow 🤩

SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
```diff
@@ -277,23 +281,47 @@ public class DDGSync: DDGSyncing {
...
+        }
+
+        do {
+            let upgradeCoordinator = dependencies.createThirdPartyAccountUpgradeCoordinator()
+            let result = try await upgradeCoordinator.upgradeThirdPartyAccountToDefaultCredential(recoveryCode,
+                                                                                                 deviceName: deviceName,
+                                                                                                 deviceType: deviceType)
+            do {
+                try updateAccount(result.account)
+                try dependencies.secureStore.persistScopedPassword(result.scopedPassword)
```
ayoy: One more question here – should we care about the case when `updateAccount` completes but `persistScopedPassword` fails, and handle it in any special way?

I realize it shouldn't normally be happening since the latter only updates keychain. If updateAccount worked, storing the scoped password should work as well, but just checking if we need any extra logic (besides firing a pixel, I suppose).

SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/ScopedAccess/ThirdPartyAccountUpgradeCoordinator.swift
```diff
@@ -0,0 +1,400 @@
...
+        }
+        let secretKey = try crypter.extractSecretKey(protectedSecretKey: protectedSecretKey,
+                                                     stretchedPrimaryKey: loginInfo.stretchedPrimaryKey)
+        let devices = try loginResult.devices.map {
+            RegisteredDevice(
+                id: $0.id,
+                name: try crypter.base64DecodeAndDecrypt($0.name, using: accountKeys.primaryKey),
+                type: try crypter.base64DecodeAndDecrypt($0.type, using: accountKeys.primaryKey)
+            )
+        }
```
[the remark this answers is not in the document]
amddg44: (Only the python demo implementation returns device type null)

SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
amddg44: Good point! Given that the scoped password cache is only a local convenience/cache and can be recovered from the server again later, I've updated this so only persisting the native account is blocking.

SharedPackages/BrowserServicesKit/Sources/DDGSync/SyncConnectionController.swift
```diff
@@ -21,26 +21,35 @@ import Foundation
 @MainActor
 public protocol SyncConnectionControllerDelegate: AnyObject {
     func controllerWillBeginTransmittingRecoveryKey() async
-    func controllerDidFinishTransmittingRecoveryKey()
+    func controllerDidFinishTransmittingRecoveryKey(shouldWaitForDevicesToChange: Bool)
```
amddg44, line 24: macOS had a UI-specific assumption here: after sending the recovery key, it waited for the device list to change before moving to “Now Syncing”. That works for native DDG-to-DDG setup because the newly connected device appears in GET /devices, but not for 3P clients, so the setup UI would get stuck waiting. I've added this temporary `shouldWaitForDevicesToChange` parameter until we include 3P clients in the device list, which should happen soon.

macOS/LocalPackages/SyncUI-macOS/Sources/SyncUI-macOS/ViewModels/RecoveryCodeViewModel.swift
```diff
@@ -35,11 +35,16 @@ final public class RecoveryCodeViewModel: ObservableObject {
     }
 
     func paste() {
-        let code = NSPasteboard.general.string(forType: .string)?
+        guard let code = NSPasteboard.general.string(forType: .string)?
             .trimmingCharacters(in: .whitespacesAndNewlines)
-            .replacingOccurrences(of: "\n", with: "")
-            .replacingOccurrences(of: " ", with: "") ?? ""
-        setCode(code)
```
ayoy, line 42: We should be fine to remove setCode (and also CharacterSet.base64) and related tests.

SharedPackages/BrowserServicesKit/Sources/DDGSync/SyncConnectionController.swift
ayoy: Ah yes, I stumbled upon it just a few minutes ago without seeing your comment and have figured this out myself 😅 glad to have my thinking confirmed though!

ayoy (approved):
@amddg44 thanks for addressing all the feedback, even though it was just minor comments anyway. You've done a fantastic job here, the code looks beautiful, is well unit-tested and easy to go through with the added comments. After fixing the one issue I reported in Asana, everything was working well in my testing with Chrome, iOS and macOS apps. And thanks for putting #5201 together that helped me inspect the messaging! Really amazing work 👏 🙇

PR #5212, opened 2026-06-01
[iOS] [Subscription] Add notification permission and scheduling for expiration reminder

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207260194172075/task/1215627915896809
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1214964411242994?focus=true
CC:

### Description

Implements the iOS side of the FE↔Native bridge for the "let trialists schedule an expiration reminder" A/B experiment.

**JS bridge (on the existing `useSubscription` channel):**
- `getUserSettings` (new, request/response): returns `{ "notificationsPermission": "granted" | "denied" | "notDetermined" }` mapped from `UNAuthorizationStatus`.
- `requestNotificationsPermission` (new, async request/response): pre-checks status and short-circuits on `.denied` without invoking `UNUserNotificationCenter.requestAuthorization` (the OS would not display a prompt in that state); otherwise triggers the system prompt and replies with `{ "granted": true | false }`.
- `subscriptionSelected` (existing, extended): now parses an optional `scheduleNotification.daysBeforeCancel` field, holds it for the duration of the purchase flow, and hands it to the scheduler only on successful completion.

**Scheduler (`DefaultSubscriptionExpirationReminderScheduler`, app-scope):**
- Schedules one local notification fired `daysBeforeCancel` days before `subscription.expiresOrRenewsAt`. Silently skips when the feature flag is off, permission is not explicitly `.authorized`, `daysBeforeCancel ≤ 0`, the subscription has no expiry, or the fire date is in the past.
- Cancels the pending reminder when the subscription becomes inactive — driven by two observers on the scheduler: `.subscriptionDidChange` (in-app changes) and `UIApplication.didBecomeActiveNotification` (force-refreshes on foreground to catch external App Store / billing changes). The foreground path early-exits when no reminder is queued so non-treatment users incur zero network cost.
- Re-cancels any prior reminder before scheduling a new one — never more than one queued.

**Feature flag:** `privacyPro.subscriptionExpirationReminderNotification` (off by default), wired through `FeatureFlagger` consistent with the existing `inactivityNotification` pattern. The scheduler takes an `isFeatureEnabled: () -> Bool` closure rather than the full flagger to keep the dependency narrow.

**Notification delegate:** the existing `NotificationServiceManager` handles tap-to-open — `segueToDuckDuckGoSubscription()` for the reminder identifier. No special `willPresent` suppression; prevention via cancellation is the only mechanism.

Pixels (`mnot_s_*`, `mnot_c_*`, `mnot_l_*`) are intentionally out of scope for this PR and will follow in a separate change.

### Testing Steps

Manual end-to-end requires the FE step to be enabled in the experiment. For native verification before that lands:

1. **Permission status mapping** (no FE needed)
   - Reset app state (delete + reinstall) so notification authorization is `.notDetermined`. Open Settings → DuckDuckGo → Notifications and toggle to confirm the three states.
   - From the subscription web view's JS console, call `webkit.messageHandlers.subscriptionPages.postMessage({ context: "subscriptionPages", featureName: "useSubscription", method: "getUserSettings", params: {}, id: "test-1" })` for each state.
   - Confirm responses match: `.authorized`/`.provisional`/`.ephemeral` → `"granted"`, `.notDetermined` → `"notDetermined"`, `.denied` → `"denied"`.

2. **Permission request short-circuit on denied**
   - Deny notifications when the system prompt first appears, then post the `requestNotificationsPermission` message above. Confirm reply is `{ "granted": false }` arrives immediately with no prompt redisplay.

3. **Schedule on purchase success**
   - With the feature flag enabled (debug menu / privacy config override), and notification permission granted, post a `subscriptionSelected` message with `params: { "id": "<plan id>", "scheduleNotification": { "daysBeforeCancel": 7 } }`.
   - Complete the App Store purchase in sandbox.
   - In Xcode, attach to the process and inspect `UNUserNotificationCenter.current().pendingNotificationRequests()` — expect one entry with identifier `com.duckduckgo.subscriptions.expiration.reminder` and a fire date 7 days before the sandbox subscription's expiry.

4. **Cancel on status change**
   - With a scheduled reminder in place, cancel the subscription via App Store Settings or trigger a `subscriptionDidChange` notification with an inactive status from the debug menu.
   - Re-foreground the app. The pending reminder list should be empty.

5. **Tap routing**
   - Schedule a near-immediate notification (e.g. set `daysBeforeCancel` so the fire date is within 1 minute), wait for delivery, tap the notification.
   - Confirm the app opens directly to the subscription settings screen.

6. **Unit tests** — `xcodebuild test` on `DuckDuckGoTests` covering the seven new test cases for the three iOS bullets from the tech design's Testing section.

### Impact and Risks

**Low.** All new behavior is gated behind the off-by-default `subscriptionExpirationReminderNotification` remote feature flag. With the flag off, the bridge methods still respond (FE relies on them), but no notifications are ever scheduled — the scheduler silently no-ops on the feature check.

#### What could go wrong?

- **Notification fires for a subscription that's no longer active.** Mitigated by `.subscriptionDidChange` observer (in-app changes) + foreground refresh + cancel (catches external App Store cancellations). Remaining edge case: subscription becomes inactive while the app is killed and never foregrounded before the fire date → notification displays. Estimated population per Kirby's framework (~0.5–1 per 75K iOS treatment users). Acceptable for an A/B test; revisit before global rollout if the variant wins.
- **Native silently ignores the `scheduleNotification` field on older app versions.** Backend should gate the experiment treatment to clients on min app version that includes this PR.
- **Permission state mismatch.** `getUserSettings` returns `"granted"` for `.provisional` and `.ephemeral`, but the scheduler only schedules for `.authorized` (because provisional notifications would deliver silently with no banner). In practice this app doesn't request provisional/ephemeral for the subscription flow, so the gap is theoretical.

### Quality Considerations

- **Performance.** The foreground observer's `pendingNotificationRequests()` guard ensures zero network cost for users without a queued reminder (the dominant case). For treatment users, one force-refresh per foreground transition — same cost shape as the existing `inactivityNotificationSchedulerService`.
- **Concurrency.** All observer handlers wrap their async work in `Task { [weak self] }` and `await` callbacks are evaluated outside `XCTUnwrap` autoclosures in tests.
- **Testability.** Reuses the existing `UNUserNotificationCenterRepresentable` protocol and `MockUNUserNotificationCenter` (extended with `pendingNotificationRequests()`). The scheduler takes injectable `subscriptionManager`, `isFeatureEnabled` closure, notification center, date provider, and notification-center observer.
- **Privacy.** No new user data leaves the device — the reminder preference is held in memory during the purchase flow and discarded after.
- **No pixels in this PR.** The tech design lists `mnot_s_*` / `mnot_c_*` / `mnot_l_*` notification pixels; they will follow in a separate change.

### Notes to Reviewer

- Three new files (`SubscriptionExpirationReminderScheduler.swift` + its mock + the existing mock's `pendingNotificationRequests` addition). The rest are surgical edits.
- The pbxproj entries for new files use hand-rolled UUIDs (`7B12…`); they were checked for non-collision but worth verifying the project still opens cleanly in Xcode before triggering CI.
- The `NotificationsPermissionStatus` enum and the three bridge value types live in a dedicated extension at the bottom of `SubscriptionPagesUseSubscriptionFeature.swift` to keep the experiment surface separate from the checkout flow in the main class body.
- `subscriptionSelected` defers clearing of `pendingScheduleNotification` so it's discarded across every exit path (success, cancellation, failure) — explicit nil-out wouldn't survive the multi-branch flow as cleanly.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches subscription checkout completion, notification permission prompts, and local notification lifecycle; behavior is remote-flagged but incorrect scheduling or cancellation could annoy trial users or miss re-engagement.
> 
> **Overview**
> Introduces an **off-by-default** remote flag (`subscriptionExpirationReminderNotification`) and a **`DefaultSubscriptionExpirationReminderScheduler`** that queues one local notification before a **free-trial** subscription’s expiry/renewal date. Scheduling runs only when notifications are **explicitly authorized**, the trial is still active, and the fire date is in the future; it **replaces** any prior reminder with the same ID. The scheduler **cancels** pending reminders when the flag is off, the subscription no longer qualifies, or the trial ends—listening to **`.subscriptionDidChange`** and **foreground** (with refresh when a reminder is already pending).
> 
> The subscription web **`useSubscription`** bridge gains **`getUserSettings`** and **`requestNotificationsPermission`** (gated by the flag for prompts), and **`subscriptionSelected`** can pass optional **`scheduleNotification.daysBeforeCancel`**, which schedules only after a **successful** purchase. **`NotificationServiceManager`** registers the reminder category (for dismiss tracking), fires pixels on tap/dismiss, and opens **subscription welcome** via a new settings deep link; subscribe flows can start from a **`landingURL`** instead of only purchase redirect URLs.
> 
> Adds **push-notification** pixel definitions and strings, subscription debug controls to trigger the scheduler, and unit tests for the scheduler and bridge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83be448f54ab69ace2851fd2c6194c34e5cd6526. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,154 @@
...
+        // External changes (App Store cancellation, billing failure) happen without our knowledge,
+        // so re-read from the backend on every foreground.
+        notificationCenterObserver.publisher(for: UIApplication.didBecomeActiveNotification)
+            .sink { [weak self] _ in
+                Task { [weak self] in await self?.cancelReminderIfInactive(forceRefresh: true) }
+            }
+            .store(in: &cancellables)
+
+        // Sign-out posts .accountDidSignOut (not .subscriptionDidChange), so cancel unconditionally.
+        notificationCenterObserver.publisher(for: .accountDidSignOut)
```
hanyutang-sandra: Pending product confirmation

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,154 @@
...
+                Task { [weak self] in await self?.cancelReminder() }
+            }
+            .store(in: &cancellables)
+    }
+
+    /// Removes the reminder from both the pending queue and Notification Center (delivered).
+    @MainActor
+    private func cancelReminder() async {
+        notificationCenter.removePendingNotificationRequests(withIdentifiers: [Self.notificationIdentifier])
+        notificationCenter.removeDeliveredNotifications(withIdentifiers: [Self.notificationIdentifier])
```
hanyutang-sandra: Pending product confirmation

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,154 @@
...
+        }
+        await cancelReminder()
+    }
+
+    private static func statusWarrantsReminder(_ status: DuckDuckGoSubscription.Status) -> Bool {
+        switch status {
+        case .autoRenewable, .notAutoRenewable, .gracePeriod: return true
+        case .inactive, .expired, .unknown: return false
+        }
+    }
```
jozsef-vesza: This should probably gate on `subscription.hasActiveTrialOffer`: we only plan to show the reminder for users in free trial.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,154 @@
...
+    func scheduleReminder(timeBeforeCancel: TimeInterval) async {
+        guard isFeatureEnabled() else { return }
+        guard timeBeforeCancel > 0 else { return }
+
+        guard await notificationCenter.authorizationStatus() == .authorized else {
+            Logger.subscription.log("Expiration reminder skipped: notifications not explicitly authorized")
+            return
+        }
+
+        guard let subscription = try? await subscriptionManager.getSubscription(forceRefresh: false),
```
jozsef-vesza, line 108: Just noting that adding the `subscription.hasActiveTrialOffer` check to `statusWarrantsReminder` will also make sure we only schedule the notification for users currently in free trial

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,154 @@
...
+
+    /// Removes the reminder from both the pending queue and Notification Center (delivered).
+    @MainActor
+    private func cancelReminder() async {
+        notificationCenter.removePendingNotificationRequests(withIdentifiers: [Self.notificationIdentifier])
+        notificationCenter.removeDeliveredNotifications(withIdentifiers: [Self.notificationIdentifier])
+    }
+
+    func scheduleReminder(timeBeforeCancel: TimeInterval) async {
+        guard isFeatureEnabled() else { return }
```
jozsef-vesza, line 100: Should we cancel scheduled reminders if the flag is turned off?

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -868,6 +892,61 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
...
+        let daysBeforeCancel: Int
+    }
+
+    func getUserSettings(params: Any, original: WKScriptMessage) async -> Encodable? {
+        let status = await userNotificationCenter.authorizationStatus()
+        return UserSettingsResponse(notificationsPermission: Self.permissionStatus(from: status))
+    }
+
+    func requestNotificationsPermission(params: Any, original: WKScriptMessage) async -> Encodable? {
+        let status = await userNotificationCenter.authorizationStatus()
```
jozsef-vesza, line 930: Should we check for the `subscriptionExpirationReminderNotification` here? E.g. could it happen that our flag is turned off, but FE still sends this message?

jozsef-vesza:
@hanyutang-sandra great job on the debug capabilities for this PR! One potential improvement would be to move the JS bridge under Subscription too, since people might test it together with the mock notification, and it's easy to see if the guards are passing.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
hanyutang-sandra: Good point! Yes we need to cancel it

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
hanyutang-sandra: Yeah make sense. I also confirmed with Product that this is free trial only

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
hanyutang-sandra: For safety it’s better to add it here

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -430,6 +446,7 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
 
             let id: String
             let experiment: Experiment?
+            let scheduleNotification: ScheduleNotificationPreference?
```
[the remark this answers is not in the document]
hanyutang-sandra: But if we proceed to purchase without a notification schedules even we say we will it will be another issue. imo it’s better to fail here early for quick fix rather than hide it later. We can add a pixel here if needed

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,153 @@
...
+        } catch {
+            Logger.subscription.log("Skipping expiration-reminder check: \(error.localizedDescription, privacy: .public)")
+            return
+        }
+
+        if let subscription, Self.subscriptionWarrantsReminder(subscription) {
+            return
+        }
+        cancelPendingReminder()
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: Under the trial-only restriction, the practical impact is very limited:

- Expiry moves earlier (shorter trial): essentially impossible
- Expiry moves later (trial extension): the pending reminder still fires at the original instant, which is now
  earlier than newExpiry. Annoying but harmless, never a missed/late reminder.

State changes that pair with an expiry shift (trial → paid, trial → expired) are already handled by the cancel-check predicate, so they're not affected.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,186 @@
...
+    /// A thrown error (e.g. transient network failure) leaves the reminder alone; a confirmed inactive/missing/no-longer-trial subscription cancels it.
+    @MainActor
+    private func cancelReminderIfInactive(forceRefresh: Bool) async {
+        let pending = await notificationCenter.pendingNotificationRequests()
+        guard pending.contains(where: { $0.identifier == Self.notificationIdentifier }) else { return }
+
+        guard isFeatureEnabled() else {
+            cancelPendingReminder()
+            return
+        }
```
jozsef-vesza, line 155: 👍

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionExpirationReminderScheduler.swift
```diff
@@ -0,0 +1,186 @@
...
+    @MainActor
+    private func cancelPendingReminder() {
+        notificationCenter.removePendingNotificationRequests(withIdentifiers: [Self.notificationIdentifier])
+    }
+
+    /// A subscription warrants a reminder iff its status is active AND it is on a free trial.
+    private static func subscriptionWarrantsReminder(_ subscription: DuckDuckGoSubscription) -> Bool {
+        switch subscription.status {
+        case .autoRenewable, .notAutoRenewable, .gracePeriod:
+            return subscription.hasActiveTrialOffer
```
jozsef-vesza, line 181: 👍

iOS/PixelDefinitions/pixels/definitions/push-notification.json5
```diff
@@ -30,5 +30,40 @@
                 "enum": ["denied", "authorized"]
             }
         ]
+    },
+    "push-notification_subscription-expiration-reminder_scheduled": {
```
jozsef-vesza, line 34: @hanyutang-sandra have you started privacy triage for the new pixels?

jozsef-vesza (approved):
LGTM, nice work! This probably still needs a privacy triage, but otherwise good to go.

iOS/PixelDefinitions/pixels/definitions/push-notification.json5
hanyutang-sandra: Good reminder! I forgot it. Starting now

hanyutang-sandra:
> Privacy Review task: https://app.asana.com/0/69071770703008/1216370828877205

Approved

PR #5217, opened 2026-06-02
[CI] Run E2E tests with internal feature flags

edulpn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203301625297703/task/1215318048361450
Tech Design URL: https://app.asana.com/1/137249556945/project/1200194497630846/task/1214400473106704
CC: @alessandroboron @ayoy 

### Description

Adds an internal user-mode leg to the iOS Maestro E2E, iOS ATB UI tests, and macOS UI tests nightlies so regressions in internally-gated features are caught before production rollout. Implements the approved proposal, addressing the gap surfaced by the [omnibar landscape incident postmortem](https://app.asana.com/1/137249556945/project/1210995391161861/task/1214172761052500).

Mechanism: a `-isInternalUser true` launch argument is forwarded to the host app by Maestro `setup.yaml` when `INTERNAL_USER_MODE=true` is on the run env, and by the XCUITest `setUp` helpers when `TEST_RUNNER_INTERNAL_USER_MODE=true` is passed to xcodebuild as a build setting. Both platforms' `LaunchOptionsHandler` follow the argument and flip the InternalUser store on at app startup. Same build, same scheme, same privacy config as the existing nightly.

iOS Maestro runs the full 8 tag groups in both modes (16 entries total). iOS ATB UI tests and macOS UI tests run external + internal with the internal leg marked `continue-on-error` while signal is calibrated. macOS exposes a `user_modes` input on `workflow_dispatch` / `workflow_call` so PR / push runs stay external-only.

Test run with 2 jobs: https://github.com/duckduckgo/apple-browsers/actions/runs/26308450954

### Testing Steps
- https://github.com/duckduckgo/apple-browsers/actions/runs/27107186268 stays 🟢 

### Impact and Risks

#### What could go wrong?
- **Maestro Cloud capacity**: matrix grows from 8 to 16 entries. The 10-job manual run took ~40 min, safely under the 300-min timeout, but worth watching the first scheduled nightly.
- **Internal-mode flakes in tags we haven't validated**: only `release` and `privacy` have been run in internal mode. The remaining 6 tags could surface internal-only flakes. Mitigated by separate failure routing and `continue-on-error: internal` on iOS `XCUITest` and macOS.
- **macOS / iOS XCUITest `-isInternalUser` plumbing**: only the iOS Maestro path was validated end-to-end. iOS XCUITest and macOS follow the same launch-argument pattern but haven't been run with internal mode on yet.

### Quality Considerations
- **Performance**: nightly wall-clock at most doubles for iOS Maestro. macOS PR / push path stays external-only.
- **Monitoring**: external pixels keep their original names; internal entries get a `_internal` suffix so existing dashboards keep working and internal analytics can filter cleanly.

### Notes to Reviewer
- ``iOS/DuckDuckGo/LaunchOptionsHandler.swift``: the new `internalUserKey` branch in `applyUITestOverrides` is independent of the existing `ff. / experiment. / config.rollout.` prefixes.
- ``macOS/DuckDuckGo/Application/AppDelegate.swift``: the flip happens after `InternalUserDeciderStore` is constructed and before `DefaultInternalUserDecider` wraps it; the `didSet` persists to disk so the value is on disk before any consumer reads it.

---

> [!NOTE]
> **Medium Risk**
> Changes how internal user and feature-flag overrides interact at app startup (privacy/feature gating) and doubles scheduled CI matrix size; production paths unchanged but mis-wiring could affect release gating or alert noise.
> 
> **Overview**
> Adds a **production vs internal user** dimension to iOS Maestro E2E, iOS ATB UI nightly, and macOS UI tests so internally-gated features are exercised before rollout.
> 
> **CI:** New `user_modes` workflow input (`production` / `internal` / `both`). Scheduled nightlies run **both** legs; PR/push defaults stay production-only on macOS. E2E gains a `setup-matrix` job that crosses device/tag rows with user mode; Maestro gets `INTERNAL_USER_MODE`, XCUITest gets `TEST_RUNNER_INTERNAL_USER_MODE`. Internal matrix legs use `continue-on-error` on iOS ATB and macOS UI. Failures upload mode-specific artifacts; **notify** creates separate Asana tasks for production vs “advance warning” internal failures. Status/duration pixels add a `_internal` suffix for internal runs.
> 
> **App plumbing:** Maestro/shared setups pass `isInternalUser: ${INTERNAL_USER_MODE}` (replacing implicit internal-user from any `ff.*` override). iOS `LaunchOptionsHandler` only sets internal user from **`-isInternalUser true`**; `ff.` / experiment / rollout overrides no longer flip internal user. iOS `DefaultFeatureFlagger` gets `allowOverrides: { internalUser || isUITesting }` and applies UI test overrides **after** init so Maestro flag overrides still work in production user mode. macOS mirrors internal user via launch args and early `InternalUserDeciderStore` setup in `AppDelegate`; macOS UI tests forward `INTERNAL_USER_MODE` like iOS ATB.
> 
> **Tests:** Updated `LaunchOptionsHandler` expectations and a `DefaultFeatureFlagger` test for `allowOverrides` when not internal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0810f6e1f6eca382e725c69e7d0557b6de684c05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/LaunchOptionsHandler.swift
```diff
@@ -211,6 +213,13 @@ extension LaunchOptionsHandler {
             guard arg.hasPrefix("-") else { continue }
             let key = String(arg.dropFirst()) // Remove leading "-"
 
+            if key == UITestOverrides.internalUserKey {
+                if userDefaults.string(forKey: key)?.lowercased() == "true" {
+                    internalUserStore.isInternalUser = true
```
samsymons: I was looking into the full logic that controls whether this flag is set. There's a complication here: it gets set any time we provide a feature flag override, which you can see below at line 225.

This _sounds_ fine at a glance, but if you go look at [`shared/setup.yaml`](https://github.com/duckduckgo/apple-browsers/blob/main/.maestro/shared/setup.yaml) you'll see the following:

```
- launchApp:
    appId: "com.duckduckgo.mobile.ios"
    clearState: true
    clearKeychain: true
    arguments:
        isOnboardingCompleted: ${ONBOARDING_COMPLETED}
        currentAppVariant: ${APP_VARIANT}LaunchOptionsHandler `currentAppVariant` key
        isRunningUITests: true
        isInternalUser: ${INTERNAL_USER_MODE}
        ff.browsingMenuSheetEnabledByDefault: "true" // 🚨 We pass this to all runs, enabling internal mode
```

This means that any test which invokes `setup.yaml` (which seems to be nearly all of them?) has been running in internal user mode for 4 months already, and while this flag is being passed there is no way to get them to run in external user mode. I think we should check whether we can remove that global feature flag given how old it is.

samsymons: I don't actually see this flag being used any more - it still exists in the `PrivacyFeature` file but it's never checked, so I'd guess that we're safe to remove it from `setup.yaml`. Best to check with @dus7 to be safe.

.github/workflows/ios_end_to_end.yml
samsymons, line 1: I think we should consider allowing the `workflow_dispatch` trigger of this workflow to pass in a user mode variable - by default it should do both, but it should be possible to run tests for either internal or external mode for situations where one flow is failing and we want to debug it.

samsymons:
The workflow provided in the Testing Steps section was seemingly cancelled and not re-run, so I've triggered a new run in order to verify that this works: https://github.com/duckduckgo/apple-browsers/actions/runs/27107186268

.github/workflows/macos_ui_tests.yml
```diff
@@ -22,6 +22,14 @@ on:
         options:
         - dmg
         - appstore
+      user_modes:
+        description: "User mode(s) to run. 'external' = production behaviour (release-gating). 'internal' = isInternalUser=true, runs internally-gated features (advance warning). 'both' = run both legs."
+        type: choice
+        default: external
+        options:
+        - external
```
dus7: I’d consider changing this option to “public” or “production” so the intent is clear right away. I know _external_ is an opposite of _internal_ and it fits the API, but we’re not using this wording to define the public feature availability.

edulpn: That’s a good point. I thought about mirroring the feature flag state enum (like internal, enabled, disabled), but that doesn’t make much sense, I think your approach has better semantics.

.github/workflows/ios_end_to_end.yml
edulpn: That’s a great point

samsymons (approved):
LGTM with the new `workflow_dispatch` argument, and the new step names are cleaner to read too - great that we now have an example of the internal job breaking while the production one still works too, couldn't have timed it better. Thanks so much!

PR #5226, opened 2026-06-02
Validate URL schemes and double extensions on file import

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214200115953388/task/1210880241624727?focus=true

### Description
Hardens the third-party file/zip import path (security-audit follow-up):
- Drops bookmarks whose URL uses a `javascript:`/`data:` scheme — activating such a bookmark would execute embedded content instead of navigating.
- Rejects files and zip entries with a disguising double extension (e.g. `foo.swift.html`).

Both rules live once in BrowserServicesKit (`String.hasUnsafeBookmarkImportScheme()` / `String.hasMultipleFileExtensions`) and are invoked from each platform's import paths — the scheme check at the shared bookmark-save chokepoint (iOS `BookmarkOrFolder.isInvalidBookmark`, macOS `LocalBookmarkStore`), and the double-extension check at both the archive readers (`ImportArchiveReader`) and the direct-file entry points (iOS `DataImportManager.importFile`, macOS `SafariArchiveImporter.readContentsForSelectedFile`). Archives themselves are exempt — their entries are validated individually; folders carrying a stray unsafe URL are kept so their children still import.

### Testing Steps
A manual smoke-test kit is attached to the linked Asana task and also here. 
[file-import-validation-tests.zip](https://github.com/user-attachments/files/28568569/file-import-validation-tests.zip)
Import each on a **clean profile** (clear bookmarks **and** passwords) on iOS and macOS, comparing a build **before** vs **after** this change.

1. **`bookmarks.html`** (import as bookmarks HTML) — safe + `javascript:` + `data:` + `<script>`/`onload`/`` + unclosed tag.
   - **Before: 6 bookmarks → After: 4.** The `javascript:` and `data:` rows are dropped (shown as "Failed to import 2" on iOS). Safe / onload / img / unclosed are kept (handlers ignored, no JS runs); a bare `<script>` never imports.
2. **`import.zip`** (import as zip) — `bookmarks.swift.html` (double-ext bookmark), `passwords.csv`, `passwords.exe.csv` (double-ext login), `cards.json`, junk `.js`.
   - **Bookmarks 1 → 0** (both platforms): the `.swift.html` double-extension entry no longer slips past the `.html` gate.
   - **Passwords 2 → 1 on macOS** (the double-extension `passwords.exe.csv` is skipped); **iOS stays 1** (it reads only the first CSV, so the decoy never imported there).
   - Credit cards need the `autofillCreditCards` flag (off by default on iOS → 0).
3. **Regression:** a normal Safari/Chrome export still imports all valid bookmarks/passwords/cards.

✅ Verified manually on **iOS and macOS** (before/after) in addition to the unit tests.

### Impact and Risks
Low–Medium — changes which imported bookmarks/files are accepted (privacy/security hardening; no change to existing stored data).

#### What could go wrong?
A legitimate item could be dropped. Mitigations: the scheme check is limited to `javascript:`/`data:` and is bookmarks-only (passwords/cards unaffected). The double-extension rule is intentionally strict (any filename with >1 dot) — an accepted trade-off agreed in the kick-off — so a legitimately double-dotted archive entry is skipped, and a directly-selected file with that shape is rejected outright (the import fails). The strictness was a deliberate decision over false-positive risk.

### Quality Considerations
- Validation centralized in BSK and reused at every per-platform import sink, so a future import source inherits both checks.
- Unit tests added in BSK (`StringFileExtensionsTests`, `BookmarkImportSchemeValidationTests`), iOS (`ImportArchiveReaderTests`, `DataImportManagerTests`), and macOS (`LocalBookmarkStoreTests`, `SafariArchiveImporterTests`), plus manual before/after smoke verification on both platforms.
- **Observability / pixels:** no new telemetry was added. The new rejections ride the import flow's existing patterns — unexpected zip entries are already skipped silently (allowlist), and invalid bookmarks already count into `failed` — so observability is unchanged from the current baseline rather than regressed. Dedicated rejection pixels were consciously skipped given the low expected frequency (P3, no demonstrated exploit) and out-of-kickoff-scope; pixel/security owners can revisit if monitoring of rejection frequency is later desired.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Security-focused import filtering can drop legitimate bookmarks (unsafe schemes) or archive entries with multiple dots in the filename; behavior is intentional but may surprise users with unusual exports.
> 
> **Overview**
> **Hardens third-party data import** so malicious or disguised payloads are rejected before they reach bookmarks or archive parsing.
> 
> Imported **bookmarks/favorites** with `javascript:` or `data:` URLs are now treated as invalid and dropped (counted as failed on macOS via `LocalBookmarkStore`; filtered via `BookmarkOrFolder.isInvalidBookmark` on iOS). **Folders** are not dropped when they carry a stray unsafe URL, so child bookmarks still import.
> 
> **Double-extension filenames** (more than one dot in the last path component, e.g. `foo.swift.html`) are rejected for **direct file picks** (iOS `DataImportManager.importFile`, macOS `SafariArchiveImporter` for non-archive files) and **skipped inside zip entries** (iOS/macOS `ImportArchiveReader`).
> 
> Shared helpers live in BrowserServicesKit: `String.hasUnsafeBookmarkImportScheme()` and `String.hasMultipleFileExtensions`. Unit tests cover BSK, iOS, and macOS import paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8958d5206e9897ffdebcc9d1de80f11c45b1c80c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/Common/Extensions/String+FileExtensions.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Common/Extensions/String+FileExtensions.swift
@@ -1,0 +1,31 @@
+//
+//  String+FileExtensions.swift
+//
+//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public extension String {
+
+    /// `true` when the path's last component contains more than one dot, e.g. `foo.swift.html`.
+    ///
+    /// Used to reject disguising double extensions in imported archive entries: a file whose
+    /// real type is masked behind an allowed extension (such as `.html`/`.csv`/`.json`).
+    var hasMultipleFileExtensions: Bool {
+        let name = (self as NSString).lastPathComponent
+        return name.filter { $0 == "." }.count > 1
+    }
+}
--- a/SharedPackages/BrowserServicesKit/Sources/Common/Extensions/String+FileExtensions.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Common/Extensions/String+FileExtensions.swift
@@ -1,7 +1,7 @@
 //
 //  String+FileExtensions.swift
 //
-//  Copyright © 2025 DuckDuckGo. All rights reserved.
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
--- a/SharedPackages/BrowserServicesKit/Sources/Bookmarks/BookmarkletExtensions.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Bookmarks/BookmarkletExtensions.swift
@@ -25,6 +25,14 @@ public extension String {
         return self.lowercased().hasPrefix("javascript:")
     }
 
+    /// URL schemes that are unsafe to import as a bookmark target, because activating the
+    /// bookmark would execute embedded content rather than navigate to a page. Used to drop
+    /// such bookmarks during third-party import (e.g. `javascript:`, `data:`).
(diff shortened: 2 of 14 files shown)


SharedPackages/BrowserServicesKit/Sources/Common/Extensions/String+FileExtensions.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  String+FileExtensions.swift
+//
+//  Copyright © 2025 DuckDuckGo. All rights reserved.
```
amddg44: Nit: new files should use 2026 for the copyright year 🙂

amddg44 (approved):
LGTM!! Nice work 👏

PR #5255, opened 2026-06-03
Add Fire button to the escape-hatch Return-to card

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211654189969294/task/1215358250572341?focus=true
Tech Design URL:
CC:

### Description

Adds a dedicated **Fire button** to the escape-hatch "Return to…" card, to the left of the three-dots menu, matching the [Figma design](https://www.figma.com/design/QUTLfN2oPR6ivG1Jwsw3Mg/?node-id=13257-116406). The button performs the existing "delete tab” action. The "Delete Tab" item is removed from the menu so the action lives in a single, more discoverable place.

All of this is gated behind a new `escapeHatchFireButton` feature flag

### Testing Steps
1. Go to general settings select NTP after idle with idle time none
2. Open a site, then open a new tab so the "Return to…" card appears on the NTP.
3. Confirm a 🔥 **Fire** glyph appears to the left of the ⋯ menu on the card.
4. Tap the Fire button on a **regular tab** → confirmation flow appears (on iPad, the popover is anchored to the Fire button).
5. Tap the Fire button on a **Fire tab** → the tab burns immediately (no confirmation).
6. Open the ⋯ menu → "Delete Tab" is **no longer present**; "Return to Tab", "Close Tab" (non-fire tabs only) and the after-inactivity picker remain.
7. Disable the `escapeHatchFireButton` flag → the Fire button disappears and "Delete Tab" returns to the menu (control behaviour).
8. Test on NTP with and without favorite, focussed state on both side (search and chat with and without favorites/chat suggestions), test with unified input mode on.

### Impact and Risks

**Low** — UI-only change on an existing, flagged surface. The destructive action wiring is unchanged (same `onBurnTabImmediately` / `onBurnTabWithConfirmation` closures); only its entry point moves from the menu to a button.

#### What could go wrong?
- The iPad burn-confirmation popover could mis-anchor. Mitigated by tracking the Fire button's window frame and passing it to the existing `FireConfirmationPresenter`; the legacy menu-anchored path is preserved for the flag-off case.
- A fire tab could be left with no destructive action if the button failed to render. Mitigated by gating both the button and the menu-item removal on the same flag, so exactly one delete affordance is always present.

### Quality Considerations
- Flag-off path is unchanged (conditional menu padding + retained menu-frame anchor), so existing behaviour is preserved for staged rollout / rollback.
- Added unit tests covering the `escapeHatchFireButton` flag wiring on `EscapeHatchModel`, and a SwiftUI preview exercising the Fire-button layout.

### Notes to Reviewer
- The long-press context menu shares `menuContentView`, so when the flag is on, "Delete Tab" is removed there too (consistent with moving the action to the Fire button).
- No new asset is committed — the provided `Fire-Tabs-24.svg` matches the existing `fireTabs` design-system glyph (only sub-pixel export rounding differed), so the change reuses the shipped asset rather than drifting from the Figma-synced source of truth.

---

> [!NOTE]
> **Low Risk**
> UI-only, feature-flagged change on the escape-hatch card; destructive actions still use the same model closures, with an updated iPad popover anchor when the flag is on.
> 
> **Overview**
> Adds remote-releasable **`escapeHatchFireButton`** (default **on**) and exposes **`isFireButtonEnabled`** on **`EscapeHatchModel`** from the feature flagger.
> 
> When the flag is on, **`ReturnToTabCard`** shows a dedicated **Fire** control beside the ⋯ menu for the existing delete-tab flow (immediate burn for fire tabs; confirmation for others, with iPad popover anchored to the Fire button). **Delete Tab** is omitted from the ⋯ menu and the shared context menu; with the flag off, behavior and menu anchoring stay unchanged.
> 
> Adds unit tests for flag wiring and a SwiftUI preview for the Fire-button layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84b05a3b5346dab89053df6d7566cc7e9a6ac87d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/ReturnToTabCard.swift
+++ b/iOS/DuckDuckGo/ReturnToTabCard.swift
@@ -27,9 +27,11 @@ struct ReturnToTabCard: View {
 
     @ObservedObject var model: EscapeHatchModel
 
-    /// Frame of the three-dots menu button in the key window's coordinate space.
-    /// Used as the popover anchor when burning a tab on iPad — the FireConfirmationPresenter
-    /// expects window-space coordinates because it attaches the popover to the key window.
+    /// Frames of the Fire button and the three-dots menu button in the key window's coordinate space.
+    /// Used as the popover anchor when burning a tab on iPad — the FireConfirmationPresenter expects
+    /// window-space coordinates because it attaches the popover to the key window. The Fire button is the
+    /// anchor when the Fire-button flag is on; the menu button is the anchor for the legacy in-menu action.
+    @State private var fireButtonFrameInWindow: CGRect = .zero
     @State private var menuFrameInWindow: CGRect = .zero
 
     var body: some View {
@@ -67,7 +69,7 @@ struct ReturnToTabCard: View {
         HStack(spacing: Metrics.innerSpacing) {
             mainView
             if model.isActionsEnabled {
-                menuView
+                actionButtonsView
             }
         }
         .padding(contentPaddingEdges, Metrics.horizontalPadding)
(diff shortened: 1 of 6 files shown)


iOS/DuckDuckGo/ReturnToTabCard.swift
```diff
@@ -358,4 +411,17 @@ private enum Metrics {
         .frame(width: 360)
 }
 
+#Preview("Return to tab — Fire button") {
+    let target = Tab(fireTab: false)
+    ReturnToTabCard(model: .preview(title: "Tokamak - Wikipedia",
+                                    subtitle: "en.wikipedia.org/wiki/Tokamak",
```
jleandroperez, line 417: Tony Stark would find this Preview quite interesting 😄

jleandroperez (approved):
✅ Verified a 🔥  button shows up
✅ Verified that, on press, a confirmation UI shows up
✅ Verified that `Delete Tab` is no longer present in the `...` Menu
✅ Verified that disabling `escapeHatchButton` restores the previous behavior
✅ Verified this works as expected with and without Favorites

One minor note, I'm having issues triggering the Escape Hatch, when restoring a Fire Tab. Would it be possible to confirm if it's just me, or we might have a regression elsewhere?

Approving nonetheless, since there are no changes in this branch that might explain a Fire Tabs regression. I'll poke you tomorrow over MM!

Nice work Sabrina!!

:shipit:

PR #5256, opened 2026-06-03
Add support for Monthly pixels in PixelKit

jdjackson:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1214036592228411?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1214950124367781?focus=true
CC: @federicocappelli, @brindy, @nshuba 

### Description
Adds support for monthly pixels to PixelKit in BSK.  Notable changes in this PR in order to make this work properly:
- Modified how Daily pixels fire, since they previously relied on `updatePixelLastFireDate` that assumed daily was the only calendar-associated firing frequency.  Now this method takes a frequency as input, and stores a map of last fired dates for any cadence used (daily, monthly, and both unique types), ensuring each cadence operates independently.
- This also necessitates a lazy migration of data for existing last fire date information, mapping it from a raw `Date` to a `[String: Date]`, storing a last fired date by frequency name.

### Testing Steps

1. All tests should pass.  This PR adds tests to cover monthly firing cases, including transition to next month and seeing the pixel fire a second time.
2. Ensure that daily pixels fired on the current day using a version _without_ these changes does not fire again after updating to a version with these changes included.  This confirms that the lazy migration of last fired data is working properly.
3. Optional: Ensure that monthly pixels fired on a day don’t fire again on the following day (by hand, may not be worth the time involved).
5. Don’t bother: Test monthly pixels fire in the next month by hand, across a month-change boundary :)

### Impact and Risks
Risk: Medium

#### What could go wrong?
Risk is primarily around the possibility of re-firing daily, unique, and unique-by-name-and-parameters, if the lazy migration code doesn’t work properly.

### Quality Considerations
- Documentation updating: We should consider updating documents if any refer specifically to Apple pixel firing types (in which case we should add reference to monthly pixels now).

### Notes to Reviewer
- Pay special attention to the lazy data migration implementation in the `hasBeenFired…` methods, as this is where most of the risk for this change comes from.

---

> [!NOTE]
> **Medium Risk**
> Changes core pixel dedupe and on-disk format; a failed lazy migration could re-fire daily or unique pixels and skew analytics.
> 
> **Overview**
> Adds a **`.monthly`** pixel frequency to PixelKit: fires at most once per UTC calendar month, appends `_monthly` to the pixel name, and skips repeats within the same month (including an early exit in `fire()`).
> 
> Last-fire persistence is refactored from a single stored `Date` to a **`[frequency mapKey: Date]`** map per pixel key. `updatePixelLastFireDate` and `pixelLastFireDate` now take a **`frequency`** argument so daily, monthly, and unique cadences do not overwrite each other. Setup/init renames **`dailyPixelCalendar`** to **`pixelCalendar`**.
> 
> **Lazy migration** on read/write converts existing raw-`Date` entries into a map with the legacy value under the `"daily"` key and persists the upgrade, so pixels already fired today should still dedupe after app update. Legacy alternate UserDefaults key lookup for daily dedupe is removed; **`pixelHasBeenFiredToday`** is renamed to **`pixelHasBeenFiredDailyToday`**.
> 
> Tests and helpers cover monthly behavior, legacy migration, and daily vs monthly independence. macOS VPN new-user flow passes **`frequency: .uniqueByName`** when reading `pixelLastFireDate`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b98888a3f950396b06a84f3a9ad6e88771f8cbb0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsymons:
Ignore the failed Asana task workflow, it's related to the recent PR task improvements - the workflow is trying to assign a review task to Cursor:

```
User cursor[bot] not found in user map
```

I'll get this fixed up with @alessandroboron.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -41,6 +41,12 @@ public final class PixelKit {
         /// Sent once per day. The last timestamp for this pixel is stored and compared to the current date. Pixels of this type will have `_daily` appended to their name.
         case daily
 
+        /// Sent once per calendar month (UTC). The last timestamp for this pixel is stored and compared to the current date. Pixels of this type will have `_monthly` appended to their name.
+        /// Note that these monthly pixels do not behave the same as ATB-based monthly measures like our official MAU stats, and thus direct comparisons between monthly pixel derrived values
+        /// and ATB-derrived monthly stats should _not_ be made. Tech Design for monthly pixels is available here:
```
samsymons:
```suggestion
        /// Note that these monthly pixels do not behave the same as ATB-based monthly measures like our official MAU stats, and thus direct comparisons between monthly pixel derived values
        /// and ATB-derived monthly stats should _not_ be made. Tech Design for monthly pixels is available here:
```

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
samsymons, line 1: With the changes in this file I believe we can now remove the `legacyUserDefaultsKeyName` function 🎉

samsymons (approved):
LGTM! Left a comment about a function that is no longer used, which conveniently already has a comment attached saying it should be eventually removed.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
jdjackson: That appears to be the case, though I hadn’t actually realized it while working on the change.  Considering it was added back in April of 2024, and intended to be only around for ~6 months, I think we’re safe at this point.  If an app actually does still have a legacy key name in use, the result will be that we simply ignore that data rather than migrating it like we used to, but given how long it’s been, I think it’s the right call at this point.

PR #5251, opened 2026-06-03
[iOS/macOS] Reduce Package.resolved churn

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1215392615473577?focus=true
Tech Design URL:
CC:

### Description

This PR reduces the amount of times that we should encounter Package.resolved being modified after Xcode resolves packages, through the following steps:
1. Removing the workspace-level Package.resolved file from source control
2. Adding a new script to allow easily resolving dependencies in both projects
3. Updating the Dependabot workflow to ask the reviewer to run the script and commit the change
4. Add a Bugbot rule for package changes

### Testing Steps

1. Check that the script works as expected, and the Bugbot rules look correct

### Notes to Reviewer

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Low Risk**
> Mostly repo hygiene and automation; lockfile bumps affect build inputs but follow existing pinned-SPM practice and are limited to dependency resolution metadata.
> 
> **Overview**
> This PR **stops tracking** the workspace-level `DuckDuckGo.xcworkspace/.../Package.resolved` (gitignore + removal from the repo) so Xcode’s regenerated aggregate lockfile no longer creates noisy diffs; **per-project and package** `Package.resolved` files remain the source of truth for CI.
> 
> It adds **`scripts/resolve-project-dependencies.sh`** to re-resolve iOS/macOS Xcode projects and `DataBrokerProtectionCore`, and updates the **Dependabot PR body** workflow so reviewers run that script and commit any changed lockfiles (or an empty ping commit if none change).
> 
> **`.cursor/BUGBOT.md`** replaces the old “compare app lockfiles to workspace” rule with a **Dependency Changes** section: exact-version pinning, when to flag missing downstream lockfile updates, and what to ignore (e.g. `originHash`-only churn).
> 
> The diff also **refreshes committed lockfiles** (e.g. `content-scope-scripts` **15.0.0** on iOS/macOS/DataBrokerProtectionCore) and pins **`swift-algorithms`** to **`exact: "1.2.1"`** in `DataBrokerProtectionCore/Package.swift`, illustrating the intended re-resolve flow.
> 
[...]

PR #5257, opened 2026-06-03
Use PR merge base for localization checks

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1215162688786291?focus=true
Tech Design URL: N/A
CC: N/A

### Description

- Updates the localization PR-check helpers to compare changed files and base file contents against the merge base of `HEAD` and the PR base branch.
- Keeps the existing `origin/<base ref>` fallback if a merge base cannot be resolved.
- Prevents localization marker issues outside the PR diff from being attributed to the PR.
- This specifically addresses the stale marker-mismatch comment seen on #5251, where the reported `Localizable.strings` file was not part of the PR diff.

### Testing Steps

1. This is an edge case that depends on PR/base-branch timing, so it is hard to reproduce consistently with a local command. It can happen when a PR branch is created from `main`, `main` later advances with localization changes, and the localization helper compares the PR against the moving `origin/main` tip instead of the PR branch's merge base.
2. In #5251, that made the marker check attribute an existing German `Localizable.strings` issue to the PR, even though `Localizable.strings` was not in #5251's changed files.

### Impact and Risks

None / Low: CI-only localization tooling. This changes only which base revision the existing translation checks compare against during PR runs.

#### What could go wrong?

If the merge base cannot be resolved, the helper falls back to the existing `origin/<base ref>` behavior. The main risk is therefore unchanged behavior in unusual checkout states, not app runtime impact.

### Quality Considerations

- `GITHUB_BASE_REF` gives the base branch name, which is useful for choosing the branch but not sufficient as a fixed diff endpoint.
- Resolving `git merge-base origin/<base ref> HEAD` gives the same comparison base as a three-dot PR-style diff, avoiding false positives from files outside the PR diff.
- This mirrors the approach used by the existing merge conflict checker (`conflict-watch`), which also compares against the PR branch's merge base to avoid moving-base false positives.
- The original #5251 marker comment was stale/noisy rather than backed by the current PR diff: #5251's changed files did not include `Localizable.strings`, and later checks for its current head were green. The bug is still valid because the workflow could attribute unrelated localization changes to the PR before this fix.

### Notes to Reviewer

This follows up on a stale localization marker comment on #5251 after #5165/#5199 landed. The check result was real for an existing German string, but the PR attribution was wrong because the helper compared against the base branch tip instead of the PR branch's merge base.

--- 
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> CI-only change to which git revision is used for comparisons; fallback preserves prior behavior when merge-base cannot be resolved.
> 
> **Overview**
> Localization PR-check helpers now resolve **`git merge-base`** between `HEAD` and `origin/<base>` (from `GITHUB_BASE_REF`, else `origin/main`) and use that commit for **`get_changed_files`** and **`get_files_content_at_base`**, instead of diffing against the moving base-branch tip.
> 
> That aligns checks with a three-dot PR diff so marker/translation failures on files **outside** the PR's changes are not blamed on the PR. If merge-base lookup fails, behavior stays the same as before by falling back to `origin/<base>`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 993f0164baf3fcb42fc0a6efc930db2cf3549021. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/translation-pr-checks/localization_utils.py
+++ b/scripts/translation-pr-checks/localization_utils.py
@@ -10,7 +10,6 @@ Copyright © 2025 DuckDuckGo. All rights reserved.
 Licensed under the Apache License, Version 2.0
 """
 
-import json
 import os
 import plistlib
 import re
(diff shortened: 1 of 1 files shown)


samsymons (approved):
Thanks for the speedy turnaround, looks good!

PR #5301, opened 2026-06-05
Make site loading success and failure performance pixels cross platform

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214769851840802/task/1215201969942746?focus=true
Tech Design URL:
CC:

### Description
Makes the success and failure site loading pixels cross platform, and added firing them on iOS.
They were already privacy triaged.

It moves the machinery for the site loading timing pixel to also be shared, but doesn't actually fire it on iOS since it is reliant on a private webkit api

### Testing Steps

1. On both iOS and macOS, check the m_site_loading_success and m_site_loading_failure pixels fire with correct params (make sure the duration looks sensible). One site that always fails to load you can test on is pigeon.com
2. Make sure the m_site_loading_timing pixel still fires on macOS with correct params

1. There's a specific edge case worth testing for on iOS. Make sure only one tab is open, and navigate to a page that will fail (e.g. pigeon.com). Relaunch the app, wait for that page load to fail, and now make one that will succeed. Originally a pixel was not fired in this case, which required a fix to some existing logic

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches main-frame navigation delegate timing and error-page edge cases; mis-gating could skew load metrics but changes are sampled and privacy-scoped.
> 
> **Overview**
> **Centralizes site-loading telemetry** in shared `SiteLoadingPixel` (BSK PrivacyDashboard), including **20% sampling**, PII-safe `navigation_type` mapping, and shared rules for when success/failure should fire (JS redirects, alternate HTML, error-page `.other` reloads).
> 
> **iOS** gains `NavigationPixelNavigationResponder` hooked into `TabViewController`’s `WKNavigationDelegate`: navigation type is captured only when policy **allows** the load (avoids races with overlapping `decidePolicyFor`), duration is stamped on `WKNavigation`, and special-error-page flows use **URL-matched** `failedURL` so unrelated navigations during error-page loads still count. **macOS** drops its local `SiteLoadingPixel` and routes success/failure through the shared type with the same sampling; timing pixels stay macOS-only.
> 
> Pixel schema updates add `m_site_loading_success` / `m_site_loading_failure` on iOS and **sample** suffixes on macOS for those events.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61b867ceb9f80515d215ef463306114f4a608627. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/PixelDefinitions/pixels/definitions/navigation.json5
```diff
@@ -1,7 +1,41 @@
 {
+    "m_site_loading_success": {
```
diegoreymendez, line 2: I think both this and the next pixel should declare `suffixes`.

Something like this:
```
"suffixes": ["platform", "form_factor", <sample%>]
```

iOS/DuckDuckGoTests/NavigationPixelNavigationResponderTests.swift
```diff
@@ -0,0 +1,198 @@
...
+            request: URLRequest(url: url),
+            navigationType: type,
+            targetFrame: WKFrameInfo.mock(isMainFrame: false, securityOriginHost: "example.com")
+        )
+    }
+
+    // MARK: - Happy paths
+
+    func test_willStart_didStart_didFinish_firesSiteLoadingSuccess() {
+        let nav = WKNavigation()
```
diegoreymendez: There's a comment in `WebViewNavigationHandling.swift:29-34` that explains that direct instantiation of `WKNavigation` crashes on deinit.

Would it make sense to use something similar to `WebViewNavigation` for our testing?  Maybe a new narrow protocol for what we're doing?

iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -2326,6 +2346,7 @@ extension TabViewController: WKNavigationDelegate {
     }
     
     func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
+        navigationPixelResponder.didFail(navigation, error: error)
```
diegoreymendez: Below this line, this method filters out non-failure cases like download handoff and `NSURLErrorCancelled`.

Since `navigationPixelResponder.didFail` runs before those guards, those cases would be miscounted as site-loading failures.

Should we move this line after the early returns, just before `showErrorNow()` is scheduled?

iOS/DuckDuckGo/NavigationPixelNavigationResponder.swift
```diff
@@ -0,0 +1,112 @@
...
+import WebKit
+
+/// iOS counterpart of macOS's `NavigationPixelNavigationResponder`. The host `WKNavigationDelegate`
+/// (e.g. `TabViewController`) just forwards its callbacks via `willStart`, `didStart`, `didFinish`, `didFail`;
+/// this class owns the pending-navigation-type, the per-`WKNavigation` start time, the gating logic
+/// (`SiteLoadingPixel.shouldFireSiteLoadingPixel`), and the firing of
+/// `SiteLoadingPixel.siteLoadingSuccess` / `.siteLoadingFailure`.
+///
+/// Error-page state is provided via closures so this class stays decoupled from
+/// `SpecialErrorPageNavigationHandler`.
```
diegoreymendez: Nit: this doc comment is closely tied to the current wiring rather than describing the class itself. It references the macOS responder, `TabViewController`, forwarded delegate methods, and `SpecialErrorPageNavigationHandler`, all of which are outside this type and could drift over time.

iOS/PixelDefinitions/pixels/definitions/navigation.json5
```diff
@@ -1,7 +1,41 @@
...
+        "triggers": ["page_load"],
+        "parameters": [
+            "appVersion",
+            "pixelSource",
+            "duration",
+            {
+                "key": "navigation_type",
+                "type": "string",
+                "description": "Safe-string representation of the WKNavigationType that initiated the navigation"
+            }
```
diegoreymendez, line 12: I think we can use the shared `navigation_type` here and below.

iOS/DuckDuckGo/NavigationPixelNavigationResponder.swift
```diff
@@ -0,0 +1,112 @@
...
+        pendingNavigationType = nil
+    }
+
+    /// Forwarded from `webView(_:didFinish:)`. Fires `.siteLoadingSuccess` for navigations that passed the `willStart` gate.
+    func didFinish(_ navigation: WKNavigation?) {
+        guard let navigation,
+              let startTime = navigation.siteLoadingStartTime,
+              let navigationType = navigation.siteLoadingNavigationType else { return }
+        let duration = Date().timeIntervalSince(startTime)
+        PixelKit.fire(SiteLoadingPixel.siteLoadingSuccess(duration: duration, navigationType: navigationType))
```
diegoreymendez: I know this isn't new, and was a mistake in the original code — but this pixel will fire unbucketed and unsampled right now.

Should we consider 5ms buckets like in our startup-performance pixels?

iOS/DuckDuckGo/NavigationPixelNavigationResponder.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Forwarded from both `webView(_:didFail:withError:)` and
+    /// `webView(_:didFailProvisionalNavigation:withError:)`. Fires `.siteLoadingFailure` for navigations
+    /// that passed the `willStart` gate.
+    func didFail(_ navigation: WKNavigation?, error: Error) {
+        guard let navigation,
+              let startTime = navigation.siteLoadingStartTime,
+              let navigationType = navigation.siteLoadingNavigationType else { return }
+        let duration = Date().timeIntervalSince(startTime)
+        PixelKit.fire(SiteLoadingPixel.siteLoadingFailure(duration: duration, error: error, navigationType: navigationType))
```
diegoreymendez: Same comment as above about the pixel being unbucketed and unsampled.

diegoreymendez (commented):
Nice work here @THISISDINOSAUR!

I added a few comments and notes.  The main one is around the site-loading duration values.

Also I saw CI is failing a few steps, so it may be worth checking that before additional testing in case the fix is impactful.

Let me know if you have any questions.

iOS/DuckDuckGo/NavigationPixelNavigationResponder.swift
THISISDINOSAUR: I do think we should do the same sampling as the other page loading pixels, but I'm less sure about the bucketing, particularly since the other performance pixels don't use them.

How about I add sampling here, and then we talk about bucketing as part of the conversation about anything else we want to do with this project?

iOS/PixelDefinitions/pixels/definitions/navigation.json5
THISISDINOSAUR: Have addressed.

Do we have any documentation for this sort of thing? I tried to find some but couldn't find any on how the pixel spec files work

iOS/DuckDuckGo/NavigationPixelNavigationResponder.swift
THISISDINOSAUR: Have added sampling

iOS/DuckDuckGo/TabViewController.swift
THISISDINOSAUR: Good catch!

iOS/PixelDefinitions/pixels/definitions/navigation.json5
diegoreymendez: Honestly, I ask AI to help me find these shared values.  The whole pixel schema is hard to manually track for me.

iOS/DuckDuckGo/NavigationPixelNavigationResponder.swift
diegoreymendez: Yeah sampling is ok... it's what we intended.

iOS/PixelDefinitions/pixels/definitions/navigation.json5
THISISDINOSAUR: To be honest I had no idea how any of this works or that we have shared values, I apparently haven't added a pixel since we started using the schemas

diegoreymendez (approved):
Works well, good job @THISISDINOSAUR .  Approved!

Just one note that I'm not seeing the `performance` pixel from CSS but I assume that's expected for the scope of this PR.  Let me know if it's not!

PR #5325, opened 2026-06-09
iOS: Native input doesn't show that subscription model is not supported

pikorddg:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1215425504065599?focus=true
Tech Design URL:
CC:

### Description
- Adds native-input recovery handling for Duck.ai chats whose selected model is no longer available to the user.
- Lets FE ask native to show the UTI model picker from the subscription recovery card's `Switch Model` CTA.
- Adds native submit-button blocking for FE recovery-card state without hiding the rest of UTI.
- Adds native -> FE model-change notification when a supported model is selected from the native picker in an active chat.
- Preserves existing FE -> native model sync through `duckAiNativeStorage` / `UnifiedToggleInputCoordinator.observeChatUpdates`.
- Adds focused unit coverage for user-script messages, UTI recovery submit blocking, model picker behavior, and model-change dispatch.

### Testing Steps
Setup
* enable the `unifiedToggleInput` flag
* Note: there are two ways to get a subscription
  1. Sandbox account
  2. Adding through Menu > Settings > I have a subscription.
  * I recommend using the latter, its faster to remove subscription this way, rather than waiting ~30 minutes for Sandbox subscription to expire

1. Basic case: you have a subscription → start a chat with a gated model (plus or pro) → remove the subscription → when you go back to that chat, the WebView shows a recovery card with two buttons: Activate Subscription and Switch Model. Don't click either button yet.
2. Tap the native text field input. With empty text, the voice button should be shown. Start typing. The voice button changes into a submit button, which should be inactive because the user hasn't selected a supported model yet. You can delete the text.
3. Tap Activate Subscription in the web recovery card → it shows a web view where you can go to purchase a subscription.
4. Tap Switch Model in the web recovery card → it opens the native model picker. Don't pick a model yet.
   1. By default, the currently available model should be selected, most likely one of the Free models.
   2. Tap outside the native model picker (not selecting any model manually)
   3. Confirm that the default-selected supported model gets applied. At this point, the web recovery card should disappear.
   4. Start typing in the input text field. The voice button changes into an active submit button.
5. Send the prompt and confirm that the response comes from the newly selected supported/free model, not from the old gated model.
6. Open new chat and tap the model picker again:
   * Tap a free/supported model → the model gets selected.
   * Tap a gated model → it opens the subscription upsell, and the model is not switched to an unsupported one.
7. Regression smoke test for a normal chat without an expired subscription / recovery card:
   * [Duck.ai](http://duck.ai/) with UTI works as before.
   * The model chip after the first prompt behaves as before.
   * Follow-up prompts work.
   * Changing the model and model selection don't cause regressions in the standard flow.
8. Regression smoke test for the current FE → native sync:
   * If the FE changes the model and saves it through `duckAiNativeStorage`, the app uses that new model. Note: the simplest way to check this:
   * Create a new chat, e.g. with `gpt-oss.`
   * In Debug settings, start the Native Storage Server, copy the server's IP address, and open it in a browser other than DDG.
   * At the very top you should see the most recently created chat with the text Assistant: <model ID> gpt-oss.
   * Go back to the app, to the chat you created. Using the web model picker (redo button), pick a different model, e.g. GPT-4o mini.
   * The FE rewrote the response to GPT-4o mini. Check the Web Server — below gpt-oss you should now see the gpt-4o-mini model.
   * Go back to the app and send a new prompt in the same chat. Make sure you get a message from gpt-4o-mini and not gpt-oss.

### Impact and Risks

**Impact Level: Medium**

#### What could go wrong?

- Native and FE model state could diverge if FE does not handle the model-change action as expected.
- Submit button state could remain blocked if FE sends `disableChatInput` without a matching unblock path.
- Recovery-card model picker routing is limited to the foreground Duck.ai tab to avoid affecting background WebViews.

### Quality Considerations

- Existing `native-input=false` flow remains unchanged.
- Existing `duckAiNativeStorage` model sync remains the source of truth for FE-driven chat updates.
- Gated model selection continues to route through existing subscription upsell logic.

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes couple native submit and model state to new FE bridge messages; mismatched enable/disable or FE handling of submitChangeModelAction could leave submit blocked or model state diverged.
> 
> **Overview**
> Adds **native Unified Toggle Input (UTI) support** when Duck.ai shows a subscription recovery card for an unsupported model, so users can switch models and aren’t left with a misleading submit affordance.
> 
> The web frontend can now signal **`showModelPicker`** (e.g. “Switch Model”), **`disableChatInput`**, and **`enableChatInput`**. Native handles these by expanding UTI, **forcing the model chip visible mid-chat** (without starting a new chat), optionally **opening the model menu** (iOS 17.4+), and **disabling only submit** while the recovery card blocks sending—not hiding the whole input. Picking a supported model clears the block and pushes **`submitChangeModelAction`** to the FE so the active chat’s model stays in sync.
> 
> Also wires **`aiChatShowModelPicker`** notification routing on the foreground Duck.ai tab, extends **`AIChatInputBoxHandling`** with **`isSubmitBlockedByRecoveryCard`**, and adds focused unit tests. Unrelated: a macOS DBP E2E test is temporarily skipped, plus small style/refactor tweaks in subscription user script and subscription state resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46580fee91178d009dfc56d6d8af73577ab5b7b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
+++ b/iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
@@ -358,11 +358,10 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
     }
 
     func getFeatureConfig(params: Any, original: WKScriptMessage) async throws -> Encodable? {
-        let response = GetFeatureConfigurationResponse(
+        return GetFeatureConfigurationResponse(
             usePaidDuckAi: subscriptionFeatureAvailability.isPaidAIChatEnabled,
             useAlternateStripePaymentFlow: subscriptionFeatureAvailability.isSupportsAlternateStripePaymentFlowEnabled,
         )
-        return response
     }
 
     // Auth V1 unused methods
@@ -825,17 +824,6 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
     func pushAction(method: SubscribeActionName, webView: WKWebView, params: Encodable) {
         let broker = UserScriptMessageBroker(context: SubscriptionPagesUserScript.context, requiresRunInPageContentWorld: true)
         broker.push(method: method.rawValue, params: params, for: self, into: webView)
-    }
-
-    private func purchaseUpdateDebugSummary(_ purchaseUpdate: PurchaseUpdate) -> String {
-        guard let data = try? JSONEncoder().encode(purchaseUpdate),
-              let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {
-            return "unavailable"
-        }
-
-        let type = json["type"] as? String ?? "unknown"
-        let token = json["token"] as? String
-        return "type=\(type) tokenPresent=\(token?.isEmpty == false) tokenLength=\(token?.count ?? 0)"
     }
 
     // MARK: Native methods - Called from ViewModels
--- a/iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
+++ b/iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
@@ -227,7 +227,6 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
 
     func handler(forMethodNamed methodName: String) -> Subfeature.Handler? {
         Logger.subscription.debug("WebView handler: \(methodName)")
-        print("🇯🇵 [subscriptionPages bridge] FE -> Native handler method=\(methodName)")
 
         switch methodName {
         case Handlers.setAuthTokens: return setAuthTokens
@@ -270,13 +269,11 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
 
     private func setTransactionError(_ error: UseSubscriptionError?) {
         transactionError = error
-        print("🇯🇵 [subscriptionPages bridge] Native transactionError=\(error.map(String.init(describing:)) ?? "nil")")
     }
 
     private func setTransactionStatus(_ status: SubscriptionTransactionStatus) {
         if status != transactionStatus {
             Logger.subscription.log("Transaction state updated: \(status.rawValue)")
-            print("🇯🇵 [subscriptionPages bridge] Native transactionStatus \(transactionStatus.rawValue) -> \(status.rawValue)")
             transactionStatus = status
         }
     }
@@ -315,10 +312,8 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
     }
     
     func setAuthTokens(params: Any, original: WKScriptMessage) async throws -> Encodable? {
-        print("🇯🇵 [subscriptionPages bridge] FE -> Native setAuthTokens host=\(original.messageHost)")
 
         guard let subscriptionValues: SubscriptionValuesV2 = CodableHelper.decode(from: params) else {
-            print("🇯🇵 [subscriptionPages bridge] Native setAuthTokens decodeFailed")
             Logger.subscription.fault("SubscriptionPagesUserScript: expected JSON representation of SubscriptionValues")
             assertionFailure("SubscriptionPagesUserScript: expected JSON representation of SubscriptionValues")
             setTransactionError(.generalError)
@@ -328,10 +323,8 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
 
         // Clear subscription Cache
         subscriptionManager.clearSubscriptionCache()
-        print("🇯🇵 [subscriptionPages bridge] Native cleared subscription cache before adopting FE tokens")
 
         guard !subscriptionValues.accessToken.isEmpty, !subscriptionValues.refreshToken.isEmpty else {
-            print("🇯🇵 [subscriptionPages bridge] Native setAuthTokens received emptyToken accessTokenEmpty=\(subscriptionValues.accessToken.isEmpty) refreshTokenEmpty=\(subscriptionValues.refreshToken.isEmpty)")
             Logger.subscription.fault("Empty access token or refresh token provided")
             markEmailAddressRestoreWideEventFlowAsFailed(with: nil)
             return nil
@@ -340,17 +333,14 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
         do {
             try await subscriptionManager.adopt(accessToken: subscriptionValues.accessToken, refreshToken: subscriptionValues.refreshToken)
             guard let subscription = try await subscriptionManager.getSubscription(forceRefresh: true) else {
-                print("🇯🇵 [subscriptionPages bridge] Native setAuthTokens adopted tokens but subscription lookup returned nil")
                 Logger.subscription.error("No subscription found after token adoption")
                 setTransactionError(.failedToSetSubscription)
                 markEmailAddressRestoreWideEventFlowAsFailed(with: UseSubscriptionError.failedToSetSubscription)
                 return nil
             }
             Logger.subscription.log("Subscription retrieved: \(subscription.isActive ? "active" : "inactive", privacy: .public)")
-            print("🇯🇵 [subscriptionPages bridge] Native setAuthTokens subscription status=\(subscription.status.rawValue) isActive=\(subscription.isActive) tier=\(subscription.tier?.rawValue ?? "nil")")
             markEmailAddressRestoreWideEventFlowAsSuccess()
         } catch {
-            print("🇯🇵 [subscriptionPages bridge] Native setAuthTokens failed error=\(error)")
             Logger.subscription.error("Failed to adopt V2 tokens: \(error, privacy: .public)")
             setTransactionError(.failedToSetSubscription)
             markEmailAddressRestoreWideEventFlowAsFailed(with: UseSubscriptionError.failedToSetSubscription)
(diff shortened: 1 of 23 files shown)


iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -1354,16 +1370,56 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
...
+            isSubmitBlockedByRecoveryCard = false
+            notifyFrontendOfActiveChatModelChange(supportedModelId)
+        }
+        // Defer to the next runloop so the toolbar (and the now-revealed chip) is laid out after the
+        // expand animation before we ask the button to open its menu.
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            self.viewController.presentModelPickerMenu()
+        }
+    }
```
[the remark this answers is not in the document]
pikorddg: I think this is a false positive. `updateModelChipVisibility()` does run after the flag flips, just indirectly. 
`presentModelPickerForActiveChat` goes through `showExpanded() → refreshToolsPresentation() → updateModelChipVisibility()`, and the supported-selection paths reach it via `updateSelectedModel() → handleModelsUpdated()`. So the chip shows on Switch Model and hides again after a supported pick — covered by `test_presentModelPickerForActiveChat_revealsModelChip_duringActiveChat` and `test_selectingSupportedModel_afterPresentingModelPicker_reHidesModelChip`, and confirmed in manual testing. Leaving as-is.

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -755,6 +767,8 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
         cancelTopOmnibarKeyboardPresentationFallback()
         isAwaitingTopOmnibarKeyboardPresentation = false
         displayState = .hidden
+        isModelPickerForcedVisible = false
```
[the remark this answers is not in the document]
pikorddg: The per-tab concern was valid, switching tabs could leave a gated chat's submit unblocked. Fix landed on the FE side: it re-emits disableChatInput/enableChatInput when a chat returns to the foreground (not only on reload), so native always receives the current state. Native intentionally stays transient. It can't refresh backgrounded tabs (only the foreground chat is bound), so caching the block per-tab would just go stale on subscription changes. Verified: cold start → switch to the gated tab → submit correctly disabled. Resolved.
Related discussion: https://app.asana.com/1/137249556945/task/1215448612535170/comment/1215558713055984?focus=true

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputToolbarView.swift
```diff
@@ -116,6 +120,18 @@ final class UnifiedToggleInputToolbarView: UIView {
...
+    /// Programmatically opens the model chip's pull-down menu. Returns `true` when the OS
+    /// exposes an API to trigger it (iOS 17.4+, where `performPrimaryAction()` lands), `false`
+    /// otherwise.
+    @discardableResult
+    func presentModelPickerMenu() -> Bool {
+        if #available(iOS 17.4, *) {
+            modelChipButton.performPrimaryAction()
+            return true
+        }
+        return false
```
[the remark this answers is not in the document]
pikorddg: Intentional, and it degrades gracefully. There's no public API to programmatically open a UIButton pull-down menu before iOS 17.4 (performPrimaryAction() lands in 17.4). So on 15.0 - 17.3 we reveal the model chip and the user taps it — acceptable tradeoff, documented on the method. 
The block-clear + submitChangeModelAction firing regardless is deliberate: it reconciles the model the user already has selected (a supported one), so even without the menu auto-opening the user lands on a usable model and the recovery card drops; the revealed chip lets them switch to another. So the picker isn't absent — it's one tap away, and the CTA's intent (get off the unsupported model) is fulfilled either way. Plus, user can still selected gated model (and be redirected to subscription payment screen) if needed.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -358,10 +358,11 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
     }
 
     func getFeatureConfig(params: Any, original: WKScriptMessage) async throws -> Encodable? {
-        return GetFeatureConfigurationResponse(
+        let response = GetFeatureConfigurationResponse(
             usePaidDuckAi: subscriptionFeatureAvailability.isPaidAIChatEnabled,
             useAlternateStripePaymentFlow: subscriptionFeatureAvailability.isSupportsAlternateStripePaymentFlowEnabled,
         )
+        return response
```
aataraxiaa: Minor - assuming this was done for temp logging, could prob undo the `let` creation here

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -265,6 +272,11 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
     private var isContentOverlaySuppressed = false
     private var pendingGatedModelId: String?
     private var pendingGatedReasoningSelection: (modelId: String, mode: AIChatReasoningMode)?
+    /// Set when the FE asks (via `showModelPicker`) to surface the model picker on the active
+    /// chat. Additively overrides only the `hasSubmittedPrompt`- driven hide in
+    /// `updateModelChipVisibility`, so the chip becomes tappable mid-chat without touching the
+    /// existing show/hide rule. Cleared once a supported model is applied or the session resets.
```
aataraxiaa: Nit - I find CC too verbose in it’s comments. Usually ask it to make them more concise. Maybe can do the same here (4 line comments per property doesnt scale)

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -227,6 +227,13 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
     }
     @Published var attachmentUsage: AIChatAttachmentUsage?
 
+    var isSubmitBlockedByRecoveryCard: Bool = false {
```
aataraxiaa, line 230: We discussed this on Zoom, but not sure where we landed - Should this be part of tab input state? As it's currently global, will this work correctly when switching between different Duck.ai tabs or switching from a Duck.ai tab to a web tab and then back to a Duck.ai tab?

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputToolbarView.swift
```diff
@@ -116,6 +120,18 @@ final class UnifiedToggleInputToolbarView: UIView {
         }
     }
 
+    /// Programmatically opens the model chip's pull-down menu. Returns `true` when the OS
+    /// exposes an API to trigger it (iOS 17.4+, where `performPrimaryAction()` lands), `false`
+    /// otherwise.
+    @discardableResult
+    func presentModelPickerMenu() -> Bool {
+        if #available(iOS 17.4, *) {
```
aataraxiaa, line 128: What are we going to do for versions under 17.4?

aataraxiaa (commented):
Tests pass but left a few comments. Also from agent:

```
 three dead-code leftovers (compiler warnings)

These are from the "Remove temp diagnostic logs" commit — each will emit an "unused" warning (CI may fail if warnings-as-errors). One combined comment, or three quick ones:
- UTIModelStore.swift:119 — accessibleCount assigned, never used
- UnifiedToggleInputCoordinator.swift:1378 — reason computed, never used
- SubscriptionPagesUseSubscriptionFeature.swift:830 — purchaseUpdateDebugSummary(_:) never called
```

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
pikorddg: Global (coordinator), not in TabInputState, that’s intentional. After recent changes it’s FE-driven: FE re-emits disableChatInput/enableChatInput when a chat returns to the foreground, so the foregrounded Duck\.ai tab always gets the current state. 
Verified both cases on device. Switching between two Duck\.ai tabs, and Duck\.ai → web tab → back. The gated chat ends up correctly blocked and the free one unblocked each time.

Why I kept it out of TabInputState: native only binds the foreground chat (the coordinator is set as the inputBoxHandler on the current tab's user script), so it can't refresh backgrounded tabs.
Caching the block per-tab would go stale on subscription changes (e.g. buy a sub while a gated chat is backgrounded → that tab reloads but native isn't bound to it, so the cached value wouldn't update). 

Keeping FE the single source of truth avoids that. The re-emit-on-foreground landed on the FE side https://app.asana.com/1/137249556945/project/1204912272578138/task/1215448612535170?focus=true

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputToolbarView.swift
pikorddg: Below 17.4 there's no API to open the menu programmatically, so the model chip is revealed and the user just taps it to open the picker manually.

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
pikorddg: Thanks! 

For this and other logs-related comments: addressed in https://github.com/duckduckgo/apple-browsers/pull/5325/changes/891faa53a1065a24d73c492584267a14a64159e2

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputToolbarView.swift
aataraxiaa: Ok. We’ll just need to surface this to stakeholders so they are aware.

aataraxiaa (approved):

PR #5337, opened 2026-06-09
iOS: Unify UTI Search + Duck.ai suggestions onto one host

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215191426890192?focus=true

### Description

On the Unified Toggle Input (UTI), Search and Duck.ai suggestions were two independent surfaces, and the dax-logo decision was computed separately from the suggestions content — so the logo could leak into the wrong state or morph at the wrong time.

This collapses both onto one config-driven SwiftUI host fed by a pure decision pipeline:
- `UnifiedSuggestionsInputsMerger` (mode + text + per-surface state) → `UnifiedSuggestionsContentResolver` → a content kind (`.list` / `.favorites` / `.logo`).
- The **same resolver** now drives dax visibility, so "which logo, when" has a single source of truth.
- One `UnifiedSuggestionsHost` / `UnifiedSuggestionsView` renders both surfaces; each owns its own loaders/runner (fixes the mutual DDG-request cancellation and removes the `retainedObjects` hack).

Folds in the parity/polish the unified surface needed:
- Crossfade on mode-switch when the content category changes (favorites↔recents, list↔logo); snap on list↔list.
- Side-swipe to switch Search↔Duck.ai, routed through the toggle/mode-change path (smooth height + toggle slide, no accidental row taps).
- Row spacing + tap-ahead arrow alignment per Figma.
- Duck.ai sync-promo card (legacy parity).
- Fires the history-delete pixel on suggestion-row deletes.

Container thinned 1035→880 lines (`DuckAISurfaceController` + `ModeSwitchSwipeGestureController` extracted; dead voice-search + orphaned `DuckAISuggestionsViewController`/`Coordinator` removed). The legacy (flag-OFF) and iPad paths are untouched.

### Testing Steps

1. Enable the `unifiedToggleInput` feature flag.
2. Focus the UTI with no text: Search shows favorites + escape hatch; toggle to Duck.ai shows recent chats (or the dax logo when empty).
3. Type in Search → suggestion rows; toggle to Duck.ai → duck.ai suggestions. Confirm rows align with the input's text + X clear button.
4. Toggle Search↔Duck.ai: content crossfades on category change (favorites↔recents, list↔logo), snaps list↔list; the dax logo never appears over favorites/recents.
5. Side-swipe over the list **and** over favorites → switches mode; toggle slides (not snaps), height animates smoothly, small swipes don't trigger row taps.
6. On a Duck.ai history row, tap X → row deletes (history-delete pixel fires).
7. Duck.ai, not typing, with recent chats + sync available → sync-promo card shows below the hatch; CTA opens sync setup, close dismisses; switching to Search snaps it away with no recents flash.
8. Focus/dismiss: dax logo slides between the NTP and focused positions on top **and** bottom address bar.
9. **Legacy regression (flag OFF), iPhone + iPad:** confirm the omnibar editing surface is unchanged vs `main` — favorites, Search/Duck.ai suggestions, dax logo, history-delete, and the interactive swipe all behave as before.
10. **iPad (flag ON):** focus the UTI and confirm Search + Duck.ai suggestions, favorites, and the dax logo render correctly — no regression vs the iPhone path.

### Impact and Risks

Medium.
- Entirely behind the `unifiedToggleInput` feature flag; legacy (flag-OFF) and iPad paths untouched.
- Large refactor of the focused-omnibar suggestions surface, but the decision pipeline is pure and unit-tested, and `DaxLogoManager`'s boolean interface is unchanged (legacy-safe).

### Notes to Reviewer

- `DaxLogoManager` still *renders* the UTI dax; only the *decision* is resolver-driven now. Migrating the rendering into the SwiftUI host is a deliberate follow-up PR (plan written) — keeps the slide via a UIKit-constraint handoff.
- iPad and the flag-OFF path deliberately reuse the shared suggestion/history components (`AIChatHistoryListViewController`, `SuggestionTrayManager`, `SwipeContainerManager`) — only the UTI-iPhone surface was unified; please sanity-check those two paths for regressions.
- **Known follow-ups (not blocking):** dax-rendering → SwiftUI host (separate PR); `toolbarCompensationOffset` de-hack (low-pri); optional `SuggestionRowView` snapshot guards.
- **CI:** host-app unit suites (`DaxLogoManagerTests`, `UnifiedInputContentContainerViewControllerTests`, `OmniBarEditingStateViewControllerDaxVisibilityTests`) couldn't run locally (CoreSimulator launchd wedge); please confirm green in CI.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Small ordering change in favorites list updates; limited to bookmark UI state refresh after save.
> 
> **Overview**
> **Fixes stale favorites for observers after remove/reorder.** In `FavoriteListViewModel`, `readFavorites` is no longer called directly from `removeFavorite` and `moveFavorite`; instead it runs inside `save()` **after** a successful Core Data save and **before** `localSubject.send()`.
> 
> That way the published `favorites` array matches persisted state when reactive consumers (e.g. another New Tab Page instance) handle the update, instead of reading the pre-change list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20f63524512d5ccd1926a878bf9413d57ec80ddc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jleandroperez:
@aataraxiaa @jaceklyp Gentleman, please, do not merge until we close this task:
https://app.asana.com/1/137249556945/project/1211150618152277/task/1215556986150983?focus=true

Thanks in advance!!

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SearchSuggestionsLoader.swift
```diff
@@ -0,0 +1,81 @@
...
+              url.isValid(usingUnifiedLogic: useUnifiedURLPrediction) else {
+            return true
+        }
+        if let scheme = url.scheme, scheme.description.hasPrefix("http"), url.isRoot, phrase.last == "/" {
+            return false
+        }
+        return true
+    }
+
+    func subscribeToTextChanges<P: Publisher>(_ textPublisher: P)
```
aataraxiaa, line 60: Chatted a bit on MM about this - This runs for chat-mode text too, and we send the full typed query to ac/ and surface all categories - legacy code filtered chat mode to URL-only. Believe this is intended but highlighting (you mention this in the type comment above)

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/UnifiedSuggestionsViewModel.swift
```diff
@@ -0,0 +1,80 @@
...
+    private var cancellable: AnyCancellable?
+    private var previousMode: TextEntryMode?
+
+    init(inputsPublisher: AnyPublisher<UnifiedSuggestionsInputs, Never>,
+         listViewModel: SuggestionsListViewModel,
+         duckAIListViewModel: SuggestionsListViewModel? = nil) {
+        self.listViewModel = listViewModel
+        self.duckAIListViewModel = duckAIListViewModel
+        cancellable = inputsPublisher
+            .sink { [weak self] inputs in
```
aataraxiaa, line 52: Should we add a `.receive(on: DispatchQueue.main).` here?

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/Suggestions/DuckAISuggestionsCoordinatorDelegate.swift
```diff
@@ -0,0 +1,28 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import AIChat
+import Suggestions
+
+protocol DuckAISuggestionsCoordinatorDelegate: AnyObject {
```
aataraxiaa: With `DuckAISuggestionsCoordinator` deleted, does anything still hold this protocol? If not, could we drop the conformance and make these plain helpers, and delete this file to finish the cleanup?

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/UnifiedSuggestionsHost.swift
```diff
@@ -0,0 +1,216 @@
...
+    /// Built once on first `.favorites` render; NTP has a heavy init, so don't rebuild per body pass.
+    private var cachedFavoritesController: NewTabPageViewController?
+
+    /// Single-host path only: the duck.ai surface's source/VM, attached lazily and detached on
+    /// disappear (mirrors the legacy per-host lifecycle). Nil on the old single-surface path.
+    private var duckAISurface: UnifiedSuggestionsDuckAISurface?
+    /// Captured on `start` so a duck.ai source attached later can drive its own text changes.
+    private var startedTextPublisher: AnyPublisher<String, Never>?
+
+    private func memoizedFavoritesController() -> NewTabPageViewController? {
```
aataraxiaa, line 52: The NTP is memoized and built once with T`ab(fireTab:)`. If fire mode toggles and the host is reused, won't the cached NTP show the stale fire-mode state? Should we invalidate on fire-mode change?

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/DuckAISuggestionsSurfaceProvider.swift
```diff
@@ -0,0 +1,183 @@
...
+
+    private func select(rowID id: String, source: DuckAISuggestionsSource) {
+        guard let selection = source.selection(forRowID: id) else { return }
+        delegate?.duckAISurfaceDidSelect(selection)
+    }
+
+    private func deleteHistory(rowID id: String, source: DuckAISuggestionsSource) {
+        guard case .url(let suggestion) = source.selection(forRowID: id),
+              case .historyEntry(_, let url, _) = suggestion else { return }
+        Task {
```
aataraxiaa: This deletion Task isn't retained/cancelled - if the UTI is dismissed mid-delete, fetchURLSuggestions runs against a detached source. We could store the handle and cancel it in detach? (Same pattern in the container's search onDeleteRow.)"

aataraxiaa (commented):
Tested and works well. Left comments. Also:

> Confirm rows align with the input's text + X clear button.
Maybe misunderstood this but row trailing edge did not align with input X.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SearchSuggestionsLoader.swift
jaceklyp: Valid call-out — now addressed: the Search source is gated to `.search` mode (commit 7ad00d8081), so the typed prompt no longer hits `ac/` while in Duck.ai. The URL suggestions you'd still expect in Duck.ai come from the surface's own `DuckAIURLSuggestionsLoader`, unchanged.

One correction to the framing: the prompt already reached `ac/` in legacy too — `showURLOnlySuggestions` applies a `.urlsOnly` filter to the *results*, not the query — so this was a redundant *extra* request rather than a new leak. Still worth removing.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/UnifiedSuggestionsViewModel.swift
jaceklyp: Valid concern, but `.receive(on: DispatchQueue.main)` here isn't the right fix: it reschedules even when already on main, adding a frame's delay to every content update (typing/mode-switch) and breaking the synchronous-emission VM tests.

Traced all four upstreams: toggle/text are main; the suggestion-loader callbacks are main (`SuggestionLoading.swift:103` wraps the completion in `DispatchQueue.main.async`); the chat VM is `@MainActor`. The only off-main source is favorites `externalUpdates`, which fires inside `context.perform { subject.send() }` (Core Data queue). Fixed by marshaling that one publisher at the source in the merger (commit 8d6adb3878), so the VM stays synchronous and the off-main mutation is gone.

iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/Suggestions/DuckAISuggestionsCoordinatorDelegate.swift
jaceklyp: Done (commit 4f64f24ae5) — verified dead (only the container's `extension … : DuckAISuggestionsCoordinatorDelegate` referenced it, and its methods are called directly rather than through the protocol), so dropped the conformance and deleted the file.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/DuckAISuggestionsSurfaceProvider.swift
jaceklyp: Fixed (commit 8d6adb3878) — store the delete `Task` and cancel it in `detach()`, with a `!Task.isCancelled` guard before the post-delete `fetchURLSuggestions`. Mirrored the same store-and-cancel (cancelled on `deinit`) for the container's search `onDeleteRow`.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/UnifiedSuggestionsHost.swift
jaceklyp: Valid, but deeper than caching. `makeSearchFavoritesController` builds the NTP from `Tab(fireTab: …shouldCreateFireTabs)`, and `shouldCreateFireTabs` is the global "open new tabs in fire mode" setting — not the current tab's fire state. Invalidating + rebuilding the cache on fire-mode change actually surfaced a fire NTP on a normal tab (reproduced on-device).

So the correct fix is to first make the NTP derive fire state from the current tab, then invalidate safely — filing that as a follow-up rather than shipping the naive invalidation. Leaving the memoize-once behavior as-is for now.

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputContentContainerViewController.swift
```diff
@@ -747,7 +832,6 @@ private extension UnifiedInputContentContainerViewController {
 
     private func refreshVisibleContent(
         suggestionRefresh: SuggestionRefreshStrategy,
```
jleandroperez: This parameter is no longer wired, and can be dropped

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SuggestionRowMapper.swift
```diff
@@ -0,0 +1,115 @@
...
+                accessibilityID: "Autocomplete.Suggestions.ListItem.Website-\(url.formattedForSuggestion())")
+
+        case .bookmark(let title, let url, let isFavorite, _):
+            return SuggestionRow(
+                id: "\(idPrefix)-bookmark-\(url.absoluteString)",
+                icon: isFavorite ? .favorite : .bookmark,
+                title: title,
+                query: nil,
+                subtitle: url.formattedForSuggestion(),
+                accessibilityID: "Autocomplete.Suggestions.ListItem.Bookmark-\(url.formattedForSuggestion())")
```
jleandroperez: Minor note, legacy code used a different Accessibility Identifier for Favorites:

```
.accessibilityIdentifier("Autocomplete.Suggestions.ListItem.Favorite-\(url.formattedForSuggestion())")
```

**Quite possibly** safe to ignore, mentioning this just in case!

Ref. https://github.com/duckduckgo/apple-browsers/blob/main/iOS/DuckDuckGo/AutocompleteView.swift#L195

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/AutocompleteRequestRunner.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  AutocompleteRequestRunner.swift
```
jleandroperez, line 2: Just a quick note, we do have a few SwiftLint errors like this

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SuggestionsSource.swift
```diff
@@ -0,0 +1,145 @@
...
+
+        if id.hasPrefix("chat-") {
+            let chatID = String(id.dropFirst("chat-".count))
+            return chats.first { $0.id == chatID }.map { .chat($0) }
+        }
+        if id == "search-searchDuckDuckGo" {
+            return .searchDuckDuckGo(q)
+        }
+        if id.hasPrefix("urls-") {
+            return urls.first { SuggestionRowMapper.row(for: $0, query: q, idPrefix: "urls").id == id }
```
jleandroperez: Perhaps it'd be good to extract all of these strings into constants?

jleandroperez (commented):
Looks good @jaceklyp, first pass on my end, all test scenarios described work well.

I did spot a couple regressions (details below), and sending you a quick quesiton.

Thanks in advance! 🙏


### Question A
```
iPad (flag ON): focus the UTI and confirm Search + Duck.ai suggestions, favorites, and the dax logo render correctly — no regression vs the iPhone path.
```

How can I enable Unified Input on iPad? (

### Regression A: Deleted URL Suggestion 
1. Tap the Address Bar
2. Start typing a URL that's in your history
3. Switch to `Duck.ai`
4. Delete a URL Suggestion 
5. Switch back to Search

As a result, the URL Suggestion deleted in (4) will still show up when switching to Search.

https://github.com/user-attachments/assets/d0013812-dd87-42eb-a4fd-f70c1021c6c3

### Regression B
1. Fresh Install
2. Enable the flags `Unified Input` + `removeChatHistory`
3. Start a new Duck.ai Chat
4. Go back to the Address Bar
5. Switch to Duck.ai

As a result, the `Remove Chat` button won't be visible (as it's cached per session).
I'll be sending you a PR tomorrow to fix this one.

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputContentContainerViewController.swift
jaceklyp: Good catch — `suggestionRefresh` was accepted but never read in `refreshVisibleContent`'s body (only `animateContentUpdates` is). Dropped it, and removed the now-dead `SuggestionRefreshStrategy` enum + `currentModeSuggestionRefresh()` helper + the call-site locals. (commit d00ca8eabb)

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SuggestionRowMapper.swift
jaceklyp: Good spot — fixed: favorites now use `Autocomplete.Suggestions.ListItem.Favorite-…` (non-favorite bookmarks keep `Bookmark-…`), matching the legacy `AutocompleteView`. (commit d00ca8eabb)

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/AutocompleteRequestRunner.swift
jaceklyp: Noted. SwiftLint is currently `skipping` in this PR's checks (not blocking), and the short header style is consistent across all the new files this branch adds, so I've left it rather than churn headers across the set — happy to do a sweep to the full Apache header if you'd prefer.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SuggestionsSource.swift
jaceklyp: Done — extracted the section ids into a `SectionID` enum (`chats`/`urls`/`search`) and reused the same constants for the `idPrefix`es. (commit d00ca8eabb)

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/AutocompleteRequestRunner.swift
jaceklyp: Update — went ahead and fixed these: applied the full Apache `file_header` block across all the new suggestion files (ran SwiftLint locally → 0 `file_header` violations now). commit e410f22451

jaceklyp:
> Looks good @jaceklyp, first pass on my end, all test scenarios described work well.
> 
> I did spot a couple regressions (details below), and sending you a quick quesiton.
> 
> Thanks in advance! 🙏
> 
> ### Question A
> ```
> iPad (flag ON): focus the UTI and confirm Search + Duck.ai suggestions, favorites, and the dax logo render correctly — no regression vs the iPhone path.
> ```
> 
> How can I enable Unified Input on iPad? (
> 
> ### Regression A: Deleted URL Suggestion
> 1. Tap the Address Bar
> 2. Start typing a URL that's in your history
> 3. Switch to `Duck.ai`
> 4. Delete a URL Suggestion
> 5. Switch back to Search
> 
> As a result, the URL Suggestion deleted in (4) will still show up when switching to Search.
> 
>  ScreenRecording_06-10-2026.17-52-07_1.MP4 
> 
> ### Regression B
> 1. Fresh Install
> 2. Enable the flags `Unified Input` + `removeChatHistory`
> 3. Start a new Duck.ai Chat
> 4. Go back to the Address Bar
> 5. Switch to Duck.ai
> 
> As a result, the `Remove Chat` button won't be visible (as it's cached per session). I'll be sending you a PR tomorrow to fix this one.

Fixed both

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputContentContainerViewController.swift
```diff
@@ -371,163 +357,255 @@ final class UnifiedInputContentContainerViewController: UIViewController {
...
+            aiChatSyncCleaner: aiChatSyncCleaner,
+            featureFlagger: featureFlagger,
+            privacyConfigurationManager: privacyConfigurationManager,
+            duckAiNativeStorageHandler: duckAiNativeStorageHandler
+        )
+        surface.delegate = self
+        surface.statePublisher
+            .sink { [weak self] in self?.duckAIStateRelay.send($0) }
+            .store(in: &duckAIRelayCancellables)
+        surface.attach(to: host, textPublisher: switchBarHandler.currentTextPublisher.eraseToAnyPublisher())
```
aataraxiaa, line 569: Does this mean that even for search query typing we fire this? And so in Search mode, typing fires a second ac/ request via the duck.ai URL loader, in parallel with the search loader's own.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SuggestionsListView.swift
```diff
@@ -0,0 +1,137 @@
...
+            Button {
+                viewModel.selectRow(id: row.id)
+            } label: {
+                SuggestionRowView(
+                    row: row,
+                    isAddressBarAtBottom: isAddressBarAtBottom,
+                    onTapAhead: { viewModel.tapAheadRow(id: row.id) },
+                    onDelete: { viewModel.deleteRow(id: row.id) },
+                    onFire: { viewModel.fireDeleteRow(id: row.id) })
+            }
```
aataraxiaa, line 72: Minor accessibility for testing -  SuggestionsListView.swift:66 — row accessibilityID is mapped but never applied. SuggestionRowMapper fills it for every row; SuggestionRowView only puts an id on the delete/fire glyph. One line: .accessibilityIdentifier(row.accessibilityID)

aataraxiaa (approved):
@jaceklyp - Great work!

 Two comments and also those bugs we discussed on Zoom. Conditionally approving based on bugs being fixed.

Note I think @jleandroperez should do a full review before we merge here.

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedInputContentContainerViewController.swift
jaceklyp: Yeah it does - but it's not new, legacy did the same: it passed the raw currentTextPublisher to the duck.ai coordinator too, so the URL loader fetched in both modes there as well.
We can gate it to .aiChat (like Search is gated to .search), but trade-off: right now duck.ai URL hits are pre-fetched so they're instant on toggle - gating means they fetch on entering Duck.ai (small delay, recents unaffected). Since this PR is parity-focused I'd leave it and file a follow-up, but can do it here if you prefer.

jaceklyp: Yeah it does — but it's not new, legacy did the same: it passed the raw `currentTextPublisher` to the duck.ai coordinator too (container `:525` at merge-base), so the URL loader fetched in both modes there as well.

We can gate it to `.aiChat` (like Search is gated to `.search`), but trade-off: right now duck.ai URL hits are pre-fetched so they're instant on toggle — gating means they fetch on entering Duck.ai (small delay, recents unaffected). Since this PR is parity-focused I'd leave it and file a follow-up, but can do it here if you prefer.

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/SuggestionsListView.swift
jaceklyp: Good catch — applied `.accessibilityIdentifier(row.accessibilityID)` to the row Button so every row is targetable, not just the delete/fire glyph. commit d12d6ee870

iOS/DuckDuckGo/UnifiedToggleInput/Suggestions/UnifiedSuggestionsView.swift
```diff
@@ -0,0 +1,135 @@
...
+    /// `.transition` (which snaps when interrupted by rapid Search↔Duck.ai toggling). Its opacity is
+    /// instant (`.animation(nil)`): the incoming list fades in, but favorites must not linger visibly
+    /// over Duck.ai while the crossfade runs. Logo is still drawn by DaxLogoManager.
+    @ViewBuilder
+    private var overlayLayer: some View {
+        if let controller = favoritesProvider() {
+            SuggestionsFavoritesView(controller: controller)
+                .opacity(isShowingFavorites ? 1 : 0)
+                .animation(nil, value: isShowingFavorites)
+                .allowsHitTesting(isShowingFavorites)
```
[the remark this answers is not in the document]
jaceklyp: Looked into this — keeping as-is. The NTP controller's `init` only builds a few view models + the SwiftUI tree (favicons load lazily) and `viewDidLoad` just registers notifications, so it's a moderate one-time cost, not heavy. More importantly, focusing into Search already builds it on the first render (the common path), so gating it would only defer the build for a Duck.ai-first focus — a narrow case not worth the extra state. The controller is memoized, so it's still built at most once per session. Mounting it (vs an insert/remove `.transition`) is required for the favorites↔recents crossfade not to snap on rapid toggling.

jleandroperez (commented):
Hello there @jaceklyp!

Confirming the fixes for the issues reported on the last review.
Unfortunately, I've found something new, details below:

### Search Row is not pressable
1. Enable Unified Input
2. Press on the Search Bar
3. Type a random text that yields no results
4. You should get a row with the 🔍 icon

If you attempt to press on such row, the Press event won't be handled.


##### Demo

https://github.com/user-attachments/assets/6a5d32d3-f7b9-4c65-a85a-12d9f853a94f

jaceklyp:
> ### Search Row is not pressable
> 1. Enable Unified Input
> 2. Press on the Search Bar
> 3. Type a random text that yields no results
> 4. You should get a row with the 🔍 icon
> 
> If you attempt to press on such row, the Press event won't be handled.

Should be fixed now 👍

jleandroperez (approved):
✅ Verified that the Escape Hatch shows below the Unified Input
✅ Verified that History and duck.ai suggestions are rendered
✅ Verified that Search / Duck.ai crossfades on category changes
✅ Verified that Side Swipe effectively switches the Mode
✅ Verified that if you delete a URL History Row, when in Duck.ai, and then switch to Search, the row doesn't come back
✅ Verified that Duck.ai history rows can be deleted
✅ Verified that the Sync Promo Card is rendered below Unified Input, and that it works as expected
✅ Verified that de Dax Logo slides between NTP and Focused Position
✅ Verified that the No-Results Row handles taps

Nice work @jaceklyp !! 

:shipit: on my end

PR #5344, opened 2026-06-10
iOS: Per-row pin/unpin in chat history sheet

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215102788845939

### Description

Per-row pin/unpin in the native chat history sheet. Leading-swipe a row to toggle; the row animates between the Pinned and Recent sections. Note sync support will be in a separate PR.

### Testing Steps

1. `aiChatNativeChatHistory` flag on.
2. Browser menu → **Chats**.
3. Leading-swipe a Recent row → tap pin → row animates into Pinned with the pinned glyph.
4. Leading-swipe the same row in Pinned → tap unpin → row animates back to Recent.
5. Background/foreground the app → pin state persists.
6. Open Duck.ai (web sheet) → same chat shows the same pin state.

### Impact and Risks

Low. Feature-flagged behind `aiChatNativeChatHistory` (internal-only). Change is additive — leading-swipe didn't exist before.

#### What could go wrong?

- **JSON round-trip drops a field.** Mitigated: `ChatPinner` uses `JSONSerialization` (not decode + re-encode), so fields the FE adds in the future survive. Covered by `testSetPinned_preservesAllOtherBlobFields`.
- **Storage write fails silently.** VM keeps the optimistic UI state and logs; the reactive `ChatHistoryReader` is the source of truth and will correct the UI on the next emission. Failure-state pixel pairs with the pending pixels-pass follow-up (task #28).
- **Concurrent FE write race.** If FE writes the same chat blob between our `getChat` and `putChat`, we'd clobber its write. Low likelihood (user has the native sheet open, not Duck.ai). Fixable later by moving read+mutate+write into a single GRDB transaction.

### Quality Considerations

- **Tests:** `ChatPinnerTests` covers JSON round-trip preservation, missing-key default, idempotent same-value, error paths. VM tests cover happy paths in both directions (incl. correct target-state passed to pinner), missing-pinner no-op, chat-not-found, error swallow, and `lastEdit`-desc insertion ordering.
- **Performance:** Pin write is off-main on `userInitiated` QoS.
- **Privacy:** No new data persisted; flipping a flag already in the chat blob.
- **Accessibility:** Same `Size24.pin` glyph for both states (matches Android), differentiated via `aiChatHistoryPinSwipeAccessibilityLabel` / `aiChatHistoryUnpinSwipeAccessibilityLabel`.

### Notes to Reviewer

[...]

PR #5348, opened 2026-06-10
iOS: Wire native chat-history pin into Sync

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215104073948290?focus=true

### Description

Adds the native-side sync push for chat-history pin/unpin and delete. 

### Testing Steps

> [!NOTE]
> Prerequisites:
> - [ ] Two devices signed into the same Sync account (e.g. simulator + a second internal-build device, or simulator + Duck.ai web on the same account).
> - [ ] `aiChatNativeStorage` flag on (gates the native chat store — without it the sheet shows "Couldn't load chats").
> - [ ] `aiChatNativeChatHistory` flag on (gates the sheet).
> - [ ] `supportsSyncChatsUpdate` flag on (gates the new push).

1. On device A: pin a Recent chat from the native history page (open Chats from the app menu) you need to swipe the cell right
2. On device B: open Duck.ai  within ~10s — confirm the same chat now appears in the Pinned section.
3. Check the native history sheet on device B matches
4. On device A: unpin the same chat → confirm device B shows it back in Recent.
5. **Offline test**: turn off network on device A, pin a chat → goes to the pending queue. Turn network back on → next sync cycle flushes the queue → device B picks it up.
6. **Delete-wins**: pin a chat, then delete it before the next sync cycle. Confirm the delete propagates and no patch is sent for the same id.

### Impact and Risks

**Low**, gated behind `supportsSyncChatsUpdate` remote subfeature. With the flag off behavior is identical to PR #5344 (pin works locally only).

#### What could go wrong?

- **Patch JSON shape mismatch** — server might reject if we drift from Android's contract. Mitigation: matched byte-for-byte against Android PR #8640.
- **Storage missing chat at flush time** — handled: such ids are dropped from the queue rather than retried forever.
- **Concurrent delete and update for same id** — handled: delete-wins arbitration skips the update.
- **Pin written but record-update Task never runs** (app killed mid-flight) — chat is locally pinned but the queue entry is absent. Mitigated by: next pin/unpin on any chat triggers a sync cycle which will read latest storage; OR a follow-up sync queue persistence beyond a single Task lifetime (out of scope here).

### Quality Considerations

- **Tests:** `AIChatSyncCleanerTests` gains update-flavoured tests (feature flag gating including the new `supportsSyncChatsUpdate` subfeature, queue add, eager trigger, patch flush, success/failure handling, delete-wins, storage-missing fallback). `ChatPinnerTests` gains 2 tests verifying the cleaner integration. `MockDDGSyncing` now exposes a stable `MockScheduling` with a call counter so the eager trigger is assertable.
- **Performance:** Patch flush runs on the existing `SyncCustomOperation` cadence. No new background work introduced.
- **Privacy:** Only `pinned` rides through this endpoint. Title/content stay client-side until JWE encryption lands (same constraint Android shipped with).
- **Feature flag:** New `supportsSyncChatsUpdate()` on `AIChatFeatureFlagProviding`, defaulting to `false` via protocol extension so existing conformances don't break. Both pin/unpin push (`supportsSyncChatsUpdate`) and delete (`supportsSyncChatsDeletion`) are now independently gated.

### Notes to Reviewer

- The PATCH endpoint and JSON-array body are reused from the existing delete path — only the per-entry shape differs.
- `MainCoordinator` and `Launching.swift` thread the storage handler and the cleaner; small reorder in `Launching.swift` to build the storage handler before `SyncService` so the cleaner has it at construction time.
- A successful history-sheet delete now calls `FireExecutor.scheduleSync()` (mirroring the chat-suggestions delete) so the deletion is flushed immediately instead of being re-pulled by the FE on the next sync cycle.
- I considered making `setPinned` async to await the cleaner record directly; the fire-and-forget Task keeps the protocol shape stable and the record is idempotent (the next sync cycle drains whatever made it into the queue).

---

> [!NOTE]
> **Medium Risk**
> Introduces a new authenticated sync PATCH contract and queue logic alongside existing delete sync; mitigated by remote kill switch, delete-wins arbitration, and broad test coverage.
> 
> **Overview**
> **Native chat-history pin/unpin now pushes to sync** when AI Chat sync and the new **`supportsSyncChatsUpdate`** remote subfeature are enabled.
> 
> After a successful local pin change, **`ChatPinner`** optionally notifies **`AIChatSyncCleaner`**, which persists chat ids in an update queue and triggers the sync scheduler. On each sync cycle, **`AIChatUpdateOperation`** drains the queue: it reads the current **`pinned`** value from native storage, applies **delete-wins** (skips ids also queued for deletion), drops ids that no longer exist locally, and calls **`patchAIChats`** through DDGSync. BrowserServicesKit adds **`AIChatUpdate`** and an authenticated **PATCH** on the AI chats endpoint (Android-aligned payload).
> 
> iOS wires the native storage handler into the cleaner at launch, passes the cleaner into **`ChatPinner`** from the history sheet, and **schedules sync after a successful history-sheet delete** so removals are flushed immediately. History delete now only schedules sync when **`burnChat`** succeeds.
> 
> Tests and mocks cover queue gating, patch flush, failure retention, delete-wins, and pinner→cleaner integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 044094311489199bae28213d314d21a3d9dc9855. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/Core/FeatureFlag.swift
```diff
@@ -757,6 +760,8 @@ extension FeatureFlag: FeatureFlagDescribing {
             Config(source: .remoteReleasable(AutofillSubfeature.onboardingDismissExperiment), cohortType: AutofillOnboardingDismissExperimentCohort.self)
         case .supportsSyncChatsDeletion:
             Config(source: .remoteReleasable(AIChatSubfeature.supportsSyncChatsDeletion))
+        case .supportsSyncChatsUpdate:
+            Config(defaultValue: .enabled, source: .remoteReleasable(AIChatSubfeature.supportsSyncChatsUpdate))
```
Bunn, line 764: Just making sure if this should be .enabled here.

Bunn (approved):

SharedPackages/AIChat/Sources/AIChat/Shared/Sync/AIChatSyncCleaner.swift
```diff
@@ -176,6 +189,52 @@ public final class AIChatSyncCleaner: AIChatSyncCleaning {
...
+        let candidates = Array(Set(pending).subtracting(pendingDeletes))
+        guard !candidates.isEmpty else { return }
+
+        let updates: [AIChatUpdate] = candidates.compactMap { chatId in
+            guard let record = try? storageHandler.getChat(chatId: chatId),
+                  let json = try? JSONSerialization.jsonObject(with: record.data) as? [String: Any] else {
+                return nil
+            }
+            let pinned = (json["pinned"] as? Bool) ?? false
+            return AIChatUpdate(chatId: chatId, pinned: pinned)
```
jleandroperez: Nit: Perhaps the json parsing could be encapsulated in a AIChatUpdate extension?

jleandroperez (approved):
✅ All test scenarios are green
✅ Code looks good

Added a minor nit, not a blocker.

:shipit:  !!

iOS/Core/FeatureFlag.swift
SabrinaTardio: I have it as enabled given its usage is still gated by the native chat history flag, but can always be turned off if necessary

PR #5350, opened 2026-06-10
Add PixelKit retry queue for failed pixel fires

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1213797000547599?focus=true
Tech Design URL:
CC:

### Description

Adds a retry queue to PixelKit: pixels that fail to fire are persisted and re-sent on app start and whenever a later pixel is sent successfully, with a 28-day expiry.

PixelKit creates and owns the queue internally, wrapping the fire-request closure it is already given and reusing the same defaults it is already given for throttling state. The public PixelKit API is therefore unchanged and every consumer gets the behaviour automatically, with no per-call-site wiring and the complexity hidden from callers.

To keep the change minimal and low-risk, it reuses the approach of the existing iOS PersistentPixel implementation (file-backed storage capped at 100 items, 28-day expiry, throttled draining) rather than introducing a new mechanism. The feature is isolated in its own internal type with a small storage abstraction.

Also fixes the DBP background agent, which previously always reported pixel sends as successful, so that failed sends are now detected and queued for retry like every other site.

### Testing Steps

1. Run the PixelKit unit tests in the BrowserServicesKit package and confirm they pass.
2. With networking unavailable, fire a pixel and confirm it is written to the retry-queue file; restore networking, fire another pixel (or relaunch), and confirm the queued pixel is sent and removed.

#### What could go wrong?

- PixelKit now routes every send through the retry queue. Mitigated by it being a thin pass-through that forwards the original completion unchanged and only adds persistence on failure and draining on success; dry-run mode skips the queue entirely.
- The DBP agent now reports send failures instead of always reporting success. Low risk and brings it in line with the other sites.
- Queue growth and processing frequency are bounded by the 100-item cap, the 28-day expiry, and throttled draining.

### Quality Considerations

Comprehensive unit tests were added for the record model, the file store (including the cap), and the queue (persist-on-failure, draining, retry marker, expiry, throttle, drain-on-success, failures received mid-drain, failed retries, and completion forwarding). The full PixelKit suite passes. App-target build verification is left to CI.

---

---

> [!NOTE]
> **Medium Risk**
> Every non–dry-run pixel path now goes through the retry decorator, which changes offline/telemetry behavior; impact is bounded by caps and tests, but duplicate or delayed pixels are possible if draining misbehaves.
> 
> **Overview**
> **PixelKit** now persists failed telemetry fires and replays them after later successes, instead of dropping them on network errors.
> 
> Non–dry-run sends go through an internal **`PixelRetryQueue`** that enqueues the full request on failure (with `originalPixelTimestamp`), drains on success (throttled, 28-day expiry, `retriedPixel=1` on replay), and uses a JSON file store capped at 100 items. **`PixelKit.setUp`** gains a required **`session`** string so browser, VPN tunnel, agents, etc. each get their own queue file and throttle keys when they share **`UserDefaults`**.
> 
> All app and test **`setUp`** call sites pass stable session ids (e.g. `ios-browser`, `macos-vpn-tunnel`). The macOS **DBP background agent** now reports pixel send success from the real fetch result so failures can be queued like other targets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed6fa970561a9a732ddb88712ee5a1a8a61d0ac6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,192 @@
...
+public final class PixelRetryQueue {
+
+    enum Constants {
+        static let lastProcessingDateKey = "com.duckduckgo.pixelkit.retry-queue.last-processing-timestamp"
+        static let expiryDays = 28
+#if DEBUG
+        static let minimumProcessingInterval: TimeInterval = 60          // 1 minute
+#else
+        static let minimumProcessingInterval: TimeInterval = 60 * 60     // 1 hour
+#endif
```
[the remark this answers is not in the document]
federicocappelli: Fixed in f2a5d5a: the drain-throttle key is now namespaced per `source` (matching the per-source queue file), so the browser, VPN agent and packet-tunnel processes sharing `UserDefaults.netP` no longer block each other's `sendQueuedPixels`. (The queue is now owned internally by PixelKit, reusing the `setUp` `defaults` and `fireRequest`.)

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,193 @@
...
+            do {
+                queuedItems = try self.store.storedItems()
+            } catch {
+                completion(.readError(error))
+                return
+            }
+
+            guard !queuedItems.isEmpty else {
+                completion(nil)
+                return
```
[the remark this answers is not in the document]
federicocappelli: Fixed in c097221: the throttle timestamp is now written only after confirming the queue is non-empty, so empty drains (the common case after a successful send) no longer advance the throttle and can't delay replay of a failure queued moments later. Added a regression test.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            self.fire(queuedItems: queuedItems) { idsToRemove in
+                do {
+                    try self.store.remove(itemsWithIDs: idsToRemove)
+                    completion(nil)
+                } catch {
+                    completion(.writeError(error))
+                }
```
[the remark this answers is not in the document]
federicocappelli: Intentional at-least-once delivery, matching the iOS PersistentPixel this is ported from. A duplicate only occurs if store.remove throws *after* a successful network send (a rare post-send disk-write failure). For anonymous measurement pixels a rare duplicate is preferable to the alternative (removing before confirming delivery, which would drop events on send failure), so leaving as-is.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,195 @@
...
+
+/// Retry queue for failed PixelKit fires.
+///
+/// `PixelRetryQueue` decorates the `PixelKit.FireRequest` network closure. When a fire fails it persists
+/// the fully-resolved request; when a fire succeeds it triggers a throttled drain that replays queued
+/// items through the *underlying* closure (so replays never re-enter this decorator). Items older than
+/// 28 days are dropped without sending.
+///
+/// `PixelKit` creates and owns this internally: it wraps the `FireRequest` passed to `PixelKit.setUp`,
+/// reuses the same `defaults` for throttling state, and drains the queue at launch and after each
```
diegoreymendez: I'd drop draining at launch from this.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,195 @@
...
+        } catch {
+            logger.error("Failed to persist pixel \(pixelName, privacy: .public) for retry: \(error.localizedDescription, privacy: .public)")
+        }
+    }
+
+    /// Drains the queue: replays each item not older than 28 days through the underlying closure, drops
+    /// expired items, and removes anything successfully sent. Throttled to `minimumProcessingInterval`.
+    func sendQueuedPixels(completion: @escaping (PixelRetryQueueStorageError?) -> Void = { _ in }) {
+        workQueue.async {
+            let storedProcessingDate = (try? self.lastProcessingDateStorage.object(forKey: self.lastProcessingDateKey)) ?? nil
```
diegoreymendez: Resiliency suggestion here.

Would you consider reading the last-run date once and keep it in memory instead of re-reading from disk on every drain?  When updating the in-memory value we'd try to write to disk.

The reason: If a read consistently fails it comes back `nil`, the throttle reads that as "never drained," and then every successful fire drains the whole queue.

Reading it once at startup would avoid that as we'd do the `nil` > fire now deduction only once, and subsequent runs would use the in-memory date.  It's just a bit more resilient to read errors.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,195 @@
...
+    var fireRequest: PixelKit.FireRequest {
+        { [weak self] pixelName, headers, parameters, allowedChars, callBackOnMainThread, onComplete in
+            guard let self else {
+                onComplete(false, nil)
+                return
+            }
+            self.underlyingFireRequest(pixelName, headers, parameters, allowedChars, callBackOnMainThread) { success, error in
+                onComplete(success, error)
+                if success {
+                    self.sendQueuedPixels()
```
diegoreymendez, line 100: Draining here is fine.  The logic looks solid with a 1-hour delay between drain attempts.  Nice!

I would consider checking for expired pixels in the failure path though, so we don't ever get into a situation where pixels are piling up forever (like if there's no connection for 3 months or whatever... unlikely, but prevention should be fairly easy).

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -222,6 +229,25 @@ public final class PixelKit {
...
         self.fireRequest = fireRequest
+
+        if dryRun {
+            self.retryQueue = nil
+        } else {
+            // Wrap the network fire-request with a retry queue so failed pixels are persisted and re-sent
+            // after the next successful fire (28-day expiry) — which, for a launching app, happens as soon
+            // as it fires its first pixel. Reuses the same `defaults` for throttling state. This is internal
+            // to PixelKit and hidden from its consumers; creating the queue performs no I/O.
+            let sourceSuffix = Self.retryQueueSourceSuffix(forSource: source)
```
diegoreymendez: Would you be open to using a dedicated identity for the retry queue instead of source?

I've two concerns with keying the identity on on source:
- `source` is a telemetry value. PixelKit is designed to allow multiple instances, but two instances in the same process with  the same `source` would silently share one store and throttle key — and overwrite each other.
- `source` is optional. A `nil` source falls back to "default", which both reads oddly in the file/key names and lumps every  nil-source instance into one shared queue.

My suggestion is: add a non-optional identifier parameter to PixelKit's init and use that for identity.  Each instance then gets its own stable / persistent identity, fully decoupled from source, and we can still run more than one instance per process safely.

diegoreymendez (commented):
This is looking great.  I haven't tested yet but completed a first round of code-only review.  I don't expect much more from the code, so once you reply I can run tests.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
diegoreymendez: Btw, instead of "identity" the best term for this could be "session", so the reader can tell it's meant to be stable for each instance and not a `UUID()` type of thing.  Don't feel strongly about the naming though, I guess there might be a case for either.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
federicocappelli: Done, the last-drain date is now seeded once from storage at init and held in memory (write-through on update), so a persistent read failure can't be misread as "never drained" and trigger a drain on every fire.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
federicocappelli: Done, now also prunes expired items on the enqueue/failure path, so a long offline stretch (only failures, no drains) can't keep dead pixels queued until the next successful send. Thanks!

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
federicocappelli: Done in 9e274fc — added a non-optional `session` to PixelKit.setUp, used for the retry-queue store filename + throttle key and fully decoupled from the telemetry `source`. init keeps an optional `session` that falls back to source only for direct (test) construction, to avoid churning ~35 test call sites — happy to make init non-optional too if you'd prefer the extra strictness.

federicocappelli: Went with `session`, as suggested.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
```diff
@@ -0,0 +1,217 @@
...
+                                       timestamp: now)
+        do {
+            // Drop already-expired items here too, so a long offline stretch (only failures, no drains)
+            // can't keep known-dead pixels queued until the next successful send.
+            if let cutoff = expiryCutoff(from: now) {
+                let expiredIDs = Set(try store.storedItems().filter { $0.timestamp < cutoff }.map(\.id))
+                if !expiredIDs.isEmpty {
+                    try store.remove(itemsWithIDs: expiredIDs)
+                }
+            }
```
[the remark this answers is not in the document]
federicocappelli: Good catch — fixed in ed6fa97: the best-effort prune now has its own do/catch, separate from the append, so a prune failure can't stop the new failed pixel from being queued. Added a regression test.

SharedPackages/BrowserServicesKit/Sources/PixelKit/RetryQueue/PixelRetryQueue.swift
federicocappelli: I agree, we send pixels at app startup anyway, a successful send should drain the queue anyway.

diegoreymendez (approved):
Works well for me.

PR #5360, opened 2026-06-11
Fix crash on provisional navigation failure before start

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201899738287924/task/1215620561014173?focus=true

### Description

Fixes the `EXC_BREAKPOINT` crash **APPLE-MACOS-BDE** (169 events / 85 users on 1.193.0).

`NavigationPixelNavigationResponder.navigation(_:didFailWith:)` evaluated `navigation.navigationAction` first in its guard. `Navigation.navigationAction` returns `navigationActions.last!` (a force-unwrap). When a provisional navigation fails *before it ever starts*, `navigationActions` is empty, so the force-unwrap traps at `Navigation.swift:44`.

The fix reorders the guard so the `siteLoadingStartTime` nil-check short-circuits first. `siteLoadingStartTime` is set only in `didStart` (after a navigation action has been recorded) and `navigationActions` is append-only, so `siteLoadingStartTime != nil` implies `navigationActions` is non-empty — the responder now bails out on never-started navigations without ever touching `navigationAction`. Behaviour is otherwise unchanged.

### Testing Steps
1. Run the added unit test `NavigationPixelNavigationResponderTests.testWhenNavigationFailsWithNoNavigationActions_thenItDoesNotCrashAndNoPixelIsFired`. It passes with the fix; reverting the guard reorder makes it crash with `Navigation/Navigation.swift:44: Fatal error: Unexpectedly found nil while unwrapping an Optional value` (the exact APPLE-MACOS-BDE signature).
2. Smoke check: browse normally and trigger failed loads (e.g. navigate to a nonexistent domain, or drop the network mid-load); the app should not crash and normal navigation pixels are unaffected.

### Impact and Risks
Low. The change is a guard-clause reorder in pixel-firing code with no behavioural change for navigations that actually started. It eliminates a crash affecting ~85 users.

#### What could go wrong?
The reorder relies on `siteLoadingStartTime` being nil exactly for navigations with an empty `navigationActions`. Verified against the code: `siteLoadingStartTime` is assigned in a single place (`didStart`, which only runs after a navigation action exists) and `navigationActions` is only ever appended to, never emptied after the first action. Worst case is that a `siteLoadingFailure` pixel is not fired for a navigation that never began loading — which is the correct outcome.

### Quality Considerations
- The empty-`navigationActions` edge case is now covered by a regression test wired into both the `Unit Tests` and `Unit Tests App Store` targets.
- No performance impact (pure read reorder); no privacy/security impact (pixel content unchanged).
- The sibling `navigationDidFinish` uses the same guard pattern but cannot reach the empty state (a finished navigation always has a recorded action), so it is intentionally left unchanged.

### Notes to Reviewer
- Targeted directly at `release/macos/1.194.0` as a crash fix.
- The underlying force-unwrap in the `Navigation` package (`Navigation.swift:44`) is a long-standing latent invariant. This PR fixes the only reachable trigger (the responder) rather than changing the shared public API on a release branch.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Guard reorder only in analytics/pixel code; behavior is unchanged for navigations that actually started loading.
> 
> **Overview**
> Fixes **APPLE-MACOS-BDE** by changing how `NavigationPixelNavigationResponder` handles `navigation(_:didFailWith:)`. The guard now checks **`siteLoadingStartTime` before `navigation.navigationAction`**, so provisional failures with no recorded navigation actions return early instead of force-unwrapping an empty `navigationActions` list.
> 
> Adds **`NavigationPixelNavigationResponderTests`** with a regression test that a never-started navigation does not crash and does not fire a site-loading failure pixel, and wires that file into both unit-test targets in the Xcode project.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7076066f9955e4ceffc4b2e2491669a0b1fd569e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo-macOS.xcodeproj/project.pbxproj
+++ b/macOS/DuckDuckGo-macOS.xcodeproj/project.pbxproj
@@ -1160,6 +1160,8 @@
 		376325C22F86964C00B6B0DB /* TabSuspensionDebugMenu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 376325C02F86964C00B6B0DB /* TabSuspensionDebugMenu.swift */; };
 		37657FC42DB8D564003D749E /* TabCrashRecoveryExtensionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37657FC32DB8D55F003D749E /* TabCrashRecoveryExtensionTests.swift */; };
 		37657FC52DB8D564003D749E /* TabCrashRecoveryExtensionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37657FC32DB8D55F003D749E /* TabCrashRecoveryExtensionTests.swift */; };
+		DBDE0002000000000000BDE0 /* NavigationPixelNavigationResponderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DBDE0001000000000000BDE0 /* NavigationPixelNavigationResponderTests.swift */; };
+		DBDE0003000000000000BDE0 /* NavigationPixelNavigationResponderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DBDE0001000000000000BDE0 /* NavigationPixelNavigationResponderTests.swift */; };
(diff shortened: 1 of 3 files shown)


diegoreymendez (approved):
Nice catch.

PR #5361, opened 2026-06-11
Add debounce frequency to PixelKit

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1213797000731646?focus=true
Tech Design URL:
CC:

### Description

Adds debounce support to PixelKit, the equivalent of the legacy Pixel.fire(debounce:) parameter, so callers migrating off the legacy API keep the same rapid-repeat de-duplication.

Debounce is expressed as a new Frequency case (.debounce(seconds:)) rather than a separate parameter, reusing PixelKit's existing per-pixel last-fire-date storage (no new storage and no change to the fire API). A pixel fired with this frequency is suppressed if it already fired less than that many seconds ago, with the window anchored to the last actual fire (a suppressed call does not extend it). A window of 0 seconds never suppresses.

This is the capability only; migrating the legacy call sites is follow-up work.

### Testing Steps

1. Run the PixelKit unit tests in the BrowserServicesKit package and confirm they pass.

#### What could go wrong?

- Additive change: a new Frequency case plus its handler. Existing frequencies, call sites, and storage are unchanged; the only non-test edit outside PixelKit is one exhaustive-switch update in the test utilities.
- Debounce reuses the existing per-pixel storage under its own independent key, so it cannot interfere with daily/monthly/unique tracking on the same pixel.

### Quality Considerations

Unit tests added for in-window suppression (anchored to the last actual fire and firing again once the window elapses) and the zero-second no-op case. The full PixelKit suite passes.

---

---

> [!NOTE]
> **Low Risk**
> Additive analytics API with isolated storage key; daily logic is a small refactor to a shared helper with the same fail-closed semantics.
> 
> **Overview**
> Adds **`Frequency.debounce(seconds:)`** to PixelKit so callers can throttle repeat fires the same way as legacy `Pixel.fire(debounce:)`, without a separate API or new storage keys beyond the existing per-pixel last-fire map (`mapKey` `"debounce"`).
> 
> A fire is **skipped** if the same pixel name fired less than `seconds` ago; the window is tied to the **last successful fire** (suppressed calls do not reset the timer). **`seconds: 0`** never suppresses. Storage read failures **fail closed** (suppress). Daily “fired today” checks are refactored to share a new **`pixelHasBeenFired`** helper with the debounce path.
> 
> Test utilities gain an exhaustive-switch case for debounce; unit tests cover in-window suppression, post-window refire, and the zero-second case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 313f97c76a0f6ddecd99a1b4a351b26550d251cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
@@ -861,18 +861,25 @@ public final class PixelKit {
         )
     }
 
+    /// Evaluates `within` against the stored last-fire date for `frequency`'s `mapKey`.
+    /// Returns `false` when there's no stored date for that frequency, and fails closed
+    /// (returns `true`, suppressing the pixel) on a storage read error.
+    private func pixelHasBeenFired(_ name: String, frequency: Frequency, within: (Date) -> Bool) -> Bool {
+        do {
+            let map = try migratedLastFireDateMap(forKey: userDefaultsKeyName(forPixelName: name))
+            guard let lastFireDate = map[frequency.mapKey] else { return false }
+            return within(lastFireDate)
+        } catch {
+            return true
+        }
+    }
+
     /// Returns `true` if the pixel was last fired less than `seconds` ago (so it should be suppressed).
     /// The stored fire date is shared across debounce intervals (mapKey `"debounce"`), and the window is
     /// evaluated against the passed `seconds`. Fails closed (suppresses) on a storage read error.
     private func pixelHasBeenFiredWithinDebounceInterval(_ name: String, seconds: TimeInterval) -> Bool {
-        do {
-            let map = try migratedLastFireDateMap(forKey: userDefaultsKeyName(forPixelName: name))
-            if let lastFireDate = map[Frequency.debounce(seconds: seconds).mapKey] {
-                return lastFireDate > dateGenerator().addingTimeInterval(-seconds)
-            }
-            return false
-        } catch {
-            return true
+        pixelHasBeenFired(name, frequency: .debounce(seconds: seconds)) { lastFireDate in
+            lastFireDate > dateGenerator().addingTimeInterval(-seconds)
         }
     }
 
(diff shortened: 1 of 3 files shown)


SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
```diff
@@ -828,6 +861,21 @@ public final class PixelKit {
...
+    /// Returns `true` if the pixel was last fired less than `seconds` ago (so it should be suppressed).
+    /// The stored fire date is shared across debounce intervals (mapKey `"debounce"`), and the window is
+    /// evaluated against the passed `seconds`. Fails closed (suppresses) on a storage read error.
+    private func pixelHasBeenFiredWithinDebounceInterval(_ name: String, seconds: TimeInterval) -> Bool {
+        do {
+            let map = try migratedLastFireDateMap(forKey: userDefaultsKeyName(forPixelName: name))
+            if let lastFireDate = map[Frequency.debounce(seconds: seconds).mapKey] {
+                return lastFireDate > dateGenerator().addingTimeInterval(-seconds)
+            }
+            return false
```
diegoreymendez: The code is fine as-is, but if you wanted to consider a small improvement, it might be possible to share some of the logic we're using here with `pixelHasBeenFiredDailyToday` through generalization.

Very quick draft of the logic that I think is shared:

```swift
  private func pixelHasBeenFired(_ name: String, frequency: Frequency, within: (Date) -> Bool) -> Bool {
      do {
          let map = try migratedLastFireDateMap(forKey: userDefaultsKeyName(forPixelName: name))
          guard let lastFireDate = map[frequency.mapKey] else { return false }
          return within(lastFireDate)
      } catch {
          return true   // fail closed
      }
  }
```

This is not a blocker to merge.

diegoreymendez (approved):
Approved, just made one suggestion but it's optional and can be ignored if you prefer.

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit.swift
federicocappelli: Implemented 👍

PR #5376, opened 2026-06-12
iOS: Fix navigation bar stuck hidden on error page

dus7:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1211689879271276?focus=true
Tech Design URL:
CC:

### Description
If a page load fails (e.g. no connectivity) while the chrome (address bar plus toolbar buttons) is hiding on scroll, it can get stuck hidden, recoverable only by force quitting or backgrounding the app.

The error path forced the chrome to show via `setBarsHidden(false)`, a reveal that bypasses `BarsAnimator` and leaves its `barsState` stale. A later scroll hid the bars again, and because the error page hides the webView there was no scrollable content left to reveal them.

Fix: route the reveal on the error page through `chromeManager.reset` so `BarsAnimator` state is synced to `.revealed`, and prevent the chrome from hiding on scroll while the error page is shown.

### Testing Steps
The underlying issue is a timing race, so the default repro takes a few tries. To make it reliable, temporarily set `ChromeAnimationConstants.duration` (MainViewController) to `1.5`.

1. Run a search so a scrollable SERP is loaded.
2. Cut the network.
3. Tap refresh, then immediately scroll to hide the chrome and keep scrolling for about 1.5 seconds.
4. Verify the error page appears with the address bar and buttons visible, and that they stay visible.
5. Regression: hiding on scroll then revealing by scrolling up still works, tapping the bottom of the screen reveals the chrome, and the new tab page and Duck Player chrome behaviour are unchanged.

### Impact and Risks
Low. Scoped to revealing the chrome on the error page. The reveal now also syncs `BarsAnimator` state, and every reveal stays routed through the current tab's `chromeDelegate`.

#### What could go wrong?
- The reveal now goes through `chromeManager.reset` (the animator reveal) rather than `setBarsHidden`. Both end at the same `setBarsVisibility(1)`, so the visual result is identical and only the animator state is additionally synced.
- `canHideBars` returns false while an error page is shown. `isError` is read live and clears on the next navigation, so no state keeps the bars permanently visible.
- Background tabs: the reveal routes through `chromeDelegate`, which is nil for tabs that are not current, so a background tab whose load fails does not move the visible chrome.

### Quality Considerations
Other reveals that bypass the animator (new tab page, Duck.ai, Duck Player) still do not sync `BarsAnimator` state, but none pair with the error page condition (a hidden webView), so none produce a user visible bug.

### Notes to Reviewer
`resetBars` is a thin addition to `BrowserChromeDelegate` so `TabViewController` (which only holds `chromeDelegate`) can trigger the reveal routed through the animator, keeping reveal symmetric with hide and scoped to the current tab.

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/BrowserChromeManager.swift
+++ b/iOS/DuckDuckGo/BrowserChromeManager.swift
@@ -22,6 +22,7 @@ import UIKit
 protocol BrowserChromeDelegate: AnyObject {
     
     func setBarsHidden(_ hidden: Bool, animated: Bool, customAnimationDuration: CGFloat?)
+    func resetBars(animated: Bool)
     func setNavigationBarHidden(_ hidden: Bool)
     func setRefreshControlEnabled(_ isEnabled: Bool)
     func setUnifiedInputContentOverlaySuppressed(_ suppressed: Bool)
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -3651,6 +3651,10 @@ extension MainViewController: BrowserChromeDelegate {
 
         setBarsVisibility(hidden ? 0 : 1.0, animated: animated, animationDuration: customAnimationDuration)
     }
+
+    func resetBars(animated: Bool) {
+        chromeManager.reset(animated: animated)
+    }
     
     func setBarsVisibility(_ percent: CGFloat, animated: Bool, animationDuration: CGFloat?) {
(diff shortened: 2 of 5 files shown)


jleandroperez (approved):
✅ Verified when the Error appears, the Address Bar is visible
✅ Verified that on scroll we still hide the Address Bar, and reveal back when scrolling upwards

:shipit:

PR #5378, opened 2026-06-12
[macOS] Update pixels to validate unclean exit detection volume

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1215586547447007?focus=true
Tech Design URL:
CC: @ayoy 

### Description

Updates our unclean exit detection with three pixel changes:

* Adds a parameter to the unclean exit detection pixel to attribute the restart to app updates when the app version has changed. If the app is killed during an app update, we can attribute the unclean exit to that update.
* Adds debug pixels if we fail to read or write the unclean exit flag (which can cause false positives or negatives in the detection), to determine if these are a significant source of detection errors.

Notes: 
* This change attributes an app restart after an unclean exit to an app update, as long as the update coincides with an unclean exit. It's possible that the two actions are separate — app crashes without being relaunched, then app is updated while app is closed, and then app is started — but this attribution should be acceptable for our purposes.
* We suppress the session restore dialog in normal runs of the debug build, to avoid interrupting development. This PR includes moving that suppression to after the detection pixel is fired, so we can debug the detection while still suppressing the prompt itself.

### Testing Steps

App Store build:

1. Build and run the app.
2. Kill the app (e.g. use the stop button in Xcode).
3. Run the app again and confirm `m_mac_unclean-exit_detected` is fired with the parameter `restart-source-attribution` with value `unknown`.
4. Kill the app.
5. In `Version.xcconfig` change the version number.
6. Build and run the app and confirm `m_mac_unclean-exit_detected` is fired with the parameter `restart-source-attribution` with value `app_update`.

Testing on Sparkle builds:

1. Build and run the app.
2. Kill the app (e.g. use the stop button in Xcode).
3. Run the app again and confirm `m_mac_unclean-exit_detected` is fired with the parameter `restart-source-attribution` with value `unknown`.
4. Go to Debug -> Simulate crash -> NSException.
5. Run the app again and confirm `m_mac_unclean-exit_detected` is fired with the parameter `restart-source-attribution` with value `crash`.
6. Kill the app.
7. In `Version.xcconfig` change the version number.
8. Build and run the app and confirm `m_mac_unclean-exit_detected` is fired with the parameter `restart-source-attribution` with value `app_update`.

### Impact and Risks

### Quality Considerations

* Privacy review: Needs privacy triage

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Changes are limited to launch-time telemetry, crash-report reads, and when the restore prompt appears in debug/test builds; core session restore and update install paths are largely unchanged.
> 
> **Overview**
> **Unclean exit analytics** now classifies why the browser restarted after an unexpected quit. On launch, `AppStateRestorationManager` asks a new `UncleanExitRestartSourceResolver` (fed by `ApplicationUpdateDetector` and, on Sparkle builds, pending-update settings plus diagnostic crash reports) and sends `restart-source-attribution` on `m_mac_unclean-exit_detected` (`crash`, `app_update`, `unknown_with_app_update`, or `unknown`).
> 
> **Operational tweaks:** a single shared `ApplicationUpdateDetector` is wired from `AppDelegate` into Sparkle update startup and state restoration; Sparkle captures a pending-update snapshot before the update controller runs. **Debug pixels** fire when the clean-termination KeyValueStore flag fails to read or write. The session-restore popover is still suppressed in normal debug/UI-test runs, but only **after** the detection pixel fires so telemetry stays visible during development.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13ab4393acafe96cf07e818caf9e395fdc98aa56. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/StateRestoration/AppStateRestorationManager.swift
```diff
@@ -231,18 +240,15 @@ final class AppStateRestorationManager: NSObject, AppStateRestorationManaging {
...
+    private func detectUnexpectedAppTermination(didRestoreRegularTabs: Bool, updateStatus: AppUpdateStatus) {
         let didCloseUnexpectedly = !appDidTerminateAsExpected
         appDidTerminateAsExpected = false // Set to false so it will be false if the app closes without terminating properly
 
         guard didCloseUnexpectedly else { return }
-        pixelFiring?.fire(SessionRestorePromptPixel.unexpectedAppTerminationDetected)
+
+        pixelFiring?.fire(SessionRestorePromptPixel.unexpectedAppTerminationDetected(reason: UncleanExitRestartReason(updateStatus)))
+
+        guard !shouldSuppressUncleanExitRestorePrompt else { return }
```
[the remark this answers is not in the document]
rachelmcr: This suppression was intentionally moved to allow the pixel to be fired for debugging. Since we do a dry run for pixels during UI tests, and this change has no behavioral changes, it should not disrupt anything in our UI test runs.

diegoreymendez (commented):
I know I suggested this, but I think we need to iterate a bit to avoid over-attribution of failures on updates.

What do you think about this approach, to seek more fidelity?

- Crash — previous session left a crash report → `crash`.
- Update:
     - Sparkle: check `pending.update` keys are present on launch → `appUpdate`
     - App Store: check version changed as our best proxy → `appUpdate`.
- Else — unclean, no crash, no update → `unknown`.

rachelmcr:
Thanks for the suggestion, Diego. I implemented it with the caveat that the `crash` source is Sparkle-only. For App Store builds, we can't access crash reports on-disk directly and the App Store crash collection is async so we may not have the crash report in time to detect it for this.

I know that still carries the risk of over-reporting on `appUpdate` for App Store builds, but the only other option I could think of would be to e.g. subscribe to the App Store crash collection processing and only fire the unclean exit pixel after we're sure crashes have been processed. But I don't see any signal for _no_ crash reports found, so I'm not sure that's feasible.

Let me know if you have ideas about how to reliably identify that; otherwise, App Store builds can report either `appUpdate` or `unknown`, and we can use our existing crash pixels to estimate actual crash volume among the unclean exits detected.

macOS/DuckDuckGo/StateRestoration/UncleanExitRestartSourceResolver.swift
```diff
@@ -0,0 +1,103 @@
...
+        if crashReportDetecting.hasNewMainBrowserCrashReport() {
+            return .crash
+        }
+
+        if buildType.isSparkleBuild, sparklePendingUpdateSnapshot {
+            return .appUpdate
+        }
+
+        if buildType.isAppStoreBuild, updateStatus == .updated || updateStatus == .downgraded {
+            return .appUpdate
```
diegoreymendez: Following up on the possibility of misattribution, this case isn't certain and I'd really explore if we can convey the exact scenario through semantics.  On Sparkle we know the source was an app update, but here it can really be a crash, a force quit, a manual quit, or the app updating — and the App was coincidentially updated.

What about having a case like `unknownButAppUpdated` or `uncertainButAppWasUpdated`?  (haven't found any name I love, but the idea is to make it semantically faithful.

Even `uncertainButLikelyAnUpdate` would work for me.  All names I try sound awful though :D

diegoreymendez (approved):
Hey @rachelmcr, I've re-tested this and it works well.  Thanks for taking the time to address my concerns.  I've one final ask that I assume we'll agree with (with some naming adjustment that I don't mind accepting).

Let me know if there's any disagreement and we can chat via DM to unblock.

Great work!

macOS/DuckDuckGo/StateRestoration/UncleanExitRestartSourceResolver.swift
rachelmcr: Yeah, I see what you mean. I think I'll go with `unknownWithAppUpdate` to convey that the restart source is unknown but occurred alongside an app update. I'll include more details in the documentation.

PR #5389, opened 2026-06-14
[macOS] Fix Next Step card UI tests

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1215708557458161?focus=true
Tech Design URL:
CC:

### Description

This PR fixes failures in the Next Steps UI test suite.

### Testing Steps

1. Check that UI tests passed

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Changes are limited to debug menu notification posting and UI test assertions; no production user-facing behavior is altered beyond debug tooling triggering the correct refresh signal.
> 
> **Overview**
> Fixes flaky **Next Steps** UI tests and aligns the debug “shift max days” menu action with how the HTML New Tab Page actually refreshes.
> 
> The debug helper `debugShiftNewTabOpeningDate` now posts **`.newTabPageWebViewDidAppear`** instead of **`NSWindow.didBecomeKeyNotification`**, matching what `NewTabPageNextStepsSingleCardProvider` and related NTP code listen for so demonstration-day state updates show up in the webview during UI tests.
> 
> **`testNextStepsCardDisappearsAfterDismissal`** no longer depends on a fixed CTA (“Get Email Protection”). It snapshots visible webview button titles before dismiss, taps **No Thanks**, then waits until a button from the pre-dismiss set disappears—handling card rotation when the list widget cycles cards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3bcf12c3204c61d125f68ee6aff0fe1ba99336e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/UITests/NextStepsListUITests.swift
```diff
@@ -52,18 +52,23 @@ final class NextStepsListUITests: UITestCase {
...
+        let dismissedCardCycledAway = webView.wait(for: NSPredicate { _, _ in
+            !buttonsBeforeDismissal.subtracting(self.visibleWebViewButtonTitles()).isEmpty
+        }, timeout: UITests.Timeouts.elementExistence)
+
+        XCTAssertTrue(dismissedCardCycledAway,
+                      "The dismissed card should no longer be visible after tapping dismiss")
     }
 
     func testNextStepsListWidgetDisappearsAfterMaxDemonstrationDays() throws {
+        throw XCTSkip("Failing test, skipping for now until we can resolve the issue")
```
samsymons: I'm not sure of the correct fix for this, or whether the test is still accurate given recent changes - open to suggestions here. I aimed Claude at the problem who went around in circles for half an hour, so I've decided to skip it for now.

macOS/UITests/NextStepsListUITests.swift
```diff
@@ -93,14 +98,17 @@ private extension NextStepsListUITests {
         webView.staticTexts["Next Steps"]
     }
 
-    var nextStepsFirstCardCTAButton: XCUIElement {
-        webView.buttons["Get Email Protection"]
```
samsymons, line 97: This was the root cause of one of these breaking - now the test will dynamically look up whatever the first card is, and assert that clicking the dismiss button causes the next card to appear.

macOS/UITests/NextStepsListUITests.swift
rachelmcr: Ah, I made a change to the debug action this test relies on, and that broke it. The previous debug action behavior was correct (it matched the logic being tested here and shouldn't have been changed) so I'll push a fix to revert back to that.

rachelmcr (commented):
Thanks, @samsymons! These failures were a side effect of the changes I made to the Next Steps debug actions at the end of the week. The dynamic lookup makes sense for robustness, and I fixed the issue with the debug action causing `testNextStepsListWidgetDisappearsAfterMaxDemonstrationDays()` to fail.

I did a [test run on CI](https://github.com/duckduckgo/apple-browsers/actions/runs/27534765344) and it looks like the dynamic lookup fails on macOS 14 so I'm looking into that.

rachelmcr:
I don't know if there's a great solution here; based on the logs, it looks like the macOS 14 runner takes long enough to do the element lookup (for visible button titles) that it times out and the assertion fails. I'm trying out a longer timeout but that has potential for flakiness: https://github.com/duckduckgo/apple-browsers/actions/runs/27547057704

Without more stable IDs for these cards/buttons (a challenge with the C-S-S integration) a dynamic lookup may not be reliable on all test runners. An alternative that still supports dynamic card order would be to hard-code a list of possible button titles, get the button title on the first card, and then assert that the title is no longer there after the card is dismissed. That is still fragile (any changes to button titles or cards could break it) but might be an option if the longer timeout doesn't work for us.

rachelmcr (approved):
The longer timeout worked for CI test runs on macOS 14, so I'll approve this with those tweaks (to the debug action and longer assertion timeout).

PR #5392, opened 2026-06-15
Add "Delete All" to Duck.ai chat history sheet

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215103550228655?focus=true

### Description
Adds a **"Delete All"** action to the Duck.ai chat-history sheet. The Fire toolbar button now presents the existing fire-confirmation modal showing the chat count ("Delete N chats?"). Confirming burns all persistent chats, flushes the deletion to sync, plays the fire animation, and keeps the sheet open so it reactively falls through to its empty state.

Implementation notes:
- The burn goes through `AIChatDeleter.deleteAllChats` (a peer to the single-chat `deleteChat` path), so it stays **off** the Fire button's `burn(request:)` orchestration — no tab/data clearing, no `FireExecutorDelegate` callbacks, and no AI-chat session teardown that would dismiss the sheet.
- Sync parity with single-delete: on success it records the clear via `recordLocalClear(date:)` and flushes immediately with `scheduleSync()`.
- Scope is persistent (normal-mode) chats only — fire-mode chats are untouched.

### Testing Steps
1. Enable Duck.ai chat history and create several chats.
2. Open the chat-history sheet and tap the Fire toolbar button.
3. Confirm the modal reads "Delete N chats?" with **Delete All** / **Cancel**.
4. Tap **Delete All** — the fire animation plays, the sheet stays open and shows the empty state, and the chats are gone.
5. With Sync enabled on a second device, confirm the chats are removed there too.
6. Tap the Fire button with **Cancel** / the X — nothing is deleted and the sheet stays.

### Impact and Risks

Medium — deletes user chat data.

#### What could go wrong?
- **Deleting more than intended.** Mitigated by an explicit confirmation modal and limiting scope to normal-mode persistent chats (fire-mode chats are not touched).
- **Deletion re-pulled by sync.** Mitigated by flushing the clear immediately via `scheduleSync()` after a successful burn, matching the single-chat delete behavior.

### Notes to Reviewer
Stacked on `sabrina/ios-ai-chat-history-sync` — please set the base accordingly.

---

> [!NOTE]
> **Medium Risk**
> Bulk deletion of persistent Duck.ai chat data; mitigated by explicit confirmation, search/edit gating, and immediate sync flush after success.
> 
> **Overview**
> Adds **Delete All** from the Duck.ai chat-history sheet: the toolbar Fire button shows a count-based confirmation (`Delete N chats?`), then clears **all persistent** chats via a dedicated `burnAllChats` / `deleteAllChats` path that avoids full `burn(request:)` tab orchestration.
> 
> The sheet wires **edit mode** (toolbar Done/Edit, compose/fire gating), disables Fire while searching or editing, and tracks `totalChatCount` so the confirmation reflects every chat—not only search matches. On success it runs the fire animation, keeps the sheet open for an empty list, and calls `recordLocalClear` plus `scheduleSync()` like single-chat delete.
> 
> Fire confirmation gains a `deleteAllChats` context with **Delete All** / neutral **Cancel**, and `AIChatSyncCleaner.updateIfNeeded()` is refactored into small helpers without changing sync behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8515e357b65b8c566aa459d0c981d5d10ab3dd0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Bunn (approved):

iOS/DuckDuckGo/Fire/AIChatDeleter.swift
```diff
@@ -76,6 +80,37 @@ struct AIChatDeleter: AIChatDeleting {
...
+    func deleteAllChats(isFireMode: Bool) async -> Result<Void, Error> {
+        let dataStore: WKWebsiteDataStore?
+        if isFireMode {
+            guard #available(iOS 17, *) else {
+                return .success(())
+            }
+            dataStore = WKWebsiteDataStore(forIdentifier: idManager.currentFireModeID)
+        } else {
+            dataStore = nil
+        }
```
jleandroperez, line 81: Perhaps we could extract this snippet (L86-L94) into a function? (As it's used in L55-L63 as well)

iOS/DuckDuckGo/Fire/ScopedFireConfirmationViewModel.swift
```diff
@@ -46,6 +46,9 @@ final class ScopedFireConfirmationViewModel: ObservableObject {
         case singleTab
         /// Search Suggestions allow deleting History Entries
         case custom(title: String, subtitle: String, action: String)
+        /// Duck.ai chat-history sheet "Delete All": title shows the count; `onDelete` is the
+        /// caller-supplied action (dismiss + fire animation + burn), off the `burn(request:)` path.
+        case deleteAllChats(count: Int, onDelete: () -> Void)
```
jleandroperez, line 51: FYI I've added `.custom` (a line above) in the hopes to making it easier to reuse.

No need to change anything though!

jleandroperez: **IF** you were to switch, you could get rid of the `onDelete` closure, as the confirmation would just run the `onConfirm` standard parameter

SharedPackages/AIChat/Sources/AIChat/Shared/Sync/AIChatSyncCleaner.swift
```diff
@@ -210,8 +209,12 @@ public final class AIChatSyncCleaner: AIChatSyncCleaning {
...
         let updates: [AIChatUpdate] = candidates.compactMap { chatId in
-            guard let record = try? storageHandler.getChat(chatId: chatId) else { return nil }
-            return AIChatUpdate(record: record)
+            guard let record = try? storageHandler.getChat(chatId: chatId),
+                  let json = try? JSONSerialization.jsonObject(with: record.data) as? [String: Any] else {
+                return nil
+            }
+            let pinned = (json["pinned"] as? Bool) ?? false
+            return AIChatUpdate(chatId: chatId, pinned: pinned)
         }
```
jleandroperez, line 232: (Noted elsewhere I believe!) It'd be great if we could extract this into a smaller method

jleandroperez (approved):
Just a few nit notes, no blockers!

:shipit:

iOS/DuckDuckGo/Fire/ScopedFireConfirmationViewModel.swift
SabrinaTardio: For now I left the current form.
We can't use the standard onConfirm → burn(request:) path because that runs the full fire orchestration (tabs/data, delegate callbacks, session kill, burning overlay). The history sheet needs a chats-only, always-on delete that keeps the sheet open, same reason single swipe-delete uses burnChat instead of burn(request:).

Happy to review later.

PR #5403, opened 2026-06-15
[macOS] Reduce favicon startup memory

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215705778162650?focus=true
Tech Design: https://app.asana.com/1/137249556945/project/1201899738287924/task/1214612370091417?focus=true

### Description

Reduces the favicon subsystem's startup memory footprint and guards the whole change behind two runtime kill switches.

Previously `FaviconStore` eagerly loaded **every** stored favicon — decoding each `NSImage` — into an unbounded in-memory dictionary at launch. On a real production profile (2,700+ favicons, ~1.2 GB of stored image blobs, individual blobs up to 154 MB) this drove a **3.3 GB peak physical footprint** at startup.

This PR (single PR for the whole effort):

- **Lazy image loading** — replaces the eager `[URL: Favicon]` cache with metadata-only `[URL: FaviconMetadata]` loaded at launch + a bounded `NSCache<NSURL, NSImage>` (32 MB) populated on demand.
- **BLOB-excluded metadata fetch** — the launch fetch uses `.dictionaryResultType` + `propertiesToFetch`, so Core Data never reads the `imageEncrypted` column into its row cache (this alone cut the startup peak from 2.8 GB → 538 MB).
- **Off-main image decode** — the on-demand favicon decode runs off the main thread; the UI refreshes via `.faviconCacheUpdated`.
- **Pixel-based cache cost** — the NSCache cost reflects real pixel bytes, so the 32 MB cap is accurate rather than ~4× off on Retina.
- **Downscale on download** — freshly downloaded favicons are downscaled to ≤64 px (the New Tab Page favorites display size at 2× Retina) before storage, so existing oversized databases self-heal as sites are revisited.

**Measured startup footprint on a production-sized DB:** peak **3.3 GB → ~0.5 GB** (−84%); resident 2.7 GB → ~0.3 GB.

**Feature flags** (both default-ON `MacOSBrowserConfigSubfeature` kill switches, remotely disable-able via privacy config):
- `faviconLazyImageLoading` — OFF reverts to the legacy eager full-image cache. Registry: https://app.asana.com/1/137249556945/project/1211834678943996/task/1215720761295352
- `faviconStoringImprovements` — OFF stores favicons at their original resolution. Registry: https://app.asana.com/1/137249556945/project/1211834678943996/task/1215720760576164

### Testing Steps
For best experience import your production database to the debug container. You'll then get all your favicons and history.
See https://github.com/duckduckgo/apple-browsers/pull/5442 PR – that one is stacked on top of this and contains Favicon Inspector and a script that copies the production database to debug.

1. Use `faviconLazyImageLoading` feature flag for lazy loading and `faviconStoringImprovements` for image downscaling and filtering before saving.
2. (Memory) With a large favicon database, observe the startup physical footprint (Instruments / `vmmap -summary`); it should be dramatically lower than before — on a ~1.2 GB favicon store, ~0.5 GB peak vs ~3.3 GB previously.
4. (Lazy-load kill switch) Set `faviconLazyImageLoading` to `disabled` and relaunch — favicons still load via the eager path and startup footprint returns to the pre-change baseline (~3.3 GB).
5. (Downscaling) Use Favicon Inspector to find a website with multiple favicons. Delete all of them from the inspector. With `faviconStoringImprovements` ON, visit that website; refresh the inspector and confirm the stored favicons are small. With it OFF, the favicon is stored at original resolution.

### Impact and Risks
Medium. Touches a broad but non-critical surface — every favicon, on both the load and store paths. Favicons are not privacy- or data-loss-sensitive and regenerate on revisit. Both behaviors are guarded by default-ON remote kill switches, so a regression can be disabled without an app update.

#### What could go wrong?
- A favicon doesn't appear on first display (cold cache). Mitigated: the cold path returns metadata immediately and refreshes via `.faviconCacheUpdated`; the tab icon (`FaviconsTabExtension`) and bookmarks bar observe it. Worst case is a one-frame placeholder before the icon.
- A favicon consumer that doesn't observe `.faviconCacheUpdated` could stay on a placeholder. Callers were audited and `FaviconsTabExtension` was wired up; other list-based consumers already reload on that notification.
- Downscaling stores a lower-resolution favicon — capped at ≤64 px, the largest size the app displays (NTP favorites at 2× Retina). If a larger display surface is added later, favicons re-fetch.
- The lazy-loading architecture itself regresses → set `faviconLazyImageLoading` off; verified to revert to the eager path (startup footprint returns to ~3.3 GB).

### Quality Considerations
- Corrupt stored bitmaps are still caught via `NSException.catch` ("bad TIFF data") and re-fetched on next visit.
- `faviconLazyImageLoading` is evaluated when the cache is built at launch, so a remote disable takes effect on next launch; `faviconImageDownscaling` is read per-download and takes effect immediately.
- The flags follow the existing `faviconWKDownload` convention (subfeature checked directly via `privacyConfig`, no `FeatureFlag` enum case), so they're toggled via remote privacy config rather than the in-app debug menu.
- No new telemetry/privacy surface; the favicon decryption-failure pixel is unchanged.

### Notes to Reviewer
- The `faviconLazyImageLoading` commit re-introduces the legacy eager `FaviconImageCache` and `FaviconStore.loadFavicons()` behind the flag so OFF is a genuine revert — that accounts for the re-added (previously-shipped) code in the diff.
- Both flags need corresponding remote privacy-config entries before they can be disabled in production; the code ships them ON by default.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Broad favicon load/store and UI refresh paths change startup memory and icon timing; mitigated by default-on remote kill switches and a legacy eager cache fallback.
> 
> **Overview**
> This PR **replaces eager favicon image loading at launch** with metadata-only startup plus a **32 MB bounded `NSCache`**, off-main decode on demand, and **host-scoped `.faviconCacheUpdated`** notifications so bookmarks, the bookmarks bar, `FaviconView`, and `LoginFaviconView` refresh when images arrive.
> 
> **`FaviconManager`** picks **`FaviconImageCache`** vs legacy **`EagerFaviconImageCache`** via **`faviconLazyImageLoading`**, adds **`resolvedCachedFavicon`** for callers that must await decode, and when **`faviconStoringImprovements`** is on applies **`faviconsToKeep`** plus **≤64 px downscaling** on download. New **`FaviconMetadata`**, debug **`FaviconsDebugInspector`**, and **`duck://favicons`** support round out the change.
> 
> Privacy config adds **`faviconLazyImageLoading`** and **`faviconStoringImprovements`** as remote kill switches (default on).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c51afc00c08e9b7ca63bbb81078e26cfd317c41. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/Favicons/Model/FaviconImageCache.swift
```diff
@@ -57,6 +64,331 @@ protocol FaviconImageCaching {
...
+            }
+
+            await MainActor.run {
+                self.inFlightImageLoads.remove(faviconUrl)
+                guard let image else { return }
+                self.cacheImage(image, for: faviconUrl)
+                // Tell UI consumers the image is now available so they re-resolve.
+                NotificationCenter.default.postFaviconCacheUpdated(
+                    faviconURLs: [metadata.url],
+                    documentURLs: [metadata.documentUrl]
```
diegoreymendez, line 247: Disclaimer: I think this has a point, but I could not reproduce any visible issue.  I do think it has merit which is why I'm reporting this, but I'm finding it hard to decide it's importance — which is what I leave you to decide, @ayoy .

Is `documentURL` the right thing to be sending here?

Semantically, I believe `documentURL` is "one site that fetched the favicon", which isn't very meaningful for notification listeners trying to understand if they need to update, as favicons can be shared across many hosts.

Should we send the hosts stored in the reference cache instead?

macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -703,6 +704,21 @@ final class MainMenu: NSMenu {
...
+        // `.faviconCacheUpdated`, so also refresh the menus on that notification.
+        faviconsCacheUpdateCancellable = NotificationCenter.default.publisher(for: .faviconCacheUpdated)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                guard let self else { return }
+                if self.isLazyMenuRebuild {
+                    self.bookmarkFaviconsNeedUpdate = true
+                } else {
+                    self.updateFavicons(in: bookmarksMenu)
+                    self.updateFavicons(in: favoritesMenu)
```
diegoreymendez: Is this `O(N)` (where N = all bookmark menu items) every time the notification fires?  I think it might even be worse than `O(N)` as the update calls might trigger additional cache update notifications — seems like this could snowball.

If you agree this might result in performance issues for really large number of bookmarks, would it make sense to localize the update based on payload?

macOS/DuckDuckGo/Favicons/Services/FaviconStore.swift
```diff
@@ -75,6 +83,73 @@ final class FaviconStore: FaviconStoring {
...
+                #keyPath(FaviconManagedObject.imageEncrypted)
+            ]
+
+            let fetchResult: [FaviconManagedObject] = try context.fetch(fetchRequest)
+
+            // If the saved bitmap is corrupt, AppKit raises an Objective-C
+            // `NSInvalidUnarchiveOperationException` ("bad TIFF data") while unarchiving.
+            // Swift's `try`/`try?` cannot catch Objective-C exceptions, so this would otherwise crash the app.
+            // `NSException.catch` bridges it to a Swift error; on failure we use a nil image, which the favicon system
+            // re-fetches on the next visit.
```
diegoreymendez: Is this statement true?  Will we re-fetch with the current logic?

I think we might be missing eviction logic for corrupted bitmaps, so that we actually trigger re-fetch.

diegoreymendez (commented):
Adding some feedback while I keep testing.

macOS/DuckDuckGo/Favicons/Model/FaviconImageCache.swift
ayoy: Yes, we can switch to hosts, good call. That struct has a helper function that retrieves hosts, but that could be just stored in the struct in place of document URLs.

Side note: I'd consider removing documentURL from the database at some point, but that's a separate piece of work.

macOS/DuckDuckGo/Favicons/Services/FaviconStore.swift
ayoy: That's right, it won't be true in the new implementation. That's because when refreshing favicons we skip those that are already fetched, but for that we don't necessarily retrieve an image from the DB at the time of filtering.

It's still not critical as a corrupted favicon would expire after a week. I could force deleting a corrupted favicon from the DB, let me try it.

macOS/DuckDuckGo/Menus/MainMenu.swift
ayoy: Oh I know what's up here. This code never gets executed (thankfully) because `lazyMenuRebuild` feature flag is on. I just spoke to @jotaemepereira about deleting that flag and keeping only lazy rebuild on, in which case this won't be a problem. We could merge as-is and then delete the flag.

macOS/DuckDuckGo/Favicons/Services/FaviconStore.swift
ayoy: Deleting corrupted favicons is added

macOS/DuckDuckGo/Favicons/Model/FaviconImageCache.swift
diegoreymendez: > I'd consider removing documentURL from the database at some point, but that's a separate piece of work.

I 100% agree with this but the reason I didn't bring it up is I didn't want to make it important for this specific PR.  I am glad we're on the same page.

macOS/DuckDuckGo/Menus/MainMenu.swift
ayoy: Actually, let me just remove the conditional code already since we'll be removing the feature flag

macOS/DuckDuckGo/Menus/MainMenu.swift
```diff
@@ -690,18 +690,24 @@ final class MainMenu: NSMenu {
...
+            }
+
+        // `faviconsLoadedPublisher` fires when favicon metadata loads, before the
+        // images are decoded. Favicon images become available lazily and post
+        // `.faviconCacheUpdated`, so also refresh the menus on that notification.
+        faviconsCacheUpdateCancellable = NotificationCenter.default.publisher(for: .faviconCacheUpdated)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                guard let self else { return }
+                self.bookmarkFaviconsNeedUpdate = true
```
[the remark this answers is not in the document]
ayoy: Yes, that's intended 🫡

diegoreymendez (approved):

macOS/DuckDuckGo/Tab/Navigation/DuckURLSchemeHandler.swift
```diff
@@ -74,12 +81,22 @@ final class DuckURLSchemeHandler: NSObject, WKURLSchemeHandler {
             default:
                 handleSpecialPages(urlSchemeTask: urlSchemeTask)
             }
+        case .favicons:
+            guard featureFlagger.internalUserDecider.isInternalUser else {
+                fallthrough
+            }
+            faviconsDebugInspector.handle(requestURL: requestURL, urlSchemeTask: urlSchemeTask)
```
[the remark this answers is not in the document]
ayoy: This is correct and expected.

macOS/DuckDuckGo/Favicons/Model/FaviconImageCache.swift
```diff
@@ -53,10 +60,411 @@ protocol FaviconImageCaching {
...
+        guard !toRemove.isEmpty else { return }
+        for metadata in toRemove {
+            entries[metadata.url] = nil
+            imageCache.removeObject(forKey: metadata.url as NSURL)
+        }
+        await removeFaviconsFromStore(toRemove)
+        NotificationCenter.default.postFaviconCacheUpdated(
+            faviconURLs: Set(toRemove.map(\.url)),
+            documentURLs: Set(toRemove.map(\.documentUrl))
+        )
```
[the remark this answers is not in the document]
ayoy: This is acceptable.

PR #5404, opened 2026-06-15
Only disable Duck Player for new users when rolling out ad-blocking

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203936086921904/task/1215650122674746?focus=true
Tech Design URL: N/A
CC:

### Description

Reworks how the Duck Player default relates to the `adBlockingExtensionEnabledByDefault` (ad-blocking defaults) rollout.

Previously the Duck Player default was **lazily computed** from the rollout flag, so the setting changed implicitly as the flag flipped. This PR:

- **Decouples** the default: the unset Duck Player default is now unconditionally `.alwaysAsk` (iOS native YouTube mode `.ask`), independent of any flag — restoring the original behavior.
- **Drives the rollout via a one-time persisted write at onboarding completion** for a new user instead: when the rollout is active (flag on **and** OS-supported), Duck Player is explicitly set to disabled. After that it's a real stored value the user owns.
  - iOS sets all three relevant settings — classic `duckPlayerMode = .disabled` (iPad), native `duckPlayerNativeYoutubeMode = .never` (iPhone), and `duckPlayerNativeUISERPEnabled = false` — so the active one is off regardless of device / native-UI flag.
  - macOS sets `duckPlayerMode = .disabled`.
- **Covers every onboarding-completion path** so the logic isn't bypassed: normal completion, the debug "Skip Onboarding" action (macOS), and the automation/UI-test bypasses (both platforms).
- Houses the logic where it belongs: an `AppSettings` protocol extension (iOS) and a static `OnboardingActionsManager` helper (macOS).
- Adds unit tests for both flag states on both platforms.

### Testing Steps
See the test scenarios in [this Asana task](https://app.asana.com/1/137249556945/task/1215722174097999).

### Impact and Risks
Medium. Changes a user-facing default (Duck Player) for **new users only**, and only when the ad-blocking defaults rollout is active on a supported OS. Gated behind the remote `adBlockingExtensionEnabledByDefault` flag and reversible by toggling it off.

#### What could go wrong?
- A new user who finishes onboarding while the rollout is active gets Duck Player off by default; they can re-enable it in Settings. There is no Duck Player choice during onboarding, so no explicit user choice is overwritten.
- Existing users are unaffected — the write only runs at onboarding completion, gated on the rollout being active.
- The automation/UI-test bypasses now also apply the default; this only triggers when the rollout flag is enabled in that test configuration, so it won't perturb tests that run with the flag off.

### Quality Considerations
[...]

PR #5405, opened 2026-06-15
Only disable Duck Player for new users when rolling out ad-blocking

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203936086921904/task/1215650122674746?focus=true
Tech Design URL: N/A
CC:

### Description

Reworks how the Duck Player default relates to the `adBlockingExtensionEnabledByDefault` (ad-blocking defaults) rollout.

Previously the Duck Player default was **lazily computed** from the rollout flag, so the setting changed implicitly as the flag flipped. This PR:

- **Decouples** the default: the unset Duck Player default is now unconditionally `.alwaysAsk` (iOS native YouTube mode `.ask`), independent of any flag — restoring the original behavior.
- **Drives the rollout via a one-time persisted write at onboarding completion** for a new user instead: when the rollout is active (flag on **and** OS-supported), Duck Player is explicitly set to disabled. After that it's a real stored value the user owns.
  - iOS sets all three relevant settings — classic `duckPlayerMode = .disabled` (iPad), native `duckPlayerNativeYoutubeMode = .never` (iPhone), and `duckPlayerNativeUISERPEnabled = false` — so the active one is off regardless of device / native-UI flag.
  - macOS sets `duckPlayerMode = .disabled`.
- **Covers every onboarding-completion path** so the logic isn't bypassed: normal completion, the debug "Skip Onboarding" action (macOS), and the automation/UI-test bypasses (both platforms).
- Houses the logic where it belongs: an `AppSettings` protocol extension (iOS) and a static `OnboardingActionsManager` helper (macOS).
- Adds unit tests for both flag states on both platforms.

### Testing Steps
See the test scenarios in [this Asana task](https://app.asana.com/1/137249556945/task/1215722174097999).

### Impact and Risks
Medium. Changes a user-facing default (Duck Player) for **new users only**, and only when the ad-blocking defaults rollout is active on a supported OS. Gated behind the remote `adBlockingExtensionEnabledByDefault` flag and reversible by toggling it off.

#### What could go wrong?
- A new user who finishes onboarding while the rollout is active gets Duck Player off by default; they can re-enable it in Settings. There is no Duck Player choice during onboarding, so no explicit user choice is overwritten.
- Existing users are unaffected — the write only runs at onboarding completion, gated on the rollout being active.
- The automation/UI-test bypasses now also apply the default; this only triggers when the rollout flag is enabled in that test configuration, so it won't perturb tests that run with the flag off.

### Quality Considerations
- Edge cases: explicit user choices are never overwritten outside onboarding; below the supported OS (macOS 15.4 / iOS 18.4) `areAdBlockingDefaultsActive` is false so nothing changes; iOS handles the iPad-classic / iPhone-native / SERP split.
- Privacy: Duck Player is a privacy feature — the default change is intentional and remotely controllable.
- Tests cover rollout-off (stays default) and rollout-on (disabled) on both platforms.

### Notes to Reviewer
- First commit (`Decouple…`) is effectively a revert of the previous lazy-default coupling; the remaining commits build the onboarding-completion approach on top.
- The iOS seam passes the resolved rollout gate as a `Bool` into the `AppSettings` extension (keeps it unit-testable without a `MainViewController`); macOS keeps a static taking the `FeatureFlagger`.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> User-facing default for new users only, gated on remote flag and OS support; existing users are unaffected unless they complete onboarding under the rollout.
> 
> **Overview**
> **Duck Player** no longer changes its unset default when the ad-blocking rollout flag flips. Unstored values always fall back to **`.alwaysAsk`** / **`.ask`** on iOS and macOS, and debug UI labels match that.
> 
> When **`adBlockingExtensionEnabledByDefault`** is active at **onboarding completion**, the app **persists** Duck Player off for new users: iOS writes **`duckPlayerMode = .disabled`** and **`duckPlayerNativeYoutubeMode = .never`** via **`AppSettings.applyAdBlockingRolloutDuckPlayerDefaultsIfNeeded`**; macOS does the same through **`OnboardingActionsManager.applyAdBlockingRolloutDuckPlayerDefaultIfNeeded`**, including a notification so live preferences refresh without relaunch.
> 
> That write runs on **every completion path** (normal finish, skip onboarding, UI-test / automation bypass), not only the main flow. macOS drops rollout-driven lazy defaults from **`DuckPlayerPreferences`** (feature-flag publisher and **`FeatureFlagger`** wiring removed from **`DuckPlayer`** init).
> 
> Tests were narrowed to the onboarding helper behavior on both platforms instead of rollout-aware getter defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9440e70b5232dc9c4325f27fcfa4b624daec452d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsymons (approved):
Approving per https://github.com/duckduckgo/apple-browsers/pull/5404

PR #5410, opened 2026-06-15
Refresh Privacy Dashboard CPM state for background tabs

muodov:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1163321984198618/task/1215712815660046?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1215747457572013?focus=true
CC: @sammacbeth 

### Description
1. Lets iOS apply dashbaord state to background tabs
2. Makes the matching condition stricter: we match domain + path now.

### Testing Steps
0. Install [this embedded extension](https://github.com/user-attachments/files/29006190/embedded-release-20260616_153744.zip) (it’s a build from [this PR](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3570))
1. open http://privacy-test-pages.site/features/autoconsent/autoconsent-index.html in the active tab
3. Long-click on the "normal cookie popup" link and **open in background**
4. wait a couple seconds
5. switch to that tab
6. verify that the reject button says "I was clicked!”
7. Open Privacy Dashboard, verify that Cookies Managed row is there
8. Go to the first page, and repeat these steps for the "cosmetic filterlist” link
9. Switch to the cosmetic filterlist test page
10. Verify that the cookie banner is NOT hidden (cosmetic filterlist is disabled for all users)
11. verify that there's NO CPM row in the privacy dashboard for that tab

### Notes to Reviewer

---

> [!NOTE]
> **Low Risk**
> Privacy UI state only; stricter URL matching reduces wrong-tab updates. No auth or data pipeline changes.
> 
> **Overview**
> Autoconsent **Privacy Dashboard (CPM) refresh** now carries the full page **URL** through the web extension stack instead of a **domain** string, and each tab decides whether to apply the update.
> 
> The shared delegate and `refreshCpmDashboardState` handler forward a `URL` in notifications (`UserInfoKeys.url`). iOS **moves** dashboard refresh handling from `MainViewController` (active tab + host match) to **`TabViewController`**, so **background tabs** can update cookie consent state before the user switches to them. macOS tab privacy dashboard code uses the same **host + normalized path** gate via `PrivacyInfo.updateCookieConsentManagedForWebExtensionDashboardState` (query strings ignored; empty path treated as `/`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f2474b06236e57f2cb661470d03475c8cd6de92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -3770,6 +3792,40 @@ extension TabViewController: AutoconsentUserScriptDelegate {
...
+              normalizedPath(url.path) == normalizedPath(refreshURL.path) else {
+            return
+        }
+
+        cookieConsentManaged = consentStatus.toCookieConsentInfo()
+    }
+
+    private func normalizedPath(_ path: String) -> String {
+        path.isEmpty ? "/" : path
+    }
```
miasma13, line 3829: Maybe we should consider trailing slash normalization here as well if the path includes it (to avoid recognising them as different URLs depending on its presence).

macOS/DuckDuckGo/Tab/TabExtensions/PrivacyDashboardTabExtension.swift
```diff
@@ -279,6 +279,22 @@ extension TabExtensions {
...
+    func updateCookieConsentManagedForWebExtensionDashboardState(url refreshURL: URL, consentStatus: ConsentStatusInfo) {
+        guard url.host == refreshURL.host,
+              normalizedPath(url.path) == normalizedPath(refreshURL.path) else {
+            return
+        }
+
+        cookieConsentManaged = consentStatus.toCookieConsentInfo()
+    }
+
+    private func normalizedPath(_ path: String) -> String {
```
miasma13, line 293: See previous comment to other instance of `private func normalizedPath`

iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -796,6 +800,24 @@ class TabViewController: UIViewController {
...
+    @available(iOS 18.4, *)
+    private func registerForWebExtensionNotifications() {
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(handleWebExtensionDashboardStateRefresh),
+                                               name: .webExtensionAutoconsentDashboardStateRefresh,
+                                               object: nil)
+    }
+
+    @available(iOS 18.4, *)
+    @objc private func handleWebExtensionDashboardStateRefresh(_ notification: Notification) {
```
miasma13: Here we have off main thread mutation, compared to previous code.

macOS/DuckDuckGo/Tab/TabExtensions/PrivacyDashboardTabExtension.swift
```diff
@@ -279,6 +279,22 @@ extension TabExtensions {
 
 // MARK: - ConsentStatusInfo to CookieConsentInfo Conversion
 
+@available(macOS 15.4, *)
+extension PrivacyInfo {
+    func updateCookieConsentManagedForWebExtensionDashboardState(url refreshURL: URL, consentStatus: ConsentStatusInfo) {
```
miasma13, line 284: Not sure if this is a regression or expected behaviour that we might already have but when replicating scenario:
1. Fire everything
2. Open new tab and visit olx.pl
3. Cookie should be managed and open PD to see info about it.
4. Open another  tab and visit olx.pl
5. Open PD, no cookies managed info
6. Switch to first tab.
7. Open PD, no cookies managed info

muodov: This is not ideal, but a known side effect. Because both tabs point to the same (domain+path), we can’t distinguish between them. So the update from the second page load overrides dashboard state for both tabs.

iOS/DuckDuckGo/TabViewController.swift
muodov: /foo and /foo/ are technically different paths, so I’d rather not do it until we see actual mismatch cases

miasma13 (approved):
LGTM 👍

PR #5429, opened 2026-06-16
[iOS] UTI: favorites on a pre-filled URL, no dismiss flash, long URL single-line

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1213881875984009/task/1215717889939551

### Description
Three related fixes to the Unified Toggle Input (UTI) omnibar. The UTI is gated by the **unified toggle input** feature flag — this is distinct from the older "experimental input" omnibar, which is a separate feature.

### Testing Steps
Enable the **unified toggle input** feature flag. (Not the "experimental input" omnibar — that's a separate feature.) Use a tab loaded on a page and ≥1 saved favorite.

A — UTI on, toggle enabled:
1. Tap the address bar → favorites show (or the Dax logo if none), NOT search suggestions.
2. Type → suggestions replace favorites; clear to empty → favorites return.
3. Pre-filled URL, don't type, tap back → no suggestion-list flash during the collapse (try a few times — it was intermittent).

B — UTI on, long URL:
4. Load a page with a long URL → tap the address bar → the URL stays on ONE line (no wrap to two).
5. Switch to Duck.ai and edit a long prompt → it can grow to multiple lines (intended; only an interacted AI-chat composer expands).

C — UTI on, toggle disabled:
6. In Settings set AI Chat to "Search only" so the UTI shows with NO toggle; relaunch.
7. Tap the address bar on a long URL → it stays ONE line. Moving the caret without typing keeps favorites; typing shows suggestions.

D — Regression (shared text entry):
8. Turn the unified toggle input flag OFF, restart app → plain omnibar focus / type / dismiss / navigate behaves normally.

### Impact and Risks
Low. UTI-only behaviour, confined to pre-filled-URL focus and dismiss.

#### What could go wrong?
The shared `SwitchBarTextEntryView` edits (`textViewDidChangeSelection` now height-only; `isUnexpandedURL` single-line gating). The shared edit path is untouched, so real typing still marks interaction. The one behavioural delta for the experimental SwitchBar omnibar: a caret-move after select-all no longer counts as interaction (favorites stay until typing) — arguably more correct. Covered by steps 8–9.

---

> [!NOTE]
> **Low Risk**
> Changes are scoped to UTI suggestion merging and shared `SwitchBarTextEntryView` selection/height logic; real typing still marks interaction, with minor shared-component edge-case risk.
> 
> **Overview**
[...]

PR #5444, opened 2026-06-17
[iOS][UTI] Restore multi-line expansion for long URLs so the end is editable

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1215798111901228?focus=true

### Description

In the Unified Toggle Input (UTI), focusing the address bar on a page with a long URL showed the URL clipped to a single, truncated line. Because that line is not horizontally scrollable, the caret could not reach the end of the URL, so the user couldn't edit the tail of a long URL — unlike the plain omnibar (a single-line `UITextField`, which scrolls horizontally for free).

The single-line behaviour was introduced by #5373 (`currentMode == .search` clause) and extended by #5429 (`!handler.isToggleEnabled` clause). This restores the original `isUnexpandedURL()`:

```swift
private func isUnexpandedURL() -> Bool {
    return !hasBeenInteractedWith && isURL
}
```

So a URL is single-line only **before** the user interacts; once the field is interacted with, the URL falls into the existing expandable path and grows to multiple lines (capped at the max height, then scrolls), making the whole URL — including its end — visible and editable.

Unit tests that asserted the single-line behaviour are updated to expect expansion (search and search-only modes), and the now-obsolete "collapses to single line on mode switch to search" test is removed.

### Testing Steps
1. Use the UTI omnibar (search-only / toggle-disabled config) on a tab loaded on a page with a long URL.
2. Tap the address bar → the URL expands to multiple lines, with its end visible.
3. Place the caret at the end of the URL and edit it → the end is reachable and editable.
4. A short URL / search query still shows on a single line; Duck.ai multi-line prompt input is unaffected.

### Impact and Risks
Low. Scoped to a single function (`isUnexpandedURL`) that gates URL height in the UTI; reuses the existing expandable layout path. No change to non-URL text, the focus animation, or multi-line prompt input.

#### What could go wrong?
A very long URL grows the card up to its max height before scrolling (the original behaviour #5373 had intentionally capped to one line). This is the accepted, intended trade-off for editability.

### Notes to Reviewer
- Targets the **release branch** `release/ios/7.225.0` for the imminent release. The same gating exists on `main` (via #5429) and will need the equivalent revert there.
- Behaviour validated manually on the simulator: a long URL now expands and its end is editable.
- Unit tests couldn't be executed locally (the `iOS Browser` scheme doesn't include the `DuckDuckGoTests` target); CI will run them.

---

> [!NOTE]
> **Low Risk**
> Single helper change in UTI text entry height logic; intentional trade-off is taller cards for very long URLs up to max height before scrolling.
> 
> **Overview**
> Reverts the **#5373 / #5429** single-line URL rules in the Unified Toggle Input so long URLs are only forced to one line **before** the user taps the field. After interaction, `isUnexpandedURL()` no longer keeps URLs collapsed in search mode or when the AI toggle is disabled—they use the existing expandable height path so the caret can reach the end of the URL.
> 
> Unit tests now expect height growth after tap in search and search-only configurations, and the obsolete test that asserted collapse when switching from AI chat to search is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b0bc6b3589667f0d09ca6d01b4ca224c3845956. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/TextEntry/SwitchBarTextEntryView.swift
+++ b/iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/TextEntry/SwitchBarTextEntryView.swift
@@ -545,12 +545,11 @@ class SwitchBarTextEntryView: UIView {
     }
 
     /// https://app.asana.com/1/137249556945/project/392891325557410/task/1210835160047733?focus=true
-    /// A URL stays single-line unless it's an interactive AI-chat composer: search mode, or
-    /// search-only (no toggle, so AI chat is unavailable), or before the user interacts — all
-    /// single-line. Only an interacted AI-chat field (toggle present) lets a long URL expand.
+    /// A URL is single-line only before the user interacts; once tapped it expands to multiple lines
+    /// so the caret can reach the end to edit (the original behaviour, restored after #5373/#5429
+    /// made long URLs single-line and therefore not editable to the end).
     private func isUnexpandedURL() -> Bool {
-        guard isURL else { return false }
-        return currentMode == .search || !handler.isToggleEnabled || !hasBeenInteractedWith
+        return !hasBeenInteractedWith && isURL
     }
 
     private func updateTextViewHeight() {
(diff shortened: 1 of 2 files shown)


miasma13 (approved):
LGTM

PR #5485, opened 2026-06-19
Add site-breakage diagnostic logging (macOS)

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215859135028061

### Description

When an internal user hits a broken site and exports their unified logs, there's currently no diagnostic trail explaining why the site broke or whether our own protections were involved. This adds an export-safe, per-page accumulator that emits a curated digest under the `Logger.siteBreakage` category on demand, so the exported logs carry that detail.

The digest is logging-only and lives entirely separate from the broken-site report path. Hosts are logged hashed so the logs are correlatable but never carry cleartext browsing history; counts, failure classes, and flags are public. State is per-tab, held in memory, and reset on each navigation.

### Testing Steps

1. Build and run the macOS app.
2. Open Console.app and filter the subsystem to the DuckDuckGo process with category `Site Breakage`.
3. Visit a site and let it load.
4. Open the site-protections popover from the address bar.
5. Confirm a site-breakage digest is written to the log.
6. Confirm page and resource hosts appear as hashes, not cleartext.

Triggering the network-connection-integrity and storage-access-quirk signals reliably requires sites that exercise those WebKit protections, which isn't feasible to reproduce on demand; their accumulation logic is covered by unit tests instead.

### Impact and Risks

Low. The change is diagnostic logging gated behind an existing UI action, with no effect on navigation, rendering, or the report payload. The new navigation-delegate SPI hooks fan out to the responder chain alongside existing ones and are no-ops when WebKit doesn't invoke them.

#### What could go wrong?

The accumulator could grow unbounded on a pathological page. Each signal group is capped, dedup-keyed, and reset per navigation, so memory stays bounded to a single page load in a single tab. Host hashing is applied at the log call via the privacy marker, so a missed marker would surface a host in cleartext — the digest is rendered in one place per group to keep that auditable.

### Quality Considerations

Per-tab in-memory state with per-navigation reset and per-group caps keeps the footprint bounded. The render-health derivation is a pure function with unit coverage; the new signal groups have unit tests for their anomaly and capping behavior.

### Notes to Reviewer

Each signal class is a `BreakageSignalGroup` that owns its accumulation, anomaly decision, and digest rendering. The emit guard derives the "did anything go wrong?" decision from the groups, so adding a future signal means adding a group rather than editing shared logic.

---
###### Internal references:


diegoreymendez:
Discontinuing this effort.

PR #5514, opened 2026-06-23
Fix VPN excluded-domain hostname matching

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215958595063372

### Description

Tightens how the VPN matches a connection's hostname against the excluded-domains list so that only the configured domain and its subdomains are matched. Previously the check could match unintended hostnames that happened to share a trailing string with an excluded domain.

### Testing Steps
1. Add `myip.com` to the VPN excluded domains.
2. Visit myip.com and confirm it is excluded from the VPN.
3. Visit whatismyip.com and confirm it still routes through the VPN.

Note: the proxy only evaluates exclusions for new connections, so it cannot re-route a connection that is already open. To avoid waiting for an existing connection to refresh, do each check in a freshly opened browser (or a different browser) rather than reusing the same window. Restarting the VPN connection also forces re-evaluation, but that is heavier than necessary for this.

### Impact and Risks

Medium. Adjusts which connections match an excluded domain.

#### What could go wrong?

N/A

### Quality Considerations

The matching logic was extracted into a pure helper and covered with unit tests for exact match, subdomains, and near-miss hostnames.

### Notes to Reviewer

I've noticed there's an issue where the proxy takes a while to refresh properly on open connections.  I've [opened an issue for it](https://app.asana.com/1/137249556945/project/1207603085593419/task/1215959562779962?focus=true), so the testing here shouldn't focus too much on whether the exclusions are being immediate.  I recommend checking before and after in different browsers so you get a fresh connection on each attempt.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes which connections bypass the VPN for excluded domains; behavior shifts for hostnames that previously matched via loose suffix rules.
> 
> **Overview**
> **VPN excluded-domain matching** no longer treats any hostname that merely ends with the configured string as excluded. Matching now requires an **exact domain** or a **proper subdomain** (`.<excludedDomain>`), so lookalikes like `evilbank.com` when excluding `bank.com` stay on the tunnel.
> 
> The check is refactored into a **static `isExcludedDomain(_:excludedDomains:)`** helper used by the instance method, with **unit tests** covering exact/subdomain hits, suffix false positives, and unrelated hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b79af85f03fba201dd36303f05d5b2cd3510155d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/LocalPackages/NetworkProtectionMac/Sources/NetworkProtectionProxy/Provider/TransparentProxyProvider.swift
+++ b/macOS/LocalPackages/NetworkProtectionMac/Sources/NetworkProtectionProxy/Provider/TransparentProxyProvider.swift
@@ -574,8 +574,17 @@ open class TransparentProxyProvider: NETransparentProxyProvider {
     }
 
     private func isExcludedDomain(_ hostname: String) -> Bool {
-        settings.excludedDomains.contains { excludedDomain in
-            hostname.hasSuffix(excludedDomain)
+        Self.isExcludedDomain(hostname, excludedDomains: settings.excludedDomains)
+    }
+
+    /// Whether `hostname` should be excluded from the VPN given the list of excluded domains.
+    ///
+    /// A hostname matches an excluded domain only when it's exactly equal to it or is a subdomain of it
+    /// (i.e. ends with `.<excludedDomain>`). This prevents an excluded `bank.com` from also matching an
+    /// unrelated `evilbank.com`, which would steer that traffic around the tunnel.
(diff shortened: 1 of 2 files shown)


SabrinaTardio (approved):
Works as expected!

PR #5520, opened 2026-06-23
[BSK] Replace legacy string-based URL fragment helpers with component-based URL.equals(_:by:)

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1215403820548743
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1216000789595049
CC: N/A

### Description
- Add `URL.EqualityComponents` OptionSet and `URL.equals(_:by:)` for component-based URL equality with named presets `.sameDocument` (scheme+host+port+path+query) and `.fuzzyIdentity` (all components, trailing-slash-normalised path)
- Add `URL.hasFragment` computed property
- Remove `String.hashedSuffixRange / .hashedSuffix / .droppingHashedSuffix` and all their tests
- Remove `URL.isSameDocument(_:)` and `URL.matches(_:)`; replace every call site with `URL.equals(_:by: .sameDocument)` / `URL.equals(_:by: .fuzzyIdentity)` across `WKNavigationActionExtension`, `DistributedNavigationDelegate`, `NavigationPixelNavigationResponder`, `FrameInfo`, and all navigation test helpers
- Remove `URL.URLProtocol` enum, which duplicated `NavigationalScheme`; replace all usages (`toHttps`, `isCustomURLScheme`, `webUrl(from:)`, `AutofillDomainNameUrlMatcher`, `AdClickAttributionFeature`, `TabURLInterceptor`) with `NavigationalScheme`
- Add `URLEqualityComponentsTests`

### Testing Steps
1. Open test page: [spammy-data-fragment.html](https://github.com/user-attachments/files/29533929/spammy-data-fragment.html), click "Copy dynamic long-fragment data: URL" button
2. Open the copied data URL in a new tab, sequentially click navigation buttons at the top, then click "Top": validate same-doc navigation is working, no assertions raised
3. iOS: enter 900000/10000 repetition values on the attached page; Click "Open data: URL in same tab" button, validate `URL.equals(_:by:)` is called and the app doesn't crash or hang. (the page won‘t load – that‘s expected)
4. Repeat the test for the iOS Browser target
5. Run BSK tests for iOS targets with `_ORIGINAL_DATA_AS_STRING_ENABLED` flag disabled in `Package.swift`, validate the tests are passing.

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- Same-document navigation classification could regress if `URL.equals(by: .sameDocument)` behaves differently from the old string comparison for unusual URL forms; mitigated by the expanded parameterised test matrix covering WebKit's own `equalIgnoringFragmentIdentifier` cases
- `URL.hasFragment` delegates to Swift Foundation's `URL.fragment`; performance on large `data:` URIs is validated by the performance test

### Quality Considerations
- All public-facing string helpers (`hashedSuffix`, `droppingHashedSuffix`, `isSameDocument`, `matches`) are removed rather than deprecated to prevent future misuse
- No behaviour change — the same-document logic gate is semantically identical to the old string-split approach

---

> [!NOTE]
> **Medium Risk**
> Changes core navigation same-document and URL equality behavior across WebKit opaque URLs and large data: URIs; broad call-site churn but heavily tested.
> 
> **Overview**
> **Replaces legacy string fragment helpers** (`hashedSuffix`, `isSameDocument`, `matches`, `URL.URLProtocol`) with **`URL.EqualityComponents` / `equals(_:by:)`** presets (`.sameDocument`, `.fuzzyIdentity`), **`URL.hasFragment`**, and **`NavigationalScheme`** across navigation, autofill, iOS URL handling, and tests.
> 
> Adds **`URLComponents(webKitUrl:)`** (macOS: `_web_originalDataAsString` via `_ORIGINAL_DATA_AS_STRING_ENABLED`) so opaque **`about:` / `data:`** URLs get correct fragments instead of scanning huge `absoluteString`s—addressing same-document classification and **SIGKILL/hang regressions** on large `data:` URLs.
> 
> **Navigation** now uses component equality for same-document checks, client redirects, and **`Navigation.willStart(isSameDocument:)`**; **`WKNavigationAction.isSameDocumentNavigation`** gates on **`hasFragment`** plus `.sameDocument` equality. Test bundles rename **`Resources` → `Res`** (iOS codesign); expanded **`URLEqualityComponentsTests`** and **`WKNavigationActionExtensionTests`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9bb6babfd5fbe05c2e2c82a8223fe7e6bcb41b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/BrowserServicesKit/Package.swift
```diff
@@ -704,7 +705,11 @@ let package = Package(
                 "WKAbstractions",
             ],
             resources: [
-                .copy("Resources")
+                // Xcode CodeSign build step fails for iOS target with "bundle format unrecognized, invalid, or unsuitable"
+                // when the folder is named "Resources"
+                //
+                // https://stackoverflow.com/questions/29271548/code-sign-error-bundle-format-unrecognized-invalid-or-unsuitable
+                .copy("Res")
```
mallexxx, line 712: iOS BSK tests were not compiling (only for me?), renaming the Resources folders and fixing some compile errors fixed the build

SharedPackages/BrowserServicesKit/Sources/Common/Extensions/URLComponentsExtension.swift
```diff
@@ -18,9 +18,42 @@
...
+    /// came through WebKit — holds the original byte string. Parsing *that* with
+    /// `URLComponents(string:)` bypasses the NSURL opaque URL limitation entirely.
+    ///
+    /// Falls back to `URLComponents(url:resolvingAgainstBaseURL:)` when the property is absent
+    /// (`URL(string:)`-created values, all hierarchical URLs).
+    init?(webKitUrl: URL) {
+#if DEBUG
+        // Link WebKit for test targets
+        _=type(of: WKWebView.self)
+#endif
```
ayoy: Nit: this could co inside `#if _ORIGINAL_DATA_AS_STRING_ENABLED`

SharedPackages/BrowserServicesKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -387,9 +387,78 @@ extension URL {
...
+    /// fragments, so the fragment is found by scanning `absoluteString` for `#` or `%23`.
+    /// Path comparison always strips a trailing '/' (except for bare root '/').
+    public func equals(_ other: URL, by components: EqualityComponents) -> Bool {
+        guard let selfParsed  = URLComponents(webKitUrl: self),
+              let otherParsed = URLComponents(webKitUrl: other) else { return false }
+
+        return EqualityComponents.allComponents.filter(components.contains).allSatisfy { component in
+            switch component {
+            case .scheme:
+               return scheme == other.scheme
```
ayoy: This doesn't need parsing URLs, so we could try and put the scheme check before parsing URLs, but also it's probably an ultra edge case so fine to keep it as-is

SharedPackages/BrowserServicesKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -387,9 +387,78 @@ extension URL {
...
+    ///
+    /// For hierarchical URLs (`http://`, `https://`, `file://`, …) components are extracted
+    /// via `URLComponents` which gives correct percent-decoding for all fields.
+    ///
+    /// For opaque URLs (`about:`, `data:`, `javascript:`, …) `URLComponents` does not surface
+    /// fragments, so the fragment is found by scanning `absoluteString` for `#` or `%23`.
+    /// Path comparison always strips a trailing '/' (except for bare root '/').
+    public func equals(_ other: URL, by components: EqualityComponents) -> Bool {
+        guard let selfParsed  = URLComponents(webKitUrl: self),
+              let otherParsed = URLComponents(webKitUrl: other) else { return false }
```
ayoy: We would return false also if both URLs fail to parse via `URLComponents(webKitUrl:)`. It's probably not a big deal as that shouldn't happen, but wanted to flag it.

SharedPackages/BrowserServicesKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -387,9 +387,78 @@ extension URL {
...
+        static let allComponents: [Self] = [.scheme, .host, .port, .path, .query, .fragment]
+    }
+
+    /// Returns true when `self` and `other` are equal for every component in `components`.
+    ///
+    /// For hierarchical URLs (`http://`, `https://`, `file://`, …) components are extracted
+    /// via `URLComponents` which gives correct percent-decoding for all fields.
+    ///
+    /// For opaque URLs (`about:`, `data:`, `javascript:`, …) `URLComponents` does not surface
+    /// fragments, so the fragment is found by scanning `absoluteString` for `#` or `%23`.
```
ayoy: Is this still valid after recent changes? We don't seem to be scanning the string.

ayoy (commented):
Dropping a few comments for now. Testing is going well and I'm not seeing any issues, but will continue later today. Thanks @mallexxx! 🙌

SharedPackages/BrowserServicesKit/Sources/Common/Extensions/URLExtension.swift
mallexxx: I assume that would be a correct behavior.
In fact, I've reworked the comparator to skip missing (nil) components and only return true if there‘s actually valuable matching components: i.e. `.host`,`.port` comparisons would return `false` for opaque urls, as well as `.fragment`-only comparisons for opaque urls on iOS.
`.path` comparisons would compare full URLComponents paths including the fragment for iOS: added tests for that: `"data:text/html,hello", "data:text/html,hello#anchor", [.path] =>  false`. for macOS that would be true.

SharedPackages/BrowserServicesKit/Tests/RemoteMessagingTests/Mappers/JsonToRemoteMessageModelMapperCardsListTests.swift
```diff
@@ -341,7 +341,7 @@ struct JsonToRemoteMessageModelMapperCardsListTests {
         #expect(items.first?.imageUrl == nil)
     }
 
-    @Test("Two-lines item has nil imageUrl when JSON value is an empty string", .timeLimit(.minutes(1)))
+    @Test("Two-lines item has nil imageUrl when JSON value is an empty string")
```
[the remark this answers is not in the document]
mallexxx: Removed the timeouts as they‘re not supported on iOS 15

ayoy (approved):
LGTM, all works great and doesn't hang for good anymore on both iOS and macOS. Thanks @mallexxx!

PR #5532, opened 2026-06-24
iOS: Fix favorites reorder double-applying move (favoritesListIndexNotMatchingBookmark spike)

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1215957790970786
Tech Design URL:
CC:

### Description

`m_d_favorites_list_index_not_matching_bookmark` (`d_favorites_list_index_not_matching_bookmark_ios_phone`) spiked +1000% in 7.225.

Cause: PR #5337 made `FavoritesListInteractingAdapter` forward the model's `localUpdates` into `externalUpdates`, and `FavoritesListViewModel.save()` refresh its published list before emitting. So `FavoritesViewModel.updateData()` now re-enters synchronously during a New Tab Page reorder, and `moveFavorites` applied the move twice — once via that re-entrant refresh, once via `allFavorites.move(...)` — desyncing the in-memory list from Core Data. The drop delegate calls `moveFavorites` repeatedly within a single drag, so once desynced the index guard `visibleChildren[fromIndex] == favorite` fails and fires the pixel during that drag. The reorder itself still works and persists (Core Data stays the source of truth; failed-guard moves are skipped, not corrupted) — the pixel is spurious. Not gated by `unifiedToggleInput`, so it affects all iOS phone NTP favorites reordering.

Fix: in `FavoritesViewModel.moveFavorites`, reload from the data source via `updateData()` instead of re-applying the move locally (matches `deleteFavorite`).

### Testing Steps

**Reproduce the bug** (on a build without this fix, e.g. current `release/ios/7.226.0`):
1. New Tab Page with 3+ favorites.
2. Open Console and filter `favorites.list.index.not.matching`.
3. Drag-reorder a favorite and release. **The move fails on the first attempt** — the favorite does not move (it returns to its original position) and `m_d_favorites_list_index_not_matching_bookmark` fires at that moment. (The index guard `visibleChildren[fromIndex] == favorite` is a single branch: it both skips the move *and* fires the pixel — the user-visible failure and the pixel are the same event.)
4. Drag-reorder the same favorite again — **the move succeeds on the second attempt** (the in-memory list has re-synced). So the symptom is "reordering only works on the second try," and each failed first attempt fires the pixel (the +1000% spike).

**Verify the fix** (this branch):
1. Repeat the steps above — the reorder **succeeds on the first attempt** and no pixel fires.
2. Run `NewTabPageFavoritesModelTests` (scheme **iOS Unit Tests**: `-only-testing:UnitTests/NewTabPageFavoritesModelTests`) → 7/7 pass.

**Smoke Test** (this branch):
1. Smoke test: adding, removing favorite, and reorder 

### Impact and Risks

Bug impact: Low for users / high telemetry noise — NTP favorites reordering still works and persists (no data loss, no crash; Core Data stays the source of truth), but the spurious index-mismatch pixel fires on essentially every reorder (the +1000% spike on a diagnostic pixel) because the in-memory list double-applies the move and desyncs mid-drag. Fix risk: Low — one line, mirrors `deleteFavorite`, covered by new + existing tests.

#### What could go wrong?
SwiftUI now sees a wholesale `allFavorites` refresh instead of a relative `.move`; that refresh already happened in 7.225, so the rendering input is simpler than today. The reorder animation is the only thing the tests don't cover (hence the manual step).

### Notes to Reviewer

Regression came from #5337 (the changed files sit below the `unifiedToggleInput` flag, which is why a UTI PR affected everyone).

---
###### Internal references:

### DIFF
--- a/iOS/DuckDuckGo/FavoritesViewModel.swift
+++ b/iOS/DuckDuckGo/FavoritesViewModel.swift
@@ -150,7 +150,8 @@ class FavoritesViewModel: ObservableObject {
         let favorite = allFavorites[fromIndex]
 
         favoriteDataSource.moveFavorite(favorite, fromIndex: fromIndex, toIndex: index)
-        allFavorites.move(fromOffsets: IndexSet(integer: fromIndex), toOffset: index)
+        // Reload from the data source; it already republishes the reordered list, so don't re-apply the move.
+        updateData()
     }
 
     // MARK: -
--- a/iOS/DuckDuckGoTests/NewTabPageFavoritesModelTests.swift
+++ b/iOS/DuckDuckGoTests/NewTabPageFavoritesModelTests.swift
@@ -20,6 +20,9 @@
 import XCTest
 import Combine
 import Bookmarks
+import CoreData
+import Common
+import Persistence
 @testable import Core
 @testable import DuckDuckGo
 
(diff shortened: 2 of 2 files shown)


pikorddg (approved):
I wasn’t able to reproduce this before the fix, but given the number of occurrences for this pixel compared to the number of users, it’s not 100% reproducible. 

At the same time, I haven’t seen this pixel firing on this branch.

The code looks good to me -> approve.

PR #5545, opened 2026-06-25
macOS: Add suggestions to sidebar

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1215970469826760?focus=true
Tech Design URL:
CC:

### Description
Wires the native side to show suggestions in the sidebar.

### Testing Steps
1. Run the app
2. Open the Debug menu and go to AI Chat and tap on Web Communication and Set Custom URL
3. Add this URL as the new one → https://euw-serp-dev-testing20.duck.ai/
4. Open the Debug menu and set the Remote Configuration URL to be: https://webhook.site/token/c69ecef0-5f06-4974-b89c-c4f4733ae9eb/request/f443e04a-e594-48e5-b1b4-113322dc8f74/raw
5. Open https://www.imdb.com/title/tt32093575 and then the sidebar
6. You should see suggestions similar to the image

movie

### Impact and Risks
Low: Minor visual changes, most of fixes are done in the FE.

#### What could go wrong?
Nothing specific, the code is behind a FF.

### Quality Considerations
Nothing specific.

### Notes to Reviewer
FE work is still ongoing.

—
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes when and how page context is collected and delivered (navigation finish, signals-only vs full attach, async user-action fetch), which could affect existing sidebar attach behavior; mitigated by feature flags and backward-compatible optional JSON fields.
> 
> **Overview**
> Adds **macOS native plumbing** so the Duck.ai sidebar can show **page-tailored suggestion chips**, behind **`sidebarSuggestedPrompts`** (with page context) and exposed to the web app as **`supportsSuggestions`**.
> 
> **Shared contract:** New **`AIChatPageTypeSignals`** and optional **`pageTypeSignals` / `attached`** on **`AIChatPageContextData`** let the FE tell full page content from **signals-only** payloads (metadata + JSON-LD/OG hints, no body) when auto-attach is off. **`getAIChatPageContext`** is now **async** and, on **`userAction`**, waits for a fresh collection when only signals are cached so tapping a suggestion or Ask-about-page can return attached content in the same response.
> 
> **Tab / sidebar behavior:** **`PageContextTabExtension`** drives signals-only collection when auto-attach is disabled, re-delivers context when the chat VC appears, and re-collects on **`navigationDidFinish`** so suggestions match the loaded page. Content-scope-scripts is bumped to **15.3.0**.
> 
> **Telemetry:** **`userDidSelectSuggestion`** metric and **`m_mac_aichat_suggestion_selected`** pixel with **`suggestionId`** and **`pageType`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a0be97a73817df56cfb9bb65a02658a5336a2c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
+++ b/macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
@@ -104,7 +104,7 @@ protocol AIChatUserScriptHandling: AnyObject {
     @MainActor func openAIChatLink(params: Any, message: UserScriptMessage) async -> Encodable?
     var aiChatNativePromptPublisher: AnyPublisher<AIChatNativePrompt, Never> { get }
 
-    func getAIChatPageContext(params: Any, message: UserScriptMessage) -> Encodable?
+    @MainActor func getAIChatPageContext(params: Any, message: UserScriptMessage) async -> Encodable?
     func getAIChatSelectionContext(params: Any, message: UserScriptMessage) -> Encodable?
     var pageContextPublisher: AnyPublisher<AIChatPageContextData?, Never> { get }
     var selectionContextPublisher: AnyPublisher<AIChatSelectionContextData, Never> { get }
@@ -277,7 +277,8 @@ final class AIChatUserScriptHandler: AIChatUserScriptHandling {
         messageHandling.getDataForMessageType(.nativePrompt)
     }
 
-    func getAIChatPageContext(params: Any, message: any UserScriptMessage) -> Encodable? {
+    @MainActor
+    func getAIChatPageContext(params: Any, message: any UserScriptMessage) async -> Encodable? {
         guard let payload: GetPageContext = DecodableHelper.decode(from: params) else {
             return nil
         }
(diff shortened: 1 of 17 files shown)


macOS/DuckDuckGo/Tab/TabExtensions/PageContextTabExtension.swift
```diff
@@ -416,8 +442,56 @@ final class PageContextTabExtension {
...
+              featureFlagger.isFeatureOn(.sidebarSuggestedPrompts),
+              case .url = content,
+              !isContextCollectionEnabled,
+              aiChatSessionStore.sessions[tabID]?.chatViewController != nil,
+              let pageContextUserScript else {
+            return
+        }
+        pendingSignalsOnlyCollection = true
+        pageContextUserScript.collect()
+    }
```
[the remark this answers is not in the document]
jotaemepereira: This is intended

macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
```diff
@@ -915,6 +924,15 @@ extension AIChatUserScriptHandler: AIChatMetricReportingHandling {
         case .userDidAcceptTermsAndConditions:
             handleTermsAccepted()
             completion?()
+        case .userDidSelectSuggestion:
+            pixelFiring?.fire(
+                AIChatPixel.aiChatSuggestionSelected(
+                    suggestionId: metric.suggestionId ?? "",
+                    pageType: metric.pageType ?? "none"
```
tomasstrba, line 969: `pageType` comes in as a plain string on purpose, so that a new page type from the FE won't break decoding. The catch is that we then forward it straight into the `aiChatSuggestionSelected` pixel, which declares a fixed enum (`recipe`…`none`, see `aichat_pixels.json5:116`). So the first time the FE adds a new type, we'll be sending a pixel value that isn't in its own enum.

Could we map it to the known set before firing — anything unrecognized falls back to `none`, or we add an `other` bucket? That keeps the metric inside its declared values.

macOS/DuckDuckGo/Tab/TabExtensions/PageContextTabExtension.swift
```diff
@@ -431,6 +504,18 @@ extension PageContextTabExtension: PageContextProtocol, TabExtension {
...
 }
 
+extension PageContextTabExtension: NavigationResponder {
+    /// Re-collect page context once the new page has finished loading. The `Tab.$content` trigger
+    /// fires at `didCommit` — before the page's markup (JSON-LD / og:type) is parsed — so page-type
+    /// signals would otherwise reflect the previous page and the sidebar's suggestions wouldn't
+    /// update on same-tab navigation. Mirrors Windows' `NavigationCompleted` re-collect.
+    func navigationDidFinish(_ navigation: Navigation) {
+        guard !isLoadedInSidebar else { return }
+        collectPageContextIfNeeded()
```
tomasstrba, line 525: Small one: the `requestSignalsOnlyCollectionIfNeeded()` call just below correctly checks that the chat VC exists, but `collectPageContextIfNeeded()` here only checks for a URL plus the auto-collect setting — there's no "is the sidebar open" check. So with auto-collect on, we now start a full page-context collection on every page load even when the sidebar is closed.

Not a correctness issue — `handle()` just caches and skips the push when there's no VC — but it is extra work per navigation that we weren't doing before (collection used to ride off `didCommit`). Is that intended? If not, gating it on an active session would skip the collection for tabs with no sidebar open.

tomasstrba (approved):
This is really great feature, nice work! 👏 

Small nitpick: Sometimes the suggestions did not refresh for me when I used back/forward buttons. They refreshed only when I manually refreshed the site.
https://github.com/user-attachments/assets/dfc22736-8ed6-49e5-9e34-6f045ed2ab58

Would be great to solve before merging

macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
jotaemepereira: @tomasstrba Why would we bucket it to an `other` parameter? With the current approach, we just sent the real value; then we will get the failure pixel, but I would prefer that instead of getting the real value. Let’s keep it as it is.

PR #5465, opened 2026-06-18
iOS: Move focused-omnibar logo + fire empty-state to SwiftUI

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215191426890192?focus=true

### Description

Moves the iPhone focused-omnibar empty state — the Dax↔Duck.ai logo and the fire-tab screen — off the UIKit `DaxLogoManager` onto the SwiftUI Unified Toggle Input (UTI) suggestions host, and fixes the host frame so content and favorites animate cleanly.

**Fixed host:** the suggestions host now has a fixed, full-height frame; the bar-height offset rides the host's safe-area inset (so the list scrolls behind the bar and favorites slide instead of snapping), and the escape hatch + Duck.ai sync-promo are pinned to the bar so they ride its UIKit animation in the same layout pass.

**Logo → SwiftUI:** `FocusedDaxLogoView` (native `Lottie.LottieView`, same asset as the NTP logo) renders the Dax↔Duck.ai morph as a content state in the host. Its mark/morph/speed live in a pure, unit-tested `FocusedLogoModel`; the content resolver stays binary. Position is screen-anchored to match the NTP logo, kept clear of the top hatch and bottom bar by a minimum-gap constraint.

**Fire empty state → SwiftUI:** the fire-tab screen (`FireModeEmptyStateView`) renders as a host content state too.

**Result:** the UTI path no longer touches `DaxLogoManager` at all — it's now used only by the legacy (non-UTI) omnibar editing state, `OmniBarEditingStateViewController`, so it can be deleted cleanly once that path is removed.

Also fixes along the way: the Duck.ai clear→recents logo flash (recents existence is now a stable fact, like search favorites), focused↔NTP hatch alignment on the bottom bar, and the sync-promo-close reflow gap. iPhone only (iPad uses the popover path).

### Testing Steps

On an **iPhone**, with the address bar set to **both top and bottom** (test each):

1. **Logo focus/dismiss:** From a new tab page with no favorites, focus the omnibar → the Dax logo appears in place (no flash/double). Toggle Search↔Duck.ai → the logo morphs. Tap the back arrow → it morphs/fades back to the NTP logo.
2. **Favorites:** With favorites present, focus → favorites show and slide (don't snap); dismiss → they slide back.
3. **Scroll-behind:** With a long suggestions list / recents, scroll → content passes behind the bar and is hidden under the escape hatch, not clipped above it.
[...]

PR #5570, opened 2026-06-26
iOS: Show after-idle RMF message in focused-omnibar mode

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1213881875984009/task/1216034379142824?focus=true

### Description

Fixes a regression where the NTP-after-idle RMF message no longer shows when the omnibar is **focused** (UTI focused mode), while the escape hatch still does. A standard (no-trigger) RMF message renders fine in focused mode — only the after-idle-triggered one was dropped.

### Testing Steps
1. Got to Settings -> General -> Make sure after idle is set to NTP and  idle time is none
2. From debug menu -> Configuration URL ->  override remote message config with https://pastebin.com/raw/kHdXWw8T
3. Debug menu -> Remote messaging ->  Press delete all and then Refresh config
4. Open a site and then background and foreground the app
5. check the RMF is visible in both focussed and unfocussed mode
6. Open a new tab manually and check the RMF is not visible.

### Impact and Risks
Low. Additive signal threaded through the existing escape-hatch path; the standalone NTP path is unchanged. The embedded focused NTP and standalone NTP share a messages configuration — both fetch with the same after-idle signal, so no divergence.

#### What could go wrong?
- If the user has **no favorites and no message yet in the shared config**, focused content can resolve to the `.logo` state, which has no message layer — out of scope for this fix (separate surface gap).

### Notes to Reviewer
- **iOS-app-only** change (no framework). The focused-mode rendering can't be verified from CLI — verified in the simulator.
- Regression most visibly introduced by #5465 (opaque focused content).

### DIFF
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -1615,7 +1615,7 @@ class MainViewController: UIViewController {
         let model = escapeHatchModelBuilder.makeTabSwitcherOnly(targetTab: currentTab, router: self)
         controller.setEscapeHatch(model)
         currentNTPEscapeHatch = model
-        configureUnifiedInputEscapeHatch(model)
+        configureUnifiedInputEscapeHatch(model, openedAfterIdle: currentTab.openedAfterIdle)
     }
 
     /// True when an escape hatch action runs in focus mode (reads the persistent omnibar-session state, which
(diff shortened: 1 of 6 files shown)


aataraxiaa (commented):
Left comment in Asana. RMF is not displayed on focus.

aataraxiaa:
Just closing this @SabrinaTardio and @Bunn has merged the fix.

PR #5596, opened 2026-06-29
iOS: Expand (and smarter) suggested prompts for contextual sheet/sidepanel - part 1

pikorddg:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1210947754188321/task/1213651518796427?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1210947754188321/task/1216117537545343?focus=true
CC:

### Description
**Part 1 — infrastructure only. The feature is not expected to be fully functional yet; full end-to-end suggestions behaviour (and its composition with the `aiChatContextualUnifiedToggleInput` path) is a follow-up.**

Groundwork for contextual suggested prompts on the contextual sheet: page-tailored prompt suggestions instead of only the static quick action. Includes
- suggestion matcher and bundled catalog,
- collecting page-type-signals,
- loading/loaded state

All of it sits behind the new `contextualSuggestedPrompts` feature flag, which is **disabled by default**.

### Testing Steps

These are **regression checks**, not feature-works checks. The suggestions feature is not expected to work end-to-end in this PR. Verify that (a) with the new flag off, behaviour is identical to before, and (b) the existing unified toggle input path is unaffected.

The easiest way: run this PR on simulator and Alpha build from the device. Compare UI after each step.

### 1.  `unifiedToggleInput` ON / automatic context collection ON
1. Open Contextual Chat
2. Tap Summarize Page
3. Send second prompt
4. Behavior is as in Alpha build

### 2.  `unifiedToggleInput` ON / automatic context collection OFF
1. Open Contextual Chat
2. Tap Ask about page
3. Tap Summarize Page
4. Send second prompt
5. Behavior is as in Alpha build

### 3.  `unifiedToggleInput` OFF / automatic context collection ON
1. Open Contextual Chat
2. Tap Summarize Page
3. Send second prompt
4. Behavior is as in Alpha build

### 4.  `unifiedToggleInput` OFF / automatic context collection OFF
1. Open Contextual Chat
2. Tap Ask about page
3. Tap Summarize Page
4. Send second prompt
5. Behavior is as in Alpha build

### 5.  `aiChatContextualUnifiedToggleInput` ON / automatic context collection ON
1. Open Contextual Chat
2. Tap Summarize Page
3. Send second prompt
4. Behavior is as in Alpha build

### 6.  `aiChatContextualUnifiedToggleInput` OFF / automatic context collection OFF
1. Open Contextual Chat
2. Tap Ask about page
3. Tap Summarize Page
4. Send second prompt
5. Behavior is as in Alpha build

### Compare with https://github.com/duckduckgo/apple-browsers/pull/5660/
7. Step 4. New First-Sheet UTI: Automatic Attach ON
8. Step 5. New First-Sheet UTI: First Submit
9. Step 3. New First-Sheet UTI: Manual Attach, Automatic Attach OFF

### Impact and Risks

Medium. The new feature is disabled by default; with the flag off, behaviour is byte-for-byte legacy. But, changes are in few context-sheet related files.

#### What could go wrong?

The `origin/main` merge restructured shared contextual-sheet coordinator/session-state methods used by both features → mitigated by the flag-off invariant (every addition is gated by `contextualSuggestedPrompts` or a helper that no-ops when it is off, verified hunk-by-hunk) plus the unified-toggle-input verification above.

### Quality Considerations

Privacy: the suggestions start surface uses a **signals-only** payload — page content is stripped, keeping only metadata and page-type signals — so no page content is attached while resolving suggestions. Feature is disabled by default.

Note to reviewer: the runtime enabler for page-type signals in local/test builds is a mock (`PageTypeSignalsConfigInjector` in `UserScripts.swift`) that is **intentionally not part of this PR**; production signal enablement comes via remote privacy config. Do not expect the feature to produce suggestions in this build.

### Notes to Reviewer

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches contextual sheet coordinator, session state, and async page-context bridge paths shared with UTI/context attach, though new behavior is mostly gated behind `contextualSuggestedPrompts`.
> 
> **Overview**
> Adds **page-tailored suggested prompt chips** on the iOS contextual Duck.ai sheet start surface, behind the new remote flag **`contextualSuggestedPrompts`** (off by default). Part 1 wires matcher, catalog, loading UI, and sheet/session plumbing; full tap-to-chat behavior is still partial.
> 
> A bundled **`PageSuggestionsCatalog.json`** plus **`ContextualSuggestionsMatcher`** pick up to four prompts from JSON-LD, `og:type`, and domain (with localized copy, translate-when-different-language, and a summarize fallback). **`AIChatContextualChatSessionState`** drives **loading/loaded** suggestion state, resolves prompts after context collection (with timeout), and when auto-attach is off pushes **signals-only** page context (metadata + page-type signals, no body text) so the web FE can tailor suggestions without attaching content.
> 
> The native start UI merges suggestions and quick actions via **`ContextualSheetAction`**, shows a pill **`AIChatSuggestionsLoadingView`** through **`AIChatQuickActionsView.setLoading`**, and adjusts quick-action chips when the flag is on. Bridge updates expose **`supportsSuggestions`** in native config, make **`getAIChatPageContext`** **async** (including timed fresh collection for FE attach), and forward **chat status** so the sheet can **delay revealing the webview** until the first prompt leaves the idle start surface (with a fallback timer). Tapping a suggestion chip currently **only prefills the input**; submit-from-suggestion is deferred.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab5ae51f1e0be67de5b9bdd637d294d28c483daf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -304,6 +323,35 @@ private extension AIChatContextualSheetCoordinator {
...
+        }
+
+        // Await with `timeout` so this FE bridge call never hangs if the JS collection never responds
+        // (web view torn down, navigation mid-request, JS error). On timeout the push still delivers.
+        return try? await withTimeout(timeout) {
+            for await result in firstResult {
+                return result
+            }
+            return nil
+        }
```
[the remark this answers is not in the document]
pikorddg: Addressed: collectFreshContextAndWait now wraps withTimeout in do/catch and calls cancelManualAttach() on timeout/cancellation. It's idempotent, and the value-nil paths already clear the flag via updateContext → cleanupFlags, so only the timeout/cancel case is affected.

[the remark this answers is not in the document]
pikorddg: The result is now correlated by page URL: collectFreshContextAndWait only accepts an emission whose contextData.url matches the current didFinish URL (via URL.matches). A concurrent collection for a different page is rejected, so we no longer return wrong-page context inline.

Full per-request correlation isn't achievable natively: collect() is a fire-and-forget broker.push and the JS collectionResult carries no request ID. Threading one through would be a cross-platform contract change. The same wait pattern ships on macOS (PR # 5545), where the equivalent race was reviewed and accepted; iOS is strictly tighter since it additionally correlates by page URL.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -162,6 +170,10 @@ final class AIChatContextualSheetCoordinator {
         startObservingContextUpdates()
 
         if sessionState.shouldAutoCollectContext {
+            sessionState.beginLoadingSuggestions()
```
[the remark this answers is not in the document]
pikorddg: Outdated — addressed after the reviewed commit. beginLoadingSuggestions() arms startSuggestionsTimeout() regardless of triggerContextCollection()'s result; after contextualContextCollectionTimeout (5s) it calls resolveSuggestionsIfLoading(from: nil), which resolves to default suggestions and clears the loading state. Worst case is a 5s fallback, not an indefinite spinner.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -442,12 +530,39 @@ private extension AIChatContextualChatSessionState {
...
+
+        let input = ResolvePageSuggestionsInput(
+            pageTypeSignals: context?.contextData.pageTypeSignals,
+            url: context?.contextData.url,
+            uiLocale: Locale.current.identifier
+        )
+
+        suggestionsResolveTask?.cancel()
+        suggestionsResolveTask = Task { [weak self] in
+            guard let resolved = await self?.suggestedPromptsProvider.resolveSuggestions(input) else { return }
```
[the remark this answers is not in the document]
pikorddg: I’ll take care of in https://app.asana.com/1/137249556945/project/1210947754188321/task/1216071255710237?focus=true

(Same-tab navigation suggestions handling). Root cause is that resolveSuggestionsIfLoading resolves on the first .loading context with no URL correlation (unlike the manual-attach path), so a stale previous-page collection can win the race. The follow-up should add URL correlation to the resolve, not just re-trigger collection on nav.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -304,6 +331,46 @@ private extension AIChatContextualSheetCoordinator {
...
+    func collectFreshContextAndWait(timeout: TimeInterval) async -> AIChatPageContextData? {
+        let expectedURL = latestDidFinishURL
+        let firstResult = AsyncStream<AIChatPageContextData?> { continuation in
+            let cancellable = pageContextHandler.contextPublisher
+                .dropFirst()
+                .first(where: { context in
+                    guard let expectedURL,
+                          let context,
+                          let emittedURL = URL(string: context.contextData.url) else { return true }
+                    return emittedURL.equals(expectedURL, by: .fuzzyIdentity)
```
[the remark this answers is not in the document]
pikorddg: Fixed in https://github.com/duckduckgo/apple-browsers/pull/5596/commits/ebb3beba14047b7c94f1777c401e2910323bca1e

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -219,6 +242,20 @@ final class AIChatContextualChatSessionState {
         emit(.submitPrompt(prompt: prompt, context: contextData))
     }
 
+    /// Call when the FE submits a prompt on its own (e.g. a tapped suggestion on the
+    /// suggestions start surface) — transition to the chat view without re-emitting a submit, since
+    /// the FE has already submitted. No-op once a chat is active.
```
aataraxiaa: I don’t understand this comment. When can FE submit a prompt on suggestion tap? Is this leftover from previous web approach?

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -371,8 +414,44 @@ final class AIChatContextualChatSessionState {
...
+    func beginLoadingSuggestions() {
+        guard featureFlagger.isFeatureOn(.contextualSuggestedPrompts), !hasActiveChat else { return }
+        suggestionsResolveTask?.cancel()
+        suggestions = []
+        suggestionsLoadState = .loading
+        rebuildViewState()
+        startSuggestionsTimeout()
+    }
+
+    private func startSuggestionsTimeout() {
```
aataraxiaa: Minor but can we move this to the private extension of this type below.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -430,12 +494,21 @@ private extension AIChatContextualSheetCoordinator {
...
+                    return nil
                 }
-                return nil
+
+                if let cached = self.sessionState.latestContext?.contextData,
+                   cached.attached != false, !cached.content.isEmpty {
+                    return cached
+                }
+                self.sessionState.beginManualAttach(fromFrontend: true)
+                return await self.collectFreshContextAndWait(timeout: Self.contextualContextCollectionTimeout)
```
aataraxiaa, line 511: Will this mean context is delivered to the FE twice (also via `submitPageContext`)?

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualWebViewController.swift
```diff
@@ -66,7 +66,10 @@ final class AIChatContextualWebViewController: UIViewController {
     private let debugSettings: AIChatDebugSettingsHandling
     private let userAgentManager: UserAgentManaging
     private let utiHostInstaller: ((AIChatContextualWebViewController) -> AIChatContextualUTIHost?)?
+    private let deferUTIInstall: Bool
```
aataraxiaa: I _think_ this is because the web VC is created/loaded while the native suggestions UI is still showing, so any web-VC UTI setup should wait until we actually transition to the web view. In the immediate UTI flow the UTI is already mounted at sheet level, while in the non-immediate flow this avoids installing/expanding it inside a hidden web VC. Is this correct? Why do we need to defer the UTI install in the web UTI case?

Basically...`deferUTIInstall` was a little hard for me to reason about.

iOS/DuckDuckGo/AIChat/ContextualMode/ContextualSheetAction.swift
```diff
@@ -0,0 +1,66 @@
+//
+//  ContextualSheetAction.swift
+//  DuckDuckGo
+//
+//  Copyright © 2025 DuckDuckGo. All rights reserved.
```
aataraxiaa: New files should be 2026, same for files below.

aataraxiaa (commented):
Mostly looks good 👍🏼. Left a few comments

iOS/DuckDuckGo/AIChat/ContextualMode/ContextualSuggestionsMatcher.swift
```diff
@@ -0,0 +1,263 @@
...
+        }
+        if contextual == nil, let hostname = hostname(from: input.url) {
+            contextual = matchByDomain(hostname, catalog.byDomain)
+        }
+
+        // "Summarize this page" is a generic fallback: when page-tailored suggestions matched, drop it
+        // so it is only offered when nothing page-specific did. It is unconditional, so it is always
+        // present otherwise — this is the never-empty floor for the start surface.
+        let defaults = contextual != nil ? catalog.defaults.filter { $0 != summarizePageID } : catalog.defaults
+        return (contextual ?? []) + defaults
```
aataraxiaa, line 112: Due to max suggestions limit I think it’s possible that translate would not be displayed for pages in another language. Maybe translate needs to take higher priority. Could be a functional test in subsequent PR.

aataraxiaa (commented):
One more comment

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
pikorddg: No, since we use native rendering,  FE requests page context with reason=.other, which we answer via the submitPageContext, single delivery.

The collectFreshContextAndWait, reason=.userAction exists to align with agreed [FE contract](https://app.asana.com/1/137249556945/task/1215655020578492/comment/1216075456188840?focus=true) an explicit .userAction request gets a direct response instead of null + subscription, while subscriptions stay for everything else (non-breaking). 

If that path were hit, the same collection could deliver both inline and via push, but from my understanding that's the accepted trade-off in the contract and matches macOS/Windows. iOS just doesn't exercise it today, so there's no double delivery.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualWebViewController.swift
pikorddg: Good catch, thank you! It was leftover from the web-view hack phase, so I removed it. 
During testing I found that Duck.ai logo could blink with contextualSuggestedPrompts ON, so I added fix to delay revealing the web view until the frontend reports an active response state via getResponseState. It's gated to the contextualSuggestedPrompts path to avoid changing the existing flow. https://github.com/duckduckgo/apple-browsers/pull/5596/commits/c30ffeadf40a9fadf093cb4ead4bd3b0036718c2

iOS/DuckDuckGo/AIChat/ContextualMode/ContextualSuggestionsMatcher.swift
pikorddg: Good catch! Makes sense, it's worth prioritising. Added to my to-do list!

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
pikorddg: https://github.com/duckduckgo/apple-browsers/pull/5596/commits/6f086941fbe99fc580c38f0912afc7d1d6edaa0a

Good catch, it’s a leftover, thanks!

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
pikorddg: https://github.com/duckduckgo/apple-browsers/pull/5596/commits/60b3bfc87e601ce0056b8624f778a67e1313e906

iOS/DuckDuckGo/AIChat/ContextualMode/ContextualSheetAction.swift
pikorddg: https://github.com/duckduckgo/apple-browsers/pull/5596/changes/07f12ecc65e9d48f184e04d21f72c237dfb6c281

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetViewController.swift
```diff
@@ -938,6 +965,43 @@ private extension AIChatContextualSheetViewController {
...
+
+private extension AIChatStatusValue {
+
+    var shouldRevealInitialPromptWebView: Bool {
+        switch self {
+        case .startStreamNewPrompt, .startStreamRestartStream, .loading, .streaming, .error, .blocked:
+            return true
+        case .ready, .unknown:
+            return false
+        }
```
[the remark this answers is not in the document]
pikorddg: https://github.com/duckduckgo/apple-browsers/pull/5596/commits/ab5ae51f1e0be67de5b9bdd637d294d28c483daf

aataraxiaa (approved):

PR #5642, opened 2026-07-02
[iOS] Return home when bailing out of onboarding subscription flow

jozsef-vesza:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1215840605898818
CC:

### Description

This PR changes the behavior of the subscription flow presented during onboarding.

Old behavior:
When users left the subscription flow without purchasing, we presented the settings root view, which is not ideal for a new user.

New behavior:
- When users navigate back from the flow without purchasing, we dismiss the modal so they end up in the browser home view.
- When users purchase a subscription during the flow, going back will present the settings root view (matches old behavior)

### Testing Steps
No purchase path:
1. Install the app and go through onboarding to reach the subscription promo (final step)
2. Dismiss the modal without purchasing -> Confirm that the Settings root view is not shown

Purchase path:
1. Install the app and go through onboarding to reach the subscription promo (final step)
2. Purchase a subscription and navigate back -> Confirm that the Settings root view is shown

### Impact and Risks
Low. Navigation-only change to the onboarding subscription entry point.

#### What could go wrong?
N/A

### Quality Considerations
N/A

### Notes to Reviewer
N/A

---

> [!NOTE]
> **Low Risk**
> Scoped to onboarding subscription navigation via origin query param and pop override; other Settings entry points unchanged.
> 
> **Overview**
> When onboarding opens Settings only to show the subscription purchase flow, **backing out without buying now dismisses the whole Settings modal** so the user returns to the browser home instead of the Settings root.
> 
> The flow is gated by a new **`isOnboardingSubscriptionFlow`** check on `SettingsDeepLinkSection` (subscription deep link whose redirect URL carries the onboarding funnel **origin**). That flag is passed into **`SettingsUINavigationController`**, which listens for **`.subscriptionDidChange`** and, on back from the pushed subscription screen (nav stack depth 2), **dismisses the modal** if no subscription was acquired; after a successful purchase, back still **pops to Settings** as before.
> 
> **`SettingsDeepLinkSectionTests`** cover onboarding vs other origins and non–subscription-flow sections. The Xcode project diff also reorders some build file entries and renames local Swift package reference labels (no behavior change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65dfb220f9722e6d25388c5ea45d06a638dd83e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/MainViewController+Segues.swift
+++ b/iOS/DuckDuckGo/MainViewController+Segues.swift
@@ -529,6 +529,9 @@ extension MainViewController {
 
                 // We are still presenting legacy views, so use a Navcontroller
                 let navController = SettingsUINavigationController(rootViewController: settingsController)
+                // When Settings is opened purely to host the onboarding subscription purchase, backing out
+                // without buying should return home rather than land on Settings (see the override).
+                navController.dismissesModalOnSubscriptionBailout = deepLinkTarget?.isOnboardingSubscriptionFlow ?? false
                 navController.navigationBar.tintColor = UIColor(designSystemColor: .textPrimary)
                 settingsController.modalPresentationStyle = UIModalPresentationStyle.automatic
                 // Opaque nav bar and matching view background so sheet top gap (if any) is visually continuous with the bar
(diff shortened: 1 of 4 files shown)


iOS/DuckDuckGo/MainViewController+Segues.swift
```diff
@@ -625,19 +628,50 @@ extension MainViewController {
...
         NotificationCenter.default.post(name: .settingsDidDisappear, object: nil)
     }
 
+    override func popViewController(animated: Bool) -> UIViewController? {
+        // Bail out to home instead of popping to the Settings root
+        // when leaving the onboarding subscription flow without a purchase.
+        if dismissesModalOnSubscriptionBailout,
+           viewControllers.count == 2,
+           !didAcquireSubscription {
+            dismiss(animated: true)
```
hanyutang-sandra, line 669: This leaves Settings still in the navigation stack, so long pressing the back button still reveals Settings. Ideally settings should be removed from the stack completely. Edge case tho but lemme know your thoughts

hanyutang-sandra (approved):

iOS/DuckDuckGo/MainViewController+Segues.swift
jozsef-vesza: Thanks @hanyutang-sandra yes, that is right. The 100% correct solution is to:
* Present the sign-up flow in its own modal -> can be dismissed without issues if no purchase
* If we detect a purchase, dynamically add the Settings view to the stack -> can navigate back to Settings (keep the existing behavior)

I played around with it for a bit, and IMO it's quite hacky, and I don't think this edge case warrants it, so I'd keep it as is for now.

PR #5661, opened 2026-07-03
Register Arabic (ar-SA) locale for iOS App Store metadata

pballart:
Task/Issue URL: https://app.asana.com/0/0/1216262151975278/f

### Description

Registers `ar-SA` (Arabic) as an iOS App Store metadata locale by adding an empty locale folder.

Arabic was enabled as an App Store localization but had no `ar-SA` folder in `iOS/fastlane/metadata`. `deliver` builds its `enabled_languages` set from the locale folders present on disk, then applies `default/release_notes.txt` only to those languages — so every other locale got release notes but Arabic was skipped entirely, leaving its "What's New" empty and blocking submit-for-review (`Arabic - What's New in This Version - This field is required`).

The folder is intentionally **empty** (just a `.gitkeep`) rather than carrying its own `release_notes.txt`:

- `deliver` keys `enabled_languages` off folder *existence*, so an empty folder is enough to bring `ar-SA` into scope.
- Any locale with no local `release_notes` inherits `default/release_notes.txt` — the same way every other locale folder works (none of them ship a `release_notes.txt`).
- `default/release_notes.txt` is updated every release, so a hardcoded `ar-SA` copy would drift and ship stale notes. Keeping the folder empty means Arabic always tracks the shared default.
- Arabic's translated description/keywords in App Store Connect are untouched — there's no `default` for those fields, so `deliver` won't overwrite them.

### Testing Steps

1. Run the **iOS – Promote TestFlight to App Store** workflow (or `bundle exec fastlane upload_metadata`).
2. In App Store Connect, confirm Arabic now has "What's New in This Version" populated from the default release notes and submit-for-review no longer flags it.

### Impact and Risks

Low — metadata-only change, no code/binary impact. Unblocks App Store submission when Arabic is an enabled localization.

---

> [!NOTE]
> **Low Risk**
> Metadata-only folder addition with no runtime or binary changes; only affects App Store metadata upload behavior for Arabic.
> 
> **Overview**
> Adds an **`ar-SA`** locale directory under **`iOS/fastlane/metadata`** so Fastlane **`deliver`** treats Arabic as an enabled on-disk language and applies the shared **`default/release_notes.txt`** to App Store “What’s New,” matching other non-English locales.
> 
> Previously Arabic could be enabled in App Store Connect without a matching folder, so release notes were never uploaded and submit-for-review failed on the required Arabic “What’s New” field. The change is metadata-only (placeholder **`.gitkeep`**); it does not alter app binaries or translated store descriptions managed in Connect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5af5f313905e00db5215cdc9ac9bd27bb1d43b9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pballart:
Not needed. Arabic should be removed as a supported language

PR #5693, opened 2026-07-07
Make AIChat model/cookie fetch chain @MainActor to fix iOS 15 SIGTRAP

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1216258354437824?focus=true

### Description

This PR marks the whole chain (AIChatCookieProviding → WKHTTPCookieStoreProvider → AIChatModelsProviding → AIChatModelsService) @MainActor, so it is now impossible to reach allCookies() from off the main actor. The hop, which caused the exception, is removed. 

I also swapped the raw WKHTTPCookieStore for the existing @MainActor DDGHTTPCookieStore abstraction, which finally makes the cookie-filtering logic testable.

### Testing Steps
1. Open the AI Chat omnibar on macOS or iOS (`onOmnibarActivated → fetchModels → cookies(for:)`) → Duck.ai model list loads with no crash.
2. `AIChatOmnibarControllerTests` / `NewTabPageOmnibarModelsProviderTests` (macOS) and `UTIModelStoreTests` / `IPadOmnibarModelPickerControllerTests` / `IPadOmnibarReasoningPickerControllerTests` / `IPadOmnibarToolPickerControllerTests` (iOS) all pass unchanged.

### Impact
Low — isolation-only refactor of a single call chain; no behavior change for any existing caller.

#### What could go wrong?
- The iOS-15-specific crash can't be reproduced locally (no deterministic repro exists, even on-device) → mitigated by the compile-time actor-isolation guarantee: the compiler statically proves there's no cross-actor call anywhere between any caller and `allCookies()`, rather than relying on a runtime hop working correctly.

### Quality Considerations
The `AIChatTests` SPM test target (`AIChatModelsServiceTests`, plus a new `WKHTTPCookieStoreProviderTests` added here to cover the previously-untestable cookie domain-filtering logic) isn't wired into either app's test plan/scheme — a pre-existing gap, unrelated to this change. Both files were hand-verified against the real APIs; worth wiring up separately.

---

> [!NOTE]
> **Low Risk**
> Isolation and abstraction swap on a single omnibar fetch path; intended behavior unchanged, with compile-time guarantees against cross-actor WebKit cookie access.
> 
> **Overview**
> Fixes an **iOS 15 SIGTRAP** when loading Duck.ai models by making the full path **`AIChatCookieProviding` → `WKHTTPCookieStoreProvider` → `AIChatModelsProviding` / `AIChatModelsService`** **`@MainActor`**, so **`allCookies()`** is never reached from a background actor.
> 
> **`WKHTTPCookieStoreProvider`** now depends on **`DDGHTTPCookieStore`** (`WKAbstractions`) instead of raw **`WKHTTPCookieStore`**, with **`nonisolated`** initializers where defaults still touch WebKit. **`HTTPCookieStoreWrapper`**’s initializer is **`nonisolated`** for the same pattern.
> 
> Adds **`WKHTTPCookieStoreProviderTests`** for host/domain cookie filtering via **`MockHTTPCookieStore`**, wires **`WKAbstractions`** into the **AIChat** SPM target and tests, and marks related **`AIChatModelsProviding`** test doubles **`@MainActor`** on iOS and macOS.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 958f8d13d954a565637f02c1fa84436a0488a93c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/AIChat/Package.swift
+++ b/SharedPackages/AIChat/Package.swift
@@ -65,6 +65,7 @@ let package = Package(
                 .product(name: "PrivacyConfig", package: "BrowserServicesKit"),
                 .product(name: "UserScript", package: "BrowserServicesKit"),
                 .product(name: "DuckAiDataStore", package: "BrowserServicesKit"),
+                .product(name: "WKAbstractions", package: "BrowserServicesKit"),
                 .product(name: "DDGSyncCrypto", package: "sync_crypto")
             ],
             resources: [
@@ -92,7 +93,8 @@ let package = Package(
                 "AIChatTestingUtilities",
                 .product(name: "BrowserServicesKitTestsUtils", package: "BrowserServicesKit"),
                 .product(name: "PersistenceTestingUtils", package: "BrowserServicesKit"),
-                .product(name: "PrivacyConfigTestsUtils", package: "BrowserServicesKit")
+                .product(name: "PrivacyConfigTestsUtils", package: "BrowserServicesKit"),
+                .product(name: "WKAbstractions", package: "BrowserServicesKit")
             ]
         )
     ]
--- a/SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelsService.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelsService.swift
@@ -19,17 +19,20 @@
 import Foundation
 import os.log
 import WebKit
+import WKAbstractions
 
 // MARK: - Cookie Providing
 
+@MainActor
 public protocol AIChatCookieProviding {
     func cookies(for url: URL) async -> [HTTPCookie]
 }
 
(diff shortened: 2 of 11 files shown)


alessandroboron (approved):
@tomasstrba I reviewed in isolatione commit dbf3e64bc1d246ce6f93587c1c7e94f08ac7916b

I believe the branch was created off main but the PR is targeting branch 
`release/ios/7.228.0`. To avoid bringing in commits that should not belong to this release I recommend rebasing dbf3e64bc1d246ce6f93587c1c7e94f08ac7916b on top of the release branch and discard all other commits.

I will approve it in the meantime

PR #5695, opened 2026-07-07
Migrate persistent iOS VPN tunnel pixels to PixelKit

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203108348835387/task/1216331332505329?focus=true

### Description

Persistent VPN tunnel events still use a separate legacy sender and retry queue. This moves those events to PixelKit's retry support so failed sends use the same delivery system as the other tunnel telemetry.

The migration preserves complete pixel names, including `_d`/`_c` and `_ios_phone`/`_ios_tablet`, and keeps retries limited to the 13 existing persistent events. The legacy queue continues draining so events queued by older builds can still be delivered.

### Testing Steps

1. Run the `PixelKit` test suite.

   Expected: The tests pass.

2. Build the `iOS VPN Extension` scheme.

   Expected: The extension builds.

3. Enable the VPN on a physical device.

   Expected: The tunnel connects.

4. Simulate a failed send for a retry-enabled tunnel event.

   Expected: Its daily and count variants enter the PixelKit retry queue with their existing names.

5. Restore pixel delivery.

   Expected: A successful pixel triggers replay of the queued events with their original names and error parameters.

Local verification:

```text
Retry-focused selection: 15 tests, 0 failures
Full PixelKit suite: 204 tests, 20 assertion failures
build succeeded — ios scheme 'iOS VPN Extension'
```

The new two-leg test and existing retry-queue tests pass. The full-suite failures are in error-parameter and naming/storage assertions outside the new test; they remain unresolved.

### Impact and Risks

Initial sends no longer carry `originalPixelTimestamp`. PixelKit adds the failure timestamp and `retriedPixel` only on replay. Local definitions for these legacy events and acceptance of the retry metadata remain unverified; confirm that contract before merging.

The existing daily-state migration runs on app-version changes. Installing a build with the same app version after migration has already run can leave stale daily state and send an extra daily variant that day.

### Notes to Reviewer

The shared retry test verifies that `.legacyDailyAndCount` with `.withRetry` queues and replays both frequency variants exactly once. It reuses PixelKit's existing test doubles and does not exercise the private provider event mapping. A separate call-site audit checks that only the 13 persistent calls change and that their events, errors, and frequency mappings remain equivalent.

---

> [!NOTE]
> **Medium Risk**
> Changes production VPN extension analytics delivery and retry semantics (e.g. retry metadata and timestamps on replay); telemetry contract should be validated before merge.
> 
> **Overview**
> Moves **13 VPN packet-tunnel telemetry events** that previously went through `persistentPixel.fireDailyAndCount` onto **`PixelKit.fire`** with `frequency: .legacyDailyAndCount` and **`options: .withRetry`**, so failed sends use PixelKit’s retry queue instead of the separate persistent pixel path. Failure cases now use **`Pixel.Event…withError(error)`** instead of passing errors into the legacy API.
> 
> Coverage spans rekey, tunnel start/update, server migration, and connection-failure-loop events. **`persistentPixel` remains only for `sendQueuedPixels`** so older builds’ queued events can still drain.
> 
> Adds **`testWhenLegacyDailyAndCountFailsWithRetryThenBothLegsAreQueuedAndReplayed`**, which asserts that a failed `.legacyDailyAndCount` + `.withRetry` fire queues both `_c` and `_d` legs and replays them (with `retriedPixel`) after a later successful pixel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f85e46d6ed9ec4ac1c18828cb7b44394615d111e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


federicocappelli (changes requested):
Hey @diegoreymendez , awkward thing first: I've already done this conversion in [duckduckgo/apple-browsers#6599](https://github.com/duckduckgo/apple-browsers/pull/6599)

Phase 3 deletes PersistentPixel and moves all 13 tunnel call sites to .legacyDailyAndCount + .withRetry, so we'd conflict. My bad for not flagging it sooner. Happy to drop it from phase 3 if you'd rather land yours, just decide.

On the tests, two things:

RetryQueueStore and FireRequestRecorder already exist as MockPixelRetryQueueStore and FireRequestMock in PixelRetryQueueTestDoubles.swift, and they're a bit richer (injectable errors, deferred completions). You couldn't get at them because they live in PixelKit's test target rather than TestingSupport/, which is also why you needed @testable import PixelKit. Shall we just promote them to TestingSupport as @_spi(Testing) public? We can do it in this PR, and it unblocks anyone else who wants retry coverage from iOS or macOS.

Also, most of these assertions are already in PixelRetryQueueTests: not-opted-in isn't queued, failures queue with a timestamp, replays send and get removed, a successful fire drains the queue. The two copies will drift.

The new bit here is multi-leg retry, nothing covers .legacyDailyAndCount with .withRetry queueing and replaying both legs. Worth keeping, and since it's platform-independent it'd slot straight into PixelRetryQueueTests with the existing doubles.

federicocappelli (approved):
LGTM

PR #5718, opened 2026-07-08
macOS: Gate Duck.ai native omnibar by subscription tier

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1215275657171787?focus=true

### Description

Brings the macOS native Duck.ai omnibar (address bar) to parity with iOS/web for subscription-tier gating of AI models and reasoning-effort options.

- **Fixes a real bug:** a user could pick a model / reasoning effort their plan doesn't include and only find out via a server error on submit. Gated options are now surfaced up front instead of failing silently.
- **Model picker:** accessible models first, then a "Subscriber Exclusive" (free users) / "Pro Exclusive" (Plus users) section listing the gated models with a PLUS/PRO tag, led by a "Try for Free" / "Upgrade" badge.
- **Reasoning-effort picker:** a gated effort (e.g. Extended Reasoning) shows the same "Try for Free" / "Upgrade" badge.
- **Confirmation dialog:** tapping a gated row opens a dialog explaining the upsell before routing to the purchase/upgrade page — neither picker navigates away silently. Copy adapts to the user: "Try for Free" for trial-eligible free users, "Upgrade" otherwise, and a distinct "Upgrade to Pro" dialog for existing Plus subscribers.
- **Anti-annoyance:** the yellow badge mutes to a neutral color after the user has seen it a few times (shared 4-view cap across both pickers), but stays in place and tappable.
- **Kill switch:** the `aiChatOmnibarSubscriptionUpsell` feature flag gates the upsell UI (tags + dialog) independently of the underlying access gating, so it can be turned off without reintroducing the original bug.

Purchase/upgrade links carry the address-bar funnel origin (`funnel_addressbar_macos__modelpicker` / `…__reasoningpicker`) + `featurePage=duckai`, matching iOS's address-bar attribution.

### Testing Steps

> Switching tiers: use a sandbox/real subscription (trial-eligible vs. ineligible is StoreKit/Apple-ID driven). The debug menu's AI Chat tier simulator was removed — see Notes to Reviewer.

**A. Reasoning-effort gating (the core fix)**
1. As a **free** or **Plus** user, select a model whose top reasoning effort is above your tier, open the reasoning picker.
   - *Expect:* the gated effort is shown with a "TRY FOR FREE" / "UPGRADE" tag on the right.
2. Click the gated effort row
   - *Expect:* the subscription  dialog opens;

**B. Model picker layout**
4. Open the model picker as a **free** user.
   - *Expect:* accessible models show first, then a "Subscriber Exclusive" header + "TRY FOR FREE"/"UPGRADE" tag, then gated models each with a PLUS or PRO tag.
5. Click an individual gated model row (below the header).
   - *Expect:* nothing happens — only the header row/badge is the entry point (matches duck.ai web).
6. Click the "Subscriber Exclusive" header row (anywhere) or its badge.
   - *Expect:* the confirmation dialog opens.

**C. Dialog copy per state**
7. **Free, trial-eligible:** primary button reads **"Try for Free"**; also shows "I Have a Subscription" and "Not Now". "Try for Free" → purchase page (`featurePage=duckai`, address-bar origin).
8. **Free, trial-ineligible:** same dialog, primary button reads **"Upgrade"**.
9. **Plus:** distinct **"Upgrade to Pro"** dialog (different title/body, **no** "I Have a Subscription" button); "Upgrade" → plan-change page.
10. In any dialog, click "Not Now".
    - *Expect:* dialog dismisses, nothing changes.

**D. Badge impression muting**
11. Open either picker ~5 times.
    - *Expect:* tag is yellow for the first 4 opens, then mutes to a neutral fill from the 5th on

**E. Kill switch**
12. Debug ▸ Feature Flag Overrides ▸ disable **"Ai Chat Omnibar Subscription Upsell"**, reopen both pickers.
    - *Expect:* gated rows fall back to a plain dimmed, non-interactive row with a PLUS/PRO label

**F. Pro user**
13. As a **Pro** user, open both pickers.
    - *Expect:* nothing gated anywhere

### Impact and Risks

Medium: gates a previously-broken flow behind correct UI and adds a new upsell surface, with a kill switch. Net risk is low — the underlying access checks (`entityHasAccess`, per-effort `isAccessible`) already existed and are additive here, and the flag can disable the new UI without touching gating.

**What could go wrong?**
- A tier mis-resolution could route to the wrong subscription page (purchase vs. upgrade) — covered by unit tests on the shared resolver and the macOS presenter for all tier combinations (free/plus/pro/internal).
- iOS now consumes the promoted shared types via `import AIChat` instead of local copies; its behavior is unchanged, verified by the migrated `ReasoningModeAccessResolverTests`.

---

> [!NOTE]
> **Medium Risk**
> Adds subscription upsell UI and changes model/reasoning selection and submit payloads on a high-traffic omnibar path; mitigated by existing access checks, a remote kill switch, and unit tests for tier routing.
> 
> **Overview**
> **macOS Duck.ai address-bar pickers** now show subscription-gated models and reasoning efforts up front instead of letting users select options that fail on submit. Accessible models stay first; gated ones appear in a subscriber section with PLUS/PRO labels and an optional **Try for Free** / **Upgrade** header badge.
> 
> Tapping a gated row (or the reasoning badge) opens a **confirmation dialog** with tier-aware copy before routing to purchase, upgrade, or subscription activation via new funnel origins and `featurePage=duckai`. **Submission** no longer sends reasoning efforts the user’s tier can’t use (e.g. after downgrade).
> 
> Shared **AIChat** gains `AIChatModelPublicAccessTier`, upgrade-flow routing, `groupByAccess` / recommended ordering, and public `ReasoningModeAccessResolver`; iOS drops duplicate types in favor of the package. **`aiChatOmnibarSubscriptionUpsell`** is a kill switch for tags and dialog only—gating stays on as dimmed, non-interactive rows when off. Badge impressions mute after a shared 4-view cap. New menu row UI, subscription dialog asset/colors, and an upsell pixel round out the change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2a0e6f15a546dd3a7c8fd42b8cb6727823e4100. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1683,42 +1697,123 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
+            menu.addItem(headerItem)
+            for gatedModel in gated {
+                menu.addItem(modelRow(
+                    for: gatedModel.model,
+                    trailingText: Self.tierBadgeText(for: gatedModel.requiredTier),
+                    isSelected: false,
+                    isDimmed: true,
+                    isInteractive: false,
+                    in: menu
+                ))
```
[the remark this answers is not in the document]
tomasstrba: Intentional, not a bug — this is a deliberate divergence from iOS, specifically requested during design review: gated model rows should be non-interactive to match the web client's model picker, where only the "Try for free"/"Upgrade" link routes to the subscription flow. The reasoning-effort picker (a separate control) does keep iOS's tap-to-upsell behavior for locked efforts, which is correct as-is.

My PR description's testing steps were worded ambiguously here ("tap a gated model or the link") — updated to make the intended behavior explicit.

macOS/DuckDuckGo/AIChat/AIChatPixel.swift
```diff
@@ -280,6 +280,10 @@ enum AIChatPixel: PixelKitEvent {
     /// Event Trigger: User selects a reasoning effort from the native omnibar picker
     case aiChatAddressBarReasoningEffortSelected
 
+    /// Event Trigger: User taps a gated model or reasoning effort in the native omnibar picker,
+    /// routing them to the subscription purchase/upgrade flow.
+    case aiChatAddressBarSubscriptionUpsellTriggered(currentTier: String, requiredTier: String, flowType: String)
```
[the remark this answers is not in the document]
tomasstrba: Good catch, thanks — fixed in 188d2f3ae0. Added `m_mac_aichat_addressbar_subscription_upsell_triggered` to `aichat_addressbar_pixels.json5` with the `current_tier`/`required_tier`/`flow_type` parameters as enums, validated via `npm run validate-pixel-defs`.

macOS/UnitTests/AIChat/AIChatOmnibarSubscriptionUpsellPresenterTests.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    @Test("Internal user has no gated flow and nothing is presented")
+    func internalUserHasNoGatedFlow() async throws {
+        let (presenter, mockTabShower, _) = createPresenter()
+
+        let routed = presenter.routeGatedSelection(requiredTier: .pro, userTier: .internal, origin: .addressBarReasoningPicker)
+
+        #expect(routed == false)
+        #expect(mockTabShower.capturedContent == nil)
+    }
```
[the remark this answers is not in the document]
tomasstrba: Thanks for flagging — checked this against the other Swift Testing suites in this codebase (e.g. `SubscriptionNavigationCoordinatorTests`) and none of them set an explicit `.timeLimit` trait for similarly fast, synchronous unit tests, so adding one here would be inconsistent with the established convention rather than fixing a real gap. Happy to revisit if there's a broader push to adopt explicit timeouts across the suite.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1734,31 +1846,122 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
+            boldTitle: effort.title,
+            regularTitle: "",
+            subtitle: effort.subtitle,
+            subtitleFontSize: 11,
+            trailingText: showsUpsellBadge ? nil : Self.tierBadgeText(for: requiredTier),
+            trailingBadgeText: badgeText,
+            emphasizesTitle: false,
+            isSelected: isSelected && !isGated,
+            isDimmed: isGated,
+            isInteractive: !isGated,
```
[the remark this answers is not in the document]
tomasstrba: Intentional, not a bug — this reflects a design-review decision from later in the PR (the reasoning picker's gated rows were changed to match the model picker's own gated-row treatment exactly: dimmed title/icon/subtitle, non-selectable row body). The lock-icon-only, full-color-title behavior this was checking against is what the PR description said at the time, but that's now stale relative to the code. I've rewritten the PR description to match the current, final behavior.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1734,31 +1846,122 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
+            action: #selector(reasoningEffortSelected(_:)),
+            badgeAction: showsUpsellBadge ? #selector(reasoningEffortBadgeSelected(_:)) : nil,
```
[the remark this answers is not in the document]
tomasstrba: Also intentional, same design-review decision as the sibling comment above — only the trailing "Try for free"/"Upgrade" badge is meant to be clickable now, via a dedicated `onTapBadge` path that's independent of the row body's `isInteractive` state. Tapping the row body itself is a no-op by design: the row is disabled (dimmed, non-selectable), matching how a gated model row already worked. PR description updated to describe this instead of the earlier whole-row-clickable behavior.

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelPublicAccessTier.swift
```diff
@@ -0,0 +1,72 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// The lowest publicly-marketed subscription tier a model or reasoning effort requires. Distinct
+/// from `AIChatUserTier`, which also includes the non-public `.internal` case. `String`-backed so
+/// the raw values ("free"/"plus"/"pro") can be used directly as pixel parameters.
+public enum AIChatModelPublicAccessTier: String, Hashable, Sendable {
```
tomasstrba, line 24: New shared home for gating types that previously lived only in iOS `UnifiedToggleInput/`: `AIChatModelPublicAccessTier`, `DuckAISubscriptionUpsellingFlow`, and `AIChatUserTier.upgradeFlow(for:)`. Moved here so macOS and iOS share one implementation

SharedPackages/AIChat/Sources/AIChat/Shared/ReasoningModeAccessResolver.swift
```diff
@@ -28,15 +27,17 @@ protocol ReasoningModeAccessResolving {
     func canSelect(modeRequiring requiredTier: AIChatModelPublicAccessTier, userTier: AIChatUserTier) -> Bool
 }
 
-struct ReasoningModeAccessResolver: ReasoningModeAccessResolving {
+public struct ReasoningModeAccessResolver: ReasoningModeAccessResolving {
```
tomasstrba, line 30: Moved from `iOS/DuckDuckGo/UnifiedToggleInput/` into the shared `AIChat` package

tomasstrba (commented):
Reviewer guide: a few pieces were promoted from iOS-only (`UnifiedToggleInput/`) into the shared `AIChat` package so macOS reuses the same gating logic instead of duplicating it. Inline notes on each new/moved shared file follow.

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
```diff
@@ -567,6 +618,59 @@ final class AIChatOmnibarController {
...
+        guard !model.entityHasAccess else { return nil }
+        return model.lowestPublicAccessTier
+    }
+
+    /// Opens the subscription activation flow, for a user who already has a subscription (e.g.
+    /// purchased on another device) and wants to sign in rather than purchase again.
+    func presentSubscriptionActivationFlow() {
+        subscriptionUpsellPresenter.presentSubscriptionActivation()
+    }
+
```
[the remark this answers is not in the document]
tomasstrba: Good catch — this was a real gap on the persisted-value path. Fixed in 2a460b585f: `effectiveReasoningEffort` now also requires `isReasoningEffortAccessible(effort)`, so an effort that survives a tier change (model still supports it, so stale-cleanup keeps it, but the current tier can't access it) is never attached at submit. Added a unit test covering the Plus→free downgrade scenario.

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelSectionBuilder.swift
```diff
@@ -84,3 +84,23 @@ public enum AIChatModelSectionBuilder {
...
+    public let model: AIChatModel
+    public let requiredTier: AIChatModelPublicAccessTier
+}
+
+public extension AIChatModelSectionBuilder {
+    /// Groups models for a picker that shows every gated model with an upsell, rather than hiding
+    /// higher-tier models: accessible models first, then every gated model paired with its required
+    /// tier (so a Plus user still sees Pro-only models, badged, alongside an upgrade path).
+    static func buildGatedSections(models: [AIChatModel]) -> (accessible: [AIChatModel], gated: [AIChatGatedModel]) {
+        let accessible = models.filter { $0.entityHasAccess }
```
jotaemepereira, line 98: 💡 Would this read better? So we do not iterate the models twice

```swift
static func buildGatedSections(models: [AIChatModel]) -> (accessible: [AIChatModel], gated: [AIChatGatedModel]) {
    var accessible = [AIChatModel]()
    var gated = [AIChatGatedModel]()
    for model in models {
        if model.entityHasAccess {
            accessible.append(model)
        } else if let requiredTier = model.lowestPublicAccessTier {
            gated.append(AIChatGatedModel(model: model, requiredTier: requiredTier))
        }
    }
    return (accessible, gated)
}
```

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelSectionBuilder.swift
```diff
@@ -84,3 +84,23 @@ public enum AIChatModelSectionBuilder {
...
+public struct AIChatGatedModel {
+    public let model: AIChatModel
+    public let requiredTier: AIChatModelPublicAccessTier
+}
+
+public extension AIChatModelSectionBuilder {
+    /// Groups models for a picker that shows every gated model with an upsell, rather than hiding
+    /// higher-tier models: accessible models first, then every gated model paired with its required
+    /// tier (so a Plus user still sees Pro-only models, badged, alongside an upgrade path).
+    static func buildGatedSections(models: [AIChatModel]) -> (accessible: [AIChatModel], gated: [AIChatGatedModel]) {
```
jotaemepereira: 🔍 I was confused by the name of this function; the name implies that all models returned are ‘gated’, which is not true. Would something like `groupByAccess` sound better to you?

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelSectionBuilder.swift
```diff
@@ -84,3 +84,23 @@ public enum AIChatModelSectionBuilder {
...
+
+/// A gated (subscriber-only) model paired with the public tier required to unlock it.
+public struct AIChatGatedModel {
+    public let model: AIChatModel
+    public let requiredTier: AIChatModelPublicAccessTier
+}
+
+public extension AIChatModelSectionBuilder {
+    /// Groups models for a picker that shows every gated model with an upsell, rather than hiding
+    /// higher-tier models: accessible models first, then every gated model paired with its required
```
jotaemepereira: 🔍 If we improve the readibility of the code below and the naming, I do not think this comment is necessary at all.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1741,42 +1741,158 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
-        )
-
-        for (index, section) in sections.enumerated() {
-            if index > 0 {
-                menu.addItem(.separator())
-            }
-            if let header = section.header {
-                let headerItem = NSMenuItem(title: header, action: nil, keyEquivalent: "")
-                headerItem.isEnabled = false
+        // Accessible models first (no header), then a "Subscriber exclusive" section listing every
```
jotaemepereira: 🔍 Could we be more concise with comments? It is too much information; the code should explain most of it.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1741,42 +1741,158 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
+
+    /// PoC-only "correct ordering" per https://app.asana.com/1/137249556945/task/1216388092784294 —
+    /// duck.ai web computes this client-side too (a hand-maintained `modelPickerLayoutByTier` per
+    /// the linked task's comments): per-tier "recommended" models come first, everything else keeps
+    /// the API's relative order. No descriptive subtitles for now — those need real backend/copy
+    /// support, not this PR's per-position guesses. Delete this whole function (and its matcher
+    /// table) once the backend ships proper ordering — nothing else in this file depends on it.
+    /// Follow-up to replace this PoC scaffolding with the final approach:
+    /// https://app.asana.com/1/137249556945/project/1211654189969294/task/1216559729471554
+    private static func pocOrderedAccessibleModels(_ models: [AIChatModel], userTier: AIChatUserTier) -> [AIChatModel] {
```
jotaemepereira: This logic and all of the poc* should not live here, given that it is not associated with any UI-related code (no AppKit code), and, for that reason, it is harder to unit test, also the VC is big enough.

You could move 
- `pocOrderedAccessibleModels` to `AIChatModelSectionBuilder`
- `pocRecommendedModelMatchers` to `AIChatModelSectionBuilder`
- `splitModelTitle` could this moved to the shared package? Maybe iOS could use it as well
- `pocTrailingLabel` only for macOS but let’s remove it from the VC

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1792,31 +1908,143 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
+        }
+        // One impression per menu-open, matching the model picker's header — not one per gated row.
+        if didShowUpsellBadge {
+            omnibarController.recordBadgeImpression()
         }
 
         menu.popUp(positioning: nil, at: NSPoint(x: 0, y: -5), in: reasoningPickerButton)
     }
 
+    /// Every reasoning-effort row — accessible or gated — goes through this one custom view (the
```
jotaemepereira: 🔍  Along with the other comments, can we be more concise?

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
```diff
@@ -65,6 +65,13 @@ final class AIChatOmnibarController {
     private let suggestionsReader: AIChatSuggestionsReading?
     private let modelsService: AIChatModelsProviding
     private let subscriptionManager: any SubscriptionManager
+    private let subscriptionUpsellPresenter: AIChatOmnibarSubscriptionUpselling
+    /// Shared across the model picker and reasoning-effort picker — reuses the same "N views" cap
```
jotaemepereira: 🔍 Same here with all comments, let’s try to be more concise or remove comments that are not needed.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1741,42 +1741,158 @@ final class AIChatOmnibarContainerViewController: NSViewController {
...
-                headerItem.isEnabled = false
+        // Accessible models first (no header), then a "Subscriber exclusive" section listing every
+        // gated model (dimmed, with a PLUS/PRO badge) — unlike AIChatModelSectionBuilder.buildSections
+        // (used by the NTP web dropdown), gated models are never hidden here, so a Plus user still
+        // sees Pro-only models. When the subscription-upsell feature flag is on, the section also
+        // gets a "Try for free"/"Upgrade" badge that opens the confirmation dialog; when it's off,
+        // the gated rows are plain dimmed, non-interactive rows with no CTA — same as before this
+        // feature shipped, so disabling the flag can't reintroduce the "select a model you don't
+        // have access to" bug even though it removes the upsell path.
+        let (accessible, gated) = AIChatModelSectionBuilder.buildGatedSections(models: omnibarController.models)
```
jotaemepereira: 💡 I want to propose a change: we are introducing a lot of logic that should not live in the UI layer. The VC currently makes at least five decisions that aren't its job: the upsell feature flag, tier -> header-title, free-trial -> badge-text, !gated.isEmpty, and ordering/labeling. 

All of that is "what should the menu contain," which is view-model work. The VC should only know "given a list of render instructions, make NSMenuItems.”

I’ve created a branch with Claude's help to move this outside the VC and make the code more readable and unit-testable.

Branch: https://github.com/duckduckgo/apple-browsers/pull/new/juan/suggestions-tom-pr

jotaemepereira (changes requested):
@tomasstrba The functionality works as expected. I’m requesting changes because we are adding a lot of different business and presentation logic to the view layer that I believe we should avoid, given that the view controllers are already big enough.

Also, what do you think of being more concise with code comments or removing them all? I’ve also had this problem with AI that adds a lot of text; we should be careful, given that most of the time the code speaks for itself, and the code comment adds more noise, and also it is something we need to maintain when the code changes.

Let me know what you think.

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatModelSectionBuilder.swift
tomasstrba: In this case, if we know there are ~ 10 models, readability is more important than micro optimization. But not a strong opinion, I'll follow your suggestion

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
tomasstrba: I think AI decided to put it there because we will be removing it in a follow-up (when backend is ready). But I agree with you, let's do it properly 👍

jotaemepereira (approved):
Thanks for addressing the changes 🙌🏼

tomasstrba:
@jotaemepereira, great feedback in this review, and thanks for the quick re-review. I trimmed the comments down even further as I did not manage to finalize them yesterday. I will wait for translations and then merge the PR.

Thanks again! 🙌🏼

PR #5720, opened 2026-07-08
Filter out crash reports older than 7 days from crash pixels

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201899738287924/task/1216384098622310?focus=true

### Description
We were sending a crash pixel for every crash report we received, including very old ones that MetricKit or the crash report collector only surfaced days or weeks after they actually happened. Those stale crashes skewed our crash metrics and triggered false SLO-breach alarms.

Crashes older than 7 days are now excluded from crash pixels across the board — iOS and macOS App Store (via MetricKit) and the macOS DMG crash report collector. Crash reports are still uploaded regardless of age; only pixel reporting is affected.

### Testing Steps
Covered by unit tests — there's no reliable manual repro because it depends on the OS delivering week-old crash reports.

1. Run `CrashesTests` (BrowserServicesKit) → `testCrashesOlderThanMaxAgeAreExcludedFromCrashPixels` and `testCrashesWithinMaxAgeAreReportedAsCrashPixels` pass.
2. Run `CrashReportingTests` (macOS/LocalPackages/CrashReporting) → the new `CrashReporterTests` and the `creationDate` tests in `CrashReportTests` pass.

### Impact
Low — this only narrows which crashes fire a pixel. Crash report uploads and CRCID handling are unchanged, so no crash data is lost.

#### What could go wrong?
- A legitimate recent crash gets filtered out → mitigated by a generous 7-day window (MetricKit normally delivers within ~24h) plus tests asserting within-window crashes still fire.
- A macOS DMG report without a readable creation date gets dropped → mitigated by failing open (undated reports still fire a pixel).

### Quality Considerations
- Pixels only: the age filter never touches the upload path, so crash reports still reach the dashboard regardless of age.
- Single source of truth: both pipelines use `CrashCollection.maxCrashReportAgeForPixels` (7 days).
- The macOS DMG path keys off the crash file's creation date (≈ crash time) rather than the in-file timestamp, because the existing Legacy `.crash` timestamp regex is end-of-string anchored and does not reliably match real multi-line reports.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Narrows analytics pixel firing only; uploads and CRCID flows are untouched, with fail-open behavior when creation dates are missing.
> 
> **Overview**
> Crash **pixels** now ignore reports older than **7 days** (`CrashCollection.maxCrashReportAgeForPixels`), so late MetricKit or collector deliveries no longer inflate crash metrics or trigger false SLO alarms. **Uploads** (Sentry / server) are unchanged.
> 
> On **iOS (MetricKit)**, pixel parameters are built only from payloads with `timeStampBegin` within the window. The **first-crash** flag is cleared only after at least one in-window crash actually fires a pixel, so a batch of only stale reports no longer consumes it.
> 
> On **macOS DMG** collection, `CrashReport` gains **`creationDate`** (file creation time on legacy `.crash` and `.ips` reports). `CrashReporter` fires pixels only for reports passing the same 7-day filter; reports with no creation date still fire a pixel.
> 
> Unit tests cover age filtering, upload vs pixel behavior, first-crash retention, and creation-date exposure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cb0d91ad805e9557dd0d6ac183310e0f8e769b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReport.swift
+++ b/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReport.swift
@@ -30,6 +30,9 @@ protocol CrashReport: CrashReportPresenting {
     var contentData: Data? { get }
     var appVersion: String? { get }
     var bundleID: String? { get }
+
+    /// The date the crash report file was created, used as a proxy for when the crash occurred.
+    var creationDate: Date? { get }
 }
 
 final class LegacyCrashReport: CrashReport {
@@ -88,6 +91,8 @@ final class LegacyCrashReport: CrashReport {
 
     let appVersion: String? = nil
     let bundleID: String? = nil
+
+    var creationDate: Date? { fileManager.fileCreationDate(url: url) }
 }
 
 final class JSONCrashReport: CrashReport {
@@ -166,4 +171,6 @@ final class JSONCrashReport: CrashReport {
         }
         return version
     }
+
+    var creationDate: Date? { fileManager.fileCreationDate(url: url) }
 }
--- a/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReporter.swift
+++ b/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReporter.swift
@@ -94,7 +94,6 @@ public final class CrashReporter: CrashReporting {
 
         if internalUserDecider.isInternalUser {
             await send(crashReports)
-            return
         } else if await promptForConsent(latest) {
             await send(crashReports)
         }
--- a/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReporter.swift
+++ b/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReporter.swift
@@ -83,7 +83,8 @@ public final class CrashReporter: CrashReporting {
             return
         }
 
-        for crash in crashReports {
+        // Only fire crash pixels for reports within the age limit; stale reports are still uploaded below.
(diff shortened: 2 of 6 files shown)


macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReport.swift
```diff
@@ -30,6 +30,9 @@ protocol CrashReport: CrashReportPresenting {
     var contentData: Data? { get }
     var appVersion: String? { get }
     var bundleID: String? { get }
+
+    /// The date the crash report file was created, used as a proxy for when the crash occurred.
+    var creationDate: Date? { get }
```
ayoy, line 35: The alternative (the most correct solution) would be to parse every crash report and read the creation date inside the report, but we can safely assume that the crash report file was created right after the crash (within seconds), which is a good enough approximation.

macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReporter.swift
mallexxx: could we cleanup this orphaned `return` as well?

mallexxx (approved):
Looks good!
Are we ok with the Cursor bot comment about loosing the `.first` parameter?

ayoy:
Thanks @mallexxx! Good catch about `isFirstCrash`, I missed Cursor's comment. Let me address it before merging.

PR #5725, opened 2026-07-08
Anchor address bar keystroke t₀ to the keyDown timestamp

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1213659555956031/task/1216392534192854?focus=true

### Description

The address bar performance coordinator stamped a keystroke's start time when the text-insertion callback ran, which lands slightly after the key was actually pressed. This anchors the start time to the originating keyDown event's timestamp when one is available, so the measured character-render latency better reflects true input-to-paint time.

The keyDown timestamp shares the same media timebase as the paint hook, so the two are directly comparable. Insertions that aren't driven by a physical keyDown — paste, IME commit, dictation, programmatic edits — have no reliable originating event, so they fall back to stamping the clock at the callback exactly as before.

### Testing Steps

This is a measurement-only change with no user-visible behavior; the effect is on the latency values reported by the AddressBar performance pixel, not something observable in the UI.

1. Run the AddressBarPerformance package tests.
2. Confirm they pass.

### Impact and Risks

Low. The change only affects the start timestamp used for the address bar latency pixel buckets.

#### What could go wrong?

If the keyDown timestamp used a different timebase than the paint hook, the computed latencies would be wrong — both are mach media-time seconds, so they are comparable. Any non-keyDown insertion path is unchanged from the previous behavior.

### Quality Considerations

The recorder already caps individual samples and drops negative deltas, so an unexpected timestamp can't skew a bucket. Existing tests exercise the fallback path (no live event under test).

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Measurement-only change to pixel timestamps; non-keyDown paths unchanged and existing sample caps guard bad deltas.
> 
> **Overview**
> **Address bar performance measurement** now starts the keystroke clock at the physical `keyDown` event’s `NSEvent.timestamp` when `NSApp?.currentEvent` is a key down, instead of only when the text-insertion path runs. That timestamp is passed into `markKeystrokeForSuggest(at:)` so char/suggest latency pixels better reflect true input-to-paint time (same mach media timebase as the paint hook). Paste, IME, dictation, and programmatic edits still pass `nil` and keep the recorder’s existing “stamp now” fallback. `NSApp` is optional-chained so unit tests without an `NSApplication` do not trap.
> 
> **CI:** bumps the Danger workflow action from `danger/danger-js@13.0.4` to `@13.0.10`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f0ee82bde652034cdbd26efd91c416f15e6b684. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/LocalPackages/AddressBarPerformance/Sources/AddressBarPerformance/AddressBarPerformanceCoordinator.swift
```diff
@@ -124,9 +124,17 @@ public final class AddressBarPerformanceCoordinator {
...
     /// still keeps the hook alive.
+    ///
+    /// For a physical keyDown, anchors t₀ to the event's `NSEvent.timestamp` — earlier than this
+    /// call, same timebase as the paint hook. Other sources (paste, IME, dictation, programmatic)
+    /// have no reliable originating event, so the recorder stamps its own clock.
     public func markKeystroke() {
         cancelPendingHookStop()
-        pendingCharStartTime = recorder.markKeystrokeForSuggest()
+        var keystrokeTime: TimeInterval?
+        if let event = NSApp?.currentEvent, event.type == .keyDown {
```
ayoy, line 137: I didn't know `NSApp` was optional? 🤔

ayoy (approved):
LGTM

macOS/LocalPackages/AddressBarPerformance/Sources/AddressBarPerformance/AddressBarPerformanceCoordinator.swift
diegoreymendez: With `NSApp.currentEvent`: every test crashed with `Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value at AddressBarPerformanceCoordinator.swift:134`.

Since it's a IUO there's no harm in unwrapping.

ayoy: Good to add a comment about it to prevent a future maintainer from trying to optimize it :)

diegoreymendez: The CI will fail if they do.

But sure I'll add a comment.

diegoreymendez:
@ayoy  - I'm reverting the danger update, as it's currently not working regardless of version, and I don't want to introduce a change I can't validate.  We can update once danger is working again.

PR #5743, opened 2026-07-09
Fix Bitwarden integration CPU drain and relaunch churn

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1216409281175249

### Description

Also fixes: https://app.asana.com/1/137249556945/project/1204912272578138/task/1216400413768936

Fixes the two internal reports (battery drain, Bitwarden Dock icon flapping) for users with Bitwarden integration enabled. Root causes, one per commit:

- Battery drain: when the Bitwarden proxy process exits, the pipe's readability handler was never uninstalled. At EOF the handler refires continuously with empty data, pegging a core — and every reconnect leaked one more permanent spinner. Monitoring now stops at EOF and on teardown.
- Relaunch churn: reconnects retried every second with no backoff, relaunching the full Bitwarden Electron binary each time. Worse, the "is Bitwarden running" check counted the browser's own spawned proxy (it registers under com.bitwarden.desktop), so after Bitwarden quit each dying proxy licensed the next spawn — a self-sustaining loop, visible as the Dock icon flapping. Retries now back off exponentially (1s → 16s), resetting when the failure reason changes, after a successful status round-trip, or on user actions (setup flow, status UI); the running check now requires an alive Bitwarden instance not spawned by us.

### Testing Steps

1. Enable Bitwarden integration (Settings → Passwords & Autofill), complete onboarding
2. Verify autofill works on a login form - https://www.fill.dev
3. Quit Bitwarden while the browser runs. Make sure CPU usage stays flat in Activity Monitor
4. Relaunch Bitwarden and make sure integration reconnects

5. Turn off `bitwardenConnectionHardening` feature flag
6. Restart the app
7. Test the Bitwarden integration again (except the CPU spike)

### Impact

Medium — limited to non-App-Store macOS builds with Bitwarden selected as password manager. Regressions could break Bitwarden autofill; mitigated by the kill switch and unit tests (new BWManagementTests target, registered in the test plan and schemes).

#### What could go wrong?

- A pipe write can fail in the ms-wide window while the proxy is dying → the message is logged and dropped; that autofill attempt silently misses and the pending completion isn't called (pre-existing behavior for lost responses; previously this window crashed the app). Recovery via reconnect.
- Backoff could delay detecting a user's Bitwarden-side onboarding step by up to 16s → mitigated by expedited retries on setup-flow and status-UI actions.
- If the hardened behavior misbehaves in the field → disable `bitwardenConnectionHardening` remotely; note the teardown fixes are deliberately not gated.

---

> [!NOTE]
> **Medium Risk**
> Changes Bitwarden autofill connection lifecycle on non–App Store macOS builds only; regressions could delay reconnects (up to 16s) or drop messages during proxy teardown, mitigated by expedited user-action retries, remote kill switch, and new unit tests.
> 
> **Overview**
> Fixes **battery drain** and **Dock icon flapping** for macOS users with Bitwarden as the password manager by changing how the native-messaging proxy reconnects and how pipes are monitored.
> 
> **Connection behavior:** Reconnects use **exponential backoff** (1s → 16s) instead of a fixed 1s interval, with backoff reset when the failure reason changes, after a successful encrypted status round-trip, or when the user opens Bitwarden, runs handshake, or refreshes status UI. The **`bitwardenConnectionHardening`** remote flag (on by default) gates backoff and EOF pipe handling; disabling it restores legacy 1s retries and EOF-tolerant monitoring while **teardown fixes stay active**.
> 
> **Proxy / pipe fixes:** On proxy exit, the stdout **readability handler is removed at EOF** so it cannot spin on empty reads. Teardown clears handlers and partial buffers, ignores late data from replaced proxies, and uses **throwing writes** instead of crashing on a dead pipe. **`RunningApplicationCheck`** no longer treats the browser-spawned Bitwarden proxy (same bundle ID) as “Bitwarden is running,” breaking a self-sustaining relaunch loop after the real app quits.
> 
> **Wiring & tests:** `BWManagementFactory` takes `isConnectionHardeningEnabled`; the app passes the feature flag from `PasswordManagerCoordinator`. Adds **`BWManagementTests`** (backoff, EOF monitoring, child-process detection) and registers the target in schemes and the BWIntegration test plan.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d7be7127bbd14b4b7b06f59632e913033fb7f34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amddg44 (approved):
LGTM!

PR #5750, opened 2026-07-09
iOS: Fix NaN crash in tab switcher ↔ web view transition

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1216390072245582?focus=true

### Description
Fixes a SIGABRT crash ("CALayer position contains NaN") that spikes on iOS 26.5 / 7.227 during the tab-switcher ↔ web-view open/close animation.

The animation sizes the preview image from its aspect ratio (height ÷ width). A degenerate preview — zero width, or a non-finite dimension — makes that division produce NaN/Inf, which lands on a view frame and aborts when Core Animation reads the layer position. The previous guard only rejected a missing preview, so a present-but-degenerate one went straight into the division.

Two sources of degenerate previews are common on iOS 26.5: keyboard/scene-inset timing can momentarily collapse the capture render size to zero, and a partial/corrupt persisted preview can decode to a zero-sized image.

The fix guards the aspect-ratio math so any zero/non-finite preview size falls back to a safe container-filling frame (never dividing), and stops capturing a preview when the render size has collapsed to zero. The pure frame math was extracted into a small value type so it can be unit-tested without a view controller.

### Testing Steps
1. Open several tabs (including a page still loading / with the keyboard up) → open and close the tab switcher repeatedly → transition animates normally, no crash.
2. Regression coverage: run `WebViewTransitionGeometryTests` (11 tests) → all pass. Removing the finite guard turns the 7 crash-guard assertions red, confirming they catch the NaN/Inf case.

### Impact
Low.

#### What could go wrong?
A degenerate preview could still reach the transition → mitigated by the aspect-ratio guard falling back to a finite container-filling frame (the pre-existing degenerate-case geometry), so at worst a briefly stretched image shows instead of a crash.

### Quality Considerations
Behaviour is unchanged for all valid previews (math preserved); the new fallback path only runs for zero/non-finite sizes that previously crashed. Covered by unit tests asserting finite frames.

---

> [!NOTE]
> **Low Risk**
> Localized UI transition and snapshot guards; valid previews keep the same math, with fallbacks only for degenerate sizes.
> 
> **Overview**
> Fixes a **CALayer NaN crash** during tab-switcher ↔ web-view transitions when tab preview images have zero or non-finite dimensions.
> 
> Preview layout math moves into **`WebViewTransitionGeometry`**, which returns `nil` for bad aspect ratios and uses safe container-filling frames instead of dividing by zero. **`WebViewTransition`** now calls that helper instead of inlined frame calculations.
> 
> **`preparePreview`** in `TabViewController` also skips capture when the inset-adjusted render size collapses to zero, so fewer degenerate previews are produced. **`WebViewTransitionGeometryTests`** cover finite-frame fallbacks and normal geometry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ea0210ac58c9de40d1d7319ee6b41481567594a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGoTests/WebViewTransitionGeometryTests.swift
+++ b/iOS/DuckDuckGoTests/WebViewTransitionGeometryTests.swift
@@ -1,0 +1,99 @@
+//
+//  WebViewTransitionGeometryTests.swift
+//  DuckDuckGo
+//
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import XCTest
+import UIKit
+@testable import DuckDuckGo
+
+final class WebViewTransitionGeometryTests: XCTestCase {
+
+    // MARK: - aspectRatio
+
+    func testAspectRatioReturnsHeightOverWidthForNormalSize() {
+        let ratio = WebViewTransitionGeometry.aspectRatio(of: CGSize(width: 100, height: 200))
+        XCTAssertEqual(ratio, 2.0)
+    }
+
+    func testAspectRatioIsNilForZeroWidth() {
+        XCTAssertNil(WebViewTransitionGeometry.aspectRatio(of: CGSize(width: 0, height: 200)))
+    }
+
+    func testAspectRatioIsNilForZeroHeight() {
+        XCTAssertNil(WebViewTransitionGeometry.aspectRatio(of: CGSize(width: 100, height: 0)))
+    }
+
+    func testAspectRatioIsNilForZeroSize() {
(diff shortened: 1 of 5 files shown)


iOS/DuckDuckGoTests/WebViewTransitionGeometryTests.swift
```diff
@@ -0,0 +1,99 @@
...
+                                                           previewSize: .zero,
+                                                           isGridViewEnabled: true)
+        assertFinite(frame)
+    }
+
+    func testPreviewFrameIsFiniteForNormalPreview() {
+        let frame = WebViewTransitionGeometry.previewFrame(for: CGSize(width: 180, height: 240),
+                                                           previewSize: CGSize(width: 300, height: 600),
+                                                           isGridViewEnabled: true)
+        assertFinite(frame)
```
bwaresiak: Two nits:
1/ Shall we test against isGridViewEnabled false as well?
2/ For cases where we have correct geometry, shall we validate actual size based off it, instead just checking the finite?

bwaresiak (approved):
Good job, tested on my device and observed no regression. Left two optional comments.

jaceklyp:
both done, pushed 👍

2/ good shout — swapped the normal `previewFrame` test to assert the exact frame (`CGRect(4, 44, 172, 344)`) instead of just finite, which also nails down that the extraction didn't change the math. `destinationImageFrame` already checked exact values.

1/ added a grid-disabled case too. one thing worth noting: that branch has no division so it can't actually produce NaN — it's really just coverage rather than a crash guard, but no harm having it.

PR #5770, opened 2026-07-10
NavigationTests: prevent WebContent process accumulation causing flaky session-restoration test

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1206506154943201
Tech Design URL: N/A
CC: N/A

### Description
- Replace per-test `killWebContentProcess()` in `tearDown` with a bounded pool-level cleanup: every ~10 tests `_terminateAllWebContentProcesses` is called on the `WKProcessPool`, preventing unbounded WebContent process accumulation that was starving the test runner and causing `testWhenSessionIsRestored_navigationTypeIsSessionRestoration` to time out
- `_webView` is nil'd **before** the termination call so `WebPageProxy.close()` fires first, emptying `m_pageMap` for that process; any deferred IPC-close callback finds no live pages and returns early, avoiding `EXC_BAD_ACCESS` in `resetStateAfterProcessTermination`
- Counter is incremented in `makeWebView` and reset to zero after each bulk termination, keeping the pool bounded to ~10 processes at any time
- The `killWebContentProcess()` extension (`kill(pid, SIGTERM)`) is left untouched so termination-specific tests continue to work

### Testing Steps
- Validate CI is green

### Impact and Risks
**Impact Level: None**

#### What could go wrong?
- `_terminateAllWebContentProcesses` is a private WebKit API; it is already used broadly throughout WebKit's own test suite (`TestWebKitAPI`) for the same cleanup purpose, so breakage risk is low
- If the pool used by nonPersistent data-store webViews differs between tests, some orphan processes may escape the bulk kill — but the pool is recovered naturally by WebKit's own process count limits

### Quality Considerations
- Change is test infrastructure only; no production code is touched
- Avoids the previous approach of raw `kill(pid, SIGTERM)` on stale PIDs, which had a PID-reuse hazard and deferred IPC crash window

### Notes to Reviewer
The root cause was that each test accumulated a WebContent process in the pool cache (nonPersistent data-store), and the prior tearDown's direct `killWebContentProcess()` call triggered an async IPC close while the `WKWebView` was still alive, causing `resetStateAfterProcessTermination` to run against live page state. Nil-ing first + bulk-terminating old ones every 10 tests sidesteps both issues.


mallexxx:
Closing in favour of correct branch (alex/) based on main.

PR #5774, opened 2026-07-10
[BSK] Fix testWhenAuthenticationChallengeReturnsCancel on macOS 26

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1215587453456948?focus=true
Tech Design URL: N/A
CC: N/A

### Description
- On macOS 26, `NSURLSessionAuthChallengeCancelAuthenticationChallenge` changed behaviour for HTTP auth challenges: instead of cancelling the task (resulting in `NSURLErrorCancelled`), the framework now delivers the 401 response — the same outcome as `rejectProtectionSpace`
- `testWhenAuthenticationChallengeReturnsCancel_responderChainReceivesFailure` was disabled because it expected `didFail(NSURLErrorCancelled)` which no longer fires on macOS 26
- Re-enable the test by building the expected history as a `var` and using `#available(macOS 26, *)` + `replaceSubrange` to swap the `didFail` entry for `response`/`didCommit`/`didFinish` events; the pre-macOS 26 failure path is preserved in the `else` branch

### Testing Steps
- Validate CI is green

### Impact and Risks
**Impact Level: None**

#### What could go wrong?
- The `#available` version floor is macOS 26; if the NSURLSession behaviour actually changed in an earlier release the `else` branch would still fail there — but we have no evidence of that

### Quality Considerations
- Pre-macOS 26 test path is preserved unchanged, ensuring no regression if run on an older SDK
- No production code was changed

### DIFF
--- a/SharedPackages/BrowserServicesKit/Tests/NavigationTests/NavigationAuthChallengeTests.swift
+++ b/SharedPackages/BrowserServicesKit/Tests/NavigationTests/NavigationAuthChallengeTests.swift
@@ -23,6 +23,7 @@ import Common
 import Swifter
 import WebKit
 import XCTest
+
 @testable import Navigation
 
 @available(iOS 15.0, *)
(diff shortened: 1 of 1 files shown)


hassaanelgarem (approved):
LGTM 👍

PR #5776, opened 2026-07-10
Fix ConfigurationStore data race causing loadData crash (Sentry 713934)

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211737940275488/task/1213468073551980

### Description
Fixes a recurring background-thread crash (EXC_BAD_ACCESS / SIGSEGV) that surfaces during configuration and remote-message refresh (Sentry 713934). The shared configuration store was reached from several concurrent refreshes without synchronization. Because the store value is heap-boxed, a copy-on-write reassignment of the store handle raced across threads and freed memory that was still referenced, so a later read faulted.

The store's "save" operations never actually mutate in-memory state — they write to files and UserDefaults, which are already thread-safe. The fix removes the spurious mutability so the store handle is immutable and only ever read concurrently, eliminating the race at its root with no behaviour change.

### Testing Steps
This is a concurrency crash with no reliable manual reproduction; it is verified with automated tests:
1. Build the iOS app → succeeds.
2. Run BrowserServicesKit `ConfigurationTests` → all pass (96/96).
3. Run the new `testConcurrentFetchesOnSharedFetcherDoNotRaceOnStore` under Thread Sanitizer → passes clean. Temporarily reverting the fix makes the same test fail with a TSan "Swift access race" report and abort, confirming it catches the regression.

### Impact
Medium–High: removes a production crash occurring on background configuration / remote-message refresh (spikes on RMF rollouts). If the fix were somehow wrong, the only affected behaviour would be config persistence — but writes still go through the same thread-safe file / UserDefaults path as before, unchanged.

#### What could go wrong?
Relaxing the store's save methods to non-mutating could break a conformer that relied on struct mutation → none do (every conformer writes to reference-backed file/UserDefaults storage); verified by building iOS + all conformers and running the Configuration suite.

### Quality Considerations
Concurrency: the configuration layer now relies on immutable handles plus already-thread-safe sinks (NSFileCoordinator, UserDefaults). Benign last-writer-wins on concurrent same-config writes is unchanged (coordinated writes, no corruption). No privacy or data-loss impact.

---

> [!NOTE]
> **Medium Risk**
> Touches shared configuration fetch/persist paths used during background privacy-config and remote-message refresh; persistence still goes through the same file/UserDefaults/NSFileCoordinator sinks, but incorrect API adoption could break a conformer that relied on in-struct mutation.
> 
> **Overview**
> Fixes a **background crash** from concurrent configuration refreshes racing on a heap-boxed `ConfigurationStoring` handle when `mutating` save methods triggered copy-on-write reassignment while other threads still read the store.
> 
> **`ConfigurationStoring`** drops `mutating` on `saveData` / `saveEtag` so saves are non-mutating at the protocol level; **`ConfigurationFetcher`** holds `store` as `let` instead of `var`. iOS app, VPN packet tunnel, and test mocks adopt the same API. VPN tunnel store uses `nonmutating set` on the etag property and `let` for `defaults` so etag writes no longer imply struct mutation.
> 
> Adds **`testConcurrentFetchesOnSharedFetcherDoNotRaceOnStore`** (500 parallel fetches, boxed existential store, TSan regression for Sentry 713934).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4804efee72924c9ffe1104dc133991530014e3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationFetching.swift
@@ -39,7 +39,7 @@ public final class ConfigurationFetcher: ConfigurationFetching {
 
     }
 
-    private var store: ConfigurationStoring
+    private let store: ConfigurationStoring
     private let validator: ConfigurationValidating
     private let configurationURLProvider: ConfigurationURLProviding
     private let sessionProvider: () -> URLSession
--- a/SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationStoring.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Configuration/ConfigurationStoring.swift
@@ -24,8 +24,8 @@ public protocol ConfigurationStoring {
     func loadEtag(for configuration: Configuration) -> String?
     func loadEmbeddedEtag(for configuration: Configuration) -> String?
 
-    mutating func saveData(_ data: Data, for configuration: Configuration) throws
-    mutating func saveEtag(_ etag: String, for configuration: Configuration) throws
+    func saveData(_ data: Data, for configuration: Configuration) throws
+    func saveEtag(_ etag: String, for configuration: Configuration) throws
 
     func fileUrl(for configuration: Configuration) -> URL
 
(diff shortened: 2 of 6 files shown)


bwaresiak (approved):
Nice one! Looks good to me

PR #5539, opened 2026-06-24
[macOS] Harden bookmarks bar favicon refresh against flicker

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201899738287924/task/1215999008194803?focus=true

### Description

Follow-up to #5534 to further limit flickering of the bookmarks bar favicons.

Instead of reloading data of the entire bookmarks bar on favicons cache update, only refresh
displayed favicons using a dedicated function that never clears the favicon (only updates existing one if needed).

### Testing Steps
1. Put several bookmarks with favicons on the bookmarks bar (e.g. a few github.com pages, plus other sites).
2. Quit and relaunch the browser (a cold start forces favicons to decode lazily off-main); for a stronger repro, open many tabs/bookmarks with favicons so the image cache churns.
3. Watch the bookmarks bar favicons during the first seconds after launch / as pages load.
   - Before: favicons flicker between the real icon and the generic placeholder.
   - After: each favicon appears once and stays — placeholder upgrades to the real icon and never reverts.
4. Confirm favicons still appear for newly-bookmarked sites and after the favicon finishes downloading (placeholder → real upgrade still works).

### Impact and Risks
Low. Bookmarks-bar rendering only — changes how the bar reacts to favicon-cache updates (in-place upgrade vs. full reload). No change to favicon fetching, storage, bookmark data, or privacy.

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Low Risk**
> Bookmarks bar UI only; favicon fetch/storage and bookmark data are unchanged.
> 
> **Overview**
> Follow-up to harden bookmarks bar favicon updates when `.faviconCacheUpdated` fires during lazy decode.
> 
> **`refreshFavicons()`** no longer calls **`reloadData()`** on the collection view. It walks **visible** `BookmarksBarCollectionViewItem` instances and calls a new **`refreshDisplayedFavicon()`** on each.
> 
> **`refreshDisplayedFavicon()`** re-resolves the bookmark favicon (stored favicon or `faviconManager` cache), sets the image view only when a decoded image exists, and **never** replaces an already-shown icon with the generic placeholder.
> 
> Placeholder → real favicon upgrades still work; full cell rebuilds on every cache notification are avoided.
> 
[...]

PR #5783, opened 2026-07-10
[macOS] Fix bookmarks bar favicon flicker between two versions

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1216447325873126?focus=true

### Description
Follow-up to the earlier bookmarks bar favicon flicker fixes (#5534, #5539). A few bookmarks still flickered — but now between two *real* icons rather than icon vs. placeholder. This affects bookmarks whose page-specific favicon differs from the site's favicon (for example a page that embeds a third-party ordering/booking widget with its own icon). While the page-specific icon was still loading, the bar briefly fell back to the site icon, then swapped back once the page icon loaded, ping-ponging between the two every time the favicon cache updated.

The bar now shows a single, stable favicon per bookmark. If the favicon isn't loaded yet, it leaves the currently shown icon in place instead of substituting a different one mid-load.

### Testing Steps
1. Add several sites to the bookmarks bar, including at least one whose page favicon differs from its site favicon (the report reproduced with **SFGATE** and **Think Coffee**).
2. Open a new window (or relaunch) so the bar's favicons load fresh.
3. Watch the bookmarks bar for ~10 seconds → each favicon settles on one icon and no longer alternates between two versions.
4. Confirm the other bookmarks' favicons still appear correctly (no missing icons / placeholders left behind).

### Impact
Low — a visual fix to bookmarks bar favicons. No data, privacy, or crash surface; worst case a bookmark shows one favicon variant instead of another.

#### What could go wrong?
A bookmark whose page-specific favicon never finishes loading now keeps the icon resolved when the cell was first shown (the site favicon) instead of swapping → this is the intended behavior; initial cell setup already resolves the site favicon as its fallback, so the correct icon is still shown.

### Quality Considerations
- Bookmarks whose page and site favicons are identical (the common case) are unaffected — they already showed one stable icon.
- Minor side benefit: the refresh path now skips a redundant cache lookup and an image copy while a favicon is still decoding.
- Not covered here: the underlying favicon-cache-update notification churn (the flicker's engine) is unchanged; this PR stops the visible symptom. A deeper change to that notification loop would be a separate follow-up.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> UI-only bookmarks bar favicon behavior plus internal debug tooling; no auth, data, or persistence changes.
> 
> **Overview**
> Fixes bookmarks bar icons **ping-ponging between page-specific and site favicons** when lazy decoding and repeated `.faviconCacheUpdated` notifications.
> 
> `refreshDisplayedFavicon()` now updates only when `bookmark.favicon(.small)` returns a decoded image—**no extra host lookup via `faviconManager.getCachedFavicon`**. If the URL favicon is still decoding, the cell **keeps whatever icon is already shown** instead of briefly swapping to the host icon.
> 
> Adds a **debug-only** path to clear the lazy cache’s in-memory decoded images (persisted favicons unchanged): **Debug → Favicons → Clear In-Memory Cache**, wired through `FaviconImageCache.clearInMemoryCache()` and `FaviconManager.clearInMemoryFaviconCache()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9bea2d020491f79ea48258c41de04f0e86126549. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/BookmarksBar/View/BookmarksBarCollectionViewItem.swift
+++ b/macOS/DuckDuckGo/BookmarksBar/View/BookmarksBarCollectionViewItem.swift
@@ -133,13 +133,18 @@ final class BookmarksBarCollectionViewItem: NSCollectionViewItem {
         titleLabel.alphaValue = isInteractionPrevented ? 0.3 : 1
     }
 
-    /// Re-resolves this bookmark's favicon and shows it in place if a decoded image is now available.
+    /// Re-resolves this bookmark's favicon and shows it in place once a decoded image is available.
     ///
-    /// Never downgrades an already-shown favicon back to the placeholder.
+    /// Only ever shows the bookmark's own (URL-resolved) favicon. `Bookmark.favicon(_:)` already falls
+    /// back to the host favicon internally when the bookmark's URL has no favicon of its own, so this
+    /// deliberately does *not* add a separate host-favicon fallback: doing so downgraded the icon to the
+    /// host favicon during the brief off-main decode window of the URL favicon (`favicon(_:)` returns
+    /// `nil` while decoding), making a bookmark whose URL favicon differs from its host favicon flicker
+    /// between the two as `.faviconCacheUpdated` fired repeatedly. While the favicon isn't decoded yet
+    /// this is a no-op, leaving the currently shown icon untouched (never downgraded to a placeholder).
     func refreshDisplayedFavicon() {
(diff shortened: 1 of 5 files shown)


jleandroperez (approved):
:shipit:

PR #5779, opened 2026-07-10
iOS: Add surface param to contextual UTI pixels

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1216049537026988

### Description

Adds a `surface` parameter (`address_bar` / `duck_ai` / `contextual_chat`) to the Unified Toggle Input (UTI) telemetry pixels so usage in the contextual chat sheet can be told apart from the Duck.ai tab and the address bar. Reuses the shared pixel helpers added in #5768 (iPhone + iPad both emit `surface`).

Also retires two orphaned pixels — `m_aichat_contextual_page_context_placeholder_shown` / `_tapped` — the tappable placeholder chip they tracked no longer exists in the UTI.

Subscription attribution is deliberately untouched: the picker-shown and upsell pixels keep folding the contextual sheet into the `funnel_duckai_*` origin (no change to `SubscriptionFunnelOrigin` or the subscription redirect URL).

### Pixels

- **`surface` added** (`pikorddg` + `aataraxiaa` co-owned): prompt_submitted, model_selected, reasoning_effort_selected, image-gen/web-search selected·deselected·submitted, customize_responses_selected, image/file attached & removed, file_validation_failed, voice_tapped (via its existing `source` key), stop_generation_tapped, show_model_picker, submit_change_model(_prompt_sent).
- **`voice_tapped`** also gains **`has_pending_page_context`** (boolean) — whether page context was attached but not yet submitted at tap time (contextual only; always false elsewhere). Measures users tapping voice expecting the attached context to carry into the voice chat.
- **Unchanged** (origin-only): model_picker_shown, reasoning_effort_picker_shown, subscription_upsell_triggered.
- **Retired**: page_context_placeholder_shown / _tapped.

### Testing Steps

Internal only. **Enable these feature flags** (Debug → Feature Flags):
- `unifiedToggleInput` — the UTI itself
- `aiChatContextualUnifiedToggleInput` — **puts the UTI inside the contextual sheet (the surface this PR is about)**
- `iPadDuckAIBarControls` — required for the iPad step only

Observe param values in the xcode console.

1. **Contextual sheet (iPhone):** open the Duck.ai sheet over a web page → select a model, toggle a tool, attach an image + a file, tap voice, submit a prompt. Confirm each pixel carries `surface=contextual_chat` (voice_tapped `source=contextual_chat`).
[...]

PR #5676, opened 2026-07-06
iOS: Contexual UTI Part 2 - Remove Placeholder, Add Menu Item, Minor UX Changes

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1216235598004458

### Description

Updates contextual Duck.ai's UTI follow-up UX:

- Removes the dashed page-context placeholder chip.
- Adds `Ask About Page` to the attachment menu.
- Shows page context in the attachments strip alongside images/files.
- Keeps attachment menu order fixed and disables unavailable menu items instead of hiding them when another item is still actionable.
- Restores pre-submit controls: model picker, keyboard Go submit, and empty-input waveform voice button.
- Preserves image/file attachments when contextual quick actions submit before the web chat script is bound.
- Shows `Customize Responses` only after first submit, when the web view exists.
- Clears stale manually attached page context when reopening on a different page, while preserving it on same-page reopen.

#### Test 1: Immediate Contextual UTI
Prerequisites:
- `ff.unifiedToggleInput = true`
- `ff.aiChatContextualUnifiedToggleInput = true`
- Restart the app after changing these flags.
- Automatic context attachment is off.

Steps:
1. Open a normal web page and open contextual Duck.ai before submitting a prompt.
   Expected: the contextual UTI appears immediately; there is no dashed placeholder chip.
2. Confirm the pre-submit toolbar.
   Expected: the model picker is visible; empty input shows the waveform voice button; typing text changes the keyboard return key to Go and submits.
3. Open the attachment menu.
   Expected: menu order is always `Take Photo`, `Add Image`, `Add File`, `Ask About Page`.

#### Test 2: Manual Page Context and Attachments
Continue from Test 1.

Steps:
1. Tap `Ask About Page`.
   Expected: page context appears in the attachments strip and visually matches image/file chips.
2. Add an image or file attachment as well.
   Expected: attachment spacing and remove buttons look consistent across page context, images, and files.
3. Tap `Summarize Page`.
   Expected: the prompt submits and attached image/file payloads are preserved.
4. Open the tools menu before and after first submit.
   Expected: `Customize Responses` is hidden pre-submit and appears post-submit only.

#### Test 3: Automatic Page Context
Continue with `ff.unifiedToggleInput = true` and `ff.aiChatContextualUnifiedToggleInput = true`.

Steps:
1. Turn automatic context attachment on.
2. Open page A, then open contextual Duck.ai before submitting a prompt.
[...]

PR #5799, opened 2026-07-13
Fix contextual UTI chip resurrection + consolidate delivery-state ownership

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1216466743733545

### Description

Two related changes to the contextual Duck.ai page-context chip, both behind the internal-only contextual-UTI flags.

**1. Fix the chip resurrecting after submit.** After submitting the first prompt with page context, the page-context chip could reappear in the unified input bar as if newly attached, and a follow-up message would then silently carry the old page's content. The cause was a late, passive re-collection of the same page (not a real navigation or attach) being mistaken for a fresh attachment.

**2. Consolidate delivery-state ownership into `AIChatContextualChatSessionState`.** The underlying cause of this whole class of bug was *dual ownership*: both the session state and the UTI chip view model independently tracked "pending vs delivered" and could disagree.

### Manual Tests

Internal only. Enable in **Debug → Feature Flags**, then **restart the app**:
- `unifiedToggleInput`
- `aiChatContextualUnifiedToggleInput` — puts the UTI inside the contextual sheet

Automatic context attachment is toggled per test in **Settings → AI Features → Duck.ai → "Automatically Send Content"**.

#### Test 1: First submit hides the chip and it does not resurrect (auto-attach ON)
Prerequisites: automatic context attachment **ON**.

Steps:
1. Open a web page, tap the Duck.ai button in the address bar.
   Expected: the contextual sheet opens with the page's context already attached.
2. Type a prompt and submit.
   Expected: chat opens; the page-context chip disappears from the input bar (marked delivered).
3. Wait a few seconds (let the page settle), then **without navigating or re-attaching**, type and submit a second prompt.
   Expected: the chip does **not** reappear beforehand, and the second message does **not** silently carry the first page's content.

#### Test 2: Deliberate re-attach of the same page still works (auto-attach ON)
Continue from Test 1 (same page, after the first submit).

Steps:
1. Open the attachment menu → tap **"Ask About Page"** (or "Summarize Page").
   Expected: the page-context chip reappears as **pending**.
2. Submit.
   Expected: the page context **is** sent with this prompt; the chip then hides again.

#### Test 3: Navigation after submit is treated as fresh (auto-attach ON)
From an active contextual chat (after a first submit with context), keep the sheet open.

Steps:
1. Navigate the underlying page to a different URL.
   Expected: the new page's context attaches as **pending** (chip visible).
2. Navigate back to the original page.
   Expected: the original page attaches **freshly as pending** — it is not wrongly treated as already sent.

#### Test 4: Manual mode, follow-up carries nothing (auto-attach OFF)
Prerequisites: automatic context attachment **OFF**.

Steps:
1. On page A, open the sheet, tap **"Ask About Page"** → context attaches. Submit.
   Expected: context sent; the chip hides after submit.
2. Without re-attaching, submit a follow-up prompt.
   Expected: the follow-up carries **no** page context and the chip stays hidden.

#### Test 5: Manual re-attach of an already-submitted page rides the follow-up
Regression test for the specific bug fixed in this PR: a deliberate re-attach of a page that was already submitted must be treated as fresh, not suppressed as an already-delivered echo.

Prerequisites: automatic context attachment can be **ON or OFF** (the re-attach here is manual either way).

Steps:
1. On page A, open contextual Duck.ai and submit a first prompt with page A's context attached.
   Expected: chat opens; the page-context chip hides (delivered).
2. **Without navigating**, open the attachment menu and tap **"Ask About Page"** for the same page A again.
   Expected: the page-context chip reappears as **pending** (attached and visible) — it is **not** left hidden.
3. Submit a follow-up prompt.
   Expected: page A's content **is** included with this follow-up (Duck.ai shows the "Page content from…" attribution), confirming the deliberate re-attach was sent and not silently dropped. The chip then hides again.

### Impact

Low: bug fix + internal refactor for a feature currently behind internal-only feature flags; no impact to users outside internal testing.

#### What could go wrong?
- A deliberate re-attach of the same page right after submitting could be mistaken for the stale case → mitigated by only applying the stale-echo check to automatic re-attachment, never an explicit attach (covered by a test).
- Navigating away and back to the same page after submitting could be wrongly treated as already sent → mitigated by treating any real navigation as resetting the tracking (covered by a test).
- A follow-up prompt could fail to mark its context delivered (chip stays visible / re-carries) → mitigated by firing the delivered transition on every submit via `onPromptDelivered` (covered by a test).

### Quality Considerations

Focused unit suites pass **124/124** (`AIChatContextualChatSessionStateTests`, `UnifiedToggleInputPageContextChipViewModelTests`, `AIChatContextualUTIHostTests`), including new coverage for `utiChipDeliveryState`, `markUTIContextDelivered`, and the `onPromptDelivered` wiring. The full `iOS Browser` app compiles clean, and the contextual-UTI Maestro flows (`test_contextual_web_uti_auto_attach_regression`, `test_contextual_immediate_uti_pre_submit_context`, and both `..._manual_context_clears_after_*navigation`) pass on iOS 26.

### Relationship to other PRs

Stacked on `pete/ios/contextual-uti-part-2` (#5676). No conflict with the sibling pixels PR #5779 — its `voice_tapped` `has_pending_page_context` reads `chipViewModel.pendingAttachedContextData`, whose semantics this change preserves; the two branches touch disjoint regions of `AIChatContextualUTIHost.swift`.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Scoped to internal contextual-UTI flags with broad unit tests; behavior change is limited to page-context chip delivery, not auth or core browsing.
> 
> **Overview**
> Fixes the contextual Duck.ai **page-context chip** reappearing after the first submit with context, and stops follow-up prompts from silently re-attaching the same page when auto-collect sends a late same-URL update.
> 
> **`AIChatContextualChatSessionState`** now tracks the last submitted page URL (until navigation) and treats passive same-document re-collections as **stale echoes**—ignored for auto-attach and pushed to the UTI as **`.delivered`** via `utiChipDeliveryState` / `isStaleEchoOfDeliveredContext`. Navigation and manual attach clear or bypass that marker so deliberate re-attaches and “navigate away and back” still show **pending** context.
> 
> **Pending vs delivered** is owned by session state: `markUTIContextDelivered` runs on every prompt delivery through **`onPromptDelivered`** (coordinator → UTI host → web VC). The chip view model no longer calls `markPromptSubmitted()`; it only reflects `deliveryState` from `setAttached`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7092bbd05a59a8faaeb029792e73cd36aa39e569. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
+++ b/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
@@ -114,10 +114,13 @@ final class AIChatContextualChatSessionState {
     private(set) var contextualChatURL: URL?
     private(set) var latestContext: AIChatPageContext?
 
-    /// URL of the context that was actually included in the last submitted prompt.
-    /// Lets a late auto-attach re-collection for that same page be delivered to the UTI chip
-    /// as already-delivered instead of resurrecting it as pending.
-    private(set) var deliveredContextURL: URL?
+    /// URL of the context that was actually included in the last submitted prompt, so long as no
+    /// navigation has happened since. Used only to recognize a passive auto-attach re-collection
+    /// (e.g. a late JS auto-update) for that same, already-submitted page as a stale echo — not
+    /// a fresh attach — so it doesn't resurrect the UTI chip. Cleared on navigation so a real page
+    /// change is always treated as fresh, and never consulted for manual attach, which is always
+    /// a deliberate user action.
+    private var deliveredContextURLWithNoNavigationSince: URL?
 
     @Published private(set) var viewState = SheetViewState(
         content: .nativeInput,
@@ -231,7 +234,7 @@ final class AIChatContextualChatSessionState {
         case .attached(let context):
             contextData = context.contextData
             frontendState = .chatWithInitialContext
-            deliveredContextURL = URL(string: context.contextData.url)
+            deliveredContextURLWithNoNavigationSince = URL(string: context.contextData.url)
             pixelHandler.firePromptSubmittedWithContext()
             Logger.aiChat.debug("[SessionState] Chat started WITH initial context (chip was attached)")
         case .placeholder:
@@ -260,7 +263,7 @@ final class AIChatContextualChatSessionState {
         switch chipState {
         case .attached(let context):
             frontendState = .chatWithInitialContext
-            deliveredContextURL = URL(string: context.contextData.url)
+            deliveredContextURLWithNoNavigationSince = URL(string: context.contextData.url)
             pixelHandler.firePromptSubmittedWithContext()
             Logger.aiChat.debug("[SessionState] UTI chat started WITH initial context")
         case .placeholder:
@@ -277,7 +280,7 @@ final class AIChatContextualChatSessionState {
         frontendState = .noChat
         chipState = .placeholder
         contextualChatURL = nil
-        deliveredContextURL = nil
+        deliveredContextURLWithNoNavigationSince = nil
         userDowngradedToPlaceholder = false
         isManualAttachInProgress = false
         isManualAttachFromFrontend = false
@@ -348,6 +351,9 @@ final class AIChatContextualChatSessionState {
     func notifyPageChanged(pageURL: URL? = nil) {
         Logger.aiChat.debug("[SessionState] Page navigation detected")
         isProcessingNavigation = true
+        // A real navigation means any subsequent context update is fresh, even if it later
+        // resolves to a URL that was already submitted (e.g. the user navigated away and back).
+        deliveredContextURLWithNoNavigationSince = nil
         if shouldAutoCollectContext, userDowngradedToPlaceholder {
             userDowngradedToPlaceholder = false
             Logger.aiChat.debug("[SessionState] Page navigation cleared temporary context removal")
@@ -516,14 +522,6 @@ final class AIChatContextualChatSessionState {
         return true
     }
 
-    /// Whether `context` is a re-collection of the page whose content was already included in
-    /// the last submitted prompt. Used so a late auto-attach emission for that same page doesn't
-    /// resurrect the UTI chip as pending — it's already been sent.
-    func isContextAlreadyDelivered(_ context: AIChatPageContextData?) -> Bool {
-        guard let context, let deliveredContextURL, let contextURL = URL(string: context.url) else { return false }
-        return contextURL.equals(deliveredContextURL, by: .sameDocument)
-    }
-
     func shouldDeliverToFrontendBridge(_ context: AIChatPageContextData?) -> Bool {
         if isUnifiedToggleInputActive, context != nil {
             Logger.aiChat.debug("[SessionState] shouldDeliverToFrontendBridge=false (non-nil context delivered to UTI)")
@@ -589,8 +587,12 @@ private extension AIChatContextualChatSessionState {
 
             case .attached:
                 chipState = .attached(context)
-                didUpdateAttachment = true
-                Logger.aiChat.debug("[SessionState] Updated attached context (setting ON)")
+                if isStaleEchoOfDeliveredContext(context.contextData) {
+                    Logger.aiChat.debug("[SessionState] Ignoring stale auto-attach echo for already-delivered context")
+                } else {
+                    didUpdateAttachment = true
+                    Logger.aiChat.debug("[SessionState] Updated attached context (setting ON)")
+                }
             }
         } else {
             Logger.aiChat.debug("[SessionState] Context updated on navigation (WebView active, chip not updated)")
@@ -599,6 +601,16 @@ private extension AIChatContextualChatSessionState {
         if didUpdateAttachment || shouldDeliverToFrontendBridge(context.contextData) {
             emitDeliveryIfNeeded(context.contextData)
         }
+    }
+
+    /// Whether `context` is a passive re-collection (e.g. a late JS auto-update) of the same page
+    /// already included in the last submitted prompt, with no navigation in between. Only consulted
+    /// from the auto-attach path — manual attach is always a deliberate user action and always
+    /// delivers as a fresh pending attachment regardless of URL.
+    func isStaleEchoOfDeliveredContext(_ context: AIChatPageContextData) -> Bool {
+        guard let deliveredContextURLWithNoNavigationSince,
+              let contextURL = URL(string: context.url) else { return false }
+        return contextURL.equals(deliveredContextURLWithNoNavigationSince, by: .sameDocument)
     }
 
     func cleanupFlags() {
--- a/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
+++ b/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
@@ -595,10 +595,10 @@ private extension AIChatContextualChatSessionState {
                 }
 
             case .attached:
-                chipState = .attached(context)
                 if isStaleEchoOfDeliveredContext(context.contextData) {
                     Logger.aiChat.debug("[SessionState] Ignoring stale auto-attach echo for already-delivered context")
                 } else {
+                    chipState = .attached(context)
                     didUpdateAttachment = true
                     Logger.aiChat.debug("[SessionState] Updated attached context (setting ON)")
                 }
--- a/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
+++ b/iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
@@ -114,6 +114,11 @@ final class AIChatContextualChatSessionState {
     private(set) var contextualChatURL: URL?
     private(set) var latestContext: AIChatPageContext?
 
+    /// URL of the context that was actually included in the last submitted prompt.
+    /// Lets a late auto-attach re-collection for that same page be delivered to the UTI chip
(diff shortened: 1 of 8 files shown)


iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -114,6 +114,14 @@ final class AIChatContextualChatSessionState {
     private(set) var contextualChatURL: URL?
     private(set) var latestContext: AIChatPageContext?
 
+    /// URL of the context that was actually included in the last submitted prompt, so long as no
+    /// navigation has happened since. Used only to recognize a passive auto-attach re-collection
+    /// (e.g. a late JS auto-update) for that same, already-submitted page as a stale echo — not
+    /// a fresh attach — so it doesn't resurrect the UTI chip. Cleared on navigation so a real page
+    /// change is always treated as fresh, and never consulted for manual attach, which is always
+    /// a deliberate user action.
```
aataraxiaa: comment too lengthy, redude to 1 line max

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -508,6 +522,25 @@ final class AIChatContextualChatSessionState {
         return true
     }
 
+    /// The delivery state the UTI chip should adopt for `context`. A passive re-collection of the
+    /// context already submitted for this page (a stale echo) is `.delivered` so the chip stays
+    /// hidden; anything else is a fresh `.pendingSubmit` attachment. This keeps the pending-vs-delivered
+    /// decision with the state owner rather than the coordinator hardcoding `.pendingSubmit`.
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -508,6 +522,25 @@ final class AIChatContextualChatSessionState {
...
+    func utiChipDeliveryState(forDelivering context: AIChatPageContextData) -> PageContextAttachmentDeliveryState {
+        isStaleEchoOfDeliveredContext(context) ? .delivered : .pendingSubmit
+    }
+
+    /// Records that the currently-attached context has been delivered in a submitted prompt, so it
+    /// stops riding subsequent prompts and the UTI chip hides. This is the sole owner of the
+    /// pending→delivered transition: the chip no longer flips its own delivery state on submit —
+    /// it is re-rendered as delivered through the normal delivery pipeline. Safe to call on every
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -541,6 +574,11 @@ private extension AIChatContextualChatSessionState {
         if isShowingNativeInput || isUnifiedToggleInputActive {
             chipState = .attached(context)
             userDowngradedToPlaceholder = false
+            // A deliberate manual attach is always a fresh pending attachment, even for a page
+            // already submitted this session. Clear the delivered marker so `utiChipDeliveryState`
+            // does not misclassify it as a stale echo and hide it — stale-echo suppression applies
+            // to passive auto re-collection only, never to an explicit user attach.
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -585,6 +627,16 @@ private extension AIChatContextualChatSessionState {
         }
     }
 
+    /// Whether `context` is a passive re-collection (e.g. a late JS auto-update) of the same page
+    /// already included in the last submitted prompt, with no navigation in between. Only consulted
+    /// from the auto-attach path — manual attach is always a deliberate user action and always
+    /// delivers as a fresh pending attachment regardless of URL.
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualUTIHost.swift
```diff
@@ -141,8 +144,11 @@ final class AIChatContextualUTIHost: UnifiedToggleInputDelegate {
         coordinator.observeChatUpdates(publisher)
     }
 
-    func markPromptSubmitted() {
-        chipViewModel.markPromptSubmitted()
+    /// Called when a prompt that carried page context has been delivered. Routes through the
+    /// session state (via `onPromptDelivered`), which owns the pending→delivered transition and
+    /// re-renders the chip — the chip is no longer told to hide itself directly.
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualUTIHost.swift
```diff
@@ -42,6 +42,9 @@ final class AIChatContextualUTIHost: UnifiedToggleInputDelegate {
     var onAttachRequested: (() -> Void)?
     var onRemoveRequested: (() -> Void)?
     var onPromptSubmitted: (() -> Void)?
+    /// Fires on every prompt delivery (first and follow-up) so the session state can mark the
+    /// attached context delivered and re-render the chip. The chip no longer hides itself.
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGo/UnifiedToggleInput/PageContextChip/UnifiedToggleInputPageContextChipViewModel.swift
```diff
@@ -56,8 +56,10 @@ final class UnifiedToggleInputPageContextChipViewModel: ObservableObject {
     private(set) var attachedContext: AIChatPageContext?
     private var attachedURL: URL?
     private var originatingURL: URL?
-    /// Whether the current attachment is waiting to be included in a prompt or has already
-    /// been delivered. `markPromptSubmitted()` flips pending attachments to delivered.
+    /// Whether the current attachment is waiting to be included in a prompt or has already been
+    /// delivered. Presentation-only: this is set solely by what the session state pushes in via
+    /// `setAttached(_:deliveryState:)` — the chip never decides delivery itself. The session state
+    /// owns the pending→delivered transition and re-pushes `.delivered` on submit.
```
aataraxiaa: 1 line comment max

iOS/DuckDuckGoTests/AIChat/AIChatContextualChatSessionStateTests.swift
```diff
@@ -1019,6 +1019,150 @@ final class AIChatContextualChatSessionStateTests: XCTestCase {
...
+                if case .deliverPageContext(let data, let targets) = effect, targets.contains(.utiChip) {
+                    delivered.append(data)
+                }
+            }
+            .store(in: &cancellables)
+        block()
+        return delivered
+    }
+
+    /// Regression test for the bug where, after submitting the first prompt WITH context via the
```
aataraxiaa, line 1037: remove all comments you added to test code, just keep bare // given/when/then

jaceklyp:
Two things flagged by an agent, ignore if you find it invalid:

AIChatContextualChatSessionState.swift:597 — stored context is overwritten before the stale-echo check. In updateContext's .attached branch, chipState = .attached(context) runs first, then the isStaleEchoOfDeliveredContext check only decides whether to set didUpdateAttachment. So a stale same-page re-collection (updated title) still mutates the stored attached context even though the intent is to ignore it. It's hidden (.delivered), so likely harmless today — but if the chip is later re-surfaced (manual re-attach, navigation) it reflects the echo's title, not what was actually submitted. Worth confirming that's intended vs. moving the assignment inside the else.

AIChatContextualUTIHost.swift:255,275 + AIChatContextualWebViewController.swift:215 — onPromptDelivered can fire twice per submission. The delegate submit methods fire onPromptDelivered?(), and if that path also routes through AIChatContextualWebViewController.submitPrompt (which calls notifyPromptDelivered() when pageContext != nil), markUTIContextDelivered runs twice. It's idempotent (same URL set, emitDeliveryIfNeeded guarded), so no observable bug — flagging only because the double-path isn't obvious.

jaceklyp (approved):
works as intended! good job

aataraxiaa:
Thanks @jaceklyp (and the agent) — both are fair.

**#1 — `chipState` overwritten before the stale-echo check:** good catch. Harmless today since the chip's already hidden (`.delivered`) at that point, but you're right that it's a latent inconsistency — the stored context could reflect the echo's title rather than what was actually submitted if it's ever re-surfaced. Moved the `chipState = .attached(context)` assignment into the `else` so a stale echo no longer mutates stored state, and added a test that the stored context is unchanged after an echo. (In a local commit — will be in my next push here.)

**#2 — `onPromptDelivered` can fire twice:** agreed, and thanks for confirming it's idempotent — `markUTIContextDelivered` sets the same URL and the re-delivery re-classifies as a stale echo, so there's no observable effect. I'm leaving the two paths as-is on purpose: they cover different submission routes (direct UTI/native delivery vs. the FE-ack via `handlePromptSubmittedFromUserScript`), and collapsing them risks dropping a route. Happy to add a short comment making the double-path + idempotency explicit if you'd prefer.

PR #5824, opened 2026-07-14
[macOS] Build once for PR test checks to avoid CopySwiftLibs flake

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216515186311636?focus=true

### Description

The macOS PR checks intermittently fail during the integration test step — not because a test failed, but because the build itself errors out with "No such file or directory," and a re-run passes. The job builds the app twice, once for each test suite, and only the second build hits this: it reuses the first build's output and just re-runs the final packaging steps, where two of them overlap and one looks for a file another is mid-way through replacing. The first, full build doesn't have the problem.

This builds the app once and runs both test suites against that single build, so the fragile second build no longer happens. The two suites still run and report their results separately, as before.

### Testing Steps
1. Verify the CI is green.

### Impact

None — CI tooling only, no app-facing change.

### Quality Considerations

Removing the redundant incremental build also trims a little CI time from every macOS PR run.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Workflow-only change to Xcode invocation order; no app or release artifact behavior.
> 
> **Overview**
> macOS PR test jobs now **compile once** with `xcodebuild build-for-testing`, then run unit and integration suites with **`test-without-building`** against the same `DerivedData` output instead of invoking a second full `xcodebuild test` for integration (which could flake during incremental CopySwiftLibs-style packaging).
> 
> The former combined build+unit step is split into dedicated **Build for testing** and **Run unit tests** steps; integration drops compile-related flags that only applied to rebuilding. A new **Upload failed build log** artifact captures `${{ matrix.flavor }}-build-for-testing-xcodebuild.log` on failure or cancel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c151f9cacf5607c78dc8fba34e824588503f8d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/macos_pr_checks.yml
```diff
@@ -259,24 +259,35 @@ jobs:
...
-          -maximum-test-execution-time-allowance 120 \
           ENABLE_TESTABILITY=true \
           ONLY_ACTIVE_ARCH=${{ matrix.active-arch }} \
           COMPILATION_CACHE_ENABLE_CACHING=YES \
           OTHER_CODE_SIGN_FLAGS=--timestamp=none \
+          | tee ${{ matrix.flavor }}-build-for-testing-xcodebuild.log \
+          | sed -e 's/CoreData: error:/CoreData: [notice]:/g' -e 's/Failed to create NSXPCConnection/[notice] NSXPCConnection unavailable/g' \
+          | xcbeautify
+
+    - name: Run unit tests
```
diegoreymendez, line 280: Runs the unit suite against the built product. It keeps the id: `run-unit-tests` the old step had, so every downstream step reading that outcome works untouched.  The test-timeout flags and `OS_ACTIVITY_MODE` we removed from the build step now live here, where they apply.

.github/workflows/macos_pr_checks.yml
```diff
@@ -259,24 +259,35 @@ jobs:
         restore-keys: |
           xcode-compilation-cache-${{ matrix.flavor }}-
 
-    - name: Build and run unit tests
-      id: run-unit-tests
+    - name: Build for testing
```
diegoreymendez, line 262: The old step compiled the app and ran unit tests.  It's now `build-for-testing`, which only compiles.  Run-time flags (test timeouts, `OS_ACTIVITY_MODE`) move to the run step; the compile/link/sign settings stay because this is now the only building step.  The `tee+xcbeautify` tail matches how the test steps already log.

.github/workflows/macos_pr_checks.yml
```diff
@@ -288,18 +299,13 @@ jobs:
     - name: Run integration tests
       id: run-integration-tests
       run: |
-        set -o pipefail && xcodebuild test \
+        set -o pipefail && xcodebuild test-without-building \
```
diegoreymendez, line 304: Was `xcodebuild test`, which triggered a second incremental build — the one that raced `CopySwiftLibs` against the framework copies and produced the "No such file or directory" flake. Now `test-without-building` reuses the existing build, so that second build never happens. The build-only flags and settings are dropped since nothing compiles here.

.github/workflows/macos_pr_checks.yml
```diff
@@ -430,6 +436,14 @@ jobs:
       with:
         pixel-name: m_ci_apple_status_mac_main_${{ matrix.flavor }}_integration_tests_${{ steps.run-integration-tests.outcome }}
 
+    - name: Upload failed build log
```
diegoreymendez, line 442: Because the build is now its own step, a compile break fails there and produces its own log. This uploads that log as an artifact on failure, mirroring the existing per-log upload steps — otherwise a build failure would leave no raw log to inspect.

samsymons:
I'm just looking into a test suite crash on this branch, it's happened two runs in a row and I can reproduce it locally. I don't think that it is the fault of this PR though.

samsymons (approved):
The overall diff looks good to me, but CI seems in very bad shape right now - after 4 attempted runs I am unable to get this branch to pass. The test suite crash seems unrelated. I don't believe the issue is caused by these changes, but I would still hesitate to merge until we can see a green run.

samsymons:
UI tests are now failing with a few new issues:

```
Failed to get matching snapshot: Lost connection to the application (pid 61256). (Underlying Error: Couldn’t communicate with a helper application. Try your operation again. If that fails, quit and relaunch the application and try again. connection to service created from an endpoint
```

samsymons:
PR is green 🎉

PR #5852, opened 2026-07-15
[iOS] Implement multi-select burn in Duck.ai chat history

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216558977091671
CC:

### Description
Adds multi-select "burn" to the native Duck.ai chat history screen.

### Testing Steps
1. Enable the `aiChatHistoryMultiselect` feature flag.
2. Open Duck.ai chat history with a few chats present.
3. Tap the overflow (•••) menu → **Select Chats** → selection mode begins.
4. Select one or more chats → tap **Delete** → confirmation reads "Delete N chats?" → confirm → only the selected chats are burned and selection mode exits.
5. Re-enter selection mode, select nothing, tap **Delete** (labelled "Delete All") → confirmation reflects the full count → confirm → all chats are burned.

### Impact
Medium: touches chat deletion, so a fault could delete the wrong chats or fail to delete selected ones. Gated behind the `aiChatHistoryMultiselect` flag and covered by unit tests at the view-model and `HistoryCleaner` layers.

#### What could go wrong?
- **Wrong or leftover chats deleted** → mitigated by keying deletion off stable `chatId`s (not index paths), and a `HistoryCleaner` test asserting only the selected files/rows are removed with a single JS clearing session.
- **Shared-package blast radius:** the batch path extends the shared `AIChat` package (`HistoryCleaning` / `AIChatJSDataCleaning` protocols, `HistoryCleaner`, `WebViewAIChatJSDataCleaner`). macOS is unaffected — it conforms to its own `AIChatHistoryCleaning` protocol and does not call the new methods. The existing single/all `clearJSData(chatID:)` behavior is unchanged (`nil` → all, `"x"` → `["x"]`), and the native↔JS `duckAiClearData` message contract is **not** touched — the batch just re-sends the existing single-chat message within one already-navigated session.
- **JS clearing failure** → the existing JS-error telemetry (`fireLoadJSFailedPixelIfNeeded`) is preserved on the batch path.

### Open question — is the batch worth it?
The work is split into two commits:
- **Commit 1** — the multi-select burn itself, using the simple sequential loop.
- **Commit 2** — the batch optimization (`deleteChats` / `burnChats` / `clearJSData(chatIDs:)` peers across four layers) so the clearing web view is built once instead of per chat.

The optimization is an **O(N) → O(1) reduction on web-view builds + page navigations**, but the per-chat JS clear messages stay O(N) — so the benefit is proportional to selection size: modest for a handful of chats, meaningful for dozens. **Is that worth the added cross-layer surface (and the shared-package change), or should we ship the simpler loop and revisit only if large multi-deletes prove common in practice?** If we drop it, it's a clean revert of commit 2.

Note: multi-select-specific analytics are intentionally deferred to the Pixels task (1216558977091677); the batch path currently reuses the existing `aiChatHistoryDelete{Successful,Failed}` operational pixels.

---

> [!NOTE]
> **Medium Risk**
> Touches chat deletion and shared AIChat clearing protocols; wrong-ID handling could delete the wrong chats, though behavior is flag-gated and covered by HistoryCleaner/view-model tests.
> 
> **Overview**
> Adds **multi-select delete and download** on the native Duck.ai chat history sheet (behind `aiChatHistoryMultiselect`). Selection mode wires **Delete** to a shared confirmation (`deleteChats` count) that either burns the chosen `chatId`s or, with nothing selected, deletes all; **Download** exports each selected chat and shows aggregate success/failure toasts.
> 
> Deletion is plumbed through **`burnChats` / `deleteAIChats`** so native storage lists files once and the headless JS cleaner runs **one WebView session** per batch (still one clear message per chat per domain). Fire confirmation’s `deleteAllChats` context is generalized to **`deleteChats`**. Export logic is refactored to **`exportChats`** with new delegate callbacks; **`MainViewController+AIChatHistoryViewModelDelegate`** holds the toast/navigation handling. Localization drops redundant `deleteAll.confirmation.title` entries from locale strings (stringsdict path unchanged).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52079555bc18305e9f64af579088fb7cfe8eba93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Bunn (approved):

iOS/DuckDuckGoTests/AIChat/AIChatHistoryViewModelTests.swift
```diff
@@ -203,6 +203,43 @@ final class AIChatHistoryViewModelTests: XCTestCase {
...
+        let done = expectation(description: "burnSelectedChats")
+        Task { await sut.burnSelectedChats(chatIds: []); done.fulfill() }
+        wait(for: [done], timeout: 1)
+
+        XCTAssertTrue(fireExecutor.burnedChatsBatches.isEmpty)
+        XCTAssertEqual(fireExecutor.scheduleSyncCallCount, 0)
+    }
+
+    func testBurnSelectedChats_noFireExecutor_isNoOp() {
+        let sut = makeSUT(chats: [chat(id: "p1", pinned: true)], fireExecutor: nil)
```
Bunn: I think we can remove this test

iOS/DuckDuckGoTests/AIChat/AIChatHistoryManagerTests.swift
```diff
@@ -305,6 +305,14 @@ private final class MockHistoryCleaner: HistoryCleaning {
         deleteAIChatCalls.append(chatID)
         return deleteAIChatResult
     }
+
+    var deleteAIChatsResult: Result<Void, Error> = .success(())
+    private(set) var deleteAIChatsCalls: [[String]] = []
```
Bunn: doesn't seem to be used for anything

SharedPackages/AIChat/Sources/AIChat/Shared/HistoryCleaner/HistoryCleaner.swift
```diff
@@ -65,6 +66,19 @@ public final class HistoryCleaner: HistoryCleaning {
...
+    /// JS clearing session, rather than repeating both per chat.
+    @MainActor
+    public func deleteAIChats(chatIDs: [String]) async -> Result<Void, Error> {
+        let nativeResult = clearLocalStorageIfAvailable(chatIDs: chatIDs)
+        let jsResult = await jsDataCleaner.clearJSData(chatIDs: chatIDs)
+
+        if case .failure = nativeResult {
+            return nativeResult ?? jsResult
+        }
+        return jsResult
```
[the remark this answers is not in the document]
SabrinaTardio: The native-before-JS ordering isn't new here; performClear already deletes native storage before JS clearing, and clear-all has the same "native fully removed even if the single JS clear fails" property. The batch reuses that established ordering rather than changing the native↔JS contract.

PR #5857, opened 2026-07-16
[macOS] Fix fire-window downloads attributed to regular session

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1214167538194692
Tech Design URL: N/A
CC: N/A

### Description
- Add `WKWebsiteDataStore+FireWindowSession.swift`: associates a `FireWindowSessionRef` with a non-persistent data store when its Fire Window is created, enabling session lookup before the originating WebView is in the window hierarchy
- Add `FireWindowSession.websiteDataStore` weak back-reference and `FireWindowSessionRef.init?(session:)` for direct session-to-ref conversion without requiring an `NSWindow`
- Link data store ↔ session in `WindowsManager.makeNewWindow` for both new and reused sessions
- Replace direct `FireWindowSessionRef(window: webView.window)` calls in `DownloadsTabExtension` with `resolveFireWindowSession(for:)`, which prefers the data store lookup and falls back to the window
- Throttle download progress updates in `DownloadsCellView` to reduce UI churn
- Re-enable `testFireWindowWithInProgressDownloadShowsWarning` in both test plans

### Testing Steps
1. Open a Fire Window (⌘⇧N)
2. Paste a direct download URL into the address bar and press Enter
3. Verify the download appears in the Fire Window's downloads panel, not the regular window's
4. Close the Fire Window while the download is in progress — confirm the warning sheet appears
5. UI Tests run: https://github.com/duckduckgo/apple-browsers/actions/runs/29482179565

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- `FireWindowSessionRef` holds only a `weak` reference to `FireWindowSession`, so the associated object on `WKWebsiteDataStore` adds no strong retain to the session's lifetime

### Quality Considerations
- No changes to `BurnerMode`, `Tab.init`, or `TabExtensionsBuilderArguments` — minimal blast radius
- The `assert(!isBurner || session != nil)` in `enqueueDownload` now correctly fires when the data store association is missing, rather than when the window is transiently nil
- Persistent data stores return `nil` from `fireWindowSession` unchanged

### Notes to Reviewer
Root cause: when a URL is pasted directly into the address bar of a Fire Window tab, `enqueueDownload` is called before the WebView has been added to the window hierarchy (`webView.window == nil`). `FireWindowSessionRef(window: nil)` returns `nil`, causing the download to be attributed to the regular session instead of the fire session.

### DIFF
--- a/macOS/DuckDuckGo-macOS.xcodeproj/project.pbxproj
+++ b/macOS/DuckDuckGo-macOS.xcodeproj/project.pbxproj
@@ -56,9 +56,7 @@
 		1D01A3D82B88DF8B00FE8150 /* PreferencesSyncView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1D01A3D72B88DF8B00FE8150 /* PreferencesSyncView.swift */; };
 		1D01A3D92B88DF8B00FE8150 /* PreferencesSyncView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1D01A3D72B88DF8B00FE8150 /* PreferencesSyncView.swift */; };
 		1D04CA182F67FECE007AB0F3 /* PageContextMultipleContextsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1D04CA172F67FECE007AB0F3 /* PageContextMultipleContextsTests.swift */; };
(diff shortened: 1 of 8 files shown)


ayoy (approved):
LGTM, thanks @mallexxx! I believe that the test failure on macOS 15 is unrelated.

mallexxx:
@ayoy, actually it indicated a regression. Fixed now

PR #5858, opened 2026-07-16
[macOS] Build UI Tests for arm64 only

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201037661562251/task/1216591205063619?focus=true
Tech Design URL: N/A
CC: N/A

### Description
- Build Review UI Test prebuilds for arm64 only by excluding x86_64 during notarized archives.
- Add the `include-x86-architecture` notarized-build input, defaulting to `false`, and forward the resulting architecture exclusion to `xcodebuild archive`.
- Keep Alpha and release DMG workflows universal by explicitly enabling x86_64.
- Configure DMG, App Store, and CI UI Test schemes to match the active run-destination architecture.

### Testing Steps
1. Ran the macOS UI Tests workflow and inspected its notarized app artifact: the main app, VPN login item, system extension, and PIR login item are arm64-only.
2. Verified the PR “Make Release Build” check passes and builds both arm64 and x86_64 Swift modules.
3. Validated workflow YAML, scheme XML, shell syntax, and whitespace.

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- Review builds used on an Intel runner require `include-x86-architecture: true`. Current UI Test runners are Apple Silicon only.
- Alpha and release callers must retain their explicit x86_64 opt-in to preserve universal DMGs.

### Quality Considerations
- No user-facing, privacy, security, analytics, or documentation changes.
- The local `archive.sh` invocation is unchanged unless an architecture exclusion is provided through its environment.

### Notes to Reviewer
- The full UI Test matrix had unrelated remote-page assertion and artifact-upload DNS failures; its notarized prebuild completed successfully.

### DIFF
--- a/.github/workflows/macos_build_notarized.yml
+++ b/.github/workflows/macos_build_notarized.yml
@@ -24,6 +24,10 @@ on:
         description: "Asana release task URL"
         required: false
         type: string
+      include-x86-architecture:
+        description: "Include x86_64 architecture in the build"
+        default: false
+        type: boolean
   workflow_call:
     inputs:
       release-type:
@@ -54,8 +58,8 @@ on:
         description: "Skip Mattermost notification"
         default: false
         type: boolean
-      build-active-architecture-only:
-        description: "Build only the active architecture"
+      include-x86-architecture:
+        description: "Include x86_64 architecture in the build"
         default: false
         type: boolean
     secrets:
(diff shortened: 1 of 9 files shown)


.github/workflows/macos_build_notarized.yml
```diff
@@ -54,6 +54,10 @@ on:
         description: "Skip Mattermost notification"
         default: false
         type: boolean
+      build-active-architecture-only:
```
ayoy: This isn't precisely working as advertised, as it just filters out intel architecture. Could you call it `build-all-architectures` and reverse the logic (default to true)? Alternatively call it perhaps `build-only-for-arm`.

ayoy (commented):
This is working well 👏  I'd like that we renamed the input to be more clear about what it actually does.

ayoy (approved):
Thanks!

PR #5869, opened 2026-07-16
[iOS] Fix Unified Toggle Input layout breaking on new-tab / modal transitions

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1216428438612098?focus=true

also: https://app.asana.com/1/137249556945/project/1206488453854252/task/1216428438612104

### Description

Fixes several ways the Unified Toggle Input (the new address-bar input, internal-only / iPhone-only) broke when moving from an active input into another surface — most visibly, launching a new tab from the tab-switcher long-press while the input was focused left the input card collapsed and floating in the middle of the screen.

### Testing Steps

Run on an iPhone with the Unified Toggle Input enabled. Repeat the input-focused steps in **both** address-bar positions (top and bottom).

1. On any tab, tap the address bar to focus the input → swipe down to reveal the toolbar → long-press the Tab Switcher button → New Tab. → The input sits at its normal resting position (top / above keyboard), full height, not floating mid-screen.
2. With the input focused on a web page, long-press Tab Switcher → New Tab. → The input focuses like a normal address-bar tap — no cursor/card sliding in from the left.
3. Repeat step 2 with the input focused on the new tab page. → Same clean animation as step 2.
4. Focus the input on the new tab page → tap the back arrow → tap the Tab Switcher. → No crash; the tab switcher opens on the first tap.
5. Focus the input → open the Tab Switcher. → The input disappears cleanly as the switcher slides in, with no overlap; closing/reselecting the tab returns you with content intact.
6. On the new tab page (logo showing), tap the address bar → open Bookmarks → dismiss. → The Dax logo is back. Repeat via Settings, and via Tab Switcher (swipe down → Tab Switcher → reselect the tab).
7. With favorites present (new tab page shows the favorites grid), repeat step 6. → The favorites grid returns.
8. New tab page → focus the input → tap the back arrow. → Logo/favorites restore smoothly with no flash, and the input collapses normally.

### Impact

Low — bug fixes to an internal-only, iPhone-only feature. The address-bar-anchor change is gated so it only alters behaviour while the input is visible and top-anchored; the shared snapshot/preview guards are defensive.

#### What could go wrong?

- New-tab teardown was consolidated (removed a redundant pre-dismiss) → mitigated: it now matches the ~17 other new-tab entry points, and the legacy omnibar still resets via its existing end-editing path.
- The instant (non-animated) dismiss is also reached by the Duck.ai-tab dismiss → mitigated: that path already ran the same layout restore, which is idempotent.

### Quality Considerations

No automated coverage — the change is UI layout/animation timing verified on-device via the steps above; the underlying container has no unit-testable seam.

---

> [!NOTE]
> **Low Risk**
> Internal iPhone-only UTI UI timing and anchoring fixes; behavior is gated to visible UTI and specific entry points, with defensive snapshot guards.
> 
> **Overview**
> Fixes **Unified Toggle Input (UTI)** breaking when leaving an active address-bar session for another surface (notably **new tab** from tab-switcher long-press and **tab switcher** while focused).
> 
> **Transition timing:** Adds optional **non-animated** teardown on `dismissOmniBar`, `performCancel`, and UTI session deactivation. **New tab** and **tab switcher** use instant dismiss so UTI collapse doesn’t race the following segue/focus animation. Tab-switcher **new-tab** shortcuts drop a redundant pre-`newTab()` cancel because `newTab()` already tears down editing.
> 
> **Layout:** When UTI dismiss runs **without** animation, the non-animated path now runs the same **resting layout restore** as the animated dismiss before cleanup, avoiding stranded content. **Bottom content anchoring** only forces the UTI top anchor when the nav bar is bottom-anchored (`navigationBarContainerTop` inactive), so top-position UTI no longer incorrectly pins content to the input.
> 
> **NTP / previews:** Shared UTI hide teardown **un-hides the NTP logo and favorites** (matching animated back dismiss). Tab **preview** updates always invoke their completion even when snapshots fail or are nil; **zero-size** snapshots are skipped to avoid graphics asserts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df73f2a2355e5356b8bf3889a9c0c5679a9fb78c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jaceklyp (commented):
A few notes on `setContentContainerBottomAnchorMode`. Centralizing the "UTI owns the bottom anchor" override in one place is the right instinct — beats repeating the guard across all ~14 callers — but the contract and the predicate can be cleaner. #3 is a genuine question, not a blocker.

iOS/DuckDuckGo/MainViewCoordinator.swift
```diff
@@ -720,9 +720,12 @@ class MainViewCoordinator {
...
     }
 
-    /// The UTI owns the bottom anchor whenever it's visible and not voice-mode-hidden, regardless of what a caller requests.
+    /// UTI owns the content bottom anchor only when bottom-anchored; forcing it for the top address bar collapses the UTI container.
     private func setContentContainerBottomAnchorMode(_ mode: ContentContainerBottomAnchorMode) {
-        let resolvedMode: ContentContainerBottomAnchorMode = (isUnifiedToggleInputVisible && !navigationBarContainer.isHidden) ? .unifiedToggleInput : mode
+        let utiOwnsBottomAnchor = isUnifiedToggleInputVisible
+            && !navigationBarContainer.isHidden
+            && !constraints.navigationBarContainerTop.isActive
+        let resolvedMode: ContentContainerBottomAnchorMode = utiOwnsBottomAnchor ? .unifiedToggleInput : mode
```
jaceklyp: **Major:** This line silently discards the caller's `mode` whenever `utiOwnsBottomAnchor` is true, but the signature (line 724) promises to set the mode unconditionally. So `setContentContainerBottomAnchorMode(.toolbar)` at e.g. `MainViewCoordinator.swift:567` / `:729`-context callers may not set toolbar mode at all, and there's no way to tell that from the call site. Same family as a mid-function state bail-out — the caller's intent is dropped based on hidden state.

Keeping the override here is correct (one chokepoint > 14 duplicated guards). The only issue is the honesty of the signature. Name the param as a request:

Current:
```swift
private func setContentContainerBottomAnchorMode(_ mode: ContentContainerBottomAnchorMode) {
    ...
    let resolvedMode = utiOwnsBottomAnchor ? .unifiedToggleInput : mode
```
Suggested:
```swift
private func setContentContainerBottomAnchorMode(requesting mode: ContentContainerBottomAnchorMode) {
    ...
    let resolvedMode = utiOwnsBottomAnchor ? .unifiedToggleInput : mode
```
Call sites then read `setContentContainerBottomAnchorMode(requesting: .toolbar)` — signals "may be overridden." Deeper (follow-up, not this PR): if UTI ownership is a hard invariant, several callers arguably shouldn't be passing a competing mode while UTI is up at all.

iOS/DuckDuckGo/MainViewCoordinator.swift
```diff
@@ -720,9 +720,12 @@ class MainViewCoordinator {
         }
     }
 
-    /// The UTI owns the bottom anchor whenever it's visible and not voice-mode-hidden, regardless of what a caller requests.
+    /// UTI owns the content bottom anchor only when bottom-anchored; forcing it for the top address bar collapses the UTI container.
     private func setContentContainerBottomAnchorMode(_ mode: ContentContainerBottomAnchorMode) {
-        let resolvedMode: ContentContainerBottomAnchorMode = (isUnifiedToggleInputVisible && !navigationBarContainer.isHidden) ? .unifiedToggleInput : mode
+        let utiOwnsBottomAnchor = isUnifiedToggleInputVisible
```
jaceklyp: **Minor:** This 3-condition boolean is a distinct concept — "UTI is currently bottom-anchored" — separate from "activate the constraints for a mode". Worth lifting to a named computed property so it self-documents and is reusable: `showNavigationBarWithBottomPosition` (`:284`) and `anchorContentContainerToInputTop` (`:716`) each reason about pieces of this same condition today.

Current:
```swift
let utiOwnsBottomAnchor = isUnifiedToggleInputVisible
    && !navigationBarContainer.isHidden
    && !constraints.navigationBarContainerTop.isActive
let resolvedMode = utiOwnsBottomAnchor ? .unifiedToggleInput : mode
```
Suggested:
```swift
private var isUnifiedToggleInputBottomAnchored: Bool {
    isUnifiedToggleInputVisible
        && !navigationBarContainer.isHidden
        && !constraints.navigationBarContainerTop.isActive
}
...
let resolvedMode = isUnifiedToggleInputBottomAnchored ? .unifiedToggleInput : mode
```

iOS/DuckDuckGo/MainViewCoordinator.swift
```diff
@@ -720,9 +720,12 @@ class MainViewCoordinator {
...
+            && !navigationBarContainer.isHidden
+            && !constraints.navigationBarContainerTop.isActive
```
jaceklyp, line 726: **Question:** This infers "address bar is at the bottom" from a constraint's `.isActive`. There's a semantic property for exactly that — `addressBarPosition.isBottom` (used at `:282` and `:746`) — which reads better:
```swift
&& addressBarPosition.isBottom
```
I checked before suggesting the swap: `navigationBarContainerTop.isActive` is toggled by `setAddressBarTopActive(_:)` (`:308`), *separately* from `addressBarPosition`, so mid-transition the live constraint state and the logical position can diverge. If you're intentionally keying off the **physical** current anchoring here (during the new-tab / modal transitions this PR fixes), the constraint check is correct and `addressBarPosition.isBottom` would be a subtle behavior change — in which case a one-line `// live constraint, not logical position — transitions` would save the next reader this detour. Which did you intend?

jaceklyp:
1/ some swiftlint issues;
2/ the collapsing of UTI when opening a modal (tab switcher) looks a bit clunky

jaceklyp (approved):

iOS/DuckDuckGo/MainViewCoordinator.swift
aataraxiaa: Addressed in df73f2a235. The helper now takes `requesting requestedMode`, and every call site uses the `requesting:` label so it is clear that live UTI ownership may override the requested anchor.

iOS/DuckDuckGo/MainViewCoordinator.swift
aataraxiaa: Addressed in df73f2a235. I extracted the predicate as `shouldUnifiedToggleInputOwnBottomAnchor`; that name describes the ownership policy without implying that the bottom constraint itself was inspected.

iOS/DuckDuckGo/MainViewCoordinator.swift
aataraxiaa: The live constraint state is intentional here. During these transitions it can diverge from the logical `addressBarPosition`, and content-anchor ownership must follow the physical layout. I kept the constraint check and named the predicate `shouldUnifiedToggleInputOwnBottomAnchor` to make that policy explicit. The PR manual-test scenarios pass with this version.

aataraxiaa:
Followed up on both points: the branch now includes current main, which contains the existing SwiftLint suppression, and fresh CI is running. I also reran the full manual-test scenarios from the PR successfully. The tab-switcher path deliberately restores the UTI layout synchronously before presentation so the collapse cannot overlap the segue; I kept that deterministic ordering rather than reintroducing concurrent animations.

PR #5871, opened 2026-07-16
[macOS] Replace flaky ErrorPageTests with error page UI tests

mallexxx:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1214797526093576?focus=true
Tech Design URL: N/A
CC: N/A

### Description
- Replaces the flaky `ErrorPageTests` integration suite (including `testWhenTabWithConnectionLostErrorActivatedAndReloadFailsAgain_errorPageIsShownOnce`, failing post-merge on main) with an end-to-end `ErrorPageUITests` suite covering all of its scenarios: error page rendering, tab-activation auto-reload (triggered and not triggered), back/forward history preservation through failing reloads, resubmit-same-URL, redirect-then-fail, pinned tabs, and session restoration.
- Adds a `failure://` debug URL scheme to `DuckURLSchemeHandler`, gated behind a new `failureURLScheme` feature flag (disabled by default). It can simulate connection-lost, not-connected, and host-not-found errors (plus an alternating mode), each stamped with an `attempt N` counter so UI tests can observe whether a reload actually re-invoked the handler — making "reloads once", "doesn't auto-reload", and "error description updates" directly verifiable.
- Adds Debug menu items (Debug → failure:// URL scheme) to open the demo variants and toggle the simulated connection error.
- Deflakes the two `AddressBarTests` focus tests by holding the scheme-handler response open until the address bar is re-activated, instead of racing the load with `Task.sleep`.

### Testing Steps
- Launch the app: the Debug → "failure:// URL scheme" submenu should be absent (feature flag off by default) and `failure://` URLs should not resolve.
- Enable the `failureURLScheme` flag via internal feature-flag overrides, then open Debug → failure:// URL scheme → "Open failure:// demo page": the demo document should load.
- Toggle "Simulate failure:// connection error" and Cmd+R: the error page should appear with the simulated description and attempt counter; toggle off and Cmd+R: demo loads again.
- Ensure [UI tests](https://github.com/duckduckgo/apple-browsers/actions/runs/32346463081) pass

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- The `failure://` handling ships in the production target but is inert unless the `failureURLScheme` flag is enabled; every code path is guarded by the flag.
- Dropping the integration suite shifts error-page regression detection to the UI test suite, which runs less frequently than PR-gate integration tests.

### Quality Considerations
- Each removed integration test has a named UI counterpart; the reload/no-reload cases are now observable via the simulated-error attempt counter rather than sleep-window `isLoading` assertions.
- The remaining unique integration coverage (restored session with a forward stack) is exercised by `testErrorPage_SessionRestoration_FailureScheme_NewTabSelectRecover`.

### Notes to Reviewer
- The UI tests launch with `featureFlags: ["failureURLScheme": true]`; production default is disabled.
- `ErrorPageUITests` uses privacy-test-pages.site and the local tests server instead of example.com.


SharedPackages/BrowserServicesKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -100,6 +100,8 @@ extension URL {
         public static let blob = NavigationalScheme(rawValue: "blob")
         public static let about = NavigationalScheme(rawValue: "about")
         public static let duck = NavigationalScheme(rawValue: "duck")
+        /// macOS: custom `WKURLSchemeHandler` demo / debug scheme (`failure://`).
```
alessandroboron, line 103: can we define the below in the macOS scheme since iOS won’t be using them?

```
public static let failureDemo = NavigationalScheme(rawValue: “failure”)
/// Raw value of ``NavigationalScheme/failureDemo`` (macOS URL scheme handler).
public static let failureDemoURLScheme = NavigationalScheme.failureDemo.rawValue

public var isFailureDemoURLScheme: Bool {
   navigationalScheme == .failureDemo
}

```

macOS/DuckDuckGo/Common/Extensions/URLExtension.swift
```diff
@@ -410,7 +410,7 @@ extension URL {
     }
 
     var isExternalSchemeLink: Bool {
-        return ![.https, .http, .about, .file, .blob, .data, .ftp, .javascript, .duck, .webkitExtension].contains(navigationalScheme)
+        return ![.https, .http, .about, .file, .blob, .data, .ftp, .javascript, .duck, .failureDemo, .webkitExtension].contains(navigationalScheme)
```
alessandroboron, line 413: Should we make the URL scheme more generic like `debug://` so that we can re-use it for different purposes in the future?

Instead of `failure://` we could have `debug://<identifier>`. That way we don’t need to add more schemes for other debug/test scenarios

alessandroboron:
@mallexxx I run the test suite on the different macOS version and we got some failures on macOS https://github.com/duckduckgo/apple-browsers/actions/runs/32221198551/job/95973567397

ayoy (approved):
LGTM, thanks @mallexxx!

PR #5895, opened 2026-07-18
macOS: Design Feedback - Fire Window

jleandroperez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1176956903599313/task/1216636964401459?focus=true
Tech Design URL:
CC:

### Description
When a window is in Fire Mode, we'll force the `Dark Aqua`  appearance in the Address Bar component.
This is also applied to both, Suggestions and Duck.ai Omnibar.

This change is gated behind a feature flag, and allows us to match the mockups, while avoiding rewiring (all) the colors, everywhere.

### Testing Steps
1. Enable the `.appRebranding` Flag
2. Relaunch
3. Open a new Fire Window

- [x] Verify the Address Bar (alone) is in Dark Mode
- [x] Verify that all of the Light Mode themes display the Address Bar, in Dark Mode

### Impact

#### What could go wrong?
We could introduce a rendering regression in the Address Bar component

### Quality Considerations
Nothing in particular

### Screenshots

Now | Before
-|-
New - First Responder | Old - First Responder
New - Suggestions | Old - Suggestions
New - Omnibar | Old - Omnibar

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches shared appearance/color resolution across the address bar, suggestions, and AI omnibar; wrong overrides could cause visible rendering regressions in those high-traffic UI surfaces.
> 
> **Overview**
> For **Fire windows** with **app rebranding** enabled, the address bar, suggestions dropdown, and Duck.ai omnibar now render with a **forced dark Aqua** appearance so they match design mocks without re-theming every color.
> 
> A new **`BurnerAppearanceStyle`** sets `darkAqua` on the relevant view hierarchies (and toggles it off on the address bar when inactive). **`ColorView`** gains **`resolvesStyleWithEffectiveAppearance`** so layer colors resolve against the view’s effective appearance instead of always using the app appearance.
> 
> Appearance helpers move to **AppKitExtensions** as **`withAppearance(from:)`** / **`withAppearance(_:)`**; call sites that draw inside overridden views now use those instead of **`withAppAppearance`**. Fire/burner mode is threaded via **`SuggestionContainerViewModel`** and **`isBurner`** on the AI chat text container; the address bar keeps an **active** look for burner tabs even when the window isn’t key.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8606123ca9ef8334afc9d8607252e14974fbed96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/AIChat/Suggestions/AIChatSuggestionsView.swift
```diff
@@ -118,7 +118,7 @@ final class AIChatSuggestionsView: NSView {
     }
 
     private func updateSeparatorColor() {
-        NSAppearance.withAppAppearance {
+        NSAppearance.withAppearance(appearance) {
```
jleandroperez, line 121: `withAppearance` resolves against a specified NSAppearance, and falls back to `NSApp.effectiveAppearance` when nil.

This allows us to preserve the original behavior, unless an explicit NSAppearance instance was set

macOS/DuckDuckGo/NavigationBar/View/AddressBarTextField.swift
```diff
@@ -763,7 +763,6 @@ final class AddressBarTextField: NSTextField {
         NSStoryboard.suggestion.instantiateController(identifier: "SuggestionViewController") { coder in
             let suggestionViewController = SuggestionViewController(coder: coder,
                                                                     suggestionContainerViewModel: self.suggestionContainerViewModel!,
-                                                                    isBurner: self.isBurner,
```
jleandroperez, line 766: I've stumbled upon a race condition, where the `suggestionsViewModel` would get initialized **before** `AddressBarTextField` had access to the correct `isBurner` value.

Rather than passing `isBurner`, we're now resolving against the `suggestionContainerViewModel.isBurner` value, which always contains the right value.

macOS/DuckDuckGo/NavigationBar/View/AddressBarViewController.swift
```diff
@@ -998,8 +1015,8 @@ final class AddressBarViewController: NSViewController {
         let navigationBarBackgroundColor = colorsProvider.navigationBackgroundColor
 
         NSAppearance.withAppAppearance {
-            // Keep selected appearance when AI chat is active, even if window loses key status
-            let shouldShowActiveState = window.isKeyWindow || selectionState == .activeWithAIChat
+            // Keep selected appearance when AI chat is active (OR) isBurner, even if window loses key status
+            let shouldShowActiveState = window.isKeyWindow || selectionState == .activeWithAIChat || isBurner
```
jleandroperez, line 1019: When in Burner Mode, we won't be switching the Address Bar colors if the app becomes inactive.
Reasoning:

Duck.ai Suggestions don't get closed when we're no longer in Foreground, and switching colors, then, would result in an unexpected color transition

macOS/DuckDuckGo/Common/Extensions/NSAppearanceExtension.swift
ayoy, line 1: Since you're touching this file, could you please move the entire file to `AppKitExtensions` module? You'll need to import `AppKitExtensions` here and there for sure, and also update access level to `public`, but that's fine – when we introduced `*Extensions` modules we agreed to not migrate anything unless touched :) thanks!

ayoy (approved):
Great stuff and nice appearance extension. Fire Window is looking great! 🙌

jleandroperez:
@ayoy I've **partially** moved `NSAppearanceExtension` to AppKit extensions (all of the generic code).

I was forced to leave an API that returns a  `ThemeAppearance`, since such enum is declared in the main app.

**Thanks a lot!!**

PR #5959, opened 2026-07-22
EventHub for Apple #1: Add EventHub shared package (core implementation + tests)

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216794817159920
CC:

### Description
First of a stack targeting `michal/event-hub` (the integration branch) — a later PR wires EventHub into the iOS/macOS app targets.

Adds `SharedPackages/EventHub`: config-driven telemetry aggregation for content-scope-scripts web events. A detector emits a named event; EventHub aggregates it per remote config, buckets counts into privacy-safe ranges over a configurable period, and fires a pixel at each period boundary — reaching parity with the Android and Windows implementations already shipped.

Built test-first against the ported Windows test suite (windows-browser PR #7898), plus new tests for pieces that don't have a Windows equivalent. 131 tests total.

Since opening, review surfaced three spec deviations, each now fixed with covering tests: events arriving after a period ended were still counted; per-tab dedup was reset at every period boundary (TD note [2] requires it survive); and a pixel disabled mid-period was torn down rather than firing its final period (TD §6). Dedup now lives in a hub-owned `DedupStore` so it outlives the per-period `Telemetry` objects that consult it.

Two items flagged below for reviewers.

### Testing Steps
1. `cd SharedPackages/EventHub && swift test` → 131/131 passing.

No app to manually test yet — this package isn't referenced by any app target until the integration PR.

### Impact
None — new, standalone package; nothing in the app imports it yet.

### Quality Considerations
- **Pixel suffix is a placeholder.** `EventHub` fires the bare governed config name (no platform suffix); per the Tech Design, the platform-specific `EventHubPixelFiring` conformance (added in the integration PR) is responsible for appending `_ios_phone`/`_ios_tablet`/`_macos`. Worth confirming this split holds before that PR lands.
- This package isn't yet wired into any Xcode scheme's test plan, so `ios_pr_checks.yml`/`macos_pr_checks.yml` won't execute its suite automatically. Only local `swift test` covers it for now — worth adding scheme/test-plan wiring or a dedicated workflow (matching `BrowserServicesKit`/`DataBrokerProtectionCore`) alongside the integration PR.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> New telemetry and consent-gating logic affect what pixels fire and what is persisted, but the package is isolated until integration; no hardcoded secrets were found in the diff.
> 
> **Overview**
> Introduces **`SharedPackages/EventHub`** as a new Swift package and registers it in the workspace. The package implements config-driven telemetry for content-scope **web events** and native ingress, aligned with Android/Windows EventHub behavior.
> 
> The **`EventHub`** runtime parses remote `eventHub` settings into period and immediate pixels, aggregates **counter** and **data** parameters with ordered bucket ranges, fires pixels at period boundaries (with **`attributionPeriod`**), and routes **`WebEventsHandler`** `webEvent` messages from UserScript. **Per-tab dedup** lives in a hub-owned **`DedupStore`** (cleared on URL change or tab close, not on period rollover), with **write-behind persistence** via **`EventHubKeyValueStore`**, a single **scheduler** for period ends and flush deadlines, and **`EventHubSettings`** stripping consent-gated telemetry entries (fail-closed on parse errors).
> 
> **131 tests** cover config parsing, parameters, persistence, scheduling, consent filtering, and end-to-end flows. **No app target imports the package yet** (integration is a follow-up PR).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75ef244fd2acd18461a9ca5cb075dcc1d5c95f49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/EventHub/Sources/EventHub/Config/EventHubConfigParser.swift
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private struct TriggerDTO: Codable {
+        let type: String?
+        let period: PeriodDTO?
+        let source: String?
+    }
+
+    private struct PeriodDTO: Codable {
+        let seconds: Int
```
bwaresiak, line 150: So yes, the spec was showing more cases (days, etc), but based on what you told me, it gets all processed into seconds, right?

SharedPackages/EventHub/Sources/EventHub/Config/EventHubConfigParser.swift
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private struct BucketDTO: Codable {
+        let gte: Int?
+        let lt: Int?
+    }
+
+    /// Decodes/encodes the `buckets` JSON object preserving key order (`BucketCounter` is
+    /// first-match-wins), relying on `KeyedDecodingContainer.allKeys` returning keys in document order.
+    /// A bucket without a lower bound (`gte`) is invalid and disqualifies the whole counter.
```
bwaresiak, line 167: I don’t think `allKeys` preserve orde... but not that it matters, since based on spec overlap should not matter - still the comment is misleading.

SharedPackages/EventHub/Sources/EventHub/Config/EventHubConfigParser.swift
miasma13: That is correct. This was introduced during Windows implementation and is highlighted as [3] in https://app.asana.com/1/137249556945/project/481882893211075/task/1215327441118832?focus=true

SharedPackages/EventHub/Sources/EventHub/Persistence/EventHubStore.swift
```diff
@@ -0,0 +1,106 @@
...
+    private func readMap() -> [String: EventHubStoredPixelState] {
+        guard let data = try? store.object(forKey: Self.storageKey) as? Data,
+              let map = try? JSONDecoder().decode([String: EventHubStoredPixelState].self, from: data) else {
+            return [:]
+        }
+        return map
+    }
+
+    private func writeMap(_ map: [String: EventHubStoredPixelState]) {
+        guard let data = try? JSONEncoder().encode(map) else { return }
```
bwaresiak, line 96: There’s quite a bit of try? calls here - shouldn’t we add error pixels (daily/count)?

SharedPackages/EventHub/Sources/EventHub/Core/EventHub.swift
```diff
@@ -0,0 +1,373 @@
...
+        }
+        for config in latestConfigs where config.isEnabled && config.trigger.isPeriod && telemetries[config.name] == nil {
+            startNewPeriodLocked(config)
+        }
+        rearmSchedulerLocked()
+    }
+
+    private func startNewPeriodLocked(_ config: TelemetryPixelConfig) {
+        guard isForeground, latestEnabled, config.isEnabled, config.trigger.period != nil else { return }
+        let telemetry = Telemetry(config: config, periodStartMillis: scheduler.nowMillis(), dedupStore: dedupStore)
```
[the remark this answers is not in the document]
miasma13: Not a bug — matches the Windows reference exactly, including this asymmetry. `TearDownPixel` leaves `dedupSeen` untouched; only `Disable()` calls `dedupSeen.Clear()`. Windows holds dedup at hub level (`Dictionary<TabId, HashSet<string>>`, keyed `pixel:param:source`), same as `DedupStore` here.

On *"previously, destroying the parameter cleared dedup automatically"* — true, but that was a side effect of the defect this PR fixes. Dedup living in the parameter is what made it reset at every period boundary, which TD note [2] forbids. Clearing on teardown reinstates a variant of that.

Same rationale applies: dedup survives a period boundary to stop a long-lived page inflating the first count of each new period. A revoke → re-grant is that situation — same tab, same page, already counted — so skipping it is the rule working, not an undercount.

Keys for removed pixels linger until tab navigation or close. It's a `Set`, so flapping can't grow it — bounded by pixels-ever-seen per live tab.

Leaving as-is for reference parity; diverging from Windows here would be a spec decision, not a fix.

SharedPackages/EventHub/Sources/EventHub/Persistence/EventHubStore.swift
miasma13: Good catch — you're right that these are silent today. Following up in #6190 rather than growing this PR.

Root cause turned out to be a port omission: Windows had `Diagnostic.cs`, an ETW `EventSource` logging every fail-safe boundary, which we didn't port and never replaced. #6190 adds `os.Logger` coverage across all of them, plus two error events via `EventMapping` for the cases that genuinely need fleet-wide visibility — persistence (read/decode/write/delete) and the consent-strip fail-closed path.

Worth noting Windows shipped zero error *pixels* — its answer was local tracing only — so two events is already a step beyond the reference. The remaining `try?` sites are `JSONEncoder` calls over our own `Codable` structs, which have no realistic failure mode; those get logging only.

PR #6005, opened 2026-07-24
Delay web-extension unloads to avoid a WebKit declarativeNetRequest crash

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201037661562251/task/1216849418420274?focus=true

### Description

Unloading a web extension while WebKit is still reading its declarativeNetRequest rules crashes the app. The bug is in WebKit ([305585](https://bugs.webkit.org/show_bug.cgi?id=305585)), fixed upstream but not in any shipping macOS 26. It also causes the macOS performance-test flakiness.

So an extension that uses declarativeNetRequest now stays loaded for 3 seconds before anything unloads or reloads it, double the longest load-to-crash gap in the crash reports. Three paths on launch would break that rule, and now wait instead:

- a scriptlet update, which reloads the extension
- an embedded upgrade, which replaces it
- the launch sync, which removes it when its type is disabled

The feature-flag handler and the fire button are synchronous, so they still unload immediately.

The wait ignores cancellation on purpose: teardown is both what cancels and when extensions unload. It suspends rather than blocking, so the UI stays responsive.

Remove this once the minimum supported macOS ships the WebKit fix.

### Testing Steps

1. Open Debug → Web Extensions → Uninstall all extensions.
2. Open Debug → Web Extensions → Clear Cached Scriptlets.
3. Run `log stream --level debug --predicate 'subsystem == "WebExtensions"'` in Terminal.
4. Relaunch the app.
5. Watch the stream for a few seconds.
   It prints `📦 Installing embedded extension`, then `⏸️ Delaying unload ...`, then `✅ Reloaded extension`, and the app does not crash.
6. Visit an ad-heavy page.
   Ads are blocked, so the extension still works after the delayed reload.

The embedded-upgrade path needs a bundled version bump, so unit tests cover it with an injected clock instead.

### Impact

Medium: a regression would delay or disrupt scriptlet updates and embedded upgrades.

#### What could go wrong?

If WebKit needs more than 3 seconds, the crash still happens. Nothing tells us when it finishes, so 3 seconds is an estimate that narrows the window rather than closing it.

A scriptlet update that arrives within 3 seconds of its extension loading applies up to 3 seconds later. Nothing else is delayed.

---
###### Internal references:

### DIFF
--- a/SharedPackages/WebExtensions/Package.resolved
+++ b/SharedPackages/WebExtensions/Package.resolved
@@ -1,13 +1,13 @@
 {
-  "originHash" : "c83b04d371e1c66148dd4d9771e4262264e9cf9c0f52cc062b940e7336600002",
+  "originHash" : "ac2a9945ad19c7da6daf61472649344613332f0fa77e9b4e44a6ce545a90bea6",
   "pins" : [
     {
       "identity" : "content-scope-scripts",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/duckduckgo/content-scope-scripts.git",
       "state" : {
-        "revision" : "dc9b1280951a22359aac33d07ffd410cbbae2b94",
-        "version" : "15.12.0"
+        "revision" : "3011e9e06cc68d3f9c2cd7b00aadabcd34d2869a",
+        "version" : "15.19.0"
       }
     },
     {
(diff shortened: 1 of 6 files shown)


jotaemepereira (approved):
👍🏼

PR #6009, opened 2026-07-24
Improved Sync relay channel pixel error reporting

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201462886803403/task/1216838070405985?focus=true
Tech Design URL:
CC:

### Description
Improves Pairing V2 failure reporting on iOS and macOS by adding two parameters to `sync_setup_ended_failed`:
- `pairing_failure_stage`: the presenter or scanner operation that failed.
- `pairing_failure_kind`: whether the relay channel was unavailable, expired, returned another HTTP error, or failed without an HTTP response.

Only terminal failures are reported, so requests that succeed after retry do not generate failure pixels. Failures that occur before the pairing screen appears are now also reported with the correct setup source and host/joiner role.

Pairing behavior and retry timing are unchanged.

### Testing Steps
1. Complete a Sync V2 setup between two clients (iOS <> macOS) → Sync setup completes normally with no user-facing behaviour change.

### Impact

---

> [!NOTE]
> **Medium Risk**
> Changes error mapping and analytics for sync pairing flows without altering successful pairing logic; misclassification could skew failure metrics but should not break setup for users.
> 
> **Overview**
> **Pairing V2 setup failures** now carry structured **stage** and **kind** metadata (presenter vs scanner steps, plus unavailable / expired / HTTP / network), surfaced through `PairingV2OperationFailure` and a new `SyncConnectionError.pairingV2OperationFailure` path.
> 
> Relay HTTP in the message exchanger is centralized so those errors are classified consistently; the coordinator tags each relay step (open channel, poll, send hello, recovery messages, key generation) before errors reach the connection controller and UI.
> 
> **Analytics:** `sync_setup_ended_failed` gains `relay_channel_failure` as a reason, plus `pairing_failure_stage` and `pairing_failure_kind` on iOS and macOS (including when presenter start fails before the barcode screen). Pixel definitions and tests are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e56d6bf804a09f0ef763863426843d0c8d6f077f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/PixelDefinitions/pixels/definitions/sync_setup.json5
```diff
@@ -6,6 +6,7 @@
         "parameters": [
             "appVersion",
             "pixelSource",
+            "channel",
```
amddg44, line 9: Adding in `channel` here as it is injected by PixelKit and keeps triggering failing pixel reports e.g. https://app.asana.com/1/137249556945/task/1216882405313891?focus=true

pballart (commented):
Codex found an uncovered edge case:

> The start-failure attribution looks correct now, but after startConnectMode() succeeds, later presenter polling or sending failures still go through .sharer, so they’re reported as exchange/host rather than connect/joiner. Could we preserve the setup source through polling and add a regression test where connect mode starts successfully before a relay operation fails?

Besides that, the PR is looking good.

amddg44:
> Codex found an uncovered edge case:
> 
> > The start-failure attribution looks correct now, but after startConnectMode() succeeds, later presenter polling or sending failures still go through .sharer, so they’re reported as exchange/host rather than connect/joiner. Could we preserve the setup source through polling and add a regression test where connect mode starts successfully before a relay operation fails?
> 
> Besides that, the PR is looking good.

Thanks for flagging @pballart, however Codex is a bit off on the role here. It’s right that these errors go through .sharer, but startConnectMode() doesn’t mean the device stays connect/joiner in V2. Host/joiner is elected later, so forcing that through polling would be misleading. I’d leave this as-is; pairing_failure_stage already gives us the presenter/scanner context.

pballart:
> > Codex found an uncovered edge case:
> > > The start-failure attribution looks correct now, but after startConnectMode() succeeds, later presenter polling or sending failures still go through .sharer, so they’re reported as exchange/host rather than connect/joiner. Could we preserve the setup source through polling and add a regression test where connect mode starts successfully before a relay operation fails?
> > 
> > 
> > Besides that, the PR is looking good.
> 
> Thanks for flagging @pballart, however Codex is a bit off on the role here. It’s right that these errors go through .sharer, but startConnectMode() doesn’t mean the device stays connect/joiner in V2. Host/joiner is elected later, so forcing that through polling would be misleading. I’d leave this as-is; pairing_failure_stage already gives us the presenter/scanner context.

Perfect, just wanted to double check ✅

pballart (approved):

PR #6010, opened 2026-07-24
iOS: Decompose UnifiedToggleInputCoordinator (part 2 — model selection, pixels, wide events)

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216832851236221

### Description
Part 2 of the `UnifiedToggleInputCoordinator` decomposition (iPhone UTI omnibar), continuing from #5984. Peels three more responsibilities off the coordinator into owned collaborators — model/reasoning selection + subscription gating (`UTIModelSelector`), pixel firing behind an injectable seam (`UTIPixelReporter`), and Duck.ai wide-event instrumentation (`UTIWideEventReporter`) — and moves the display-state/intent value types into their own file. Behaviour-preserving; the coordinator shrinks from ~2,570 to ~2,265 lines.

`UTIAttachmentController` is the one remaining collaborator and is intentionally **not** in this PR — it collides head-on with the open native image/file paste work (#5860) and will be extracted once that merges.

### Testing Steps
1. Focus the omnibar and toggle Search ↔ Duck.ai → both render and submit as before.
2. On a Duck.ai tab, change the model and the reasoning effort → the selection applies; picking a gated model/reasoning routes to the subscription upsell.
3. Submit a Duck.ai prompt, with and without a file attachment → the prompt sends and the model chip / tools behave as before.

### Impact
Low — internal, behaviour-preserving refactor with no functional or UI change. Guarded by the full UTI unit suite plus new `UTIPixelReporter` tests.

---

> [!NOTE]
> **Low Risk**
> Internal refactor with explicit behavior-preservation goal; analytics paths are reorganized but should fire the same events via UTIPixelReporter and existing pixel helper helpers.
> 
> **Overview**
> **Part 2** of the `UnifiedToggleInputCoordinator` split peels three collaborators out of the coordinator (~2,570 → ~2,265 lines) without intended behavior change.
> 
> **`UTIModelSelector`** now owns model/reasoning picker menus, chip visibility, subscription gating, and pending gated-selection retries; the coordinator wires it via `ViewSurface`, `Environment`, and `Callbacks` instead of inline `pendingGated*` and menu factories.
> 
> **`UTIPixelReporter`** centralizes omnibar UTI pixel firing behind injectable **`UTIPixelFiring`** and live **`UTIPixelContext`**, replacing scattered `Pixel`/`DailyPixel` calls. **`UnifiedToggleInputCoordinatorPixelHelper`** accepts optional `firing` for the same seam.
> 
[...]

PR #6028, opened 2026-07-27
iOS: Name the UTI display-state questions once (part 1 — predicates + call-site migration)

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216878797836710

### Description

The unified toggle input describes what it's showing with a small enum, and code all over the app opened that enum up inline to ask questions of it — so the same question was re-derived in several places, each free to drift.

- Names four of those questions once each, next to the state they read, and points every caller at the name — so only the state's owner and its own file now look inside the enum.
- Stops handing a whole display state to the tools menu, which only needed "is the input showing at all". That removes a state the iPad tool picker was inventing to satisfy the signature.
- Renames an address-bar-focus check to say what it tests, and folds an identical copy of it in the same file into it.
- No behaviour change: every replacement swaps an expression for a provably equal one.
- Adds the first tests for the state owner, which had none — a table over all six display states plus the invariants call sites now depend on.

Part 1 of 4. Later parts add a test suite for the state owner, split an overgrown test file, and run the Maestro verification.

### Testing Steps

Internal build with the unified toggle input enabled. iPhone, address bar at bottom, except the last step.

**On a Duck.ai tab**
1. On a duck.ai tab submit a prompt → the input collapses to the chat-input footer: centred placeholder, fire and menu buttons either side.
2. Tap the collapsed footer → it expands to the full input; the fire/menu buttons and the separator above it go away.
3. Tap the chat web view while the input is expanded → the input collapses back to the footer.
4. Rotate to landscape and back while collapsed → the footer stays pinned to the bottom in both orientations.
5. On a web page focus the address bar, then dismiss it → it is expanded when focused and returns to rest with no jump.
6. Focus the address bar and swipe down over the area below the input → the keyboard drops and the input stays expanded.
7. On iPad open the omnibar tool picker → the tools button appears, the menu opens, tool selection works.

### Impact

**Low** — a naming and ownership refactor behind an internal-only feature, with no intended behaviour change. It does touch keyboard and layout paths, which is why the steps above concentrate there.

#### What could go wrong?

- One call site now reads live state where it previously read a snapshot passed in by its caller. All ten callers were traced and each computes that snapshot immediately beforehand from the same source → mitigated by tracing every caller, and by a doc comment recording which half of the function still reads the snapshot.
- A predicate could be defined slightly wider or narrower than the expression it replaced, changing when a branch fires → mitigated by checking each of the 13 substitutions against its original individually, including the deliberately narrow search-on-Duck.ai-tab definition, which must not match collapsed tabs.

### Quality Considerations

**Left raw** wherever one question doesn't capture the original test: the composite that also reads card position, a comparison against a previously captured state, the transition-pair matches, and three multi-case dispatch switches.

**Verified**: clean build, no new warnings; 9 new state-machine tests plus 342 existing unit tests pass across the coordinator, render-state, per-tab-state and fire-onboarding suites; no test referenced a changed signature.

**Rebased onto #6010** after it merged. That PR moved the enum into its own file and split the coordinator further; its new pieces consume derived booleans through their environment rather than the raw state, so the two reinforce each other.

—
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Refactor-only with equivalence checks and new unit tests; touches keyboard and layout branches but does not change intended UTI behavior.
> 
> **Overview**
> Centralizes repeated questions about unified toggle input **display state** into `UTIStateMachine` and routes callers through `UnifiedToggleInputCoordinator` instead of inline enum pattern matching.
> 
> New predicates include **`isAITabCollapsed`**, **`isOmnibarEditing`**, **`omnibarState`**, and **`isSearchOnAITab`** (intentionally tied to expanded Duck.ai tab + search mode, not any AI-tab state). **`dismissOmnibarKeyboard`** now keys off **`isInputPaneExpanded`** instead of a three-case switch. The iPad tool picker and tools presentation API take **`isActive`** instead of a fabricated omnibar display state or full **`displayState`**.
> 
> **`MainViewController`** renames the escape-hatch focus check to **`isAddressBarFocused`** and reuses it where chrome morph and burn flows duplicated the same condition. Collapsed-footer layout reads live **`isAITabCollapsed`** rather than deriving collapse from render **`isExpanded`**.
> 
> Adds **`UTIStateMachineTests`** with an exhaustive predicate matrix, transition checks, and disjointness tests that document assumptions at call sites. No intended behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b546ac5dc1dc52471224dca07df8c20c9ccbe132. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -1896,9 +1896,7 @@ class MainViewController: UIViewController {
         configureUnifiedInputEscapeHatch(model)
     }
 
-    /// True when the address bar owns editing focus. Reads the persistent omnibar-session state, which
-    /// survives a card tap that dismisses the keyboard. Narrower than `isKeyboardOwnedByOmnibar`, which
-    /// also treats a first responder anywhere inside the omnibar view hierarchy as omnibar-owned.
+    /// True when the address bar owns editing focus; survives a card tap that dismisses the keyboard.
     private var isAddressBarFocused: Bool {
         omniBar.isTextFieldEditing || unifiedToggleInputCoordinator?.isOmnibarSession == true
     }
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -1635,7 +1635,7 @@ class MainViewController: UIViewController {
         updateUnifiedToggleInputKeyboardVisibility(keyboardVisible)
 
         let coordinator = unifiedToggleInputCoordinator
-        let isAITabCollapsed = coordinator?.displayState == .aiTab(.collapsed)
+        let isAITabCollapsed = coordinator?.isAITabCollapsed == true
         let isBottomExpandedUTIKeyboardAnchored = coordinator?.isInputEditing == true
             && coordinator?.cardPosition == .bottom
             && viewCoordinator.isNavigationBarContainerBottomKeyboardBased
@@ -1896,9 +1896,10 @@ class MainViewController: UIViewController {
         configureUnifiedInputEscapeHatch(model)
     }
 
-    /// True when an escape hatch action runs in focus mode (reads the persistent omnibar-session state, which
-    /// survives the card tap that dismisses the keyboard).
-    private var isEscapeHatchInFocusMode: Bool {
+    /// True when the address bar owns editing focus. Reads the persistent omnibar-session state, which
+    /// survives a card tap that dismisses the keyboard. Narrower than `isKeyboardOwnedByOmnibar`, which
+    /// also treats a first responder anywhere inside the omnibar view hierarchy as omnibar-owned.
+    private var isAddressBarFocused: Bool {
         omniBar.isTextFieldEditing || unifiedToggleInputCoordinator?.isOmnibarSession == true
     }
 
@@ -3172,7 +3173,7 @@ class MainViewController: UIViewController {
     /// no-op once correct — no loop, no per-frame work beyond the bool check.
     private func reanchorAITabCollapsedFooterIfNeeded() {
         guard let coordinator = unifiedToggleInputCoordinator,
-              coordinator.displayState == .aiTab(.collapsed),
+              coordinator.isAITabCollapsed,
               coordinator.cardPosition == .bottom,
               !viewCoordinator.isNavigationBarContainerBottomKeyboardBased else { return }
         viewCoordinator.setNavBarContainerBottomToKeyboard()
@@ -4154,7 +4155,7 @@ extension MainViewController: BrowserChromeDelegate {
         chromeMorphAnimator.cancel()
 
         if percent < 1 {
-            if omniBar.isTextFieldEditing || unifiedToggleInputCoordinator?.isOmnibarSession == true {
+            if isAddressBarFocused {
                 dismissOmniBar()
             }
             _ = findInPageView?.resignFirstResponder()
@@ -5726,7 +5727,7 @@ extension MainViewController: EscapeHatchActionRouter {
         }
 
         // Captured before the confirmation dialog steals focus, so we can restore the keyboard afterwards.
-        let wasInFocusMode = isEscapeHatchInFocusMode
+        let wasInFocusMode = isAddressBarFocused
         let tabViewModel = tabManager.viewModel(for: tab)
         let presenter = FireConfirmationPresenter()
         presenter.presentFireConfirmation(
@@ -5756,7 +5757,7 @@ extension MainViewController: EscapeHatchActionRouter {
             return
         }
 
-        let wasInFocusMode = isEscapeHatchInFocusMode
+        let wasInFocusMode = isAddressBarFocused
         let tabViewModel = tabManager.viewModel(for: tab)
         let request = FireRequest(
             options: .all,
--- a/iOS/DuckDuckGo/UnifiedToggleInput/Coordinator/UTIStateMachine.swift
+++ b/iOS/DuckDuckGo/UnifiedToggleInput/Coordinator/UTIStateMachine.swift
@@ -58,8 +58,24 @@ final class UTIStateMachine {
         displayState == .aiTab(.expanded)
     }
 
+    var isAITabCollapsed: Bool {
+        displayState == .aiTab(.collapsed)
+    }
+
     var isContextualChatState: Bool {
         displayState == .contextualChat
+    }
+
+    /// The omnibar session is being edited. Distinct from `isOmnibarSession` (any omnibar state)
+    /// and from `isActive` (any state other than `.hidden`).
+    var isOmnibarEditing: Bool {
+        displayState == .omnibar(.active)
+    }
+
+    /// The omnibar sub-state, or `nil` when the current state isn't an omnibar one.
+    var omnibarState: UnifiedToggleInputDisplayState.OmnibarState? {
+        if case .omnibar(let state) = displayState { return state }
+        return nil
     }
 
     /// Folds contextual + Duck.ai tab into one "Duck.ai surface" bucket — used only for the funnel `origin`.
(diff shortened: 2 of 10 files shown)


iOS/DuckDuckGo/UnifiedToggleInput/Coordinator/UTIStateMachine.swift
```diff
@@ -58,10 +58,26 @@ final class UTIStateMachine {
         displayState == .aiTab(.expanded)
     }
```
aataraxiaa, line 60: Add single line doc here

iOS/DuckDuckGo/UnifiedToggleInput/Coordinator/UTIStateMachine.swift
```diff
@@ -98,6 +114,13 @@ final class UTIStateMachine {
         isToggleEnabled && !(hidesToggleOnDuckAITab && isAITabState)
     }
 
+    /// Search mode on an expanded Duck.ai tab — the state where the toggle has been flipped to Search
+    /// but the chat web view is still the tab's content. Requires `.aiTab(.expanded)`, not any AI-tab
+    /// state: on a collapsed AI tab there is no input pane for a mode to apply to.
```
aataraxiaa: collapse to one line

aataraxiaa: also note that this state is not possible with default UTI behavior. on duck.ai tab we don’t currently show a toggle (but it is possible, it’s just not a feature that is enabled)

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -1896,9 +1896,10 @@ class MainViewController: UIViewController {
         configureUnifiedInputEscapeHatch(model)
     }
 
-    /// True when an escape hatch action runs in focus mode (reads the persistent omnibar-session state, which
-    /// survives the card tap that dismisses the keyboard).
-    private var isEscapeHatchInFocusMode: Bool {
+    /// True when the address bar owns editing focus. Reads the persistent omnibar-session state, which
+    /// survives a card tap that dismisses the keyboard. Narrower than `isKeyboardOwnedByOmnibar`, which
+    /// also treats a first responder anywhere inside the omnibar view hierarchy as omnibar-owned.
```
aataraxiaa: collapse to one line

iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
```diff
@@ -851,11 +851,14 @@ extension MainViewController {
         updateUnifiedInputContentVisibility(for: coordinator, renderState: coordinator.computeRenderState())
     }
 
+    /// The collapsed-AI-tab pose reads the coordinator's live display state, so a caller passing an
+    /// already-stale snapshot can't drive a pose the coordinator has moved past. Everything below it
+    /// still trusts `renderState` — every current caller computes that fresh from this same coordinator.
```
aataraxiaa: collapse, ensure it’s clear what pose means here

iOS/DuckDuckGo/UnifiedToggleInput/Coordinator/UTIStateMachine.swift
aataraxiaa: Done — added: `/// A Duck.ai tab showing the collapsed chat-input footer rather than the expanded input pane.`

iOS/DuckDuckGo/UnifiedToggleInput/Coordinator/UTIStateMachine.swift
aataraxiaa: Done — down to one line for the definition, plus a second one-liner for your point below.

aataraxiaa: Good catch, added — `/// Not reachable with default behaviour: the mode toggle isn't shown on Duck.ai tabs today.`

Checked it against the code: `isToggleVisible` drops the toggle when `hidesToggleOnDuckAITab && isAITabState`, and that flag snapshots the `aiChatTabHideToggle` feature flag — so with it on there's no way for the user to flip to Search there. Phrased as "today" rather than an invariant since it's flag-driven.

Worth flagging that this makes the narrow definition load-bearing rather than cosmetic: the local it replaced in `computeRenderState` literally read `isAITabState`, which is only equivalent because it sits inside `case .aiTab(.expanded)`. Promoting that broader form would have made the search-suggestions sink fire on collapsed tabs.

iOS/DuckDuckGo/MainViewController.swift
aataraxiaa: Done — one line: `/// True when the address bar owns editing focus; survives a card tap that dismisses the keyboard.`

Dropped the `isKeyboardOwnedByOmnibar` comparison to keep it to one line. That one stays separate deliberately — it adds a first-responder-hierarchy check, so folding them would be a behaviour change.

iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
aataraxiaa: Done — now: `/// The collapsed-footer pose — how the Duck.ai chat input and its separator sit at the bottom of the screen — reads the coordinator's live display state; the chrome below it still trusts `renderState`.`

Defined "pose" inline as the footer + separator arrangement. Scoping it to the footer also fixes an over-claim in my original wording: only those two lines moved to live state, while `applyTopChromeState` below still reads the passed snapshot. Equal today — the two callers of this overload both compute `renderState` immediately beforehand from the same coordinator.

jaceklyp (approved):
all good from my side;

one comment from agentic review:

One thing to flag (PLAUSIBLE, likely intentional) — reported inline at MainViewController+UnifiedToggleInput.swift:860: this is the only non-mechanical change. isAITabCollapsed went from isAITabState && !renderState.isExpanded (mixes live state + the passed renderState) to purely live displayState == .aiTab(.collapsed), now ignoring the renderState argument for the footer pose. The author added a doc comment saying that split is deliberate — so it's probably fine — but it's the one place behavior can differ if a caller ever passes a renderState that disagrees with the live display state mid-transition. Worth the author confirming; everything else is a pure equivalence.

aataraxiaa:
Thanks @jaceklyp — confirming that one is intentional, and worth spelling out because the staleness risk actually runs the *opposite* way from what the agentic review assumed.

**The old form was the one that could diverge.** `isAITabState && !renderState.isExpanded` mixed live state with the snapshot, so if the two ever disagreed it produced a contradictory pose — in both directions:

| live `displayState` | stale `renderState` from | old result | new result |
|---|---|---|---|
| `.aiTab(.collapsed)` | `.aiTab(.expanded)` | `true && !true` = **false** → no footer pose while collapsed ❌ | `true` ✅ |
| `.aiTab(.expanded)` | `.aiTab(.collapsed)` | `true && !false` = **true** → footer pose while expanded ❌ | `false` ✅ |

(`.aiTab(.collapsed)` maps to `.flanked`, and `UnifiedToggleInputCardLayout.isExpanded` is only true for `.expanded` — `UTIViewConfig.swift:36`.)

Reading live `displayState` makes the pose self-consistent by construction, so this substitution removes a latent staleness bug rather than adding one. That is why it is the only non-mechanical change in the diff.

**Why it can't bite today either way.** The 2-arg overload has exactly two callers, and both compute the snapshot immediately beforehand from the same coordinator:
- `MainViewController+UnifiedToggleInput.swift:851` — the no-arg overload, recomputes inline (this is the path the other 8 callers funnel through)
- `MainViewController+UnifiedToggleInputIntentHandling.swift:196` — computed at `:187`; the calls in between (`isAITabState`/`cardPosition` reads, nav-bar anchor, chrome background, `anchorContentContainerToInputTop`, `apply(viewConfig:)`) are all synchronous and none reassign `displayState`

So there is no mid-transition window where they disagree. Worth noting the plan I was working from claimed this function had "both current callers" — it actually has **10**, which is why I traced all of them and added the doc comment rather than taking that on trust.

**What the residual risk applies to.** A caller passing a genuinely stale snapshot would now only affect `applyTopChromeState` below the pose lines — content overlay, chat header, chrome background. That half is unchanged by this PR and was already purely snapshot-driven; the doc comment at `:857` names the split so a future caller can't assume the whole function is live.

Also FYI since it post-dates your review: pushed `b546ac5dc1` adding `UTIStateMachineTests` (9 tests) after @aataraxiaa's pass — the state machine had no coverage at all. Two of them pin invariants relevant here: the exact `isInputPaneExpanded` case set that `dismissOmnibarKeyboard` now depends on, and `isSearchOnAITab` staying false on collapsed Duck.ai tabs.

PR #6030, opened 2026-07-27
macOS: open the Prompt Bar from the global shortcut and menu bar icon

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201011656765697/task/1216915638786612?focus=true
Tech Design URL:
CC:

### Description
Adds the floating Prompt Bar itself: a borderless panel with a single prompt field that opens what the user types in a new Duck.ai tab. The user's configured shortcut is now registered system-wide, and the menu bar icon's placeholder click toggles the bar. The field is deliberately minimal for now — the address bar's Duck.ai controls (attach, tools, model and reasoning pickers, voice) land once they are extracted into a component both surfaces share.

### Testing Steps
1. On an internal build with `macosPromptBar` enabled, press ⌥Space from another app → the bar appears near the top of the screen holding the pointer, focused.
2. Type a prompt and press Enter → the bar closes and the prompt opens in a new selected Duck.ai tab. Reopen and press Escape, or click another window, → the bar dismisses without submitting.
3. In Settings → AI Features, change the shortcut and confirm the new combination opens the bar and the old one no longer does; switch the shortcut off and confirm neither works and the menu bar icon disappears.
4. Click the menu bar icon, check if the prompt bar correctly appears.

### Impact
Low

#### What could go wrong?
The Carbon hot key could collide with a combination another app already owns → registration fails and returns false rather than throwing, leaving the menu bar icon as a working entry point, and the settings recorder already rejects the reserved system combinations.

---

> [!NOTE]
> **Medium Risk**
> New global hotkey registration and window-less Duck.ai tab routing affect core navigation and multi-display behavior; failures are mostly localized (hotkey collision, wrong window reuse) rather than data or auth.
> 
> **Overview**
> Adds the **floating Prompt Bar** behind `macosPromptBar`: a borderless `NSPanel` with a minimal prompt field, opened by the user’s **Carbon global hotkey** or the **menu bar icon** (no longer a no-op).
> 
> **`PromptBarCoordinator`** wires shortcut registration to **`effectiveKeyboardShortcutPublisher`** (nil when Duck.ai is off or the shortcut is disabled) and toggles **`PromptBarPresenter`**, which places the bar via **`PromptBarPlacement`**, dismisses on resign-key unless auxiliary UI is up, and does not activate the full app on show.
> 
> Submitting a prompt runs **`PromptBarPromptSubmitter`**, which reuses a visible browser window on the same display/space when possible, otherwise opens Duck.ai via new **`AIChatTabOpening`** / **`AIChatTabManaging`** paths (new tab in a chosen window or a new window at a screen-centered drop point), then activates the app.
> 
> **Preference defaults** for shortcut and menu bar icon change from **on** to **off** when nothing is stored; **`effectiveKeyboardShortcutPublisher`** is added so the global shortcut stays unregistered when Duck.ai is unavailable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1704b9f26092fcde42746ab8ef695a0c526b1850. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
@@ -107,13 +107,11 @@ protocol AIChatTabOpening {
     @MainActor
     func openVoiceSession(inSourceCollection sourceCollection: TabCollectionViewModel?, behavior: LinkOpenBehavior)
 
-    /// Opens an auto-submitted query in a new tab of a specific window, for window-less entry points
-    /// that resolve their own target rather than taking `lastKeyMainWindowController`.
+    /// For window-less entry points that resolve their own target rather than taking `lastKeyMainWindowController`.
     @MainActor
     func openAIChatTab(withQuery query: String, inNewTabOf windowController: MainWindowController)
 
-    /// Opens an auto-submitted query in a new window placed by its top-center point. Without one a
-    /// new window cascades off the last key window, possibly onto another display.
+    /// Placed by top-center point; without one a new window cascades off the last key window.
     @MainActor
     func openAIChatTab(withQuery query: String, inNewWindowAt droppingPoint: NSPoint)
 }
@@ -221,11 +219,9 @@ protocol AIChatTabManaging {
     @MainActor
     func openAIChat(_ url: URL, with behavior: LinkOpenBehavior, hasPrompt: Bool)
 
-    /// Bypasses last-key-window resolution.
     @MainActor
     func openAIChat(_ url: URL, inNewTabOf windowController: MainWindowController, hasPrompt: Bool)
 
-    /// Positioned by its top-center point rather than cascaded off the last key window.
     @MainActor
     func openAIChat(_ url: URL, inNewWindowAt droppingPoint: NSPoint, hasPrompt: Bool)
 
--- a/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
@@ -107,19 +107,13 @@ protocol AIChatTabOpening {
     @MainActor
     func openVoiceSession(inSourceCollection sourceCollection: TabCollectionViewModel?, behavior: LinkOpenBehavior)
 
-    /// Opens a chat with a pre-filled, auto-submitted query in a new selected tab of a specific window.
-    ///
-    /// For entry points that own no window and so must resolve their own target — the Prompt Bar
-    /// picks the browser window on the screen and Space the user is actually looking at, which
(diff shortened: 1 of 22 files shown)


macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
```diff
@@ -253,6 +279,14 @@ extension WindowControllersManager: AIChatTabManaging {
         }
     }
 
+    func openAIChat(_ url: URL, inNewTabOf windowController: MainWindowController, hasPrompt: Bool) {
```
tomasstrba, line 282: Not a blocker: `hasPrompt` is dead in both new overloads — this one and `inNewWindowAt` below never read it. The prompt is delivered via `promptHandler.setData(...)`, and unlike the `.currentTab` path these always open a fresh tab/window, so there's no reuse decision for it to gate. Harmless but misleading, since a reader expects it to matter. Given how much this codebase leans on intent-revealing signatures, I'd drop the parameter (or add a one-line comment that it's kept only for API symmetry).

macOS/DuckDuckGo/PromptBar/Shortcut/CarbonGlobalShortcutRegistrar.swift
```diff
@@ -0,0 +1,124 @@
...
+        static let promptBarHotKeyID: UInt32 = 1
+    }
+
+    private var hotKeyRef: EventHotKeyRef?
+    private var eventHandlerRef: EventHandlerRef?
+    private var handler: (() -> Void)?
+
+    private(set) var registeredShortcut: PromptBarShortcut?
+
+    deinit {
```
tomasstrba, line 39: Not blocking, just a note: this type isn't `@MainActor`, but its lifecycle is effectively main-thread-bound — `register`/`unregister` are driven by the `@MainActor` coordinator, and Carbon `kEventHotKeyPressed` events dispatch on the main run loop, so `handler` is only ever touched on main. The one soft spot is this `deinit` calling `UnregisterEventHotKey`/`RemoveEventHandler`: if the registrar were ever released off-main, those Carbon calls would run off-main. It's owned by a `@MainActor` type so in practice it's main-released — worth a mental note rather than a change.

tomasstrba (approved):
LGTM! ✔️ Works well 👍 Tried many desktop variations or window states, it always creates new tab in a correct window. I also like the position of the input.

Just two codebase related comments found by Claude

Bunn:
> LGTM! ✔️ Works well 👍 Tried many desktop variations or window states, it always creates new tab in a correct window. I also like the position of the input.
> 
> Just two codebase related comments found by Claude

Thanks, I'll address them in the stacked PR

PR #6054, opened 2026-07-28
EventHub for Apple #2: macOS EventHub integration

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216947041711145
CC:

### Description
Second in the stack, targeting `michal/event-hub-1`. Wires the EventHub package from #5959 into macOS and moves the five YouTube ad-blocking detection events onto it, retiring the hardcoded path (opt-in check plus event-type-to-pixel mapping compiled into the app). Which events are measured, and how, now comes from remote config instead of an app release.

Two behavioural changes: detections are aggregated into a bucketed daily count fired once per period rather than a pixel per detection, and the analytics opt-in now strips gated entries from the config before EventHub sees them — so a withheld opt-in means the measurement doesn't exist rather than being filtered later.

macOS only — iOS still uses the old path.

### Testing Steps
Use the setup described in https://app.asana.com/1/137249556945/project/414235014887631/task/1216994836433629?focus=true and the sample test page to trigger the web detector event. Filter the logs by "EventHub" subsystem. Please verify the following:

- [ ] On app launch the feature is reported as enabled telemetry definitions are being loaded
- [ ] Triggering an event starts the period
- [ ] When app is in foreground the pixel is fired at the period end
- [ ] Triggering an event on a single tab on a single page and reloading it will register it only once (de-dup)
- [ ] Triggering an event on the same page but opened in separate tabs is registered individually
- [ ] Triggering an event on a page, then in the same tab navigating elsewhere, then navigating again to the first page will record the event twice (navigation clears de-dup state)
- [ ] Recorded counts are not cleared by the Fire button
- [ ] Recorded counts are persisted between app relaunch

### Impact
High — touches per-tab lifecycle code that runs regardless of the feature flag, and the consent gate deciding whether telemetry is collected.

#### What could go wrong?
- Gated config names not byte-matching the server-side telemetry keys → the consent strip silently no-ops and telemetry is collected without opt-in. Names come from the shipped Windows config; nothing on-device validates them, so worth a second pair of eyes.
- The per-tab component is created for every tab, so a fault there affects all browsing → mitigated by following the existing tab-extension registration pattern.
- Pixel names drifting from the definitions, since PixelKit rewrites macOS names → mitigated by a test asserting the name PixelKit actually requests.
- Counts recorded shortly before quit being lost → mitigated by a flush in the termination sequence.

### Quality Considerations
- Resolves the open question from #5959: the platform-suffix split holds — the package fires the bare config name, macOS appends `_macos`. Matching the declared names required opting out of PixelKit's automatic `m_mac_` prefix.
- Privacy: consent stripping fails closed, and navigation URLs are never logged. Fired parameters (including YouTube login state) are logged at info level — flagging in case that's too permissive for a persisted system log.
- Not addressed here: the now-unused ad-blocking pixel mapping and its five stale definitions (kept for the iOS transition), measurement being suspended while the app is inactive, and fire-window tabs counted like any other.
- As in #5959, the package suite isn't in a scheme test plan, so CI won't run its 128 tests — local `swift test` only.

---

> [!NOTE]
> **High Risk**
> Per-tab EventHub wiring runs for all tabs regardless of the feature flag, and consent/config key mismatches could collect telemetry without opt-in; changes also affect launch, termination, and pixel naming paths.
> 
> **Overview**
> **macOS** integrates the shared **EventHub** package and moves five YouTube ad-blocking detection signals off the retired **`WebEventsSubfeature`** path (per-detection daily pixels with hardcoded opt-in and name mapping). Telemetry is now driven by remote **`eventHub`** privacy config, aggregated into **bucketed daily counts** fired once per period, with pixels sent through PixelKit as `*_macos` using `doNotEnforcePrefix`.
> 
> `MacOSEventHubIntegration` owns the hub: dedicated **`EventHubKeyValueStore`**, privacy-config publishers, **`YouTubeAdBlockingTelemetryConsentRequirement`** (both YouTube ad-blocking and analytics must be on—consent strips gated config keys before EventHub sees them), and app lifecycle hooks (foreground/background/terminate flush). **`EventHubTabExtension`** registers **`WebEventsHandler`** per tab and reports navigation/tab close for dedup. **`UserScripts`** no longer registers **`webEventsSubfeature`**.
> 
> The **EventHub** package gains cold-start safety (`hasConsumedInitialSettings` defers config apply until after initial publisher replay), automatic re-apply on settings/enablement changes without `onConfigChanged()`, optional nil backing store, public **`Logger.eventHub`**, and debug logging for period fires. **`PrivacyFeature.eventHub`** is added. New **`event_hub.json5`** pixel definitions document the successor metrics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68800aeb363b06a78bd33e173303b820a148d5d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -1525,6 +1525,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
         startAutomationServerIfNeeded()
 
+        // This is also called in applicationDidBecomeActive, but we're also calling it here, since
+        // applicationDidBecomeActive returns early on `didFinishLaunching` when it's delivered during
+        // startup (e.g. if a modal alert is shown first). EventHub cannot start any measurement period
+        // until it has been foregrounded at least once, so missing that first activation would leave it
+        // silently inert. Calling it twice is harmless — it re-checks already-settled state.
+        eventHubIntegration.applicationDidBecomeActive()
+
         PixelKit.fire(GeneralPixel.launch, doNotEnforcePrefix: true)
         profilerToken.stop()
     }
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -1847,10 +1847,15 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 stateRestorationManager?.applicationWillTerminate()
             },
 
-            // 6. Auto-clear (burn on quit)
+            // 6. EventHub pending telemetry flush
+            .perform { [eventHubIntegration] in
+                eventHubIntegration.applicationWillTerminate()
+            },
+
+            // 7. Auto-clear (burn on quit)
             autoClearHandler,
 
-            // 7. Privacy stats cleanup
+            // 8. Privacy stats cleanup
             .terminationDecider { [privacyStats] _ in
                 .async(Task {
                     await privacyStats.handleAppTermination()
@@ -1858,7 +1863,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 })
             },
 
-            // 8. Close windows before quitting while waiting for ⌘Q release
+            // 9. Close windows before quitting while waiting for ⌘Q release
             .perform {
                 NSApp.visibleWindows.forEach { $0.close() }
             }
--- a/macOS/DuckDuckGo/Application/AppDelegate.swift
+++ b/macOS/DuckDuckGo/Application/AppDelegate.swift
@@ -1049,8 +1049,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                                                                                 uiHosting: { windowControllersManager.activeViewController },
                                                                                 isOnboardingCompletedProvider: { onboardingManager.state == .onboardingCompleted },
                                                                                 dockCustomization: dockCustomization)
-        eventHubIntegration = MacOSEventHubIntegration(privacyConfigurationManager: privacyConfigurationManager,
-                                                        keyValueStore: keyValueStore)
+        // Dedicated store: EventHub flushes often and `KeyValueFileStore.set` rewrites the whole file.
+        // `try?` — telemetry runs in memory rather than failing launch.
+        eventHubIntegration = MacOSEventHubIntegration(
+            privacyConfigurationManager: privacyConfigurationManager,
+            keyValueStore: try? KeyValueFileStore(location: URL.sandboxApplicationSupportURL, name: "EventHubKeyValueStore"))
 
         if AppVersion.runType.requiresEnvironment {
             remoteMessagingClient = RemoteMessagingClient(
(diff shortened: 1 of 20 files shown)


bwaresiak:
Can you confirm one thing for me: how this should behave for same-document navigations? E.g. browsing shorts or videos on yt is not triggering `eventHub.onNavigationStarted` so any events from youtube may be greatly undercounted.

miasma13:
> Can you confirm one thing for me: how this should behave for same-document navigations? E.g. browsing shorts or videos on yt is not triggering `eventHub.onNavigationStarted` so any events from youtube may be greatly undercounted.

This is related to SPA (single page application) navigations. It was deliberately descoped from the initial implementation on all platforms (to be followed up later on). While in theory it could directly impact the YouTube ad blocking analytics pixels (now migrated to EventHub) it is not - as they are modelled as 0-1 count telemetry pixels with 1 day period.

See: https://app.asana.com/1/137249556945/task/1215256061408150/comment/1217091538120617?focus=true

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1047,6 +1049,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                                                                                 uiHosting: { windowControllersManager.activeViewController },
                                                                                 isOnboardingCompletedProvider: { onboardingManager.state == .onboardingCompleted },
                                                                                 dockCustomization: dockCustomization)
+        eventHubIntegration = MacOSEventHubIntegration(privacyConfigurationManager: privacyConfigurationManager,
+                                                        keyValueStore: keyValueStore)
```
bwaresiak: Question: I think my impression from TD was that we will have separate (new) store to hold these pixels - wasn’t that the case?

macOS/DuckDuckGo/EventHub/MacOSEventHubIntegration.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    /// The `telemetry` config names (keys in the remote eventHub settings) gated behind the YouTube
+    /// analytics opt-in. These must byte-match the server-side config keys: a mismatch fails open — the
+    /// entry is never stripped, so its telemetry would be collected without consent.
+    private static let youTubeAdBlockingTelemetryConfigNames: Set<String> = [
+        "webTelemetry_youtube_adBlocker_day",
+        "webTelemetry_youtube_playabilityError_day",
+        "webTelemetry_youtube_videoAd_day",
+        "webTelemetry_youtube_staticAd_day",
+        "webTelemetry_youtube_buffering_day"
```
bwaresiak, line 38: Shouldn’t these be config-provided, so in case we ever change the telemetry, we could update these? Otherwise any new ones will be impossible to add as older client would fire telemetry without consent.

miasma13: Not sure if this could be supported by remote updates. The main intent is to wire the client-side stored consent for the predefined set of pixels/telemetries. Since the consent was given via specific UI flow I don't think we are setup to handle more generic flows. The main driver here was to backport the YT ad blocking analytics pixels to EventHub.

PR #6063, opened 2026-07-28
Add SnapshotTestingSupport package

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214200115953388/task/1216948720664505?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/task/1216266676088027

First of a stacked series landing snapshot testing incrementally: **package (this PR) → iOS adoption → macOS adoption.**

### Description

`SnapshotTestingSupport` is a thin DuckDuckGo wrapper around [Point-Free's swift-snapshot-testing](https://github.com/pointfreeco/swift-snapshot-testing) that gives shared UI snapshot tests one consistent home: rendering SwiftUI views for iOS and macOS, validating the recording environment (OS version / scale), naming reference images, and reusing SwiftUI preview states between Xcode previews and automated tests. App-facing preview helpers stay separate from the test-only assertion helpers, so production targets can share preview configuration without depending on snapshot-testing libraries.

Nothing consumes the package yet — no app, project, or submodule changes, and no reference images — so there's no behavioural impact on the apps. Adoption lands in the follow-up PRs.

See [`SharedPackages/SnapshotTestingSupport/README.md`](https://github.com/duckduckgo/apple-browsers/blob/pau/snapshot-support-package/SharedPackages/SnapshotTestingSupport/README.md) for full details — size modes, environment requirements, recording, conventions, reference storage, and worked examples.

### Testing Steps
1. In `SharedPackages/SnapshotTestingSupport`, run `swift test` → 35 tests across 7 suites pass.

### Impact
None — internal test tooling; no target consumes it yet.

---

> [!NOTE]
> **Low Risk**
> New internal test infrastructure with no app or production target adoption in this PR; changes affect only future snapshot test workflows.
> 
> **Overview**
> Introduces **`SharedPackages/SnapshotTestingSupport`**, a DuckDuckGo wrapper around Point-Free **`swift-snapshot-testing`**, and registers it in the Xcode workspace. Nothing in iOS/macOS app targets links it yet, so there is no runtime impact.
> 
> The package ships two libraries: **`PreviewSnapshots`** (SwiftUI-only, safe for app targets) to share preview state lists between Xcode previews and tests, and **`SnapshotTestingSupport`** (test-only) with **`assertImageSnapshot`** / **`assertImageSnapshots`**, size modes, light/dark and iPhone/iPad strategies, and **`snapshotBackground()`**.
> 
[...]

PR #6080, opened 2026-07-29
Adopt snapshot testing on iOS

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216991493996407
Tech Design URL: https://app.asana.com/1/137249556945/task/1216266676088027

Second of the stacked snapshot-testing series — **base: `pau/snapshot-support-package` (#6063).** Package → **iOS adoption (this PR)** → macOS adoption (#6081). ⚠️ Do not merge before #6063.

### Description

Adopts snapshot testing on iOS: mounts the `SnapshotReferences` submodule, links `SnapshotTestingSupport` into the iOS `UnitTests` target, and adds the first iOS snapshot tests (AIChatSyncPromoView, VoiceSearchFeedbackView) with their preview mocks.

The new snapshot tests are marked **`.disabled(...)` for now**, so they don't run in the default scheme or CI until the sync automations land (they need the pinned simulator + the submodule). Re-enable later by removing the trait. View edits are `#if DEBUG` preview plumbing only — no runtime or visual change.

### Reference images

Stored in the `SnapshotReferences` submodule (pinned `1c63aa9`) — review the actual PNGs here:
- [AIChatSyncPromoViewTests](https://github.com/duckduckgo/apple-browsers-snapshots/tree/1c63aa9549136c0a419f05950166d9c86177d68f/iOS/DuckDuckGoTests/AIChat/InputBox/SwitchBar/Suggestions/__Snapshots__/AIChatSyncPromoViewTests)
- [VoiceSearchFeedbackViewTests](https://github.com/duckduckgo/apple-browsers-snapshots/tree/1c63aa9549136c0a419f05950166d9c86177d68f/iOS/DuckDuckGoTests/__Snapshots__/VoiceSearchFeedbackViewTests)

### Testing Steps
1. Build the iOS `UnitTests` (via the `iOS Browser` scheme) → compiles; the new snapshot suites report as skipped (disabled), so the default run is unaffected.

### Impact
None — test-only additions (currently disabled) plus DEBUG preview plumbing; no runtime or visual change.

---

> [!NOTE]
> **Low Risk**
> Changes are test infrastructure and DEBUG preview plumbing; new snapshot suites are disabled and there is no production behavior change.
> 
> **Overview**
> Adopts the shared **snapshot testing stack** on iOS: adds the `SnapshotReferences` submodule, ignores inline `__Snapshots__` (with a macOS exception), and wires **UnitTests** to the local `SnapshotTestingSupport` package instead of direct Point-Free dependencies.
> 
> **SnapshotTestingSupport** gains preview/snapshot helpers—`PreviewSnapshots` void convenience init, per-configuration `interfaceOrientation`, and landscape-aware device/size handling in `assertImageSnapshots`—plus matching unit tests.
> 
> **First iOS image snapshot suites** for `AIChatSyncPromoView` and `VoiceSearchFeedbackView` use shared `PreviewSnapshots` definitions and `assertImageSnapshots`; both suites are **`.disabled`** until sync automations and the submodule are ready. DEBUG-only preview mocks (`VoiceSearchFeedbackView_PreviewMocks`) and preview refactors support those tests without changing production UI.
> 
> The **iOS Browser** scheme now includes `SnapshotTestingSupportTests`; `SuggestionJsonScenarioTests` imports `SnapshotTestingSupport` for inline snapshots via re-exports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b919633bff4ef7c4c5c28b79b6c02af8441534c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


frosty (approved):
I think this looks fine, however we do lose alternative previews here – dark mode for both, and a landscape orientation for the voice search feedback view.

I don't know what the shape of other snapshots will be, but I also wonder if we could add a convenience initializer to `PreviewSnapshots`, for the case where there's no state / configuration info:

```
public init<Content: View>(
    configure: @escaping () -> Content
) where State == Void {
    self.init(
        configurations: [.init(name: "", state: ())],
        configure: { _ in configure() }
    )
}
```

Then the call site for that preview would be:

```
static let snapshots = PreviewSnapshots {
    AIChatSyncPromoView(onCTATap: {}, onCloseTap: {})
}
```

pballart:
@frosty thanks for the review, let me address your comments:
1. Dark mode loss: true i skipped that since snapshots will automatically capture light and dark mode and for the previews itself we also have this option in the menu to toggle dark mode on and off.
2. Landscape orientation: this one was mistakenly removed. The good thing about this is that we can have multiple previews, they are not 1-1 mapped to snapshots so we can have 1 preview which is used for the snapshot tests and then extra previews with other configurations. I’ll add this back as a new preview.
3. Convenience init: yes this looks a good addition. My assumption was that majority of views will have state but it’s good to provide this convenience for the cases where there is no state.

pballart:
> @frosty thanks for the review, let me address your comments:
> 
> 1. Dark mode loss: true i skipped that since snapshots will automatically capture light and dark mode and for the previews itself we also have this option in the menu to toggle dark mode on and off.
> 2. Landscape orientation: this one was mistakenly removed. The good thing about this is that we can have multiple previews, they are not 1-1 mapped to snapshots so we can have 1 preview which is used for the snapshot tests and then extra previews with other configurations. I’ll add this back as a new preview.
> 3. Convenience init: yes this looks a good addition. My assumption was that majority of views will have state but it’s good to provide this convenience for the cases where there is no state.

Edit: I added support for snapshotting in landscape since it might be useful in some views.

PR #6090, opened 2026-07-29
[iOS] [Subscription] Implement Key Subflows

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216916193033428?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1215971357218019?focus=true

### Description
- Adds the post-subscription onboarding flow that hands users into the two key subflows after purchase: VPN activation and Duck.ai.
- VPN subflow: guides the user through enabling the VPN, shows live connection status, VPN tips, and widget education, with a tap-to-allow hint overlay for the system VPN permission prompt.
- Duck.ai subflow: lets the user pick an AI model and hands off into a Duck.ai chat, preselecting that model in the input bar before any prompt is sent.
- Introduces reusable onboarding scaffolding (section/section-delegate navigation and a prefetcher that warms up model/connection data ahead of time).
- Extends the shared `CardItem` component to support secondary inline text/color.
- Adds debug tooling to launch and iterate on the onboarding subflows independently.

### Testing Steps
1. In the debug menu, open the Subscription onboarding debug screen.
2. Launch the VPN activation subflow → the activation screen appears with connection info.
3. Enable the VPN → confirm the system permission prompt shows the tap-to-allow hint, and status updates to connected once allowed.
4. Continue through VPN tips and widget education → each screen advances correctly.
5. Launch the Duck.ai subflow → select a model and continue.
6. Confirm onboarding dismisses and Duck.ai opens with the chosen model preselected in the input bar (before typing anything).
7. Repeat step 5–6 starting from a fresh NTP tab → confirm the model is still preselected (in-place load path).

### Impact
**Impact Level: Medium**

#### What could go wrong?
- VPN activation could stall if the connection-info service fails → the flow surfaces status and the user can retry; failures are contained to the onboarding UI.
- Duck.ai model preselection could be dropped on the in-place load path → mitigated by preselecting the model after `load(...)` in `openAIChatInTab`, covered by view-model tests.

### Quality Considerations
- Edge cases: onboarding launched onto an NTP (no current link) vs. an existing tab; empty/failed model list handling in the prefetcher.
- Performance: prefetcher warms model and connection data off the critical path to avoid UI stalls.
- Tests: added coverage for the prefetcher, Duck.ai view model, VPN activation view model, and connection info service.
- Privacy/Security: VPN activation uses the standard tunnel controller; no new data collection introduced.

### Notes to Reviewer
- The contextual-sheet AI chat code and its tests were reverted to `main` since the new flow uses the native full contextual UI instead.

---

> [!NOTE]
> **Medium Risk**
> Touches VPN tunnel start, system permission prompts, and AI model persistence across onboarding dismissal and chat entry; failures are mostly contained in onboarding UI but incorrect connection or model state could confuse users.
> 
> **Overview**
> Adds **post-subscription onboarding** for **VPN activation** and **Duck.ai model selection**, wired through section navigation and a prefetcher for models and connection info.
> 
> The **VPN path** turns on the tunnel via the existing controller, shows pre-VPN vs VPN egress IP/location (including **(Nearest)** on info cards), handles denial and start failures, and chains **widget education** and **VPN tips**. A **tap-allow** overlay and coordinator guide the system VPN permission prompt when no config is installed. Tunnel plumbing gains **configuration-denied** signaling, **`isConfigurationInstalled`**, and **`connectionErrorObserver`** on the app dependency provider.
> 
> The **Duck.ai path** lists accessible models (tier badges, iPhone-only selection), persists the choice, and launches chat via **`openAIChat`** with **`updateSelectedModel`** so the omnibar reflects the pick on new-tab and in-tab loads. **`AIChatUserTier.resolve(from:)`** ties model access to subscription tier.
> 
> **`CardItemText`** adds optional **secondary inline text/color** and **`styledText`** for body rows (used on VPN location lines). Older standalone carousel/model-picker views are reorganized into the new VPN/DuckAI feature folders with added tests and debug playground for the hint overlay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1799c511e0b9e91154518a90a08aeab4d5e9abc4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIChatLauncher.swift
```diff
@@ -0,0 +1,35 @@
...
+@MainActor
+struct SubscriptionOnboardingDuckAIChatLauncher {
+
+    func launch(from presentingViewController: UIViewController, modelID: String?) {
+        guard let mainViewController = presentingViewController.view.window?.rootViewController as? MainViewController else {
+            assertionFailure("Expected MainViewController as rootViewController when launching Duck.ai from onboarding")
+            return
+        }
+        mainViewController.dismiss(animated: true) {
+            mainViewController.openAIChat(flowType: .mobileAppOnboarding, modelId: modelID)
```
alessandroboron, line 32: ust double-checking we don’t need to use `.default` instead of `.mobileAppOnboarding`.

`.mobileAppOnboarding` won’t show the T&C on the first prompt, but it will show after the first prompt.

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIView.swift
```diff
@@ -0,0 +1,198 @@
...
+        static let contentInsetVertical: CGFloat = 16
+    }
+
+    @StateObject private var viewModel: SubscriptionOnboardingDuckAIViewModel
+    private let title: String?
+
+    @State private var isShowingInfoSheet = false
+
+    @MainActor
+    init(viewModel: @autoclosure @escaping () -> SubscriptionOnboardingDuckAIViewModel,
```
alessandroboron, line 39: Who owns `SubscriptionOnboardingDuckAIViewModel` once the real flow exists?  If the coordinator will own the section VMs, should this be `@ObservedObject` instead of `@StateObject`? Happy to keep `@StateObject` if the view is meant to own it.

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIView.swift
```diff
@@ -0,0 +1,198 @@
...
+        static let contentInsetHorizontal: CGFloat = 16
+        static let contentInsetVertical: CGFloat = 16
+    }
+
+    @StateObject private var viewModel: SubscriptionOnboardingDuckAIViewModel
+    private let title: String?
+
+    @State private var isShowingInfoSheet = false
+
+    @MainActor
```
alessandroboron: Do we need to explicitly mark the init with `@MainActor`?

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIView.swift
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Whether rows show/report a selection at all — false on iPad, where model preselection has no way
+    /// to reach a fresh chat session, so neither the checkmark nor its accessibility value should appear.
+    var isSelectable: Bool {
+        !DevicePlatform.isIpad
+    }
+
+    var cardItems: [CardItem] {
+        viewModel.availableModels.map { model in
+            let nameParts = model.name.split(separator: " ", maxSplits: 1)
```
alessandroboron: `AIChatModel` exposes `tileAndComponents` that you can use to replace split separation logic

```
/// The display name split into a bold family part and a regular remainder, e.g.
/// "GPT-5.4 mini" → ("GPT-5.4", "mini"). Best-effort first-token split for menu rows.
var titleComponents: (bold: String, regular: String) {
   guard let spaceIndex = name.firstIndex(of: " ") else { return (name, "") }
   return (String(name[..<spaceIndex]), String(name[name.index(after: spaceIndex)…]))
}
```

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+
+import Foundation
+import Combine
+import AIChat
+import Subscription
+
+/// A seam over ``UTIModelStore`` so the model fetch and persisted selection can be mocked in tests. The
+/// default implementation wraps a live `UTIModelStore` (which fetches `/models`)
+@MainActor
+protocol SubscriptionOnboardingAIModelProviding: AnyObject {
```
alessandroboron, line 27: Interestingly, I’m doing something very similar to this (without the subscription part)

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+}
+
+@MainActor
+final class DefaultSubscriptionOnboardingAIModelProvider: SubscriptionOnboardingAIModelProviding {
+    /// Upper bound on how long the underlying `/models` fetch may run before it's treated as failed
+    private static let fetchTimeout: TimeInterval = 10
+
+    private let subscriptionManager: any SubscriptionManager
+
+    private lazy var store = UTIModelStore(modelsService: AIChatModelsService(),
```
alessandroboron: If we initialise `UTIModelStore` in line we can’t control this dependency in tests. Should we create a protocol and inject it?

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+    var persistedModelID: String? { get }
+    /// Fetches the available models, resolving once. An empty result signals "no models"; `UTIModelStore` swallows the underlying error, so failure is inferred from emptiness.
+    func fetchModels() async -> [AIChatModel]
+    func updateSelectedModel(_ modelID: String)
+}
+
+@MainActor
+final class DefaultSubscriptionOnboardingAIModelProvider: SubscriptionOnboardingAIModelProviding {
+    /// Upper bound on how long the underlying `/models` fetch may run before it's treated as failed
+    private static let fetchTimeout: TimeInterval = 10
```
alessandroboron: What does the screen show during load? I wonder if we should lower the timeout.

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/TapAllowHintOverlayPlayground.swift
```diff
@@ -0,0 +1,234 @@
...
+import SwiftUI
+import DesignResourcesKit
+import UIComponents
+
+/// Debug harness that renders a hand-built mock of the system "Add VPN Configurations" permission dialog in
+/// both the iOS 26 and iOS 18 visual styles, with ``FloatingPointerBubble`` laid on top using the same
+/// screen-centre-relative offsets as `TapAllowHintOverlayWindow`. The real dialog can't be summoned or styled
+/// on demand, so this lets us eyeball and tune the hint alignment without triggering the real permission
+/// prompt. Reached from the Subscription debug menu's Onboarding section.
+struct TapAllowHintOverlayPlaygroundView: View {
```
alessandroboron, line 26: Does this need a Preview?

[Edit], I can see this in the DebugMenu. good idea!

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+                                           subscriptionManager: subscriptionManager)
+
+    nonisolated init(subscriptionManager: any SubscriptionManager = AppDependencyProvider.shared.subscriptionManager) {
+        self.subscriptionManager = subscriptionManager
+    }
+
+    var persistedModelID: String? { store.persistedModelId }
+
+    /// Bridges `UTIModelStore`'s single-slot `onModelsUpdated` callback
+    func fetchModels() async -> [AIChatModel] {
```
alessandroboron, line 51: I believe two calls to `fetchModels()` can hang. `UTIModelStore` has a single `onModelsUpdated` slot, and `store.fetchModels()` cancels the prior internal task, so if a second `fetchModels()` starts before the first resolves, the second overwrites the slot, the first never gets its callback, and (since the single-shot guard keys off the shared slot) the first's continuation is never resumed which causes a permanent hang. Safe when called from the prefetcher serializes via `shouldStartFetch` but as safety net in case in the future we will use outside of it we could save the inflight taks and return that if there are subsequent calls.

Also given the prefetcher can cancel the task should we use `withTaskCancellationHandler`?

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    func updateSelectedModel(_ modelID: String) { store.updateSelectedModel(modelID, isNewChatContext: true) }
+}
+
+/// Backs the Duck.ai onboarding screen: reads the available AI models from the shared
+/// ``SubscriptionOnboardingPrefetcher`` (retrying on appearance if the fetch hasn't resolved yet), tracks the
+/// selected one, and persists it so the launched chat opens with it.
+final class SubscriptionOnboardingDuckAIViewModel: ObservableObject {
```
alessandroboron, line 71: Could this be @MainActor on the class instead of on each method?

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+
+    @Published private(set) var models: [AIChatModel] = []
+    @Published private(set) var selectedModelID: String?
+
+    private let prefetcher: SubscriptionOnboardingPrefetcher
+    private(set) weak var delegate: SubscriptionOnboardingSectionDelegate?
+    private var cancellables = Set<AnyCancellable>()
+
+    var availableModels: [AIChatModel] {
```
alessandroboron: Suggestion: since nothing reads the raw models (the view and the selection default both use availableModels), could we apply the entityHasAccess filter + ordering once when the sink receives `.loaded` and publish the display-ready list directly? That'd make it a single source of truth, avoid re-running the filter/sort on every body pass, and as a bonus turn availableModels into plain stored state instead of the current non-@MainActor computed var.

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+/// Backs the Duck.ai onboarding screen: reads the available AI models from the shared
+/// ``SubscriptionOnboardingPrefetcher`` (retrying on appearance if the fetch hasn't resolved yet), tracks the
+/// selected one, and persists it so the launched chat opens with it.
+final class SubscriptionOnboardingDuckAIViewModel: ObservableObject {
+
+    @Published private(set) var models: [AIChatModel] = []
+    @Published private(set) var selectedModelID: String?
+
+    private let prefetcher: SubscriptionOnboardingPrefetcher
+    private(set) weak var delegate: SubscriptionOnboardingSectionDelegate?
```
alessandroboron: Suggestion: Could we make delegate private and add a `goBack()` method so the view calls `viewModel.goBack()` instead of `viewModel.delegate?.sectionDidRequestGoBack()`? Keeps all delegate calls funneling through the VM and avoids the view being able to reach other delegate methods that are meant to be VM-internal.

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,146 @@
...
+        self.delegate = delegate
+    }
+
+    @MainActor
+    private func observePrefetcher() {
+        guard cancellables.isEmpty else { return }
+        prefetcher.$models
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] state in
+                guard let self, case .loaded(let models) = state else { return }
```
alessandroboron, line 93: What happens if the fetch fails? And do we need to show any progress view while it’s loading?

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingConnectionInfoService.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// Reads `https://duckduckgo.com/connection.json`
+struct DefaultSubscriptionOnboardingConnectionInfoService: SubscriptionOnboardingConnectionInfoService {
+    private static let connectionInfoURL = URL(string: "https://duckduckgo.com/connection.json")!
+    private static let requestTimeout: TimeInterval = 10
+
+    private let urlSession: URLSession
+
+    init(urlSession: URLSession = .shared) {
```
alessandroboron: Should we use DDG `DefaultAPIService`? it handles error codes, timeout, retry etc.

SharedPackages/UIComponents/Sources/UIComponents/iOS/CardItem.swift
```diff
@@ -99,12 +99,33 @@ public struct CardItemText {
...
+        self.secondaryColor = secondaryColor
+    }
+}
+
+extension CardItemText {
+    /// The text as a SwiftUI `Text`: the main run in `font` coloured `color` (falling back to
+    /// `defaultColor`), plus — when `secondaryText` is set — an inline second run in the same `font`
+    /// coloured `secondaryColor` (or `defaultColor`). The caller applies the modifier.
+    func styledText(defaultColor: Color) -> Text {
+        let main = Text(verbatim: text).font(font.font).foregroundColor(color ?? defaultColor)
```
alessandroboron: nip: no need to use Text(verbatim: when using a variable. Also this line `return main + Text(verbatim: " " + secondaryText)` could become `return main + " " + secondaryText`

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -4070,6 +4070,9 @@ class MainViewController: UIViewController {
                 AIChatPromptHandler.shared.setData(prompt)
             }
             loadUrlInNewTab(chatURL, inheritedAttribution: nil) { [weak self] in
+                if let modelId {
```
alessandroboron, line 4134: should we do this before the if condition so we do only once?

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationView.swift
```diff
@@ -0,0 +1,341 @@
...
+    init(viewModel: @autoclosure @escaping () -> SubscriptionOnboardingVPNActivationViewModel,
+         title: String? = nil) {
+        _viewModel = StateObject(wrappedValue: viewModel())
+        self.title = title
+    }
+
+    var body: some View {
+        SubscriptionOnboardingBaseView(
+            title: title,
+            navigationButton: .back({ viewModel.delegate?.sectionDidRequestGoBack() }),
```
alessandroboron: as https://github.com/duckduckgo/apple-browsers/pull/6090/changes#r3679644751

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationView.swift
```diff
@@ -0,0 +1,341 @@
...
+        static let featureRowSpacing: CGFloat = 10
+    }
+
+    @StateObject private var viewModel: SubscriptionOnboardingVPNActivationViewModel
+
+    private let title: String?
+
+    @State private var isShowingInfoSheet = false
+    @State private var tapAllowHint = TapAllowHintOverlayWindow()
+    @State private var didTapStartVPN = false
```
alessandroboron: suggestion: Right now we show/hide the hint based on different event that happen throughout the file and the logic is within the view. Have you considered extracting the logic into a small `TapAllowHintCoordinator: ObservableObject` that takes events in (`startTapped`, `appWillResignActive`, `appDidBecomeActive`, `connected`, `permissionDenied`, `disappeared`) and publishes a single `shouldShowHint` bool?

The View then just forwards events and maps the bool to t`apAllowHint.show()/hide()`. That way the part of the screen that requires quite a bit of manual testing becomes testable with unit tests.

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingPrefetcher.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    func fetchModelsIfNeeded() {
+        guard models.shouldStartFetch else { return }
+        modelsTask?.cancel()
+        models = .loading
+        modelsTask = Task { @MainActor [weak self] in
+            guard let self else { return }
+            let fetched = await self.modelProvider.fetchModels()
+            guard !Task.isCancelled else { return }
+            self.models = fetched.isEmpty ? .failed : .loaded(fetched)
```
alessandroboron: Do we want any fallback when the models fetch fails? or we show the users a particular message to retry onboarding later?

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingPrefetcher.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// Kicks off both fetches at flow start.
+    @MainActor
+    func prefetch() {
+        fetchConnectionInfoIfNeeded()
+        fetchModelsIfNeeded()
+    }
+
+    func fetchConnectionInfoIfNeeded() {
+        guard connectionInfo.shouldStartFetch else { return }
+        connectionInfoTask?.cancel()
```
alessandroboron: I think `connectionInfoTask?.cancel()` (and `modelsTask?.cancel()`) can never actually cancel a live task as `shouldStartFetch` is only true for `.idle` or `.failed`. Not a big deal but at first sight it could read that we cancel tasks and restart when concurrent calls are made.

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingPrefetcher.swift
```diff
@@ -0,0 +1,122 @@
...
+            guard !Task.isCancelled else { return }
+            self.models = fetched.isEmpty ? .failed : .loaded(fetched)
+        }
+    }
+
+    var persistedModelID: String? { modelProvider.persistedModelID }
+
+    func updateSelectedModel(_ modelID: String) {
+        modelProvider.updateSelectedModel(modelID)
+    }
```
alessandroboron, line 109: Should we cancel the tasks in deinit for the case where the user exits the flow mid-fetch?

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationViewModel.swift
```diff
@@ -0,0 +1,390 @@
...
+import Combine
+import Foundation
+import VPN
+
+/// Drives the VPN activation screen. It reads the original
+/// connection info from the shared ``SubscriptionOnboardingPrefetcher`` while off, retrying the fetch on appearance if it hasn't resolved yet;
+/// starts the VPN through the injected controller; and describes the new (egress) IP and location
+/// from the shared server-info observer and reports
+/// completion up to the flow via ``SubscriptionOnboardingSectionDelegate``.
+final class SubscriptionOnboardingVPNActivationViewModel: ObservableObject {
```
alessandroboron, line 26: Should we mark the ViewModel @MainActor ?

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationViewModel.swift
```diff
@@ -0,0 +1,390 @@
...
+        connectionObserver.recentValue.isConnected
+    }
+
+    var isConnectedPublisher: AnyPublisher<Bool, Never> {
+        connectionObserver.publisher
+            .map(\.isConnected)
+            .eraseToAnyPublisher()
+    }
+
+    var configurationDeniedPublisher: AnyPublisher<Void, Never> {
```
alessandroboron, line 264: he VM only observes configurationDeniedPublisher, not controllerErrorPublisher, so any non-denial activation failure fails silently: `connectionState` never flips and the screen stays as-is with no feedback. Should we surface a generic "something went wrong / Try Again" state here too? Right now the user won't know anything happened so may be worth to check with product about this

alessandroboron (commented):
@hanyutang-sandra I left some comments with questions/suggestions. Let me know what you think. Happy to sync if anything does not make sense :)

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIView.swift
hanyutang-sandra: I don't think the coordinator should own the section VMs. The coordinator owns only the
cross-section state (eg, Navigation, prefetcher and progress store) and a section
VM owns only screen state. So I’d lean towards keeping `@StateObject`.

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
hanyutang-sandra: I used `MockAIModelProvider` (which conforms to `SubscriptionOnboardingAIModelProviding`) for testing, and `UTIModelStore` is only used on prod. 

But maybe you mean testing `DefaultSubscriptionOnboardingAIModelProvider` instead? In that case imo we can use a stub service with the actual `UTIModelStore` without creating a new protocol. Lemme add the tests

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
hanyutang-sandra: It shows nothing (see screenshot). There are no separate loading or failure-state designs, so the
only role of the fetch timeout is to flip the prefetcher's state and allow a re-entry retry. Happy
to take it to 5s if you think that's better.

Screenshot 2026-08-04 at 3 04 53 PM

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
hanyutang-sandra: Hardened it. And added tests to make sure things work

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
hanyutang-sandra: See here:
https://github.com/duckduckgo/apple-browsers/pull/6090#discussion_r3715504763

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIChatLauncher.swift
hanyutang-sandra: Is T&C terms & conditions? I think they both show after the first prompt:

`.mobileAppOnboarding`
Simulator Screenshot - iPhone 17 - 2026-08-04 at 16 44 46


`.default`
Simulator Screenshot - iPhone 17 - 2026-08-04 at 16 47 13

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIView.swift
hanyutang-sandra: Yeah it can be removed. I think it’s leftover from some previous changes

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingConnectionInfoService.swift
hanyutang-sandra: Didn’t notice this service. Thx!

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingPrefetcher.swift
hanyutang-sandra: No. There’s no handling for failure state. See https://github.com/duckduckgo/apple-browsers/pull/6090#discussion_r3715504763

SharedPackages/UIComponents/Sources/UIComponents/iOS/CardItem.swift
hanyutang-sandra: I don’t think we can add String and Text() together but updated the verbatim

iOS/DuckDuckGo/MainViewController.swift
hanyutang-sandra: No unfortunately. The model updates has to stay after the load

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationViewModel.swift
hanyutang-sandra: Lemme check with product for this. We can launch the user to retry screen if there’s anything wrong, same as if the user denied the config. But the copy should probably be different

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingPrefetcher.swift
hanyutang-sandra: Oh yeah good catch. Lemme fix it

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
```diff
@@ -0,0 +1,187 @@
...
+    /// or the timeout arrives first wins; the other becomes a no-op.
+    private func resolvePending() {
+        guard isFetching else { return }
+        isFetching = false
+        timeoutTask?.cancel()
+        timeoutTask = nil
+        let continuations = Array(pendingContinuations.values)
+        pendingContinuations.removeAll()
+        continuations.forEach { $0.resume(returning: store.models) }
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: hmmm actually I think we can afford to wait longer, given both loading and failure UI shows the same thing which is an empty list

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationViewModel.swift
hanyutang-sandra: Still in discussion, but it now observes both controller error publisher and the connection error observer publisher (also include tunnel errors) and shows the retry screen. We are good with using the retry screen, but the subtitle copy may change later

hanyutang-sandra: Actually as I was typing the last comment we have agreed on the copy. Updated

iOS/DuckDuckGo/Subscription/Onboarding/Features/DuckAI/SubscriptionOnboardingDuckAIViewModel.swift
alessandroboron: May be worth discussing these scenarios with Product. Additionally, we could add pixels (count and daily) to measure how many users experience timeouts, and implement a solution if needed.

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/TapAllowHintOverlayWindow.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    private var window: UIWindow?
+
+    /// Layers the hint over the frontmost scene.
+    func show() {
+        guard window == nil, let scene = Self.foregroundWindowScene else { return }
+        let hostingController = UIHostingController(rootView: Self.hintView)
+        hostingController.view.backgroundColor = .clear
+        let window = UIWindow(windowScene: scene)
+        // The system permission dialog dims the screen with a scrim that sits above `.alert`, so `.alert + 1`
```
alessandroboron: nip: stale comment says .alert + 1 but we use greatestFiniteMagnitude :)

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/TapAllowHintOverlayWindow.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    func appDidEnterBackground() {
+        shouldShowHint = false
+    }
+
+    func turnOnFinished() {
+        reset()
+    }
+
+    func connected() {
```
alessandroboron: connected is not used, should we remove it?

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationViewModel.swift
```diff
@@ -0,0 +1,437 @@
...
+            .store(in: &cancellables)
+
+        // Both sources replay their current value on subscription, so each drops its own
+        Publishers.Merge(vpnController.controllerErrorPublisher.dropFirst(),
+                         errorObserver.publisher.dropFirst())
+            .compactMap { $0 }
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.didFailToStartVPN = true
+                self?.didDenyVPNPermission = false
```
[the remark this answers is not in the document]
alessandroboron: @hanyutang-sandra can you please check this is a real issue?

alessandroboron (approved):
@hanyutang-sandra Approving as this is in good shape and you've addressed the earlier feedback nicely.

I only left a couple of minor comments. 

I don’t believe this comment by [cursor](https://github.com/duckduckgo/apple-browsers/pull/6090/changes#r3723549460) is a real bug given the user has no VPN running yet and an error can’t surface with no activation attempt but worth a sanity check.

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationViewModel.swift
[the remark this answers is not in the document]
hanyutang-sandra: Replied in Asana https://app.asana.com/1/137249556945/task/1216050449851047/comment/1217249905808250?focus=true

iOS/DuckDuckGo/Subscription/Onboarding/Features/VPN/SubscriptionOnboardingVPNActivationView.swift
```diff
@@ -0,0 +1,353 @@
...
+            let startVPN: () -> Void = {
+                tapAllowHint.startTapped()
+                Task {
+                    await viewModel.turnOnVPN()
+                    tapAllowHint.turnOnFinished()
+                }
+            }
+            guard viewModel.didFailActivation else {
+                return .single(.init(UserText.subscriptionOnboardingVPNActivationTurnOnButton, action: startVPN))
+            }
```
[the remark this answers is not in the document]
hanyutang-sandra: It does not cause any bug or crash since the controller & tunnel tolerate it well. If the user taps it while it’s being started, the worst case is that it will get stopped to allow for recovery or corrupt telemetry. I don’t think it worths disabling the button or do anything additional to add more complexity to this layer but I will leave a note for product input. This is also something that can happen for other vpn entries

PR #6124, opened 2026-07-31
Exclude os_log_fault diagnostics from macOS crash reporting

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201037661562251/task/1217076871494606?focus=true
Tech Design URL:
CC:

### Description

macOS writes a diagnostic report for every `os_log_fault` a framework emits, flagged `is_simulated: 1` — but the process never terminated. `CrashReportReader` filtered only on file extension and filename, so these counted toward the `m_mac_crash` pixel, uploaded as crash reports, and showed the crash-report consent prompt to users whose app hadn't crashed. Now dropped at the reader, which covers all three. Driven by WebKit faulting on macOS 26.6 via `NetworkProcessProxy::didReceiveInvalidMessage`; Safari faults identically on the same build, so the fault itself is an OS regression, not ours.

### Testing Steps
1. Run the `CrashReportingTests` suite in Xcode — 3 new reader tests cover the `is_simulated` flag and the prompt path.
2. On macOS 26.6, put an `ExcUserFault_DuckDuckGo-*.ips` carrying the app's bundle ID into `~/Library/Logs/DiagnosticReports/`.
3. Relaunch — no consent prompt, no crash pixel. A genuine crash report still prompts.

### Impact
Low. Narrows which diagnostic files count as crashes; genuine crash reports are untouched.

#### What could go wrong?
A real crash report misclassified as simulated would go unreported → `is_simulated` is only set by the OS on non-fatal faults, and the seven pre-existing reader tests still assert real reports pass through.

### Quality Considerations
Only covers the Sparkle/DMG path. The App Store build collects via MetricKit (`CrashCollection`), which has no `is_simulated` field and needs its own predicate — not in this PR.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Narrows which diagnostic files count as crashes; real reports without `is_simulated` still pass through, with regression tests for existing reader behavior.
> 
> **Overview**
> macOS **`.ips`** diagnostics from **`os_log_fault`** (often **`ExcUserFault_*`**, header **`is_simulated: 1`**) were treated like real crashes because **`CrashReportReader`** only matched extension and filename. That drove **`m_mac_crash`** pixels, uploads, and the crash-report consent prompt when the app had not actually terminated.
> 
> This PR adds **`isSimulated`** on **`CrashReport`** — **`JSONCrashReport`** reads **`is_simulated`** from the IPS header; legacy **`.crash`** reports stay **`false`** — and **`CrashReportReader.getCrashReports`** drops simulated reports. That single path covers **`CrashReporter`** and **`hasNewCrashReport`** (consent / unclean-exit detection). Three reader tests cover **`ExcUserFault_`** filenames, the flag regardless of name, and the empty prompt path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a704f98bc32b78b04786bea66b5ea9c6a31c526d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReport.swift
+++ b/macOS/LocalPackages/CrashReporting/Sources/CrashReporting/CrashReport.swift
@@ -33,6 +33,10 @@ protocol CrashReport: CrashReportPresenting {
 
     /// The date the crash report file was created, used as a proxy for when the crash occurred.
     var creationDate: Date? { get }
+
+    /// `true` when the report is a non-fatal diagnostic filed via `os_log_fault` rather than a crash.
+    /// The process kept running, so these must not be counted, uploaded, or prompted for as crashes.
+    var isSimulated: Bool { get }
 }
 
 final class LegacyCrashReport: CrashReport {
@@ -91,6 +95,9 @@ final class LegacyCrashReport: CrashReport {
 
     let appVersion: String? = nil
     let bundleID: String? = nil
+
+    // `os_log_fault` diagnostics are only ever written in the JSON (`.ips`) format.
+    let isSimulated = false
 
     var creationDate: Date? { fileManager.fileCreationDate(url: url) }
 }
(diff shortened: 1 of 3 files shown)


samsymons (approved):

diegoreymendez (approved):
Fantastic work @jotaemepereira !

PR #6135, opened 2026-08-03
iOS: Fix invalid & missing Search Token (Dindex) requests

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217093200725389

### Description

The Search Token (Dindex) experiment attaches a short-lived, UA-bound token to variant-b SERP requests so the backend can serve a faster combined Index/Deep response. Backend analysis surfaced two issues degrading the experiment — **invalid tokens** (token minted under one User-Agent but presented on a SERP request sending a different UA) and **missing tokens** (variant-b SERP requests arriving with `dindexexp=b` but no token). This PR addresses both and resets the experiment.

**Invalid tokens (UA mismatch)**
- **Empty iPad tabs:** the warm fell back to a mobile UA (`?? false`) when there was no current tab, but a fresh iPad tab navigates the SERP in desktop mode — so the token was minted mobile and the SERP went out desktop. Warm now falls back to `AppWidthObserver.shared.isLargeWidth`, the same default `Tab.init` uses, so the warmed UA matches the tab the SERP will use.
- **UA-aware fetching:** the fetcher now tracks the UA a token was minted for and refetches when the warm UA changes (desktop/mobile switch), keeping the cached token aligned to the current mode.
- **Always attach the token:** retrieval no longer withholds the token when the current UA differs from the mint UA. The live token is always attached, so the backend keeps visibility on any residual mismatch (invalid-token) cases instead of the client silently dropping them. Only *fetching* is UA-aware; retrieval is unconditional.

**Missing tokens (cold cache on non-omnibar SERP paths)**
- The token was only ever warmed on address-bar focus, so every non-omnibar path to a SERP (external links, widgets, share extension, bookmarks/NTP favorites, session restore) plus the cold-start race hit the interceptor with a cold cache and attached nothing. Warming now also runs **on app foreground** (covers cold launch and every resume, preceding those entry points) and **reactively on a cache miss** at attach time (warm-on-miss, so the next request finds a token).

**Reset experiment**
- Renamed the subfeature/flag `searchTokenExperiment` → `searchTokenExperimentV2` so the corrected client starts fresh cohort assignment and analysis rather than mixing with previously contaminated data.

### Testing Steps

Prereq: a proxy (Proxyman/Charles) with SSL enabled for `duckduckgo.com` and the token host (`links.duckduckgo.com`). Force the treatment cohort via **Settings → Debug → Search Token → Experiment cohort override → treatment** (the flag is now `searchTokenExperimentV2`).

1. **iPhone, omnibar search:** fresh new tab → tap the address bar → type a query → search. Confirm the token request and the SERP request carry the **same** `User-Agent`, and the SERP request includes the `X-DDG-Search-Token` header and `dindexexp=b`.
2. **iPad (full screen), empty tab:** fresh new tab → search. Confirm both the token request and the SERP request go out with the **desktop (Macintosh)** UA and match, and the token is attached.
3. **Desktop mode:** on any page toggle **Desktop Site** on, then search. Confirm token and SERP UAs are both desktop and match.
4. **Foreground warming:** force-quit the app, relaunch, and reach a SERP via a **non-omnibar** path (external `duckduckgo.com` link, widget, share extension, or a bookmark/NTP favorite). Confirm the **token request goes out** on foreground.
5. **Control cohort:** switch the override to control and search. Confirm the SERP request has `dindexexp=a` and **no** `X-DDG-Search-Token` header.

### Impact and Risks

**Low.** All behavior is cohort-gated to the Search Token experiment's treatment arm; control and non-enrolled users are unaffected. The only always-on change is the `searchTokenExperimentV2` rename, which resets cohort assignment (intended). Extra warming is bounded by the fetcher's coalescing (skips in-flight/fresh tokens), so `/search-token` traffic stays low. A missing or mismatched token is a safe no-op on the client — the SERP still loads, just without the speed-up — and mismatches are now intentionally surfaced to the backend rather than hidden.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Changes are gated to the search-token experiment treatment cohort; the v2 rename intentionally resets assignment. Extra token fetches are coalesced by the fetcher, and SERP loads safely without a token.
> 
> **Overview**
> Fixes **invalid** (UA-mismatched) and **missing** search tokens on treatment SERP navigations for the Dindex experiment, and resets the experiment as **`searchTokenExperimentV2`** so cohorts and metrics start clean.
> 
> **UA alignment:** `SearchTokenFetcher` records the UA used to mint the token and refetches when the warm UA changes (e.g. desktop/mobile), while **`retrieveToken()`** still returns any live cached token so mismatches stay visible to the backend. Token warming uses **`AppWidthObserver.shared.isLargeWidth`** when there is no current tab (iPad empty-tab desktop SERP). On SERP attach, **`TabViewController`** sets **`webView.customUserAgent`** to match the warmed token path.
> 
> **Missing tokens:** Warming runs on **app foreground** in addition to omnibar focus, and **`searchToken(for:)`** triggers **`warmSearchTokenIfEligible()`** on a cache miss so the next navigation can attach a token. Experiment pixels and enrollment now target **`searchTokenExperimentV2`** / **`iOSBrowserConfigSubfeature.searchTokenExperimentV2`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7216feb82f293d378a9fc82a62657a85500217d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/SearchToken/SearchTokenFetcher.swift
+++ b/iOS/DuckDuckGo/SearchToken/SearchTokenFetcher.swift
@@ -29,7 +29,7 @@ import os.log
 ///
 /// - `fetchIfNeeded(userAgent:)` fetches only when there is no live token or the current token is within
 ///   `window` seconds of expiry (refresh-ahead), and coalesces concurrent triggers into one request.
-/// - `retrieveToken()` returns the cached token synchronously while it is still within its TTL, else `nil`.
+/// - `retrieveToken(matching:)` returns the cached token if it was minted for the given UA and is still within its TTL, else `nil`.
 final class SearchTokenFetcher {
 
     private let requester: SearchTokenRequesting
@@ -39,6 +39,7 @@ final class SearchTokenFetcher {
 
     private let lock = NSLock()
     private var cachedToken: String?
+    private var cachedUserAgent: String?
     private var fetchedAt: Date?
     private var isFetching = false
 
@@ -53,9 +54,9 @@ final class SearchTokenFetcher {
     }
 
     /// The cached token while it is still within its TTL, otherwise `nil`. Synchronous, non-blocking.
-    func retrieveToken() -> String? {
+    func retrieveToken(matching userAgent: String) -> String? {
         lock.lock(); defer { lock.unlock() }
-        guard let cachedToken, let fetchedAt else { // Token Exists
+        guard let cachedToken, let fetchedAt, cachedUserAgent == userAgent else { // Token exists for this UA
             return nil
         }
         guard now().timeIntervalSince(fetchedAt) < ttlProvider() else { // Token is valid
@@ -70,12 +71,12 @@ final class SearchTokenFetcher {
     /// All locking is confined to the synchronous helpers below; the lock is never touched across the
     /// `await`, keeping this async-safe.
     func fetchIfNeeded(userAgent: String) async {
-        guard beginFetching() else { return }
+        guard beginFetching(userAgent: userAgent) else { return }
         defer { endFetching() }
 
         do {
             let token = try await requester.requestToken(userAgent: userAgent)
-            store(token: token)
+            store(token: token, userAgent: userAgent)
             Logger.general.debug("SearchToken: fetched (len=\(token.count, privacy: .public))")
         } catch {
             Logger.general.debug("SearchToken: fetch failed: \(error.localizedDescription, privacy: .public)")
@@ -86,10 +87,10 @@ final class SearchTokenFetcher {
 
     /// Marks a fetch as in flight and returns whether the caller should proceed. Skips if a fetch is
     /// already running or the current token still has more than `window` seconds of life.
-    private func beginFetching() -> Bool {
+    private func beginFetching(userAgent: String) -> Bool {
         lock.lock(); defer { lock.unlock() }
         if isFetching { return false }
-        if let fetchedAt, cachedToken != nil, // Token exists
+        if let fetchedAt, cachedToken != nil, cachedUserAgent == userAgent, // Fresh token already cached for this UA
            ttlProvider() - now().timeIntervalSince(fetchedAt) > windowProvider() { // Token is fresh
             return false
         }
@@ -97,9 +98,10 @@ final class SearchTokenFetcher {
         return true
     }
 
-    private func store(token: String) {
+    private func store(token: String, userAgent: String) {
         lock.lock(); defer { lock.unlock() }
         cachedToken = token
+        cachedUserAgent = userAgent
         fetchedAt = now()
     }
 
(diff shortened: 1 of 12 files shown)


iOS/DuckDuckGo/SearchToken/SearchTokenFetcher.swift
```diff
@@ -84,22 +87,24 @@ final class SearchTokenFetcher {
...
+        if let fetchedAt, cachedToken != nil, cachedUserAgent == userAgent, // Fresh token already cached for this UA
            ttlProvider() - now().timeIntervalSince(fetchedAt) > windowProvider() { // Token is fresh
             return false
         }
         isFetching = true
         return true
     }
 
-    private func store(token: String) {
+    private func store(token: String, userAgent: String) {
```
dus7: Nitpick: now, looking at it I think calling this method `store` makes it a bit confusing - there’s no storage involved, but since both properties are called `cached*` maybe lets rename this method to `cache(token:userAgent:)`?

iOS/DuckDuckGo/SearchToken/SearchTokenFetcher.swift
```diff
@@ -39,6 +41,7 @@ final class SearchTokenFetcher {
 
     private let lock = NSLock()
     private var cachedToken: String?
+    private var cachedUserAgent: String?
```
dus7, line 44: Wondering if this could be optimized by keeping it as `[UAString: TokenData]` where TokenData is `var token, date` and then look up by UA? It would prevent excessive token exchange and invalidation when switching back and forth between small form factor and/or desktop version when having multiple tabs. Not sure how that affects the experiment though, but I assume two (or more) tokens can live side-by-side as long as we’re within their TTL?

dus7 (approved):
Works as expected. Tested on a device and simulator. 
I left one nitpick and one optimization option, though I’m not sure it is feasible with existing contract.

iOS/DuckDuckGo/SearchToken/SearchTokenFetcher.swift
hassaanelgarem: Great idea and makes sense, and you're right multiple tokens can live side-by-side. However, right now, we always attach the token even when the UA doesn't match, on purpose, so the backend can see the mismatch rate. Looking up strictly by UA would mean a SERP with no cached token for its UA gets nothing, which quietly kills the exact signal we're using to validate these fixes.

I would do this as a follow-up once we validate that these fixes are working in production.

iOS/DuckDuckGo/SearchToken/SearchTokenFetcher.swift
hassaanelgarem: e119f6846a8bb7521bf9577bd64c219c6b7643c4 👍

PR #6157, opened 2026-08-04
[macOS] Stop asserting on expected too-long-filename download errors

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202406491309510/task/1213975525679527
CC: @mallexxx

### Description
Downloads whose server-suggested filename is too long for the filesystem fail with `NSFileWriteInvalidFileNameError` (underlying POSIX `ENAMETOOLONG`). This is an expected, non-actionable condition, but it was missing from the write-error allow-list added in #4483, so it tripped the "unexpected error" pixel assertion.

This surfaced as the `m.mac.debug.assertion.failure` anomaly on 2026-08-03 (+420%), which tracked the 1.200.1 rollout rather than any real download regression. Adding the code to the allow-list stops the false assertion; the download still fails with the real underlying error as before.

### Testing Steps
1. Trigger a download whose suggested filename exceeds the filesystem limit (or rely on the added unit test).
2. Confirm the download is marked failed with the real error and no debug assertion pixel fires.

### Impact
Low — removes a false-positive debug assertion; no user-facing behavior change.

---

> [!NOTE]
> **Low Risk**
> Only adjusts which Cocoa write errors skip a debug assertion; download failure handling and user flows are unchanged.
> 
> **Overview**
> Stops false **`m.mac.debug.assertion.failure`** pixels when a download’s server-suggested name exceeds the filesystem limit (`NSFileWriteInvalidFileNameError` / `ENAMETOOLONG`). That case was missing from the legitimate write-error allow-list in `FileDownloadManager.defaultDownloadLocation`, so placeholder creation failures were treated as “unexpected” even though the download already fails with the real Cocoa error.
> 
> **`NSFileWriteInvalidFileNameError`** is now included alongside disk-full, permission, and similar codes so the assertion is skipped while the task still ends in **`.failed`** with the underlying error—same user-visible outcome, no save dialog fallback.
> 
> Adds **`testWhenFileCreationFailsWithInvalidFileNameThenDownloadIsMarkedFailed`** to lock in: no `chooseDestination`, nil destination to WebKit, and failed state with the invalid-filename code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f3e2541e03bea7827ee84a8a4e56c823c6b337d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/FileDownload/Model/FileDownloadManager.swift
+++ b/macOS/DuckDuckGo/FileDownload/Model/FileDownloadManager.swift
@@ -256,6 +256,7 @@ extension FileDownloadManager: WebKitDownloadTaskDelegate {
                 NSFileWriteVolumeReadOnlyError,   // EROFS   — read-only filesystem
                 NSFileWriteNoPermissionError,     // EACCES/EPERM — no write permission
                 NSFileNoSuchFileError,            // ENOENT  — parent dir removed / volume unmounted
+                NSFileWriteInvalidFileNameError,  // ENAMETOOLONG — suggested filename too long for the filesystem
                 NSFileWriteUnknownError,          // catch-all for unexpected write failures
             ]
             Logger.fileDownload.error("Failed to create file in the Downloads folder: \(error)")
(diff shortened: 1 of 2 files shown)


jleandroperez (approved):
:shipit:

PR #6159, opened 2026-08-04
[macOS] Load pinned tabs before extracting Duck.ai page content

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1217124903028827?focus=true
CC:

### Description

Attaching a pinned tab that hadn't been selected since browser start sent no page content to the conversation. The attachment card showed up as normal, so it looked like it had worked.

The check for "does this tab need loading before I read its content?" missed two cases:

- **A pinned tab restored at launch.** It's already a full tab, but its page isn't loaded until you first click it — so content got read from an empty web view.
- **A tab that's still loading.** The page URL is set the moment a load starts, which made the tab look ready, so content got read half-way through the load. This one affects suspended tabs on `main` too; it's hidden by the second or two people take to type a prompt.

Both now load (or wait for the in-flight load) before reading. Also waits on "loading finished" instead of "navigation finished", because restoring a saved page doesn't produce the navigation the old signal needed.

Covers both attach paths — the omnibar and the sidebar `@`-mention picker share the same extraction code.

### Testing Steps

1. Pin a content-heavy page. Quit the browser, relaunch, and **do not** click the pinned tab.
2. Switch the address bar to Duck.ai mode and attach that pinned tab.
3. Ask something answerable only from that page → the answer uses the page content. On `main` the same steps answer with no page context.
4. Repeat steps 1–2, then submit immediately instead of pausing → content still arrives (covers the in-flight-load case).
5. Repeat step 4 with a *suspended, unpinned* tab → content still arrives. This case is also broken on `main`.
6. Attach a normal already-loaded tab, and the current tab → both attach with no reload flicker and no added delay.
7. Attach a pinned tab pointing at an unreachable host → resolves promptly rather than hanging for ~5s.
8. Repeat step 3 through the sidebar's `@`-mention picker → content arrives there too.

### Impact

Low. Limited to attaching a tab in Duck.ai; already-loaded tabs take the same no-wait path as before.

#### What could go wrong?

- A load that never reports finishing would leave the attachment dropped, as it is today → bounded by the existing timeout, and the signal now used is the one the lazy loader already relies on for these tabs.
- Attaching a tab now triggers a load for pinned tabs that hadn't loaded → only when the web view has no page at all, so no visible page is reloaded under the user.

### Quality Considerations

Unit tests cover the new check for both shapes that need a load, including the restored-pinned-tab shape paired with an assertion that the old check reads false for it. The negative case and the in-flight-load branch aren't unit-testable — tabs never navigate under unit tests, so every tab has a nil web view URL. Those rely on the manual steps above.

---

> [!NOTE]
> **Low Risk**
> Scoped to Duck.ai tab attachment and page-context extraction; already-loaded tabs keep the no-wait path. Bounded timeouts limit hang risk if a load never finishes.
> 
> **Overview**
> Fixes **empty or partial page content** when attaching tabs in Duck.ai (omnibar and sidebar `@` picker) by treating “needs a real page in the web view” more broadly than “was just materialized from suspended.”
> 
> **`ResolvedTab.needsLoad`** is true when the tab was materialized **or** `webView.url` is still nil — covering **pinned tabs restored at launch** that are already `.loaded` but never navigated until first click.
> 
> **Submit-time extraction** now waits on an **in-flight load** before starting another (`isLoading` first), then reloads only when `needsLoad` is still true. Prewarm on attach uses the same `needsLoad` gate.
> 
> The wait path switches from **navigation finished** to **`loadingFinishedPublisher`** (same signal as `TabLazyLoader`), so restored interaction state without a navigation event still completes correctly.
> 
> Unit tests cover `needsLoad` for materialized suspended tabs and restored-but-never-navigated tabs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c8d5ad0207708097c7097b9a8f9a72f68e8bc1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
@@ -856,13 +856,15 @@ final class AIChatOmnibarController {
         persistTabAttachmentsToActiveTab(current)
     }
 
-    /// Wakes a just-attached tab if it's suspended so its content is loaded by the time the user
-    /// submits, avoiding a submit-time wait. Fire-and-forget — `extractPageContextsForOmnibarSubmit`
-    /// re-resolves and wakes regardless, so this is purely a latency optimization.
+    /// Wakes a just-attached tab whose page isn't loaded yet — a suspended tab, or a pinned tab
+    /// restored at launch and not yet selected — so its content is ready by the time the user submits.
+    /// Fire-and-forget: `extractPageContextsForOmnibarSubmit` re-resolves and waits regardless, so this
+    /// is purely a latency optimization. Without it a pinned-tab attach pays the full navigation +
+    /// collect budget on the submit critical path.
     private func prewarmAttachedTab(id: String) {
         guard let originTabCollection = origin?.originTabCollectionViewModel,
               let resolved = AIChatTabPickerSource.materializeAttachableTab(withId: id, forOrigin: originTabCollection, in: Application.appDelegate.windowControllersManager),
-              resolved.wasMaterialized else {
+              resolved.needsLoad else {
             return
         }
         resolved.tab.reload()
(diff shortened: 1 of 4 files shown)


jotaemepereira (approved):
LGTM

PR #5683, opened 2026-07-06
Remove eager full-screen tab snapshot that crashed in drawHierarchy

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214294661819890/task/1216237756104396?focus=true
Tech Design URL:
CC:

### Description
The Unified Toggle Input feature captured a full-screen snapshot of the current tab on every tab activation, and that snapshot rendered the whole view — including the live web view — via a synchronous bitmap render that crashed while reading the web content's graphics surface. This removes that eager per-activation capture. The swipe-between-tabs overlay is unaffected: it still captures the current tab the moment a swipe begins, and adjacent tabs fall back to the existing tab preview.

### Testing Steps
1. With Unified Toggle Input enabled, open several tabs and switch between them rapidly → no crash, tabs switch normally.
2. Swipe left/right between tabs → the tab you're swiping from animates correctly; adjacent tabs slide in (incoming tab may show the plain page preview without chrome the first time).
3. Load heavy/media-rich pages, background/foreground the app, and switch tabs → no crash.

### Impact
Low

#### What could go wrong?
On the first swipe toward a tab, the incoming tab shows a webview-only preview without chrome until it has been a swipe source once — cosmetic and transient, mitigated by the source tab always being captured fresh at swipe start.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Targeted removal of a crash-prone capture path with no auth or data changes; main tradeoff is a possible first-swipe preview without chrome until that tab is captured at swipe start.
> 
> **Overview**
> **Removes the per-tab-activation full-screen snapshot** that ran on every Unified Toggle Input refresh via `refreshUnifiedToggleInput`. That path deferred a `UIGraphicsImageRenderer` + `view.drawHierarchy` capture of the entire `MainViewController` (including the live web view) into `previewsSource.updateFullScreenSnapshot`, which could crash when reading the web content graphics surface.
> 
> The **`captureCurrentTabScreenSnapshotIfPossible`** helper is deleted from `MainViewController`, along with the **`defer` block** that invoked it on each tab activation. Tab switching no longer triggers that synchronous render.
> 
[...]

PR #6162, opened 2026-08-04
iOS: Capture tab previews with takeSnapshot instead of blocking drawHierarchy

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1206488453854252/task/1217093719249437
Tech Design URL:
CC:

### Description

Tab thumbnails were captured by rendering the web view synchronously on the main thread, which blocks until the render server responds — and since web content lives in a separate process, that wait is cross-process. It ran on every completed page load.

Capture now uses WebKit's asynchronous snapshot API, which this same file already uses for its other two preview paths, so the work no longer occupies our main thread. The old synchronous renderer is kept for the one case the snapshot API cannot cover: a JavaScript alert on screen.

Apple's hang diagnostics list this call site as a hang point in every release since 7.227.0, growing to roughly 2 reports/day in 7.230.0. The same API caused the 7.226.0/7.227.0 hang spike, fixed then by #5683.

### Testing Steps

Regression testing only. The hang is a 95th-percentile field effect whose cost varies with page complexity and memory pressure, so it will not reproduce reliably on a development device and a clean local run would not demonstrate the fix — that will be confirmed from Organizer hang data after release. What matters here is that tab thumbnails are unchanged.

Throughout: load a content-heavy page (e.g. a news site), open the tab switcher, and check the thumbnail shows the right page correctly framed — no toolbar chrome captured, no black bands, no vertical offset, no stretching, not blank.

1. Address bar at the top.
2. Address bar at the bottom (Settings → Appearance → Address Bar).
3. Scrolled down far enough that the toolbars have auto-hidden.
4. Open 4–5 tabs on different sites → each thumbnail shows its own tab and none are blank.
5. On a page that shows a JavaScript alert, with the alert on screen → the thumbnail is still captured and includes the alert. This is the one case that deliberately still uses the old synchronous renderer.

Steps 1–3 matter most: thumbnails are cropped using the web view's content insets, and those differ in each of those states.

### Impact

Medium. Tab thumbnails are user-visible, so a capture regression would show up as blank or badly framed thumbnails. No privacy or data-loss risk.

#### What could go wrong?

- The snapshot API returns nil where the old renderer returned an image → the thumbnail is simply not updated; `updatePreview()` already skips nil images, so the previous thumbnail stays.
- Framing shifts, because the old path cropped top and bottom content insets while the existing geometry helper returned full bounds → a new inset-aware overload reproduces the original framing, covered by unit tests and by steps 1–3.

### Quality Considerations

Only the inset-cropping path is live in production. The `floatingUI` flag selecting the full-bounds path is absent from the live privacy config and its flag default is `.disabled`, so every user takes the cropped path — which is also why the synchronous renderer was universal rather than a rare fallback.

This is a partial fix. The call site accounts for roughly 8% of hang reports in 7.230.0. Apple's diagnostics carry no duration field, so its share of hang *seconds* is unknown and plausibly larger, since a render-server round trip is a long stall. Unrelated subsystems also roughly doubled in that release, which points to broader main-thread contention; that is tracked separately on the linked task.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> User-visible tab thumbnails depend on snapshot success and inset math; failures keep the previous image, but framing regressions would show in the tab switcher.
> 
> **Overview**
> Tab switcher thumbnails are now captured with WebKit’s asynchronous **`takeSnapshot`** instead of synchronously **`drawHierarchy`** on the main thread, reducing main-thread stalls tied to cross-process render-server waits on page load.
> 
> When **floating UI** is off (the live default), snapshot framing matches the old cropped thumbnails via a new **`WebViewPreviewSnapshotGeometry.visibleRect`** overload that subtracts top/bottom scroll **content insets**; horizontal insets are ignored. With floating UI enabled, full web view bounds are still used.
> 
> **`preparePreviewSync`** remains only when a JavaScript alert overlay is visible, because **`takeSnapshot`** does not include native overlays. Unit tests cover inset edge cases (empty height, horizontal-only insets).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 434e1120d54779fb4531a4cb6a26fa6715c78bd4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/TabViewController.swift
+++ b/iOS/DuckDuckGo/TabViewController.swift
@@ -59,6 +59,17 @@ enum WebViewPreviewSnapshotGeometry {
         guard webViewBounds.width > 0, webViewBounds.height > 0 else { return nil }
         return webViewBounds
     }
+
+    /// The web content rect with the top and bottom content insets removed, matching the framing the
+    /// synchronous `drawHierarchy` capture produced.
+    static func visibleRect(webViewBounds: CGRect, contentInset: UIEdgeInsets) -> CGRect? {
+        let rect = webViewBounds.inset(by: UIEdgeInsets(top: contentInset.top,
+                                                       left: 0,
+                                                       bottom: contentInset.bottom,
+                                                       right: 0))
+        guard rect.width > 0, rect.height > 0 else { return nil }
+        return rect
+    }
 }
 
 class TabViewController: UIViewController {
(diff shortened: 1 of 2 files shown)


brindy (approved):
LGTM! - thanks

PR #6165, opened 2026-08-04
iOS: Add native input editing bridge (editPrompt)

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1217249867328887
Tech Design URL: https://app.asana.com/1/137249556945/task/1216352541975427

### Description
Adds the messaging layer for native-owned message editing in Duck.ai. The frontend can hand an edit off to the native input, which opens prefilled with the message's text and attachments, waits for the user to submit, and returns the edited result for the frontend to apply. The frontend can also cancel an in-progress edit from its side (`cancelEdit`). It's advertised to the frontend behind the `nativeAIPromptEditing` flag; when the flag is off, the frontend keeps its current inline editing (no regression).
### Testing Steps
Prerequisites: 
enable **Settings → Debug → Feature flags → `nativeAIPromptEditing`** (default off); open a **Duck.ai full-tab** chat;
set custom Duck.ai url **Settings → Debug → AI Chat  → Set Custom AI Chat URL → https://use-serp-dev-testing1.duck.ai/**

1. Go to Duck.ai and start a conversation
2. Tap on the edit button below the latest sent message
3. This should trigger edit more, test submission, removing attachments and cancelling
4. Disable `nativeAIPromptEditing`, tap on the edit button on the FE, old inline experience should be visible

### Impact
Low — the capability is behind a default-off flag and there is no frontend consumer yet, so there is no user-facing change.

#### What could go wrong?
- Nothing user-facing: gated by the default-off `nativeAIPromptEditing` flag, and no frontend sends the message yet → no path reaches users.

### Quality Considerations
Known gaps, tracked and intentionally out of scope for this PR:
- Teardown-to-cancel (e.g. tab switch) so a pending edit can't leak across tabs — will be handled with the edit-mode UI.
- Attachment double-counting against the per-conversation limit during edit — https://app.asana.com/1/137249556945/task/1217151500573898
- The `#if DEBUG` E2E harness (the `+ → New chat` repurpose, `debugSimulateEditPromptRoundTrip`, and preset payload) is a temporary internal hook and should be removed once the frontend wires the real `editPrompt` / `cancelEdit` calls.

---

> [!NOTE]
> **Low Risk**
> Default-off feature flag with no production FE consumer yet; changes are additive bridge/API with stub defaults on `AIChatInputBoxHandling`.
> 
> **Overview**
> Adds the **native-owned message editing** bridge so the Duck.ai web app can hand off edits to the iOS unified toggle input (UTI) instead of inline editing.
> 
> **Bridge & capability:** New user-script messages `editPrompt` and `cancelEdit`, plus `supportsNativePromptEditing` on native config (gated by `nativeAIPromptEditing` and native chat input availability). `EditPromptRequest` / `EditPromptReply` define the async contract: submit with prompt, images, and files, or `{ cancelled: true }`.
> 
> **UTI behavior:** `UnifiedToggleInputCoordinator` implements `editPrompt` by suspending until the user submits or cancels—prefilled text, decoded attachments, edit UI, and the “replace response” disclaimer when `hasResponsesToLose` is true. Submit resolves the continuation with encoded attachments; cancel/background ends edit mode with `.cancelled`.
> 
> **Attachments:** iOS `AIChatImageAttachment` can retain original base64 data/format for round-trip encoding; the image encoder prefers that over re-JPEG/PNG from the display thumbnail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c290845eb08f24a3a795b4f1c00a00cd00317588. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AIChat/UserScript/AIChatUserScript.swift
+++ b/iOS/DuckDuckGo/AIChat/UserScript/AIChatUserScript.swift
@@ -332,7 +332,7 @@ final class AIChatUserScript: NSObject, Subfeature {
             "prompt": "Debug edit: change me and tap send",
             "hasResponsesToLose": true,
             "images": [["data": Self.debugPNGBase64, "format": "png"]],
-            "files": [["data": Self.debugTextBase64, "fileName": "notes.txt", "mimeType": "text/plain"]]
+            "files": [["data": Self.debugPDFBase64, "fileName": "notes.pdf", "mimeType": "application/pdf"]]
         ]
         Task { @MainActor in
             let reply = await handler.editPrompt(params: params, message: DebugEditUserScriptMessage())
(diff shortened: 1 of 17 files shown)


iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -749,23 +749,64 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
...
+    private func makeAttachments(from request: EditPromptRequest) -> [UnifiedToggleInputAttachment] {
+        var attachments: [UnifiedToggleInputAttachment] = []
+        for image in request.images ?? [] {
+            guard let data = Data(base64Encoded: image.data), let uiImage = UIImage(data: data) else { continue }
+            let fileName = "image.\(image.format == "png" ? "png" : "jpg")"
+            attachments.append(.image(AIChatImageAttachment(image: uiImage, fileName: fileName)))
+        }
+        for file in request.files ?? [] {
+            guard let data = Data(base64Encoded: file.data) else { continue }
+            attachments.append(.file(AIChatFileAttachment(data: data, fileName: file.fileName, mimeType: file.mimeType)))
```
aataraxiaa, line 806: Attaching a PDF through this path (same pdf that is there when editing begins) blocks submit with "We can't read one of the files attached. Please check that it isn't corrupted and try again.” Removing the pdf but keeping the image submitted without error.

Reason -> FileDescriptor copies pageCount: nil, and pageValidationError reads nil-for-a-PDF as "inspection failed" rather than "never inspected", so it returns fileUnreadable.

UnifiedToggleInputAttachmentPresenter.makeFileAttachment could be used here:

```
attachments.append(.file(UnifiedToggleInputAttachmentPresenter.makeFileAttachment(
     data: data,
    fileName: file.fileName,
   mimeType: file.mimeType )))
```

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
```diff
@@ -758,14 +774,35 @@ final class UnifiedToggleInputCoordinator: NSObject, AIChatInputBoxHandling {
...
+        }
+        return attachments
+    }
```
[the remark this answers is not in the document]
aataraxiaa: Think this is valid.

iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
```diff
@@ -1436,9 +1436,14 @@ extension MainViewController: AIChatTabChatHeaderViewDelegate {
...
 
     func aiChatTabChatHeaderDidTapNewChat() {
+#if DEBUG
+        // Temporary internal E2E hook: borrow this button to drive the editPrompt round-trip.
+        currentTab?.debugSimulateEditPrompt()
+#else
         unifiedToggleInputCoordinator?.startNewChat()
         unifiedToggleInputCoordinator?.showExpanded(inputMode: .aiChat)
         currentTab?.submitStartChatAction()
+#endif
```
aataraxiaa: Should this swap be gated on the flag? (so we dont’ impact all debug builds). Or, maybe you are going to remove before merge.

iOS/DuckDuckGo/AIChat/UserScript/AIChatUserScript.swift
```diff
@@ -256,6 +256,12 @@ final class AIChatUserScript: NSObject, Subfeature {
             return handler.enableChatInput
         case .focusChatInput:
             return handler.focusChatInput
+        case .editPrompt:
+            return { [weak self] params, message in
```
aataraxiaa, line 260: If self has gone this closure returns nil, so a teardown hands the FE an empty reply instead of the { cancelled: true }. Unsure of impact.

aataraxiaa (approved):
Left a few comments @SabrinaTardio. Most important one re: pdf attachment submit failure. However, approving as you can address them here or in follow-up PR

iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputCoordinator.swift
SabrinaTardio: Ah Thanks! I did not notice… Fixed in the follow up PR

iOS/DuckDuckGo/AIChat/UserScript/AIChatUserScript.swift
SabrinaTardio: fixed in the follow up PR

PR #6087, opened 2026-07-29
iOS Promo Queue: Add lease arbitration and modal-side coordination

bkunat:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671677432066/task/1216967536463627?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1208671677432066/task/1216624996708289?focus=true
CC:

### Description
Introduces the first of three promo queue changes behind a disabled-by-default flag (`modalPromptCoordination`). This PR establishes shared cross-surface admission and integrates the launch-modal side; lifecycle and NTP rendering follow in later PRs.

`modalPromptCoordination` OFF (same as `main`)

Screenshot 2026-07-29 at 17 30 32

`modalPromptCoordination` ON

Screenshot 2026-07-29 at 17 30 40

### Testing Steps
- Check for regressions to current behavior (`main`) with `modalPromptCoordination` flag off

### Impact
Low: the new behavior is disabled by default.

#### What could go wrong?
Existing modal prompts could be incorrectly blocked → mitigated by the legacy feature-off path and focused state-machine tests.

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Medium Risk**
> Changes how launch modals interact with NTP promos and feature-flag transitions; flag defaults off limits production exposure, but coordination bugs could suppress modals or promos when enabled.
> 
> **Overview**
> Adds **disabled-by-default** `promoPresentationCoordination` (remote `PromoQueueSubfeature.iOSPromoPresentationCoordination`) and replaces the launch-modal coordination stack with a promo-queue model that can arbitrate **launch modals** against **visible NTP promos** when the flag is on.
> 
[...]

PR #6194, opened 2026-08-05
iOS Promo Queue: Add lifecycle-safe scheduling and NTP render gating

bkunat:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1215172677539195/task/1217045769951384
Tech Design URL: https://app.asana.com/1/137249556945/project/1208671677432066/task/1216624996708289?focus=true
CC:

### Description

PR 2 of 3 in the Promo Queue stack. It builds on [PR 6087](https://github.com/duckduckgo/apple-browsers/pull/6087), and [PR 6291](https://github.com/duckduckgo/apple-browsers/pull/6291) is stacked on this PR.

This extends Promo Queue to coordinate lifecycle-safe launch modals and New Tab Page cards supplied by the Remote Messaging Framework (RMF). When enabled, one modal attempt or one logical RMF session can own the app-wide promo slot. This prevents modal/RMF overlap and duplicate RMF cards across NTP hosts.

The feature remains default-disabled through `promoPresentationCoordination`. Existing behavior remains unchanged while the feature is disabled.

Screenshot 2026-08-13 at 00 50 44

### Testing Steps

- Check for regressions to current behavior (`main`) with the `promoPresentationCoordination` flag off.

### Impact

Low—the feature flag defaults to disabled.

#### What could go wrong?

When enabled, a promo could remain blocked or overlap another promo → covered by lifecycle, admission, retry, and integration tests.

### Quality Considerations

Existing behavior remains unchanged while the feature flag is disabled.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Large launch and NTP promo coordination change, but default-disabled flag and extensive tests limit production exposure; mis-coordination when enabled could block or overlap promos.
> 
> **Overview**
> When **`promoPresentationCoordination`** is on (still default-off), promo behavior is fixed at startup as **`PromoCoordinationMode`** instead of mid-session flag transitions. **One app-wide slot** is shared between launch modals and a **single logical NTP remote-message session**, replacing per-surface visible-promo leases.
> 
> **`PromoCoordinationService`** now drives modal presentation from app readiness (foreground token, active/deferred presenter), reconciles RMF through an **idle → owned → draining** state machine, and registers weak NTP **renderer** generations with appearance and removal terminals.
> 
> **`ModalPromptCoordinationManager`** adds app lifecycle hooks, cancellable scheduling, UIKit attachment verification, **pending** prompt retries after failed preflight, and **`shouldSuppressOtherSessionPromos`** (replacing same-session “recent modal” bookkeeping). Providers gain **`isModalPromptStillValidForPresentation`**; What’s New can supply replacements.
> 
> **NTP** renders RMF only when centrally authorized in coordinated mode; **`MainViewController`** toggles promo surface active/visible around suggestion trays, popovers, and onboarding so eligibility matches what the user actually sees. Duck.ai sync promos use the same suppression signal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42c7b42b6cc2db99b65afd467bc3d6e49f7a9261. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/ModalPromptCoordination/ModalPromptCoordinationManager.swift
```diff
@@ -61,120 +65,237 @@ enum ModalPromptAttemptPhase: Equatable {
...
+              case .verifying(let presentation) = activeAttempt.state else {
             return
         }
 
-        guard case .acquired(let lease) = promoQueueLeaseArbiter.acquireModalLease() else {
+        scheduleAttachmentVerification(
+            attemptIdentity: activeAttempt.attempt.lease.attemptIdentity,
+            exactRoot: presentation.exactRoot
+        )
+    }
```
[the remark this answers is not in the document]
bkunat: Correct observation, intended behavior. Pending work is deliberately checkpoint-based: the tech design specifies that a failed pre-presentation attempt is retained "as in-memory pending work for a later eligible foreground," and the retry checkpoint is the next foreground's modal evaluation, where the pending item is consumed before cooldown/provider evaluation

iOS/DuckDuckGo/ModalPromptCoordination/ModalPromptCoordinationManager.swift
```diff
@@ -276,39 +448,55 @@ private extension ModalPromptCoordinationManager {
...
+            case .ready(let route):
+                self.beginPresentation(committedAttempt, using: route)
+            case .alreadyAttached:
+                self.adoptAttachedRoot(committedAttempt)
+            case .recoverableFailure:
+                self.finishPreVisibleAttempt(
+                    committedAttempt,
+                    retainAsPending: true,
+                    notifyRelease: true
+                )
```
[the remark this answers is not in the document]
bkunat: This chain is specified, not accidental. The implementation plan requires the registry-wide NTP retry when committed work releases before anything is visible ("trigger the same guarded registry-wide retry"), and the tech design states that pending work "does not hold the RMF arbitration slot, but it continues to suppress the AI Chat sync promo until it is retried or discarded."

iOS/DuckDuckGo/ModalPromptCoordination/Providers/ModalPromptProvider.swift
```diff
@@ -40,7 +40,7 @@ struct ModalPromptConfiguration {
 /// A type that can provide a prompt to be presented to the centralised modal prompts coordination system.
 /// Providers act as lightweight adapters between feature-specific modal prompt logic and the centralised `ModalPromptCoordinationManager`.
 @MainActor
-protocol ModalPromptProvider {
+protocol ModalPromptProvider: AnyObject {
```
alessandroboron, line 43: A few thoughts on this:

1. We've gone from a simple contract "give me a modal, I'll tell you when it was presented” to one where a provider author now has to understand three subsystem concepts: prepared, retained, and what an invalid prompt is. It would be nice if these concepts were transparent to the promo author.
    - Could we add extra doc comments so an author knows when they can ignore them?
    - Is it possible to fold `isPreparedModalPromptStillValid` and `isRetainedPreparedModalPromptStillValid` into one, e.g. isModalPromptStillValidForPresentation(_:)?

2. The 0.1s is a UX beat so we don't present the instant the app becomes active. Right now it sits between acquiring the lease and presenting, which creates the "prepared but not yet presented” window and I believe a lot of the logic exists to manage this window.
    - If we move this delay to the outset, e.g. a quiescent gate before anyone acquires a lease ( delay -> acquire the lease -> select -> present synchronously) would it simplify the system?

iOS/LocalPackages/SetDefaultBrowser/Sources/SetDefaultBrowserCore/PromptDecider/DefaultBrowserPromptTypeDecider.swift
```diff
@@ -154,6 +169,25 @@ package final class DefaultBrowserPromptTypeDecider: DefaultBrowserPromptTypeDec
...
     }
 
+    package func isPreparedPromptStillValid() -> Bool {
+        guard let storedInfo = defaultBrowserManager.storedDefaultBrowserInfo() else {
+            return true
+        }
+        return storedInfo.isEligibleToShowDefaultBrowserPrompt()
+    }
+
+    package func isRetainedPreparedPromptStillValid() -> Bool {
```
alessandroboron, line 179: How often can the retained path actually run, and could it use the cached status like the prepared path? If a default-browser prompt goes decision → retain → revalidate, are we burning two checks for one prompt? We only have up to 4 per year

iOS/LocalPackages/SetDefaultBrowser/Sources/SetDefaultBrowserCore/PromptDecider/DefaultBrowserPromptTypeDecider.swift
```diff
@@ -58,6 +58,21 @@ package extension DefaultBrowserPromptType {
...
     func promptType() -> DefaultBrowserPromptType?
+    /// Revalidates prepared work from cached browser status without consuming selection state.
+    func isPreparedPromptStillValid() -> Bool
+    /// Revalidates retained work using a fresh browser-status check before retrying presentation.
+    /// A failed check falls back to the stored status rather than invalidating the retained work.
+    func isRetainedPreparedPromptStillValid() -> Bool
+}
+
+package extension DefaultBrowserPromptTypeDeciding {
+    func isPreparedPromptStillValid() -> Bool {
```
alessandroboron, line 69: Probably this `isPreparedPromptStillValid` extension is not needed since we have implemented it

iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
```diff
@@ -877,6 +877,7 @@ extension MainViewController {
         )
 
         if renderState.isContentVisible {
+            newTabPageViewController?.setPromoSurfaceActive(false)
```
alessandroboron, line 880: I may be wrong but I noticed we record every time we cover/uncover the new tab page? Does it mean that any future feature that covers the NTP has to know to call setPromoSurfaceActive? What happens if someone forgets to implement those, coordination silently breaks?

iOS/DuckDuckGo/ModalPromptCoordination/Providers/ModalPromptProvider.swift
bkunat: 1. I agree that immediate/retained provider vocabulary was too broad. After the last update the provider API now has one presentation-validity hook. Rare invalid replacement is an optional capability.

2. Moving the delay to the outset would require another barrier to prevent RMF win during standard-launch modal coordination. I ended up keeping the 0.1s beat after global ownership.  The delay is now properly documented tho

iOS/LocalPackages/SetDefaultBrowser/Sources/SetDefaultBrowserCore/PromptDecider/DefaultBrowserPromptTypeDecider.swift
bkunat: You're right. selection plus retained retries could consume multiple fresh system checks. 

I've updated it so that selection remains the single fresh check; later presentation validation uses stored status. 

The trade-off now is now is that cache can be stale after an external default-browser change but given the 4x/y limit there's no other way around it.

iOS/DuckDuckGo/UnifiedToggleInput/MainViewController+UnifiedToggleInput.swift
bkunat: we do update whether the NTP promo surface is currently visible. 
we can detect when the NTP is attached to a window, but not when another view covers it.

A future feature that covers an attached NTP must also use this handoff. If it bypasses it, coordination could incorrectly treat a hidden NTP promo as active and block another promo. 

I could not come up with a way to enforce this for arbitrary overlays at compile time. 
Documenting it as an integration requirement is the second best option.

iOS/DuckDuckGo/NewTabPageViewController.swift
```diff
@@ -795,7 +888,9 @@ extension NewTabPageViewController {
         }
     }
 
-    private func dismissHostingController(didFinishNTPOnboarding: Bool, updateUnifiedInputContentOverlaySuppression: Bool = true) {
+    private func dismissHostingController(didFinishNTPOnboarding: Bool,
```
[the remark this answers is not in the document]
bkunat: False positive. On the asynchronous UTI path, dismissHostingController removes overlay suppression before clearing coverage, which synchronously deactivates the NTP promo surface.

iOS/DuckDuckGo/ModalPromptCoordination/Factory/PromoCoordinationFactory.swift
```diff
@@ -35,6 +36,9 @@ enum PromoCoordinationFactory {
     ) -> PromoCoordinationService {
 
         let isIPad = DevicePlatform.isIpad
+        let promoCoordinationMode = PromoCoordinationMode(
```
bkunat, line 39: We only read this flag once when we build the app's promo coordination graph. Keeping the mode fixed means modal prompts and RMFs cannot end up using different paths after a privacy-config refresh. When it is off, we keep the old direct-render flow and do not register renderers or arbitrate.

iOS/DuckDuckGo/NewTabPageMessagesModel.swift
```diff
@@ -78,95 +242,379 @@ final class NewTabPageMessagesModel: ObservableObject {
...
+        if !isRemoteMessageRendererAttachedToWindow {
+            remoteMessageHostDidDetach()
+            return
+        }
+
+        switch removalPath {
+        case .synchronousSourceClear:
+            removeWithoutAnimation(presentation: presentation, removalID: removalID)
+        case .automatic:
+            if #available(iOS 17, *) {
```
bkunat, line 536: This is the SwiftUI split I settled on. iOS 17 gives us a completion callback for removal, so that path stays animated. On iOS 15/16, coordinated mode clears the source synchronously with animations disabled so we have a reliable point to hand off. With the flag off, the old UI behavior is unchanged.

alessandroboron (commented):
@bkunat 

The coordination needs almost every UI area that can sit on top of the NTP to report back its state. We call `setPromoSurfaceActive` / `setPromoSurfaceRenderable` / `setPromoSurfaceVisible` / `setPromoSurfaceCovered` from a bunch of spots such as the `UnifiedSuggestionsHost `, `SuggestionTrayViewController`, and even the onboarding path where we flip `setPromoSurfaceVisible` / `setPromoSurfaceCovered` as the Dax dialogs show and hide.

NewTabPagePromoSurfaceHandoff comment states:
> "Any future host that covers an attached NTP must use this handoff; arbitrary overlays are not detected automatically.”

That's the part that mainly concerns me. If someone later adds something that covers the NTP and forgets to call into this, coordination breaks silently. It's spread over a lot of files and it's easy to get wrong.

One idea that might remove most of it. `HomePageConfiguration` is already a single shared instance, and it's injected into all three NTP renderers. `remoteMessageToShow` is the one place they all go through to get the message. If the lease check lived there, all three would be coordinated at a point they already share, and the individual surfaces wouldn't need to push their visibility at all. Would this help simplify the system?

We'd still need to release when the card actually leaves the screen, but the card already reports that (its onDisappear), so it's one existing signal instead of tracking spread everywhere. It will probably be a less precise system but simpler and more maintainable.

This touches a few entry points and the RMF rendering, so it might be worth having a second reviewer take a look too.

bkunat:
@alessandroboron I fully agree. I investigated it and it looked like it could work on paper, but it fell apart once I got into the details.

`HomePageConfiguration` is shared, but `remoteMessageToShow` only selects the candidate. Each renderer copies that candidate into its own state, including cached or hidden renderers, so putting the lease there still wouldn’t tell us which physical renderer should show it. We’d have to move renderer identity, attachment, handoff and removal state into `HomePageConfiguration`. this would effectively recreate the coordinator one level up.

onDisappear tracks a particular SwiftUI mount, not whether the RMF is actually visible. We already have paths that keep the NTP mounted while hiding it with alpha or covering it with another view, and those changes don’t reliably trigger the card’s onDisappear. The current system handles them by explicitly making the surface ineligible and waiting for an exact removal signal, although that manual integration is also the weakness you pointed out.

Today I spent most of the day looking for a solution to the issue you flagged that wouldn’t require rewriting the whole system or making the scope of the existing changes even wider.

I also tried using one shared controller to track which NTP is currently visible. Anything that temporarily hides or covers it (Dax dialog or a transition) would register with the controller before appearing and unregister after disappearing, preventing promos in the meantime. This reduced the scattered state calls, but every new host or overlay still had to remember to integrate with it correctly. So it mostly put the same manual responsibility behind a cleaner API rather than fixing the underlying issue. 

tl;dr it still felt like patchwork.

The clean end state would be one central UI component that always knows which version of the NTP is currently visible (the standard NTP, suggestion tray or unified-input view) and whether something like a Dax or onboarding dialog is covering it. The promo coordinator would use that information to decide when and where a promo can appear. Each NTP would only provide a placeholder where the selected promo is rendered, and adding a new promo would mean registering its content and rules instead of adding visibility calls across the UI.

I don’t think we should implement that as part of these PRs tho. It would widen the scope considerably. My suggestion is to merge the current version, then prepare a refactor plan next week and implement it incrementally in atomic PRs. I could work on that outside of my TP since my next few TPs will already be focused around promos.

The current setup isn’t ideal for adding or changing NTP promo surfaces, but right now I don’t see a smaller solution to the issue you found that doesn’t end up looking like patchwork.

bkunat:
Closing as we decided to lean more on the direction Android took

PR #6213, opened 2026-08-06
Add resting outline to the Duck.ai model picker pill

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217146848886708

### Description

This PR matches model picker in the address bar's Duck.ai input with other platforms by adding a border in its resting state

### Testing Steps
1. Focus the address bar and switch it to Duck.ai mode → the model pill has a visible outline at rest, matching the same pill on the New Tab Page input below it.
2. Hover the pill, then press and hold → the fill tints as before, outline stays visible in both states.
4. Switch the app to dark mode, then open a Fire Window → the outline follows the appearance in both.
5. Click the pill and pick a different model → the fill clears as the menu disappears, with no lag and nothing left
7. Click the image-generation and web-search buttons → no stuck fill, and their active-pill styling is unchanged.

### Impact

Low — visual only. No behaviour change beyond the fill no longer sticking.

#### What could go wrong?

The fill reset hooks `NSMenu.didEndTrackingNotification`, which every menu in the app posts → filtered to the button's own root menu (a submenu has a `supermenu`), and the pre-existing reset on the action's return is kept as a fallback for actions that pop no menu at all.

### Quality Considerations

The outline colour is written to a `CGColor`, which freezes whichever appearance is current when it's assigned. It's resolved against the view's own `effectiveAppearance` rather than the app's, so a Fire Window's forced dark override gets the right colour — the pre-existing fill colours on this button resolve against the app appearance and have that latent bug, left alone as out of scope here.

---

> [!NOTE]
> **Low Risk**
> Visual and local AppKit interaction only in AI Chat omnibar controls; menu notification handling is scoped to root menus with a synchronous fallback reset.
> 
> **Overview**
> Adds a **persistent resting outline** on the address-bar Duck.ai **model picker pill** via a dedicated `borderLayer` (design-system `.lines`), resolved against the view’s `effectiveAppearance` so Fire Window dark overrides stay correct.
> 
> **Interaction fix:** model picker and omnibar tool buttons that open menus now route clicks through `sendMenuOpeningAction`, which clears pressed/hover fill when root menu tracking ends (`NSMenu.didEndTrackingNotification`) and re-syncs hover from the pointer. That stops the press fill sticking after model/menu dismissal; `refreshHoverState` also runs when tracking areas are recreated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e28d8fbe112d094ccc37a6603834b9fee91bfeb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatOmnibarToolButton.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatOmnibarToolButton.swift
@@ -34,6 +34,42 @@ protocol FocusRingControlling: AnyObject {
     func takeKeyboardFocus()
 
     var isFocusRingSuppressed: Bool { get set }
+
+    /// Drops the press fill and re-derives hover from where the pointer actually is.
+    func resetTransientFillState()
+}
+
+extension FocusRingControlling {
+
+    /// Sends a button action that pops a menu, clearing the button's transient fill in step with
+    /// the menu disappearing.
+    ///
+    /// Two problems make the naive "reset after `sendAction` returns" wrong. Menu tracking is
+    /// modal, so no `mouseExited` reaches the button's tracking area while the menu is up and its
+    /// hover state is stale by the time it closes. And `popUp` only returns once AppKit has
+    /// flashed the selected item and animated the menu away, so resetting on its return trails
+    /// the menu's disappearance by a visible beat. `didEndTrackingNotification` lands when the
+    /// user commits to an item, ahead of both animations.
+    func sendMenuOpeningAction(_ sendAction: () -> Void) {
+        let observer = NotificationCenter.default.addObserver(
+            forName: NSMenu.didEndTrackingNotification,
+            object: nil,
+            queue: nil
+        ) { [weak self] notification in
+            // Submenus post this too as they close under a still-open parent. Only the root menu
+            // going away means the button is done being "active" — a submenu has a `supermenu`.
+            guard (notification.object as? NSMenu)?.supermenu == nil else { return }
+            // `queue: nil` posts synchronously on the posting thread, and menu tracking runs on
+            // the main thread, so this is already main-actor in practice.
+            MainActor.assumeIsolated {
+                self?.resetTransientFillState()
+            }
+        }
+        sendAction()
+        NotificationCenter.default.removeObserver(observer)
+        // Belt and braces: covers an action that popped no menu at all, so posted no notification.
+        resetTransientFillState()
+    }
 }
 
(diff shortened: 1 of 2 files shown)


macOS/DuckDuckGo/AIChat/AIChatOmnibarToolButton.swift
```diff
@@ -34,6 +35,33 @@ protocol FocusRingControlling: AnyObject {
...
+            object: nil,
+            queue: nil
+        ) { [weak self] notification in
+            // Submenus post this too, under a still-open parent; only the root menu means done.
+            guard (notification.object as? NSMenu)?.supermenu == nil else { return }
+            // `queue: nil` posts synchronously, and menu tracking runs on the main thread.
+            MainActor.assumeMainThread {
+                self?.resetTransientFillState()
+            }
+        }
```
jleandroperez, line 59: I'm not 100% sure we still need the Notification Observer, as we're invoking `resetTransientFillState` imperatively, right after the `sendAction` call on L60

jleandroperez (approved):
Just one minor note, other than that, looks great! nice work!

:shipit:

macOS/DuckDuckGo/AIChat/AIChatOmnibarToolButton.swift
tomasstrba: It's load-bearing, and it's the reason this isn't just the imperative call. `NSMenu.popUp` runs a modal tracking loop and doesn't return until AppKit has finished flashing the selected item and animating the menu away, so the reset on L60-63 lands roughly half a second after the menu visually disappears — you see the fill sitting there after the menu is gone. That lag was the original report; `didEndTrackingNotification` fires when the user commits to an item, ahead of both animations, which is what makes it look instant. The trailing call is the fallback for an action that pops no menu at all (image-gen and web-search just toggle a mode), so neither path is redundant.

Happy to go simpler if you'd prefer: resetting *before* `sendAction` instead of after would drop the observer entirely and is unconditionally immediate, at the cost of the button no longer showing its pressed fill while its menu is open. Your call — I've left this as-is for now rather than pick for you.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tomasstrba: Jorge, as Claude reported, I had a simpler solution before, but there was a visual lag with the background after closing the menu. This approach helped, but technically it's more complicated, I agree

PR #6215, opened 2026-08-06
EventHub for Apple

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203936086921904/task/1212993781157341?focus=true
CC:

### Description
This is the feature branch containing merge of previously separate stacked PRs.

Each part below was reviewed and merged individually into this branch, so the diff here has already been through review in smaller pieces:

| Part | PR | Title |
|------|----|-------|
| 1 | #5959 | Add EventHub shared package (core implementation + tests) |
| 1b | #6190 | Add diagnostics — logging at every fail-safe boundary, plus two error events |
| 2 | #6054 | macOS EventHub integration |
| 3 | #6128 | iOS EventHub integration |
| 4 | #6164 | Retire the old detection pixel path and run the package tests in CI |
| 5 | #6221 | Post-review fixes and final polish |

### Testing Steps
See #6054 and #6128.

### Quality Considerations

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches analytics consent gating, persisted aggregation state, and replaces the prior web-events pixel path—mis-wiring or consent-strip bugs could affect telemetry volume or privacy, though the design fails closed on consent and gates counting on feature enablement.
> 
> **Overview**
> Introduces a new **EventHub** Swift package and wires it into the workspace, replacing the BSK **`WebEventsSubfeature`** path with a config-driven telemetry hub that fires pixels from remote `eventHub` settings.
> 
> **Remote config & privacy:** Adds **`PrivacyFeature.eventHub`** so the feature can be toggled and supplied with a `telemetry` map (period vs immediate triggers, counter buckets, data parameters).
> 
> **Runtime:** **`EventHub`** aggregates web and native events on a serial queue—per-tab dedup for web events (cleared on URL change / tab close, survives period rollovers), period windows with frozen config snapshots, write-behind persistence, foreground-gated period starts, and a single scheduler for period end + flush deadlines. **`EventHubSettings`** strips consent-gated telemetry keys (fail-closed if stripping is impossible). **`WebEventsHandler`** forwards CSS `webEvent` messages into the hub with caller-provided tab IDs.
> 
> **Pixels:** Firing goes through an injected **`EventHubPixelFiring`** seam (platform suffixes stay in app wiring). Debug **`EventHubDebugEvent`** cases cover persistence and consent-strip failures.
> 
> The diff is largely the package plus tests (functional, integration, unit); macOS/iOS app integration and retiring the old detection pixel path are described in the PR as separate merged parts and may not all appear in this snapshot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82a865d6fc4837c62d209773b1bec09cc9a98a3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/EventHub/Sources/EventHub/Core/EventHub.swift
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    private func flushDirtyLocked() {
+        guard !dirtyNames.isEmpty else { return }
+        for name in dirtyNames {
+            if let telemetry = telemetries[name] {
+                store.savePixelState(telemetry.snapshot())
+            }
+        }
+        dirtyNames.removeAll()
```
bwaresiak, line 399: Tho thoughts here:
1/ How often we will have more than one „dirty” here? Potentially a bit when going to background right? If so, I’d create `savePixelStates` taking array of the changes instead of re-serializing one by one.
2/ If storing fails, error is not propagated - thus we clean dirty names even if these were not persisted - I don’t think that fail should happen often, so question is hoe recoverable that would be on next try (that would be like next app cycle right? if app survives in memory that long...)

miasma13: Both done in #6221 — https://github.com/duckduckgo/apple-browsers/commit/abc5cd9af90c501e97c448004e3a901d96f0a151

**1/** More often than just backgrounding, as it turns out: `startNewPeriodLocked` marks dirty on every period start, and both `applyConfigLocked` and `checkPixelsLocked` start *all* configured periods in a loop before flushing — so it's one per configured pixel on every cold start.

The cost was also worse than one serialisation each: every pixel's state lives under a single composite key, so each save did a full decode **and** encode of the whole map plus a write. N dirty names meant N of those. `savePixelStates` now takes the batch and does one read-modify-write.

**2/** Agreed, fixed. `savePixelStates` reports whether the store accepted the write and `flushDirtyLocked` only clears the names on success, so the next flush retries them. No new machinery needed — `rearmSchedulerLocked` already arms the flush timer for as long as anything is pending.

On how recoverable it was: not at all, across an app cycle. Within a session a name was only re-dirtied if another event happened to arrive for that pixel; across a restart the in-memory counts are gone too, so there was nothing left to retry *from*. The realistic trigger is iOS refusing writes while the device is locked — the app suspends right after, so the retry now lands on the next foreground, i.e. roughly one per app session rather than a tight loop.

One deliberate asymmetry worth flagging: a state that cannot be **serialised** is skipped and does not fail the batch. That failure is permanent, so counting it as "retry" would spin forever. Only a rejected store write is retried.

Side effect of 1/ that's relevant to 2/: a failed flush now reports `pixelStatePersistenceFailed(.write)` once instead of once per pending pixel.

bwaresiak (approved):

PR #6219, opened 2026-08-06
iOS: Support message edit mode in the contextual sheet

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217249867328887

### Description
Brings the message edit-mode experience to the contextual Duck.ai sheet, matching the full-tab behaviour. While editing a previously sent message the sheet swaps its header for the minimal "Edit Message" header (with a ✕ that cancels the edit), and the chat content fades out so the edit input stands alone. Cancelling the edit or closing the sheet exits edit mode, so its chrome never lingers when the sheet reopens. The edit header view is shared with the full-tab surface.

### Testing Steps
Prerequisites:
enable Settings → Debug → Feature flags → nativeAIPromptEditing (default off); open a Duck.ai full-tab chat;
set custom Duck.ai url Settings → Debug → AI Chat → Set Custom AI Chat URL → https://use-serp-dev-testing1.duck.ai/

1. Go to Duck.ai to login through Duo
2. Visit a site and open the contextual mode
3. Perform a chat (without attaching the page content
4. Tap on the edit button below the latest sent message
5. This should trigger edit more, test submission, removing attachments and cancelling
6. Disable nativeAIPromptEditing, tap on the edit button on the FE, old inline experience should be visible

### Impact
Low — scoped to the contextual Duck.ai sheet. Edit mode is entered via a `#if DEBUG` command (and, later, the FE bridge), so it isn't user-facing yet.

---

> [!NOTE]
> **Low Risk**
> Changes are UI and edit-mode wiring in the contextual sheet and shared header, with no auth or data-model changes; behavior is gated behind native prompt editing.
> 
> **Overview**
> Brings **native message edit mode** to the contextual Duck.ai sheet so it matches the full-tab flow: when the unified toggle input enters edit, the sheet swaps its normal header for the shared **Edit Message** header (cancel ends edit), and the chat **content fades out** while editing.
> 
> `AIChatEditHeaderView` now supports **host embedding** (`preferredHeight: nil`): no fixed height, clear background, and no bottom separator so it fits the sheet chrome. `AIChatContextualUTIHost` forwards edit-mode changes via `onEditModeChange` and exposes `endEditMode()`; dismiss and header cancel both clear edit so state does not stick when the sheet reopens.
> 
> Smaller aligned fixes: `editPrompt` in `AIChatUserScript` returns **cancelled** if the script is deallocated mid-await; full-tab edit dimming also restores the **web view container background**; edit file attachments use the shared presenter; the replace-response disclaimer uses a smaller info glyph.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69aa79f8f04d9c01ce83adad4103273055cae9ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualModeFeature.swift
```diff
@@ -42,20 +42,27 @@ struct AIChatContextualModeFeature: AIChatContextualModeFeatureProviding {
...
     /// Whether Duck AI contextual chat mode is available.
     var isAvailable: Bool {
         featureFlagger.isFeatureOn(.contextualDuckAIMode)
             && isPageContextEnabled
-            && aiChatURLProvider().isStandaloneDuckAIURL
+            && isEligibleDuckAIURL
+    }
+
+    private var isEligibleDuckAIURL: Bool {
+        aiChatURLProvider().isStandaloneDuckAIURL || debugSettings.customURL != nil
```
aataraxiaa: FYI - this enabled feature availability with a custom URL, but didn’t allow conteuxtual debug URL to load. 

`AIChatContextualWebViewController:491` (nav policy) and `:323` (isSupportedURL) both gate on `matchesCustomURL`, which compares scheme + host + path — so any debug FE that redirects within itself (/ → /chat) fails them. 

So when testing the sheet spinned with `isPageReady` false. I worked around this locally:

```
 private var debugCustomHost: String? {
     debugSettings.customURL.flatMap { URL(string: $0)?.host }
 }

 // decidePolicyFor(navigationAction:) — before the existing check
 if let debugHost = debugCustomHost, url.host == debugHost { return .allow }

 // chatURLForLoading
 isSupportedURL: url.isDuckAIURL || debugSettings.matchesCustomURL(url)
    || (debugCustomHost != nil && url.host == debugCustomHost)
```

So unsure if we want this change on it’s own as it doesn’t fully enable debug URL testing. Not a blocker here as this is really an improvement on debug testing.

aataraxiaa (approved):
Tested and looks good @SabrinaTardio. Left one comment re: debug code.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualModeFeature.swift
SabrinaTardio: I was able to load the debug URL in the contextual page. When testing, it redirects to the full page the first time and then working as expected after that.
Anyway I’ve removed it for now.

PR #6223, opened 2026-08-06
Ship Review - Polish tab attachments for macOS

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1217202913297218?focus=true
Tech Design URL:
CC:

### Description

Reworks Duck.ai tab attachments. A card now follows its tab's navigation and disappears when the tab closes; the Add Tabs dialog gets keyboard navigation and a confirm button that reads Add or Update depending on what was attached when it opened; Recent Tabs rows toggle off; and a PDF dropped on the prompt attaches instead of pasting its path. Copy moves from "Attach" to "Add".

### Testing Steps
1. Attach tabs from the paperclip menu and from Add Tabs — Recent Tabs rows show a checkmark and toggle off.
2. Navigate an attached tab, then close it — the card follows the new page, then disappears.
3. Open Add Tabs with tabs attached — the button reads Update; with none attached it reads Add and is disabled until you pick one.
4. Drag a PDF onto the prompt — it attaches as a file card.

### Impact

Medium. Changes which page reaches Duck.ai when tabs are attached.

#### What could go wrong?

- A card follows a navigation the user didn't drive — a redirect or a script sending the tab elsewhere.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes what page content Duck.ai receives when tabs navigate, close, or are picked from stale lists—core prompt context behavior, though guarded by policy, instance IDs, and tests.
> 
> **Overview**
> **Tab attachments now stay aligned with live tabs.** New `AIChatAttachedTabsTracker` and `AIChatAttachedTabNavigationPolicy` watch attached tabs across windows: cards **follow navigation** (title/favicon/URL), **drop** when the tab closes or lands on an unattachable page, and submit only ships page context for attachments still present with matching `instanceID` (new on `AIChatTabAttachment`).
> 
> **Add Tabs modal and attach menu** get keyboard navigation (arrows, space), a selection cap counter, **Add** vs **Update** confirm, and stale-pick handling via `togglePickedTabAttachment` with user-visible error. Recent Tabs rows toggle detach and show checkmarks; copy shifts from “Attach” to “Add” with new localizations.
> 
> **Drag-and-drop** on the omnibar accepts PDFs (and images) through the same validation as the file picker instead of pasting paths. **Tab picker origin** resolution improves for detached Duck.ai windows. Carousel/tab cards refresh in place on navigation; attachment views use `effectiveAppearance` for theming.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cae9fd94ec894e23147585a65d076ffcfce6864d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/AIChat/AIChatAttachTabsModal.swift
```diff
@@ -168,3 +248,20 @@ struct AIChatAttachTabsModal: ModalView {
...
+
+    let onWindow: (NSWindow?) -> Void
+
+    func makeNSView(context: Context) -> NSView {
+        let view = NSView()
+        DispatchQueue.main.async { onWindow(view.window) }
+        return view
+    }
+
+    func updateNSView(_ nsView: NSView, context: Context) {
```
tomasstrba, line 296: This reports the window on every update, even when it hasn't changed. Writing the `@State` invalidates the view, `body` re-runs, the representable gets a fresh (non-equatable) closure, `updateNSView` fires again — a loop that keeps re-rendering the sheet for as long as it's open. SwiftUI's `State` setter doesn't equality-gate, so nothing breaks the cycle.

Only call back when the window actually changed:

```swift
.background(WindowReader { if sheetWindow !== $0 { sheetWindow = $0 } })
```

or guard inside `updateNSView` against the last reported window.

macOS/DuckDuckGo/AIChat/AIChatAttachTabsModal.swift
```diff
@@ -114,21 +130,84 @@ struct AIChatAttachTabsModal: ModalView {
...
+    /// the field consumes arrow keys itself.
+    private func startKeyMonitor() {
+        guard keyMonitor == nil else { return }
+        keyMonitor = NSEvent.addLocalMonitorForEvents(matching: .keyDown) { event in
+            // The monitor is app-wide; only this sheet's keys are this view's business.
+            guard let sheetWindow, event.window === sheetWindow else { return event }
+            switch Int(event.keyCode) {
+            case kVK_DownArrow: return moveHighlight(by: 1) ? nil : event
+            case kVK_UpArrow: return moveHighlight(by: -1) ? nil : event
+            case kVK_Space where highlightedId != nil: return toggleHighlighted() ? nil : event
```
tomasstrba, line 179: `highlightedId` is never cleared once set, so after the first arrow press space can no longer be typed into the search field — which still holds focus. Multi-word filtering is dead for the rest of the session.

The doc comment on `highlightedId` says "Nil until the first arrow press, so typing a space still types", which is only true before that first press. Clearing it on `searchQuery` change would keep both behaviours.

macOS/DuckDuckGo/AIChat/AIChatAttachTabsModal.swift
```diff
@@ -114,21 +130,84 @@ struct AIChatAttachTabsModal: ModalView {
...
-                    Text(UserText.aiChatAttachTabsModalAttachButton)
+                    Text(opensWithAttachments ? UserText.aiChatAttachTabsModalUpdateButton : UserText.aiChatAttachTabsModalAttachButton)
                         .frame(maxWidth: .infinity)
                         .frame(height: 28)
                 }
-                .buttonStyle(DefaultActionButtonStyle(enabled: true, topPadding: 0, bottomPadding: 0, pillShape: true))
+                // Nothing selected means nothing to add or update, whether it started that way or
+                // ended that way; cancelling is all that's left.
+                .buttonStyle(DefaultActionButtonStyle(enabled: !selectedIds.isEmpty, topPadding: 0, bottomPadding: 0, pillShape: true))
+                .disabled(selectedIds.isEmpty)
```
tomasstrba: Opening with tabs attached and then unchecking them all is a real edit — remove everything — but Update is disabled, so the only way out is Cancel, which keeps them. Dead end for what looks like a supported action.

The comment above argues "nothing selected means nothing to add or update", but with `opensWithAttachments == true` an empty selection is meaningful. `selectedIds.isEmpty && !opensWithAttachments` would cover it.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1498,8 +1497,9 @@ final class AIChatOmnibarContainerViewController: NSViewController {
     @objc private func recentTabClicked(_ sender: NSMenuItem) {
         guard let candidate = sender.representedObject as? AIChatTabAttachment else { return }
         didMutateDuringAttachMenuSession = true
-        omnibarController.toggleTabAttachment(candidate)
-        omnibarController.pixelHandler.fire(.tabChosen)
+        let wasAttached = omnibarController.activeTabAttachments.contains { $0.id == candidate.id }
+        guard omnibarController.togglePickedTabAttachment(candidate) else { return }
```
tomasstrba, line 1514: When the tab has moved on since the menu was built, this click does nothing at all — no card appears, no message, no explanation. The staleness check itself is right; the user just isn't told why their click was ignored.

Also, `didMutateDuringAttachMenuSession = true` on line 1499 is set before we know anything changed, so a rejected toggle still counts as a mutation for the session.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
```diff
@@ -1530,25 +1530,59 @@ final class AIChatOmnibarContainerViewController: NSViewController {
             omnibarController.removeTabAttachmentFromActiveTab(id: id)
             omnibarController.pixelHandler.fire(.tabAttachmentRemoved)
         }
-        for tab in diff.add {
-            omnibarController.toggleTabAttachment(tab)
+        for tab in diff.add where omnibarController.togglePickedTabAttachment(tab) {
```
tomasstrba, line 1549: Rejected tabs are skipped silently, so a confirm can half-apply: some cards appear, others never do, with nothing said about which or why. Worse than the menu case since the user explicitly selected a set and expects it applied as one.

Worth collecting the rejections and surfacing them through `lastAttachmentError`, the way the file/image picker path does.

tomasstrba (commented):
This works really well! I found one issue when testing the Ship Review build:

1. Attach a tab in omnibar using @ 
2. Click on attach button → Add Tabs
3. Unselect the tab
4. "Update" button is disabled, even though I edited the list and should be able to confirm

Plus adding a couple of code related comments below found by Claude:

macOS/DuckDuckGo/AIChat/AIChatAttachTabsModal.swift
jotaemepereira: Good catch — fixed in a8c815e879 with the guarded form you suggested, so the window is only reported when it actually changes.

macOS/DuckDuckGo/AIChat/AIChatAttachTabsModal.swift
jotaemepereira: Fixed in a8c815e879 by clearing the highlight on searchQuery change as you suggested, so space types again once you start filtering; the doc comment is updated to match.

macOS/DuckDuckGo/AIChat/AIChatAttachTabsModal.swift
jotaemepereira: Fixed in 25639a616c — confirm is now disabled only when the dialog opened with nothing attached, so unchecking everything still applies the removal.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
jotaemepereira: Fixed in a14e24ec60 — a rejected pick now explains itself through the same error line the file picker uses, and the mutation flag moved after the successful toggle in a8c815e879.

macOS/DuckDuckGo/AIChat/AIChatOmnibarContainerViewController.swift
jotaemepereira: Fixed in a14e24ec60 — the rejection is reported from togglePickedTabAttachment, so a half-applied confirm now says which picks were dropped and why; copy is still pending Ship Review.

macOS/DuckDuckGo/AIChat/AIChatAttachedTabsTracker.swift
```diff
@@ -0,0 +1,229 @@
...
+//
+
+import AppKit
+import Combine
+import Foundation
+
+/// Keeps each prompt's tab attachments in step with the tabs they point at: cards follow or drop on
+/// navigation per `AIChatAttachedTabNavigationPolicy`, and closed tabs lose their attachment.
+@MainActor
+final class AIChatAttachedTabsTracker {
```
tomasstrba, line 26: I like this encapsulation 👍

tomasstrba (approved):
LGTM! ✔️ 
Tested multiple scenarios with regular windows, Fire Windows, and edge cases like pinned tabs, etc..

PR #6225, opened 2026-08-06
Register account_info key after Sync login

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203822806345703/task/1217213092722099?focus=true
Tech Design URL:
CC:

### Description
The unified device list needs an `account_info` key to encrypt device names and types. This makes sure the account has one. 

When unified device-list writes are enabled, a successful Sync login now ensures the account has an `account_info` key wrapped for each available credential. Existing server keys are adopted safely, and registration failures remain non-blocking so they never prevent login.

Adds an iOS Sync debug action for manually ensuring the key and checking its wrapper count.

### Testing Steps
Test on iOS as the debug screen changes are iOS-only. The underlying key-registration functionality is shared by both platforms
1. Settings → Debug → Feature Flags, enable `syncCanWriteUnifiedDeviceList`.
2. Settings → Sync & Backup, set up a new sync account → sync enabled.
3. Settings → Debug → Sync, tap "Ensure account_info key" → alert reports "Ensured 1 wrapper".
4. Tap it again → same count, so no duplicate key.
5. Log out, log back in with the recovery code → login completes normally.
6. Tap "Ensure account_info key" → count unchanged, so login already registered it.

---

> [!NOTE]
> **Medium Risk**
> Touches post-login crypto key registration and server set-if-absent behavior; failures are non-blocking for login but incorrect handling could affect unified device list encryption.
> 
> **Overview**
> When **unified device list writes** are enabled (`canWriteUnifiedDeviceList`), a successful Sync **login** now ensures the account has an **`account_info`** protected key (for encrypting device name/type), merges those wrappers into the local protected-keys cache, and **does not block login** if registration fails.
> 
> **`ScopedAccessCredentialManager`** gains `ensureAccountInfoProtectedKeys`: reuse server keys when present, otherwise build wrappers via an injectable **`AccountInfoKeyFactory`** (`DefaultAccountInfoKeyFactory`) and register them with **`setKeysIfAbsent`** (batch upload per purpose). **`set-if-absent`** handling is updated to accept **200/201**, treat **409** as “already exists” by refetching keys for that purpose, and validate purpose/consistency instead of matching by wrapping identity.
> 
> Adds **`ensureAccountInfoKeyForDebug()`** on sync debugging support and an iOS **Sync debug** row to trigger it manually. Tests and mocks cover login caching, flag gating, and the new key-registration paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd46f098e63827758756a743487e80ef917d6010. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
@@ -293,6 +293,14 @@ public class DDGSync: DDGSyncing {
         initializeIfNeeded()
     }
 
+    public func ensureAccountInfoKeyForDebug() async throws -> Int {
+        guard let account = try dependencies.secureStore.account() else {
+            throw SyncError.accountNotFound
+        }
+
+        return try await dependencies.scopedAccess.ensureAccountInfoProtectedKeys(for: account).count
+    }
+
     public func prepareThirdPartyRecoveryCode(purpose: String) async throws -> String {
         guard let account else {
             throw SyncError.accountNotFound
--- a/SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
@@ -122,7 +122,8 @@ public class DDGSync: DDGSyncing {
         let result = try await dependencies.account.login(recoveryKey, deviceName: deviceName, deviceType: deviceType)
         try updateAccount(result.account)
         persistRecoveredThirdPartyScopedPasswordIfAvailable(from: result.accessCredentials, account: result.account)
-        updateProtectedKeysCache(with: result.keys)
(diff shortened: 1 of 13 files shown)


pballart (approved):
Tested it and works fine. The code also looks good but there is one edge case I want to raise although it might be a false positive:

> If an account's account_info key is created before its third-party credential exists, it's stored with only the ddg wrapper — and because both the login-time check and setKeysIfAbsent treat any existing wrapper as complete, the third-party credential created later never gets its own account_info wrapper and can't decrypt unified-device metadata.

amddg44:
I intentionally kept that out of scope here as this was already pretty sizeable; this follow-up [PR](https://github.com/duckduckgo/apple-browsers/pull/6268) will handle wrapper reconciliation when credentials change, including adding a third-party wrapper to an existing `account_info` key without replacing it

PR #6283, opened 2026-08-11
iOS: Duck.ai address bar menu and floating contextual input

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215757651854873
Tech Design URL:
CC:

### Description

Starting a contextual Duck.ai interaction from the address bar now offers a choice, and asking about the page no longer opens a sheet to do it.

The Duck.ai button in the address bar shows a menu with **New Chat** and **Ask About Page**. New Chat opens Duck.ai empty, so people who use it as a general assistant keep a one-tap path with no page attached. Ask About Page opens a light surface floating over the page — suggested prompts and the input, with the page already attached — leaving the page itself readable and scrollable underneath. Submitting a prompt opens the chat.

iPhone only, behind `aiChatContextualFloatingInput`, internal by default. iPad keeps today's sheet whether the flag is on or off, and so does iPhone with the flag off.

### Testing Steps

With the feature flag on (internal builds have it on by default):

1. Open any web page → tap the Duck.ai button in the address bar → menu appears with New Chat and Ask About Page
2. Tap **New Chat** → Duck.ai opens with no page attached
3. Back on the page, tap Duck.ai → **Ask About Page** → the input rises with the keyboard, page dimmed but still visible, page attached in the input, suggestions appear once the keyboard settles
4. Scroll the page → suggestions collapse into a deck; stop scrolling → they deal back out after half a second
5. Tap a suggestion → the chat opens full height showing your prompt and the response — no half-height step, no pre-submit pills
6. Reopen the surface and dismiss it four ways, each of which should feel the same: tap the page, swipe the input down, tap the address bar, VoiceOver escape
7. Tap a **text field** on the page while the surface is up → the surface leaves, the keyboard stays put for the page's field
8. Tap a **link** while the surface is up → the surface leaves and the link opens on that same tap
9. Long-press the page to select text → the surface leaves with the keyboard
10. Remove the page attachment with its ×, then tap **Ask About Page** in the input → the page attaches again
11. Submit a prompt, dismiss the sheet, tap Duck.ai → the sheet reopens on the existing chat rather than the menu

With the flag off, and on iPad with it on: the Duck.ai button opens the contextual sheet exactly as before.

### Impact

**Medium** — a new entry point to Duck.ai on iPhone, behind an internal-only flag. Nothing changes for anyone until the flag is enabled.

#### What could go wrong?

- The surface could be left on screen with no keyboard under it → it dismisses whenever a keyboard it was sitting above goes away, including a long press starting a text selection; covered by tests
- With a hardware keyboard attached no software keyboard ever appears → the dismissal rule waits for a keyboard to have appeared before reacting to one leaving, so the surface is not torn down on sight; covered by tests
- A dismissal finishing after the next surface has mounted the shared input could take that input away → removal is scoped to the parent that owns it; covered by a test
- The address bar button could offer a menu it has not attached, doing nothing when tapped → every runtime input to that decision now re-attaches it, including session resets that no navigation or surface change would notice

### Quality Considerations

- The page stays interactive throughout: touches pass through everywhere the surface does not claim them, so scrolling and links keep working while it is up
- A held finger is not a tap — resting mid-scroll no longer dismisses the surface
- Accessibility: `accessibilityPerformEscape` dismisses. The surface deliberately does not set `accessibilityViewIsModal`, since the page is meant to stay reachable; worth a look from someone who tests with VoiceOver
- The suggestion pills use `UIGlassEffect` with content in the effect view's `contentView`, so the material adapts to the page rather than being held light by a tint. One known artifact: glass renders a neutral fill on its first frame, so pills appearing over a dark page flash light briefly before settling

---

> [!NOTE]
> **Medium Risk**
> New address-bar entry and complex floating-input lifecycle (keyboard, gestures, shared UTI host) on a core chrome path, but gated to iPhone behind an internal feature flag.
> 
> **Overview**
> **Adds an address-bar Duck.ai menu and a floating pre-submit input** on iPhone so *Ask About Page* no longer opens the contextual sheet first.
> 
> When the flag is on and there is no chat to reopen, the Duck.ai button offers **New Chat** (empty Duck.ai) or **Ask About Page** (floating input over the live page with context attached). Submitting a prompt or suggestion **promotes** that surface into the existing sheet, reusing the same UTI host so draft state survives. Dismissing without submit (page tap, drag, address-bar focus, VoiceOver escape, keyboard loss) tears the floating surface down cleanly.
> 
> Also restyles context/quick-action chips for the floating design (glass pills, re-attach *Ask About Page* chip after removal) and wires new analytics for the menu and floating-input lifecycle. iPad and flag-off keep today’s sheet behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 524868f699cfe505c5689c61be2eea1302deda66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -725,6 +869,42 @@ private extension AIChatContextualSheetCoordinator {
...
+        case .summarize, .summarizePage:
+            pixelHandler.fireQuickActionSummarizeSelected()
+            promoteFloatingInputToSheet()
+            persistentUTIHost?.submitQuickActionPrompt(action.prompt)
+        }
+    }
+
+    func contextualInputViewController(_ viewController: AIChatContextualInputViewController, didSelectSuggestion suggestion: ContextualSuggestedPrompt) {
+        promoteFloatingInputToSheet()
+        sheetViewController?.submitSuggestion(suggestion)
```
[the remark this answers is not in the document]
aataraxiaa: +1

SharedPackages/AIChat/Sources/AIChat/iOS/Views/QuickActions/AIChatQuickActionChipView.swift
```diff
@@ -98,41 +124,149 @@ public final class AIChatQuickActionChipView: UIView {
...
             highlightOverlay.bottomAnchor.constraint(equalTo: bottomAnchor),
             highlightOverlay.leadingAnchor.constraint(equalTo: leadingAnchor),
             highlightOverlay.trailingAnchor.constraint(equalTo: trailingAnchor),
         ])
     }
 
+    /// Glass only applies its legibility treatment to content in its `contentView`; the translucent
+    /// style has no effect view, so there the chip hosts its own.
+    func hostContent(in host: UIView) {
+        guard iconView.superview !== host else { return }
```
aataraxiaa, line 244: I think this guard means the `.translucent` chips never get their content constraints. `setupUI()` does `addSubview(iconView)` (so `iconView.superview === self`), then `setupConstraints()` — which no longer carries the icon-leading / icon-label-spacing / label-trailing / centerY constraints — then `applyBackgroundStyle()`, which for the default style calls `hostContent(in: self)`. `host` is `self`, so this guard returns before `contentConstraints` is built, and `applyHorizontalPadding` then sets constants on three nil optionals.

As far as I can tell nothing else creates them, and reassigning `backgroundStyle = .translucent` in `AIChatQuickActionsView.configure` hits the same guard, so the icon and label end up with no positional constraints and the chip contributes no intrinsic width.

`.glass` is fine because the host is the effect view's `contentView`. But `useGlassStartActionBackgrounds()` only runs on the floating surface's chips controller, so the sheet's own `contextualInputViewController` stays translucent on every device and in both flag states. Since promoted sheets open straight onto the web view ("no pre-submit pills"), I suspect a translucent pill just never got rendered in testing.

Would `guard contentConstraints.isEmpty || iconView.superview !== host else { return }` cover it, or is there a reason the first pass should be a no-op that I'm missing?

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -795,6 +798,14 @@ private extension AIChatContextualChatSessionState {
         }
         // No "Ask about page" for pages that can't be attached — it would no-op on tap.
         guard isCurrentPageAttachable() else { return [] }
+        // Dropped only where the attachment strip shows its own re-attach button, off the same gate
+        // that drives it, so the two can't disagree.
+        let actions = quickActionsIgnoringReattachAffordance()
+        guard floatingInputFeature.isAvailable else { return actions }
+        return actions.filter { $0 != .askAboutPage }
```
aataraxiaa, line 806: The comment above says the two can't disagree because they're off the same gate, but I think there's a gap: the strip's button also needs `isOfferingReattach`, and `tapToRemove()` is its only setter, while `prepareForNewChat()` clears it.

So with auto-attach off, tapping "New chat" inside the sheet lands in `.placeholder` with `clearReattachOffer()` just having run — no `.askAboutPage` chip and no pill. The `+` menu's "Ask About Page" still gets you there, so it's not a dead end, but it's a change to the sheet on iPhone that the flag name doesn't really advertise.

Would gating this filter on the floating surface specifically (rather than on feature availability) be closer to what you want? Or widening the pill's condition to also cover a plain `.placeholder`?

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -474,6 +614,10 @@ private extension AIChatContextualSheetCoordinator {
         host.onPromptSubmitted = { [weak self] in
             guard let self else { return }
             self.sessionState.beginChatForUTISubmission()
+            if self.isFloatingInputPresented {
+                self.pixelHandler.fireFloatingInputPromotedToSheet()
```
aataraxiaa: I think this pixel only ends up counting typed prompts. Both chip paths call `promoteFloatingInputToSheet()` first, which nils `floatingInputViewController`, so by the time `onPromptSubmitted` runs `isFloatingInputPresented` is already false — a suggestion or Summarize tap fires neither this nor `fireFloatingInputDismissedWithoutSubmission` (promotion bypasses `dismissFloatingInput` by design). Those sessions would be missing from both counters, so the pair won't reconcile against surface impressions.

The definition ("a prompt sent from the floating contextual input opens the contextual sheet") reads like it's meant to cover suggestion taps, which I'd guess are the common path. Firing it inside `promoteFloatingInputToSheet()` would cover all three routes in one place — does that work for you?

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualWebViewController.swift
```diff
@@ -89,8 +89,10 @@ final class AIChatContextualWebViewController: UIViewController {
     private var isContentHandlerReady = false
     /// The Duck.ai initialization handshake is signaled by `getAIChatPageContext`.
     private let frontendReadinessGate = AIChatFrontendReadinessGate()
+    /// The frontend counts too: the page finishes navigating before the chat app starts listening, and
+    /// anything dispatched into that gap is dropped. Early prompts queue until `markFrontendAsReady()`.
     private var canDeliverPrompt: Bool {
-        isPageReady && isContentHandlerReady
+        isPageReady && isContentHandlerReady && frontendReadinessGate.isReady
```
aataraxiaa, line 95: This looks like the right call — dispatching into a frontend that isn't listening lost the prompt anyway, just less visibly.

One question on the unbounded part: every flush point (`markFrontendAsReady()`, and the two `submitPendingIfReady()` calls) depends on the same `getAIChatPageContext` handshake, so a load where the frontend accepts prompts but never issues it would keep a queued prompt forever rather than dropping it late. That applies to everyone with contextual mode on rather than just this flag.

Is the handshake guaranteed for every URL this VC loads? If not, would a bounded fallback — dispatch anyway after N seconds, the way `waitUntilFrontendReady` already bounds the suggestions path — be worth adding?

aataraxiaa (commented):
Nice work — the floating surface reads really well, and the choreography comments made this a pleasure to follow.

Four things I noticed while tracing it. The first one I think is a real break on the sheet's chips, so worth a look before merge; the rest are questions.

aataraxiaa (approved):
Tested and looks great so far @jaceklyp. Multi-agent review found some issues. I posted only those issues that I validated/agreed with.

The most pressing issue is that this breaks the older UX. Happy for you to fix here or in a follow-up

jaceklyp:
Thanks for the review @aataraxiaa — all four of yours plus both Bugbot findings are fixed in 898b7fc203. Details below, since a couple of them turned out to be deeper than the reported symptom.

### 1. `.translucent` chips got no positional constraints — confirmed, and it was a regression

Exactly as described. This PR moved the icon/label leading, spacing, trailing and centerY constraints out of `setupConstraints()` and into `hostContent(in:)`, but `setupUI()` had already run `addSubview(iconView)`, so the `guard iconView.superview !== host` returned on the very first translucent pass and `contentConstraints` stayed empty.

Rather than widen the guard, I removed the ownership split that caused it — `setupUI()` no longer parents the icon and label, so `hostContent(in:)` is the only thing that decides where content lives and the plain superview check is correct again.

That surfaced a second, latent problem: `highlightOverlay` was on top only by accident of `setupUI()`'s ordering, so **any** glass→translucent switch re-parented the content above the overlay and silently stopped it tinting the icon and label. `hostContent` now calls `bringSubviewToFront(highlightOverlay)` unconditionally to state that invariant.

### 2. Remove button tap swallowed (Bugbot, high) — confirmed

The chip-level `UITapGestureRecognizer` is new in this PR and spans the whole chip, so with default `cancelsTouchesInView` it cancelled the ×'s touch in the attached state.

`chipTapped` only ever meant anything in `.placeholder`, so instead of `cancelsTouchesInView = false` the recognizer is now simply not live outside that state — `placeholderTapRecognizer.isEnabled` is set per state in `updateUI(for:)`, and the now-redundant state guard inside `chipTapped` is gone. One mechanism expressing the rule instead of two.

### 3. Promotion pixel only counted typed prompts — confirmed

Both chip paths call `promoteFloatingInputToSheet()` before submitting, which nils `floatingInputViewController`, so the `isFloatingInputPresented` check in `onPromptSubmitted` was already false by the time it ran. Suggestion and Summarize taps fired neither this pixel nor `fireFloatingInputDismissedWithoutSubmission`.

Moved into `promoteFloatingInputToSheet()`, which is the single funnel for all three routes — mirroring how `dismissFloatingInput` already owns its own pixel.

### 4. Promoted sheet could stall with no input bar (Bugbot, medium) — confirmed

`viewDidLoad` with `opensOntoSubmittedChat` skips the mount and calls `transitionToWebView()`, but nothing enters the waiting state at load. The only thing that mounts the input afterwards is `revealInitialPromptWebViewIfNeeded()`, which requires that wait to have started — and for suggestions that only happens if the submission Task reaches `submitSuggestionPrompt`. A 5s `waitUntilFrontendReady` timeout therefore left a sheet on the web view with no input at all and no way back.

Fixed deterministically rather than with a longer timer: `deliverSuggestionPrompt` now reports whether anything is still going to bring the sheet a chat — a prompt that went out, or another surface that took over. Only when nothing is does `abandonAwaitedSubmittedChat()` run, which drops out of the awaiting state and re-applies the view state so the sheet becomes an ordinary pre-submit surface the user can retype into. Cancellation deliberately counts as "something else is driving", so a superseded submission doesn't tear down a sheet whose chat is genuinely in flight.

The doc comment on `beginWaitingForInitialPromptResponseStateIfNeeded` claiming the sheet "carries the fallback that guarantees the reveal happens" was overstating things and has been corrected.

### 5. "New chat" with auto-attach off lost both affordances — confirmed, and the comment was wrong

You were right that the two gates could disagree. The filter dropped `.askAboutPage` on `floatingInputFeature.isAvailable` alone, while the strip's button additionally needs `isOfferingReattach`, which `prepareForNewChat()` clears — so after a new chat, neither was left.

The filter now also requires `userDowngradedToPlaceholder`, the session-state fact for the same explicit removal that sets `isOfferingReattach`. `resetToNativeInputState` pairs `resetToNoChat()` with `prepareForNewChat()`, so the two move together.

This makes `.askAboutPage` reachable on the floating surface in the case where the page is attachable but the attach didn't land, where it previously hit a `break` and did nothing — it now performs the attach.

Two regression tests added for exactly the reported sequence: remove → chip stands down, new chat → chip returns; plus the flag-off case where nothing else offers re-attach so the chip must survive removal.

One thing left as follow-up rather than half-done here: this is still **two** flags for one fact, set by the same user action in two places. Collapsing them to one source of truth needs a decision about where that state belongs, since `UnifiedToggleInputPageContextChipViewModel` is shared with the omnibar UTI, which has no contextual session to read from. Filed with the other state-duplication work.

### 6. Unbounded prompt queue — not changed, and I don't think it should be here

No, the handshake isn't guaranteed for every URL this VC loads, so you've identified a real hole. But every flush point depending on `getAIChatPageContext` predates this flag and applies to all of contextual mode, so a bounded fallback belongs in its own change where it can be reasoned about for the restore and new-chat paths too, rather than riding in behind `aiChatContextualFloatingInput`. Happy to pick it up next if you agree.

Build clean, 236 unit tests green.

aataraxiaa (approved):
Smoke tested latest changes, look good 👍🏼

PR #6290, opened 2026-08-11
Define DEBUG for SPM packages in CI so package snapshot tests compile

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214200115953388/task/1217362000068446

### Description
CI builds with `-configuration CI`, which SwiftPM maps to `.release`, so a package's `.define("DEBUG", .when(configuration: .debug))` never fires. As a result the `#if DEBUG` PreviewProviders that `SyncUI-macOS`'s snapshot suites reference aren't compiled, and the package test target fails to build under CI.

This gates an **unconditional `DEBUG` define in `SyncUI-macOS/Package.swift` behind a `SPM_FORCE_DEBUG_FOR_SNAPSHOTS` env var**, set only on the macOS PR `Build for testing` step. DEBUG is scoped to the single local package that needs it — vendored dependencies are never touched — and release/notarized workflows (which don't set the var) keep building as `.release`, so no debug code ships. The now-unnecessary `#if DEBUG` guards on the two snapshot suites are dropped.

### Testing Steps
1. CI green — `Build for testing` compiles the `SyncUI-macOSTests` target (the snapshot suites remain `.disabled`, so they compile without running).

### Impact
Low — CI test-build configuration only; release/notarized builds are untouched, so no debug code ships.

### DIFF
--- a/.github/workflows/macos_pr_checks.yml
+++ b/.github/workflows/macos_pr_checks.yml
@@ -276,6 +276,8 @@ jobs:
 
     - name: Build for testing
       id: build-for-testing
+      env:
+        SPM_FORCE_DEBUG_FOR_SNAPSHOTS: "1"
       run: |
         echo "Runner ${RUNNER_NAME} (${RUNNER_TRACKING_ID})"
 
--- a/macOS/LocalPackages/SyncUI-macOS/Package.swift
+++ b/macOS/LocalPackages/SyncUI-macOS/Package.swift
@@ -2,6 +2,12 @@
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
+import Foundation
+
+// Set by the "Build for testing" step in .github/workflows/macos_pr_checks.yml. Under CI the package builds
+// in release configuration (SPM maps the CI configuration to .release), so `.when(configuration: .debug)` below
+// doesn't fire and the DEBUG-only snapshot tests wouldn't compile. This forces DEBUG on for that CI build.
+let forceDebugForSnapshots = ProcessInfo.processInfo.environment["SPM_FORCE_DEBUG_FOR_SNAPSHOTS"] == "1"
 
 let package = Package(
     name: "SyncUI-macOS",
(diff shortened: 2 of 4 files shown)


tomasstrba (approved):
Approving.

Just a couple of non-blocking nit-picks found by Claude:

1. Add a comment in `Package.swift` next to the `snapshotTestBuild` check saying which workflow step sets `PACKAGE_SNAPSHOT_TESTS` and why. Right now nothing connects the manifest to `macos_pr_checks.yml` — whoever next tidies that `env:` block has no way to know a package manifest reads it.

2. `environment["PACKAGE_SNAPSHOT_TESTS"] != nil` treats `PACKAGE_SNAPSHOT_TESTS=0` as on. `== "1"` is more honest.

3. The name reads like it selects tests, but what it does is define `DEBUG`. Something like `SPM_FORCE_DEBUG_DEFINE` would describe the effect.

4. Defining `DEBUG` in an optimized build is fine today — the only four `#if DEBUG` sites in `SyncUI-macOS` are preview code — but it's a latent trap. The day someone adds debug-only *behaviour* to this package, CI tests will exercise a path that shipping builds don't. A dedicated define (`SNAPSHOT_PREVIEWS`, set the same way, with the four `#if DEBUG` sites changed to match) keeps the same mechanism with a much narrower blast radius.

pballart:
> Approving.
> 
> Just a couple of non-blocking nit-picks found by Claude:
> 
> 1. Add a comment in `Package.swift` next to the `snapshotTestBuild` check saying which workflow step sets `PACKAGE_SNAPSHOT_TESTS` and why. Right now nothing connects the manifest to `macos_pr_checks.yml` — whoever next tidies that `env:` block has no way to know a package manifest reads it.
> 2. `environment["PACKAGE_SNAPSHOT_TESTS"] != nil` treats `PACKAGE_SNAPSHOT_TESTS=0` as on. `== "1"` is more honest.
> 3. The name reads like it selects tests, but what it does is define `DEBUG`. Something like `SPM_FORCE_DEBUG_DEFINE` would describe the effect.
> 4. Defining `DEBUG` in an optimized build is fine today — the only four `#if DEBUG` sites in `SyncUI-macOS` are preview code — but it's a latent trap. The day someone adds debug-only _behaviour_ to this package, CI tests will exercise a path that shipping builds don't. A dedicated define (`SNAPSHOT_PREVIEWS`, set the same way, with the four `#if DEBUG` sites changed to match) keeps the same mechanism with a much narrower blast radius.

Thanks @tomasstrba ! I fixed the first 3 while acknowledging the 4th as a controlled risk.

PR #5782, opened 2026-07-10
macOS: Updates Settings Sidebar DRK Icons

jleandroperez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1176956903599313/task/1216457879063564?focus=true
Tech Design URL:
CC:

### Description
This PR updates the Icons used in the macOS Settings UI, with the latest assets.

### Testing Steps
1. Enable the `.appRebranding` Feature Flag
2. Relaunch

- [ ] Verify the Settings UI matches the [latest mockups](https://www.figma.com/design/DBZiXJRgTTJTuBeWAPyE5h/%F0%9F%96%A5%EF%B8%8F-Desktop-Components?node-id=34524-70629&m=dev)

### Impact
Low: Minor visual changes, small bug fixes, improvement to existing features

#### What could go wrong?
The incorrect Icon could be picked up

### Quality Considerations
Nothing in particular.

Thank you!!

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Low Risk**
> Visual-only asset and accessor changes gated by `AppRebrand`; main risk is a wrong legacy/rebrand mapping or missing asset name at runtime.
> 
> **Overview**
> Adds **app-rebrand–aware** 16px color icons for macOS Settings (and shared DRK callers): many `DesignSystemImages.Color.Size16` accessors now pick **new** assets when `AppRebrand.isAppRebranded()` is true and **`*-legacy`** imagesets otherwise.
> 
> **Assets:** Refreshed SVGs in the primary imagesets, duplicated prior art into new `*-legacy` catalogs, and added/repointed resources such as `Browser-Default-Color-16`, `DuckDuckGo-Ai-Color-16`, and `Subscription-Color-16`. Rebrand builds also swap a few symbol names (e.g. default browser → `browserDefaultColor16`, private search → `searchFindColor16`, subscription → `subscriptionColor16`, paid AI → `duckDuckGoAiColor16`).
> 
> **macOS:** `CurrentSettingsIconProvider` uses **`shield`** instead of **`shieldCheck`** for web tracking protection so the sidebar matches the updated DRK icon.
> 
[...]

PR #6292, opened 2026-08-11
iOS: Deliver text selections to Duck.ai (4/5)

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1216945875759265
Tech Design: https://app.asana.com/1/137249556945/project/1206488453854252/task/1217363923358702?focus=true
CSS PR: https://github.com/duckduckgo/content-scope-scripts/pull/2953
CC:

> [!IMPORTANT]
> Release gate: do not merge or ship this PR until [Privacy Configuration PR #5782](https://github.com/duckduckgo/privacy-configuration/pull/5782) is deployed and rollout ordering is confirmed. It enables C-S-S `textSelection` and native `aiChat.textActions` from iOS 7.235.0; macOS remains on the disabled base state. This is not a staged user-facing rollout—the feature should become available when iOS 7.235.0 ships.

### Description

Layer 4 of 5. Behind `aiChatTextActions` (`internalOnly`), iPhone only. Attached selections now reach the model. A selection is attached only through the explicit **Ask Duck.ai** text-selection menu item; opening Duck.ai from the omnibar never reads page selection state.

- **Delivery.** Selections travel on the prompt's own `selections` key, *alongside* `pageContext` rather than replacing it, so "the page plus N selections" is expressible. Payloads with no selection are byte-identical to before (`encodeIfPresent`). Selections are cleared only after dispatch, and only the IDs actually sent — so one attached mid-flight survives, and a dropped push clears nothing.
- **Suggestion taps.** "Summarize / Translate this selection" now submit, acting on the selection without also attaching the page.
- **iframes.** A selection made inside an iframe is readable: the Content Scope Scripts `textSelection` feature reports only *whether* a selection exists, never its text, so native can retain the reporting frame and read it only when the user picks a menu item.

The read is bound to the document that reported the selection and fails closed rather than falling back to the main frame. The threat model, the alternatives considered and the residual risks are in the tech design linked above — not repeated here.

Not included: pixels and wide events (layer 5).

**Known frontend gap, not an iOS change:** a delivered selection currently renders as a separate item in the conversation rather than as an attachment on the submitted prompt. Already raised with the frontend — please don't file it.

### Steps to test

**Setup** — iPhone. Flags on: `aiChatTextActions`, `aiChatContextualUnifiedToggleInput`, `contextualDuckAIMode`, `pageContextFeature`, `contextualSuggestedPrompts`. Settings → AI Features → Duck.ai enabled. Production frontend (no `aichat.debug.customURL`).

The branch consumes production Content Scope Scripts 16.10.0, containing the generated build from [Content Scope Scripts PR #2953](https://github.com/duckduckgo/content-scope-scripts/pull/2953).

**Selection entry point** — Unless a step explicitly says **omnibar**, “Ask Duck.ai” means the item in the text-selection edit menu. The omnibar opens ordinary contextual chat and never implicitly attaches selected text.

**Auto-attach** — Settings → AI Features → **Automatically Send Content**. Its default follows the `aiChatAutoAttachContextByDefault` flag rather than being fixed, so check the toggle rather than assuming. Sections A and B must each be run in their stated state; C and D are unaffected by it.

**A page with an iframe** — [MDN's `<iframe>` reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe) renders live examples in iframes with selectable content. Note its frames are `about:blank`, so they are same-origin with the parent, which is all step 8 needs: nothing in the read path branches on origin (`ancestorOrigins` is only checked for *whether* the frame is framed), so a same-origin subframe exercises the identical code.

#### A. Auto-attach ON

1. **The page is not attached on top of a selection.** Select a distinctive paragraph → **Ask Duck.ai** → the sheet shows **only the selection chip, no page chip**, even though auto-attach is on. This is deliberate: the selection is what the user asked about. Ask "what does this say?" → the answer is about **that paragraph**, not the whole page, and the chip clears on send.
2. **Page plus selection, and suggestions hide at two attachments.** Open the sheet on a page **without** a selection so the page chip auto-attaches, then select text → **Ask Duck.ai**. Both chips present, selection suggestions **hidden**, and the answer reflects **both** — the page was not replaced. Remove the page chip → back to one attachment, suggestions appear.

#### B. Auto-attach OFF

3. **No page chip; the selection still delivers.** Select → **Ask Duck.ai** → selection chip only, answer about the selection, and the selection suggestions **are** offered (one attachment).
4. **Manual page attach still composes.** Attach the page manually, then add a selection → both ride, as in step 2.

#### C. Independent of the setting

5. **Suggestions submit.** With one selection attached, **Summarize this selection** submits, answers about the selection, and does **not** attach the page. Repeat **Translate this selection** on a foreign-language page. Then tap a suggestion and remove the chip while it loads → nothing is submitted. (Suggestions only appear before a chat's first prompt.)
6. **Multiple selections.** Attach 2–3 from different pages, navigating between them (they persist) → all are reflected. A 6th shows the limit banner (cap 5).
7. **Attach during dispatch.** Attach a selection, send, and immediately attach a second while the first dispatches — the second chip **survives**.
8. **Alternate main frame ↔ iframe on the first attempt.** On one page with distinctive selectable text in both places, alternate **main → iframe → main → iframe → main**. Each time, make a fresh selection and immediately choose **Ask Duck.ai** from that selection's edit menu without reselecting. Verify the chip/answer uses the newly selected frame's exact text on the first attempt; remove the previous chip or burn the chat between attempts so the result is unambiguous. Opening the menu naturally moves focus away from an iframe, and its selection must still attach. This is the diagnostic-free acceptance test for both the deferred focus claim and repeated ownership reassertion in CSS PR #2953.

    *Optional, opportunistic:* if you happen to hit a page with a genuinely cross-origin embed (CodePen, a comment widget — check the origin in Web Inspector), repeat the first part there. It exercises no new code of ours, but it is the only way to observe whether `document.hasFocus()` and native frame-targeted `evaluateJavaScript` behave the same across origins. Cross-origin *support* is an open decision for the PR5 privacy triage, so this is an observation to feed that, not a gate on this PR.
9. **The omnibar never attaches hidden or stale text; bfcache restores.** Attach and submit a selection, burn the chat, return to the page, then tap the **omnibar** Duck.ai icon → the sheet opens with **no selection chip**, even if WebKit retained the DOM range after hiding the blue selection UI. Repeat after selecting text, switching tabs, and returning after the blue highlight disappears. Then select text, navigate forward and go back → if the restored selection is visibly highlighted, choosing **Ask Duck.ai** from its text-selection menu uses it again.

#### D. Regression

10. **Search with DuckDuckGo** — in scope because it shares the changed read path. It searches the selection, and with **Duck.ai switched off** in Settings it still appears and works while **Ask Duck.ai** is **absent**.
11. **No selection.** Page attach/remove, page suggestions and normal Duck.ai chat behave as before, in both auto-attach states: with it **on** a page-context-only prompt is unchanged; with it **off** the page is still not attached and page shortcuts still work from the signals-only payload.
12. **Kill switch and OS.** With `aiChatTextActions` off both items disappear. Spot-check one older OS (16 or 17) that the menu still behaves — no OS gate is enforced, and `UIMenuBuilder` selection menus arrived in iOS 16, so on iOS 15 the items are expected simply not to appear.

### Definition of Done

- [ ] Tested with auto-attach **on** and **off**, and the page is never attached on top of a selection
- [ ] Selections reach the model alongside page context, on the production frontend
- [ ] Selections attached during dispatch are not cleared by that dispatch
- [ ] Summarize and Translate suggestions submit without attaching the page
- [ ] Alternating main-frame and iframe selections attach the newly selected frame on every first attempt
- [ ] iframe selections survive focus moving to the text-selection menu
- [ ] The omnibar never implicitly attaches hidden or stale selected text
- [ ] A stale or mismatched frame delivers nothing rather than the wrong text
- [ ] No regression to page context, Search with DuckDuckGo, or existing suggestions

### Internal references

[Software Engineering Expectations](https://app.asana.com/0/1201621942593206/1201517092knowledge)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971970628)

---

> [!NOTE]
> **Medium Risk**
> Touches WebKit frame-targeted selection reads and Duck.ai prompt payload semantics; wrong-frame or premature consumption could drop or mis-send user-selected text, though the design fails closed on stale frames and only consumes after successful dispatch.
> 
> **Overview**
> iOS layer 4 ships **attached text selections through to Duck.ai** on the native prompt bridge, alongside existing `pageContext`, gated behind `aiChatTextActions` on iPhone.
> 
> **Payload & bridge.** `AIChatNativePrompt` gains an optional `selections` array (omitted when empty). `AIChatUserScript` inlines selections on `submitPrompt`, clears chips only after a successful bridge push, and only for the IDs that were sent—so queued prompts snapshot selections and late attachments survive dispatch.
> 
> **Contextual UX.** Selection entry suppresses auto page attach; selection-scoped suggestion taps submit without forcing a page attach. **Ask Duck.ai** can open the **floating contextual input** when that surface is enabled (not only the sheet). Omnibar Duck.ai no longer implicitly reads page selection.
> 
> **iframe-safe reads.** New `SelectionFrameUserScript` (Content Scope Scripts **16.10.0** `textSelection`) tracks which frame has a selection; the edit menu reads via isolated-world JS in that frame with timestamp binding, and menu items require the frame machinery to be available.
> 
> **Tests** cover delivery, consumption, floating input routing, and frame claim logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3273517f218e8b57c11cf6b389fe7f5c105a0d91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/TextSelection/selectionframe.js
```diff
@@ -0,0 +1,106 @@
...
+    // user has committed, by which point the page has had time to swap the range or rewrite the text in it.
+    function publish(text) {
+        if (text.trim().length > 0 && canClaim()) {
+            snapshot = text;
+            post(true);
+            return;
+        }
+        snapshot = '';
+        post(false);
+    }
```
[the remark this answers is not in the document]
aataraxiaa: Good catch — this one is real, fixed.

`publish()` was treating "non-empty selection in a frame that may not claim" identically to "no selection", so it wiped its own `snapshot` and posted `false`. The focus gate is only meant to stop an unfocused subframe *acquiring* a claim; it should never destroy a selection the frame legitimately holds. As written, any `selectionchange` arriving after focus moved out of a claiming iframe dropped the text Ask/Search would read.

Now: empty selection releases the claim as before; non-empty in an ineligible frame is a no-op, leaving the stored selection intact. This is also strictly safer than before — if a frame that legitimately claimed later re-selects while unfocused, we keep the text the user selected rather than the replacement.

Not unit-testable here: this file executes no JavaScript (the cases assert on `sut.source` substrings and a mocked `WKFrameInfo`), so it needs the iframe harness or a manual pass.

iOS/DuckDuckGo/WebView.swift
```diff
@@ -153,29 +156,40 @@ final class WebView: WKWebView {
...
                 completion(nil)
                 return
             }
             completion(Self.normalizedSelection(text))
         }
+
+        frame.evaluateJavaScript(SelectionFrameUserScript.readSelectionScript,
+                                 in: self,
+                                 contentWorld: .defaultClient,
+                                 completionHandler: handler)
```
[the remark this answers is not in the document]
aataraxiaa: Fail-closed is deliberate here, not a regression — so no fallback is going back in.

Returning main-frame text when the tracked frame is missing or the token mismatches would be a silent substitution requiring no attacker at all: the user selects in an iframe, we hand Duck.ai or Search unrelated text from the top document, and nothing signals it. A dead tap is recoverable by selecting again; wrong text sent to a model is not. The PR does not claim stale frames "degrade rather than dead-tap" — it claims the read fails closed.

One part of this is fair, though: "the script message has not landed yet" is a genuine race, not a fallback question. The script posts on selection *appearance* rather than on clear specifically to narrow the window between the action and the message identifying its frame, but both are asynchronous so it is narrowed, not eliminated. That is recorded as a known limitation in the tech design rather than papered over with a fallback.

iOS/DuckDuckGoTests/TextSelection/SelectionFrameUserScriptTests.swift
```diff
@@ -0,0 +1,198 @@
...
+    /// An out-of-view third-party iframe could otherwise select its own text and have that read instead.
+    func testASubframeOnlyClaimsTheSelectionWhileFocusIsInsideIt() {
+        XCTAssertTrue(sut.source.contains("ancestorOrigins"))
+        XCTAssertTrue(sut.source.contains("document.hasFocus()"))
+    }
+
+    /// The text is read later, only once the user picks a menu item.
+    func testNeverSendsTheSelectedText() {
+        XCTAssertFalse(sut.source.contains("postMessage({ text"))
+        XCTAssertTrue(sut.source.contains("hasSelection: hasSelection"))
```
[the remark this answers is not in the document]
aataraxiaa: Leaving these as they are, having counted them: across the three files cited there are 17 `///` lines, and **15 are single lines** stating why a test exists — "a rename on one side only would silently stop tracking", "chips are reconciled and removed by id, so identical text must not collide". A bare `// given / when / then` cannot carry that, and it is the part a future reader actually needs.

Two are multi-line, and neither is a doc comment on a test method: one is a class-level rationale on the `AIChatUserScriptSelectionDeliveryTests` suite, the other a caveat noting that its test covers the mechanism rather than proving it sufficient.

On the `macOS` references in `AIChatSelectionContext.swift` — those justify constants rather than narrate code (the 5 cap and the 9500-char `maxContentLength` both match macOS and content-scope-scripts `page-context.js`), and `macOS/DuckDuckGo/AIChat/AIChatSelectionContextAttacher.swift` does exist, so the cross-reference is live rather than stale. Dropping them would lose the only record of where those numbers come from.

Happy to trim if a human reviewer disagrees — but Danger already flags this PR as large, and rewriting 17 accurate one-liners would add diff noise without adding information.

iOS/DuckDuckGoTests/AIChat/AIChatSelectionContextTests.swift
```diff
@@ -154,3 +193,92 @@ final class AIChatSelectionContextBuilderTests: XCTestCase {
         XCTAssertFalse(AIChatSelectionContextBuilder.displayTitle(for: selection).contains("…"))
     }
 }
+
+/// Consumption clears the chips and the session's selections, so it has to happen exactly when they
+/// reach the frontend — never on a submit that carried none, and never on one that was dropped before
+/// it got there, or the user's text is destroyed unsent.
```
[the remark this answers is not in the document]
aataraxiaa: Same point as the newer thread on `SelectionFrameUserScriptTests` — answered there in full, in short: 15 of the 17 `///` lines are single-line statements of *why*, the two multi-line ones are a suite-level rationale and a mechanism caveat rather than method docs, and the `macOS` references justify the cap and truncation constants against a file that does exist. Leaving them.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualChatSessionState.swift
```diff
@@ -520,6 +523,20 @@ final class AIChatContextualChatSessionState {
         beginLoadingSuggestions()
     }
 
+    /// Suppresses auto-attach for a sheet opened from a text selection, until the user navigates, attaches
+    /// the page themselves, or opens the sheet again without a selection.
+    ///
+    /// `pendingSignalsOnlyCollection` alone is not enough: it is consumed by the first context update to
+    /// arrive, so an emission already in flight when the sheet opened would spend it and let the collection
+    /// we actually triggered fall through to auto-attach, attaching the page a moment after the selection.
```
aataraxiaa: Make this more concise

iOS/DuckDuckGo/TextSelection/SelectionFrameUserScript.swift
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    func selectedText(from value: Any) -> String? {
+        guard let result = value as? [String: Any],
+              result["eventTimestamp"] as? Double == eventTimestamp else { return nil }
+        return result["selectedText"] as? String
+    }
+}
+
+/// Tracks which frame holds the page's text selection, so a selection made inside an iframe can be read.
```
aataraxiaa: “Track” is not accurate here. Remove any usage of this word from this PR.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -419,7 +433,16 @@ final class AIChatContextualSheetCoordinator {
             sessionState.clearAttachedSelections()
         }
 
-        await presentSheet(from: presentingViewController, restoreURL: restoreURL, skippingAutoAttach: true)
+        if floatingInputFeature.isAvailable,
```
Bunn, line 436: should this also require floatingInputViewController == nil? Right now it nulls persistentUTIHost and then presentFloatingInput bails on its own guard, so the chip never renders.

iOS/DuckDuckGo/TextSelection/SelectionFrameUserScript.swift
```diff
@@ -0,0 +1,117 @@
...
+//
+
+import WebKit
+import UserScript
+
+/// A frame that can be evaluated in, without exposing the `WKFrameInfo` it wraps.
+///
+/// Reading `WKFrameInfo.request` terminates the app, so the value is kept unreachable rather than
+/// documented as unsafe.
+struct SelectionFrame {
```
Bunn, line 27: should this be @MainActor? All its state is main-only by convention today.

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatConsumableDataHandling.swift
```diff
@@ -311,6 +311,8 @@ public struct AIChatPageContextData: Codable, Equatable {
 /// ones it missed (via `getAIChatSelectionContext`), mirroring how `AIChatPageContextHandler` backs
 /// `getAIChatPageContext`. Unlike the single-slot page-context handler, selections append and the
 /// list is read non-destructively (the FE dedupes by `id`); it's cleared when a prompt is submitted.
+///
+/// Not used where prompts are composed natively — those carry selections on `AIChatNativePrompt.selections`.
 public final class AIChatSelectionContextHandler {
```
Bunn, line 316: will this become dead code once the changes are shipped?

Bunn (approved):

SharedPackages/AIChat/Sources/AIChat/Shared/AIChatConsumableDataHandling.swift
aataraxiaa: No — this remains live macOS code. `AIChatMessageHandler` owns this handler and uses it to append selections pushed before the frontend is ready, return them from `getAIChatSelectionContext`, and clear them after prompt submission. iOS composes selections directly into `AIChatNativePrompt`, which is why the comment calls out that distinction.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
aataraxiaa: The floating surface is dismissed before another page selection can invoke this path. Page activation requests dismissal synchronously, and `dismissFloatingInput()` clears `floatingInputViewController` before the dismissal animation begins. The subsequent selection action therefore creates a fresh host; it cannot reach this branch with the old floating controller still present.

iOS/DuckDuckGo/TextSelection/SelectionFrameUserScript.swift
aataraxiaa: Done

PR #6303, opened 2026-08-12
Add PixelKit.Options and reshape the fire API around it

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1217419466538070?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1217412374194171
CC:

### Description

Replaces PixelKit's eight-parameter `fire` with two entry points over a single protocol requirement.

```swift
pixelKit.fire(event, frequency: .daily)                      // fire and forget, as today
pixelKit.fire(event, options: .parameters(["source": "x"]))  // when the request needs shaping
try await pixelKit.fireAsync(event, frequency: .daily)       // when the outcome matters
```

The six parameters that describe *how the request is shaped* collapse into `PixelKit.Options`. The two that describe *what the pixel is*, event and frequency, stay where they are. `options` is defaulted, so the majority of call sites, the ones passing only an event and a frequency, are untouched by this PR and read exactly as before.

`Options` ships presets derived from measured usage rather than from what is theoretically expressible: `.unenforcedPrefix` (133 sites), `.parameters(_:)`, `.namePrefix(_:)` (17 sites, all DBP), `.withoutAppVersion` (19 sites). Combinations nobody uses deliberately get no preset; it is a value type, so those compose.

Top of a 2-PR stack, on top of the `PixelKit.Event` rename.

**Call sites that reach PixelKit through the `PixelFiring` protocol are migrated here** (28 sites across 8 files), because removing the 8 convenience overloads leaves them no other way to pass those parameters. Call sites using a concrete `PixelKit` or the static `PixelKit.fire` are untouched: each keeps its own legacy method, so they migrate later.

### Testing Steps

No user-facing behaviour to exercise. Pixels must fire exactly as before. CI must be green.

### Impact

Medium. Nothing user-facing changes, but pixel names and parameters are contractual, and this PR moves every pixel through a rewritten code path.

#### What could go wrong?

- `doNotEnforcePrefix: false` becomes `enforcePrefix: true`, and a botched polarity inversion would silently rename production pixels → mitigated by 6 parity tests asserting the legacy and options paths produce byte-identical names and parameters. These were checked by mutation: deliberately inverting the polarity turns 0 failures into 18, so they hold the invariant rather than passing vacuously.
- A conformer could fail to implement the new requirement and silently recurse forever → mitigated by giving the requirement an `event:` label the sugar does not have. A same-signature extension method would otherwise become the requirement's own default implementation. Omitting it is now a compile error naming the expected labels.
- `fireAsync` could resume its continuation twice under `.dailyAndCount` / `.dailyAndStandard`, which complete more than once → mitigated by the existing single-resume latch, with a regression test.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> All pixels flow through a rewritten path where wrong `enforcePrefix` polarity or header handling could change contractual names or wire payloads; parity tests and the distinct protocol witness mitigate this, but telemetry regressions would be silent until validated.
> 
> **Overview**
> **PixelKit firing is reshaped** so request-shaping knobs live in **`PixelKit.Options`** while **`event`** and **`frequency`** stay on **`fire`**. Callers use **`fire(_:frequency:options:)`** for fire-and-forget or **`fireAsync`** when they need the outcome; **`fireAsync`** now returns **`FireResult`** (`.sent` / `.suppressed`) instead of a boolean.
> 
> The **`PixelFiring`** protocol has one requirement, **`fire(event:frequency:options:onComplete:)`** (the **`event:`** label avoids accidental infinite recursion via protocol extensions). The old eight-parameter overloads on the extension are removed; **`PixelKitMock`** and protocol call sites are updated to **`options:`** presets such as **`.unenforcedPrefix`**, **`.parameters(_:)`**, and **`.withoutAppVersion`**.
> 
> **Legacy wide-parameter `fire` on `PixelKit`** remains as a forwarding shim for unmigrated concrete/static call sites. **`PixelKit.fireAsync`** on the shared instance throws **`PixelKitError.notConfigured`** when **`setUp`** was not called. Docs (**`pixels.mdc`**, **`BUGBOT.md`**) describe the two **`PixelFiring`** protocols and how extra parameters appear in **`options`**.
> 
> Tests add **legacy vs options parity** (names, parameters, headers), **`fireAsync`** multi-completion handling, and static/instance header behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 006f52afabd8e1ea80312db9fbb167be4b96e4bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Options.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Options.swift
@@ -27,7 +27,12 @@ extension PixelKit {
     /// call reads `fire(event)` or `fire(event, frequency: .daily)` with no options in sight.
     public struct Options: Equatable, Sendable {
 
-        /// Extra HTTP headers merged over the instance's default headers.
+        /// HTTP headers for the request.
+        ///
+        /// Note this *replaces* the instance's `defaultHeaders` rather than merging with them: see
+        /// `fire(pixelNamed:)`, which does `headers ?? defaultHeaders`. Leave it `nil` (the default)
+        /// to send `defaultHeaders`. Long-standing PixelKit behaviour, preserved here deliberately;
+        /// whether it should merge instead is a separate question from this API reshape.
         public var headers: [String: String]?
 
         /// Extra query parameters merged over the event's own parameters. On key collision the
--- a/SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Options.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Options.swift
@@ -1,0 +1,110 @@
+//
+//  PixelKit+Options.swift
+//
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+extension PixelKit {
+
+    /// How a pixel request is shaped, as opposed to what the pixel *is*.
+    ///
+    /// `event` and `frequency` stay as parameters on `fire` because they describe the telemetry
+    /// itself. Everything here describes the outgoing request, and is defaulted, so the common
+    /// call reads `fire(event)` or `fire(event, frequency: .daily)` with no options in sight.
+    public struct Options: Equatable, Sendable {
+
+        /// Extra HTTP headers merged over the instance's default headers.
+        public var headers: [String: String]?
+
+        /// Extra query parameters merged over the event's own parameters. On key collision the
+        /// value supplied here wins.
+        public var additionalParameters: [String: String]?
+
+        /// Overrides the platform prefix applied to the pixel name.
+        public var namePrefix: String?
+
+        /// Characters left unescaped when building the query string.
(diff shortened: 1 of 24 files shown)


SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Options.swift
```diff
@@ -0,0 +1,110 @@
...
+extension PixelKit {
+
+    /// How a pixel request is shaped, as opposed to what the pixel *is*.
+    ///
+    /// `event` and `frequency` stay as parameters on `fire` because they describe the telemetry
+    /// itself. Everything here describes the outgoing request, and is defaulted, so the common
+    /// call reads `fire(event)` or `fire(event, frequency: .daily)` with no options in sight.
+    public struct Options: Equatable, Sendable {
+
+        /// Extra HTTP headers merged over the instance's default headers.
```
brindy: Not sure if this is an existing bug in PixelKit or just incorrectly documented but in PixelKit.swift:426 supplying headers replaces the default rather than merging them. (Transparency: GPT spotted this)

brindy (approved):
LGTM - just one comment to clarify

SharedPackages/BrowserServicesKit/Sources/PixelKit/PixelKit+Options.swift
federicocappelli: Pat your GPT on the head, I think this is a bug other than a doc error. I'm investigating

federicocappelli: Claude:
---
Good catch, and GPT was right — the doc comment was wrong, not the code. Fixed in 46a7e6d.

If you pass headers, they replace the default ones. They do not merge. That is how it already worked, and this PR does not change it. I fixed the comment and added a test.

One extra thing I found: the old static fire was losing the default headers, while the instance one kept them. Both are the same now. This only shows up in the macOS VPN tunnel, as everywhere else the default headers are empty.

PR #6321, opened 2026-08-12
[iOS] [Subscription Onboarding] Add Core Flow Orchestration and Entry Points (2/2)

hanyutang-sandra:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217384771427831?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1215971357218019?focus=true

### Description
Second and final of a 2-PR split of the post-checkout subscription onboarding flow, built on PR 1. Adds the flow controller and its view factory, which stitch PR 1's individual screens into one navigable flow (order confirmation → welcome → VPN → widget → IDTR → Duck.ai → progress); out-of-flow activation recording — completing the VPN or Duck.ai checklist step when a customer turns those on outside the flow entirely (from Settings, from the AI Chat entry points); PIR availability; and the feature flag plus the two real entry points that make the flow reachable by customers — automatically after finishing a purchase, and via a "Continue Setup" card in Subscription Settings if they didn't finish everything the first time. Also restores the debug menu's "Full Flow" entry that PR 1 intentionally held back, since it needs this PR's flow controller to exist.

(Originally split into two PRs — Core Flow Orchestration and Feature Flag/Entry Points — but merged into one after finding that `SettingsViewModel`'s new required init param and its only caller needed to land together to keep each PR independently buildable; splitting further didn't add reviewable value once that was resolved.)

### Testing Steps
1. Debug menu → Subscription → Onboarding → "Full Flow After Subscription" → step through the whole flow end to end, confirming forward/back navigation and the progress summary update as steps complete.
2. Turn on VPN or Duck.ai from outside the flow (Settings, AI Chat) → re-open the flow's progress screen → confirm the corresponding checklist item is marked complete.
3. Enable the subscription onboarding flag locally (internal-user override) with it otherwise off.
4. Complete a purchase → the onboarding sheet should appear automatically after the post-checkout confirmation page.
5. Exit partway through, then open Subscription Settings → confirm the "Continue Setup" card appears and resumes at the first unfinished step.
6. Complete the flow → confirm the onboarding-flow pixels fire as expected.
7. Turn the flag back off → confirm neither entry point presents the flow anymore.
8. Run the updated flow view model, instrumentation, and post-checkout trigger unit tests → all pass.

### Impact
Medium — could disrupt the post-purchase experience for real subscribers once the flag is turned on, since this is the PR that makes the flow live.

#### What could go wrong?
The post-checkout trigger fires once, off the purchase-completion signal → a bug here could show the onboarding sheet at the wrong time, more than once, or never — mitigated by dedicated trigger tests and the flag defaulting off, so a bad trigger ships dark until explicitly released.

### Quality Considerations
Both entry points are gated by the same flag, checked only when the flow is about to be entered rather than continuously — so a remote flag flip mid-flow can't strand a customer partway through.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes the post-purchase and subscription-settings paths for paying users once experiments are enabled; incorrect gating or duplicate presentation could affect subscriber onboarding, though flags default off and launch is latched after first show.
> 
> **Overview**
> Adds **post-purchase subscription onboarding orchestration** so entitled subscribers can walk a guided setup (order confirmation → welcome → VPN/widget/IDTR/Duck.ai → progress) from **post-checkout** or **Subscription Settings**, gated by new **Sep 2026 A/B experiments** (US English, free-trial vs paid cohorts).
> 
> **Flow & entry:** Introduces `SubscriptionOnboardingFlowView` / `FlowViewModel`, view factory, and launcher that build entitlement-aware checklists, backfill VPN/PIR progress, and refuse launch on an empty checklist (with failure pixels). Purchase completion now fires `onPurchaseCompleted` to present onboarding once; settings passes onboarding KV store, PIR locale, and a Duck.ai handoff closure.
> 
> **Out-of-flow progress:** `SubscriptionOnboardingActivationRecorder` persists checklist completion when users enable **VPN** (first config install), **PIR** (profile saved), or **paid Duck.ai** outside the flow; experiment activation metrics fire from those hooks.
> 
> **Supporting:** New privacy-pro subfeatures, `SubscriptionManager.isOnFreeTrial()` / `isActiveSubscription()`, funnel pixels/instrumentation, progress fields (`postCheckoutFlowStartedAt`), entitlement-filtered checklist items, shared `PIRAvailability`, and debug-menu full-flow support.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a09fc98a603d1ecc5a6c035cb4cae27b9d6185d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/Subscription/Onboarding/Models/SubscriptionOnboardingActivationRecorder.swift
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    func recordPIRActivated() {
+        markComplete(.pir)
+    }
+
+    private func markComplete(_ item: SubscriptionOnboardingChecklistItem) {
+        var persistor = SubscriptionOnboardingProgressPersistor(keyValueStore: keyValueStore)
+        persistor.markComplete(item)
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: No need. It will show vpn connected in flow and record the completion at the moment. Users do not need to re-activate the vpn to get completion status

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingFlowViewModel.swift
```diff
@@ -0,0 +1,247 @@
...
+            // An incomplete VPN means the customer skipped.
+            if !progress.completedItems.contains(.vpn) {
+                instrumentation.stepSkipped(.vpnActivation)
+            }
+        case .duckAI:
+            // Advancing means user skipped the step
+            instrumentation.stepSkipped(.duckAI)
+        default:
+            break
+        }
```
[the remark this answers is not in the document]
hanyutang-sandra: I think this is expected and can happen in rare case. They did skip it in the end but their intention was to chat with DuckAI

iOS/DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -598,6 +600,7 @@ final class DefaultSubscriptionPagesUseSubscriptionFeature: SubscriptionPagesUse
             if let preference = pendingScheduleNotification, let scheduler = expirationReminderScheduler {
                 await scheduler.scheduleReminder(daysBeforeCancel: preference.daysBeforeCancel)
             }
+            onPurchaseCompleted?()
```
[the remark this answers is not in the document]
hanyutang-sandra: No stripe flow on iOS

iOS/DuckDuckGo/Subscription/Views/SubscriptionFlowView.swift
```diff
@@ -199,6 +219,32 @@ struct SubscriptionFlowView: View {
...
+        guard let persistor = viewModel.onboardingPersistor else { return }
+        Task { @MainActor in
+            guard let flow = await SubscriptionOnboardingFlowViewModel.postCheckout(
+                persistor: persistor,
+                isPIRAvailable: viewModel.isPIRAvailable,
+                subscriptionManager: viewModel.subscriptionManager,
+                onFinish: { onboardingFlow = nil },
+                pirScreen: { pirDestination }) else { return }
+            onboardingFlow = OnboardingFlowPayload(flow: flow)
+        }
```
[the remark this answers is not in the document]
hanyutang-sandra: I think this is fine. If this is latched the onboarding flow can also be entered from settings

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingFlowViewModel.swift
```diff
@@ -0,0 +1,278 @@
...
+extension SubscriptionOnboardingFlowViewModel: SubscriptionOnboardingSectionDelegate {
+
+    func sectionDidComplete(_ section: SubscriptionOnboardingSection) {
+        if case .activation(let item) = section.kind {
+            guard !progress.completedItems.contains(item) else { return }
+            progress.markComplete(item)
+        }
+        // `.vpnTips` completes nothing of its own
+        guard section != .vpnTips else { return }
+        instrumentation.stepCompleted(section)
```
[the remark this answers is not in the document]
hanyutang-sandra: This is only for instrumentation, and we bundle vpnTips and vpn Widget together so no separate vpn tip pixel will be fired

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1436,19 +1450,37 @@ extension SettingsViewModel {
...
+    /// Backfill only: profiles saved from now on record themselves via `BrokerProfileJobEventsHandler.onProfileSaved`.
+    private func recordPIRActivationIfNeeded() {
+        guard isPIRActivated else { return }
+        SubscriptionOnboardingActivationRecorder(keyValueStore: keyValueStore).recordPIRActivated()
+    }
+
+    private func reportVPNActivatedExperimentMetricIfNeeded() {
+        guard state.networkProtectionConnected else { return }
+        SubscriptionOnboardingExperiment.fireVPNActivatedMetric(isSubscriptionActive: state.subscription.hasActiveSubscription)
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: Actually we only care about in flow activation

iOS/DuckDuckGo/Subscription/Views/SubscriptionFlowView.swift
```diff
@@ -199,6 +219,33 @@ struct SubscriptionFlowView: View {
...
+        guard let persistor = viewModel.onboardingPersistor else { return }
+        Task { @MainActor in
+            guard let flow = await SubscriptionOnboardingFlowViewModel.postCheckout(
+                persistor: persistor,
+                isPIRAvailable: viewModel.isPIRAvailable,
+                subscriptionManager: viewModel.subscriptionManager,
+                onFinish: { onboardingFlow = nil },
+                pirScreen: { pirDestination }) else { return }
+            viewModel.didPresentOnboarding()
+            onboardingFlow = OnboardingFlowPayload(flow: flow)
```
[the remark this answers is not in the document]
hanyutang-sandra: The purchase can be done in multiple places so we cannot pop to settings and then launch the flow. The subscription onboarding sheet is presented immediately after purchase is successful so the window is extremely small. Even if the user is able to click back, they cannot dismiss the subscription flow v by swiping down and there is no back button at the root either, and as long as the subscription flow view is present, the onboarding will show. This is a false positive

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingLauncher.swift
```diff
@@ -0,0 +1,119 @@
...
+                                              isPIRAvailable: Bool,
+                                              subscriptionManager: any SubscriptionManager,
+                                              onFinish: @escaping () -> Void,
+                                              vpnController: SubscriptionOnboardingVPNControlling = DefaultSubscriptionOnboardingVPNController(),
+                                              @ViewBuilder pirScreen: @escaping () -> PIRScreen) async
+    -> SubscriptionOnboardingFlowViewModel? {
+        var persistor = persistor
+        if await vpnController.isVPNConfigured() {
+            persistor.markComplete(.vpn)
+        }
```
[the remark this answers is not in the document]
hanyutang-sandra: This is a false positive. PIR completion is recorded the moment a profile is saved, and that state is persisted on the user's device. At the start of flow construction, completion status is freshly read from that same store, so it already reflects PIR completion.

However, there is a narrow edge case where the user saved a profile before the hook is shipped. This can be covered by backfill again at postCheckout launch point

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -181,6 +182,7 @@ final class AIChatContextualSheetCoordinator {
          duckAiNativeStorageHandler: DuckAiNativeStorageHandling? = nil,
          duckAiFireModeStorageHandler: DuckAiNativeStorageHandling? = nil,
          debugSettings: AIChatDebugSettingsHandling = AIChatDebugSettings(),
+         onboardingActivationRecorder: SubscriptionOnboardingActivationRecording = NullSubscriptionOnboardingActivationRecorder(),
```
alessandroboron: What's the reasoning behind using `NullSubscriptionOnboardingActivationRecorder`? In production we want to record while in test we want to assert the behaviour works as expected.

Could we drop the defaults so the real recorder has to be injected, and have the tests pass a mock instead?

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualWebViewController.swift
```diff
@@ -158,6 +158,7 @@ final class AIChatContextualWebViewController: UIViewController {
          pixelHandler: AIChatContextualModePixelFiring,
          debugSettings: AIChatDebugSettingsHandling = AIChatDebugSettings(),
          userAgentManager: UserAgentManaging = DefaultUserAgentManager.shared,
+         onboardingActivationRecorder: SubscriptionOnboardingActivationRecording = NullSubscriptionOnboardingActivationRecorder(),
```
alessandroboron: See above comment https://github.com/duckduckgo/apple-browsers/pull/6321/changes#r3932187177

iOS/DuckDuckGo/AIChat/AIChatContentHandler.swift
```diff
@@ -196,6 +198,8 @@ final class AIChatContentHandler: AIChatContentHandling {
          featureDiscovery: FeatureDiscovery,
          productSurfaceTelemetry: ProductSurfaceTelemetry,
          freeTrialConversionService: FreeTrialConversionInstrumentationService = AppDependencyProvider.shared.freeTrialConversionService,
+         onboardingActivationRecorder: SubscriptionOnboardingActivationRecording = NullSubscriptionOnboardingActivationRecorder(),
```
alessandroboron: See above comment https://github.com/duckduckgo/apple-browsers/pull/6321/changes#r3932187177

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
```diff
@@ -67,6 +67,7 @@ protocol DependencyProvider {
     var subscriptionExpirationReminderScheduler: SubscriptionExpirationReminderScheduling { get }
     var dbpSettings: DataBrokerProtectionSettings { get }
     var syncAutoRestoreDecisionManager: SyncAutoRestoreDecisionManaging { get }
+    var subscriptionOnboardingSession: SubscriptionOnboardingSessionStateManaging { get }
```
alessandroboron, line 72: Do we need to access subscriptionOnboardingSession from other part of the app or just from `DBPService`? If only in DBPService we can avoid creating this reference and injecting it directly from `Launching` where DBPService is created.

iOS/DuckDuckGo/Subscription/Onboarding/Features/OrderConfirmation/SubscriptionOnboardingOrderConfirmationViewModel.swift
```diff
@@ -120,6 +120,28 @@ private extension SubscriptionOnboardingOrderConfirmationViewModel {
     }
 }
 
+// MARK: - Debug Menu
```
alessandroboron, line 123: Would it make sense to move this extension to private in the relevant DebugMenu section so there's no risk of using it outside of that scope (if possible)?

iOS/DuckDuckGo/Subscription/Onboarding/Pixels/SubscriptionOnboardingInstrumentation.swift
```diff
@@ -0,0 +1,112 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import PixelKit
+import AIChat
+
+/// Reports the onboarding funnel.
+@MainActor
```
alessandroboron: Any particular reason why we need to fire pixel on the main thread?

iOS/DuckDuckGo/Subscription/Onboarding/Pixels/SubscriptionOnboardingInstrumentation.swift
```diff
@@ -0,0 +1,112 @@
...
+        fire(.subscriptionOnboardingStepSkipped(step: section.pixelStepName, entryPoint: entryPoint.pixelValue))
+    }
+
+    private func fire(_ pixel: SubscriptionPixel) {
+        pixelFiring?.fire(pixel)
+    }
+}
+
+/// Reports nothing, for previews and tests.
+struct NullSubscriptionOnboardingInstrumentation: SubscriptionOnboardingInstrumenting {
```
alessandroboron: See https://github.com/duckduckgo/apple-browsers/pull/6321/changes#r3932187177

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/EntitlementStatus.swift
```diff
@@ -58,6 +58,15 @@ public struct EntitlementStatus: Equatable, CustomDebugStringConvertible {
                           paidAIChat: false)
     }
 
+    /// Returns a status with all entitlements enabled
+    public static var allEnabled: EntitlementStatus {
```
alessandroboron: EntitlementStatus.allEnabled reads as "this customer has paid for everything", and it's public in BSK. Could we find another way if used in previews or tests not to expose this to the main target unless is available only via #if DEBUG?

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -3882,12 +3882,26 @@ class MainViewController: UIViewController {
...
+                dbpIOSPublicInterface?.meetsLocaleRequirement ?? false
+            },
+            onRequestDuckAIChat: { [weak self] modelID in self?.requestOnboardingDuckAIChat(modelID: modelID) ?? false }
         ))
         viewController.view.backgroundColor = UIColor(designSystemColor: .surface)
         return viewController
     }
 
+    /// Dismisses whatever's presented, then opens Duck.ai chat from the subscription onboarding flow.
+    /// - Returns: `false` if there's nothing to dismiss into, so a caller can fall back.
```
alessandroboron: `Returns: false if there's nothing to dismiss into` doesn't match the body, it always returns true

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingViewFactory.swift
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    func screen(for section: SubscriptionOnboardingSection) -> AnyView {
+        AnyView(content(for: section)
+            .onFirstAppear { reportShown(section) })
+    }
+
+    /// `.vpnTips` is bundled with `.vpnWidget` and shares its pixel name — firing here too would
+    /// double-count a single "vpn_widget shown" impression.
+    func reportShown(_ section: SubscriptionOnboardingSection) {
```
alessandroboron, line 56: nip: `reportShown` can be private?

iOS/DuckDuckGo/Subscription/Views/SubscriptionContainerViewFactory.swift
```diff
@@ -198,7 +211,9 @@ enum SubscriptionContainerViewFactory {
                                                        userScript: SubscriptionPagesUserScript(),
                                                        userScriptsDependencies: userScriptsDependencies,
                                                        subFeature: subscriptionPagesUseSubscriptionFeature,
-                                                       dataBrokerProtectionViewControllerProvider: dataBrokerProtectionViewControllerProvider)
+                                                       dataBrokerProtectionViewControllerProvider: dataBrokerProtectionViewControllerProvider,
+                                                       onboardingKeyValueStore: nil,
+                                                       meetsPIRLocaleRequirement: { false })
```
alessandroboron, line 217: just double checking this should hardcoded to `false` here?

iOS/DuckDuckGo/Subscription/Views/SubscriptionSettingsView.swift
```diff
@@ -650,3 +669,75 @@ private var resubscribeWithWinBackOfferView: some View {
...
 }
+
+// MARK: - Onboarding
+
+extension SubscriptionSettingsViewV2 {
+
+    /// The "Continue Setup" re-entry card.
+    var onboardingSetupSection: some View {
+        Section {
+            if let onboardingProgress {
```
alessandroboron: if `onboardingProgress` is nil we show an empty section? should this let be at caller site?

```
if isOnboardingEnabled, onboardingProgress {
  onboardingSetupSection
}

```

iOS/DuckDuckGo/Subscription/Views/SubscriptionSettingsView.swift
```diff
@@ -650,3 +669,75 @@ private var resubscribeWithWinBackOfferView: some View {
...
+                persistor: onboardingPersistor,
+                isPIRAvailable: isPIRAvailable,
+                subscriptionManager: settingsViewModel.subscriptionManager,
+                onFinish: { onboardingFlow = nil },
+                onRequestDuckAIChat: settingsViewModel.onRequestOnboardingDuckAIChat,
+                pirScreen: { pirDestination }) else { return }
+            onboardingFlow = flow
+        }
+    }
+
```
alessandroboron, line 705: Should `onboardingProgress` / `onboardingPersistor` / `isOnboardingEnabled` (:708–:723) live on SubscriptionSettingsViewModel? Two reasons:
1. They're deriving domain state in the view, which checklist this customer has, whether the card should show, whether they're enrolled in the experiment, and there's no way to test any of it in the view. `SubscriptionSettingsViewModel` already has a test file.

2. Also seem we do the work at every body recomputation. We create a new  SubscriptionOnboardingProgressPersistor which does the lookup for feature flags etc.

Is it possible to have the view model expose a Published property that we observe that changes based on some events?

iOS/DuckDuckGo/Subscription/Views/SubscriptionFlowView.swift
```diff
@@ -165,6 +159,16 @@ struct SubscriptionFlowView: View {
         .onChange(of: viewModel.state.shouldGoBackToSettings) { _ in
             dismiss()
         }
+
+        .onChange(of: viewModel.state.shouldPresentOnboarding) { shouldPresent in
+            if shouldPresent {
+                startOnboarding()
```
alessandroboron: any chance that the view could be dismissed while the flow is suspended in the `await` in `startOnboarding()`? If so `viewModel.didPresentOnboarding()` would still be called.

should we use 

```
.task(id: viewModel.state.shouldPresentOnboarding) {
    guard viewModel.state.shouldPresentOnboarding else { return }
    await startOnboarding()
}
```

and latch  on view first appear?

.```
sheet(item: $onboardingFlow, onDismiss: { viewModel.onboardingFinished() }) { flow in
    SubscriptionOnboardingLauncher.launch(flow: flow)
        .onFirstAppear { viewModel.didPresentOnboarding() } 
}
```

File: iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingFlowViewModel.swift [hunk 23]
```diff
@@ -0,0 +1,288 @@
...
+            self.sequence = Self.makePostCheckoutSequence(checklist: self.progress.checklist,
+                                                          completedItems: self.progress.completedItems)
+        case .subscriptionSettings:
+            self.sequence = Self.makeSubscriptionSettingsSequence(checklist: self.progress.checklist,
+                                                                  completedItems: self.progress.completedItems)
+        }
+    }
+
+    /// Reports the flow's start, records the post-checkout started marker, and kicks off prefetching.
+    func flowDidStart() {
```
alessandroboron, line 129: `flowDidStart()` has no test  should we write tests for it? We could assert that `recordPostCheckoutFlowStartedIfNeeded` is firing for `.postCheckout` and not for `.subscriptionSettings`. The Settings card only appears if that marker is set, `isSettingsReEntryEnabled(hasStartedFlow:)` reads it,  so if the write ever stops happening, "Continue Setup" quietly disappears for everyone.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
hanyutang-sandra: In prod code a real `SubscriptionOnboardingActivationRecorder` is passed in `TabViewController`. `NullSubscriptionOnboardingActivationRecorder` is used as a fallback following the pattern above.  Removed it here but we still need to have it for the tests (`AIChatContextualSheetCoordinatorTests`, `AIChatContextualWebViewControllerTests`) because these classes have no tests touching activation logic.

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
hanyutang-sandra: Hmm I don’t think it has anything to do with `DBPService`. This is used to know the launch session so that we can remove the `SubscriptionOnboardingSetupCard` in the next one once the pct is at 100%. It’s used in Settings via `launchSettings` in `MainViewController + Segue` (https://github.com/duckduckgo/apple-browsers/pull/6321/changes#diff-980ce62fb1f666d05ac31546359aa7debd86a51bac334a23dbe7700fa8830a7aR536). But yeah I can inject it in Settings VM

iOS/DuckDuckGo/Subscription/Onboarding/Features/OrderConfirmation/SubscriptionOnboardingOrderConfirmationViewModel.swift
hanyutang-sandra: Hmm yeah it’s possible. The debug menu could build the debug-only VM itself and thread it down instead of an `Int`. But `forcingFreeTrial` also writes directly to `state`, which has a private setter. And moving it out of this file means either exposing that setter more broadly, or duplicating the logic elsewhere. I'm hesitant to loosen state's setter just for this.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/EntitlementStatus.swift
hanyutang-sandra: Moved it out of BSK and updated the name to `mockAllEnabled` to make it clear

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingViewFactory.swift
hanyutang-sandra: No it’s also used in tests

iOS/DuckDuckGo/Subscription/Views/SubscriptionContainerViewFactory.swift
hanyutang-sandra: Yes they should. The 3 call sites with it belong to restore flow, plan update flow and email flow, which has nothing to do with onboarding.

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingFlowViewModel.swift
hanyutang-sandra: Test added

iOS/DuckDuckGo/Subscription/Views/SubscriptionSettingsView.swift
hanyutang-sandra: Yeah it’s reasonable. Move them to vm instead

iOS/DuckDuckGo/Subscription/Views/SubscriptionFlowView.swift
hanyutang-sandra: Yeah good point. Added

iOS/DuckDuckGo/Subscription/Onboarding/Pixels/SubscriptionOnboardingInstrumentation.swift
hanyutang-sandra: Moved it to the test file but it’s still needed for stubbing the unit tests that do not test instrumentation. We have `SpyInstrumentation` for the tests focusing on instrumentation

iOS/DuckDuckGo/Subscription/Onboarding/Navigation/SubscriptionOnboardingFlowViewModel.swift
```diff
@@ -0,0 +1,288 @@
...
+    let pirScreen: () -> AnyView
+
+    /// Both edges of the PIR sheet, reported from the summary's own observation of `pirLaunch`
+    func reportPIRPresentation(_ isPresenting: Bool) {
+        if isPresenting {
+            instrumentation.stepShown(.pir)
+        } else if progress.completedItems.contains(.pir) {
+            instrumentation.stepCompleted(.pir)
+        }
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: This is false positive. Once the user has completed pir step, regardless of in flow or out of flow completion, they cannot reopen the pir sheet again.

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1416,19 +1434,51 @@ extension SettingsViewModel {
...
+        let persistor = SubscriptionOnboardingProgressPersistor(keyValueStore: keyValueStore)
+        guard !persistor.completedItems.contains(.vpn) else { return }
+        guard await vpnController.isVPNConfigured() else { return }
+        SubscriptionOnboardingActivationRecorder(keyValueStore: keyValueStore).recordVPNActivated()
+    }
+
+    /// Reads the live connection status rather than `state.networkProtectionConnected`, which isn't loaded
+    /// yet the first time this runs on `onFirstAppear`.
+    private func reportVPNActivatedExperimentMetricIfNeeded() {
+        guard vpnController.isConnected else { return }
```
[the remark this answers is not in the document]
hanyutang-sandra: Mark it pending for now because I need to clarify what does VPN activation mean. It could mean:
- VPN feature activated (aka VPN config installed)
OR
- VPN is activated (aka is Using VPN, VPN usage)

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionFlowViewModel.swift
```diff
@@ -172,7 +239,11 @@ final class SubscriptionFlowViewModel: ObservableObject {
                 self.setTransactionStatus(.idle)
             }
         }
-        
+
+        subFeature.onPurchaseCompleted = {
+            Task { await self.requestOnboardingIfNeeded() }
+        }
```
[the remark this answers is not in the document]
alessandroboron: @hanyutang-sandra `onBackToSettings (:237)`, `onActivateSubscription (:243)` and `onFeatureSelected (:255)` capture self strongly, and `subFeature` is also retained strongly on the VM so `self` → `subFeature` → `closure` → `self` is a cycle. deinit's `cleanUp()` can't break it because it won't be called if there's a cycle.

iOS/DuckDuckGo/SettingsViewModel.swift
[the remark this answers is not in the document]
hanyutang-sandra: Confirmed we will go with VPN feature activated (aka VPN config installed, first time activation)

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionFlowViewModel.swift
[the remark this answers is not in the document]
hanyutang-sandra: Thanks for bring it back to my attention. Don’t know why it’s not updated but it was updated in one previous commit to 
```
subFeature.onPurchaseCompleted = { [weak self] in
            guard let strongSelf = self else { return }
            Task { await strongSelf.requestOnboardingIfNeeded() }
        }
```
https://github.com/duckduckgo/apple-browsers/pull/6321/changes#diff-a8c9de88fa69dc2b242508652a89d5372bddde0dfd9534d14c36af31fc39c830R250

Yes this retain cycle is still happening for other neighboring callsites but I didn’t update those given the size of the PR is already big enough. But since we are on it lemme create an asana task for it to be fixed later

iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -1410,19 +1428,51 @@ extension SettingsViewModel {
...
+    }
+
+    /// Backfill only: a config being installed is considered vpn step completed. Already-complete customers
+    /// are filtered out above, so this only repeats the IPC round-trip for customers still incomplete.
+    private func recordVPNActivationIfNeeded() async {
+        let persistor = SubscriptionOnboardingProgressPersistor(keyValueStore: keyValueStore)
+        guard !persistor.completedItems.contains(.vpn) else { return }
+        guard await vpnController.isVPNConfigured() else { return }
+        SubscriptionOnboardingActivationRecorder(keyValueStore: keyValueStore).recordVPNActivated()
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: This is no longer needed for backfill, given we want to check if feature activation has a correlation with retention and we only care about feature activated after experiment started.

iOS/DuckDuckGo/Subscription/Views/SubscriptionFlowView.swift
```diff
@@ -199,6 +202,33 @@ struct SubscriptionFlowView: View {
...
+    private func startOnboarding() async {
+        guard viewModel.state.shouldPresentOnboarding, onboardingFlow == nil else { return }
+        guard let persistor = viewModel.onboardingPersistor else { return }
+        guard let flow = await SubscriptionOnboardingFlowViewModel.postCheckout(
+            persistor: persistor,
+            isPIRAvailable: viewModel.isPIRAvailable,
+            subscriptionManager: viewModel.subscriptionManager,
+            onFinish: { onboardingFlow = nil },
+            onRequestDuckAIChat: viewModel.onRequestDuckAIChat,
+            pirScreen: { pirDestination }) else { return }
```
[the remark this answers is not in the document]
hanyutang-sandra: Post subscription onboarding is a on-shot event after subscription purchase so if something is wrong post purchase there’s no way to retry it. imo we can add a pixel to monitor how frequent this happens and if there’s a lot (eg. entitlement read right after purchase), we can think of a way to retry. For the experiment tho it should be expected

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionFlowViewModel.swift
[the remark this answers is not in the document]
alessandroboron: Given they are retain cycle I would fix them as part of this PR before merging it rather than later :)

iOS/DuckDuckGo/Subscription/Onboarding/Models/SubscriptionOnboardingActivationRecorder.swift
```diff
@@ -0,0 +1,115 @@
...
+    var isPIRActivated: Bool { get }
+    /// Whether the VPN step was already recorded before this call.
+    var isVPNActivated: Bool { get }
+}
+
+extension SubscriptionOnboardingActivationRecording {
+    /// Records the activation and reports whether it was already recorded beforehand, so a caller that also
+    /// fires a one-time experiment metric can guard it without risking the read/write ordering itself.
+    @discardableResult
+    func recordDuckAIActivatedIfNeeded() -> Bool {
```
alessandroboron, line 47: nip: `ifNeeded` not really correct in this case because we always record it. Same for `recordPIRActivatedIfNeeded` and `recordVPNActivatedIfNeeded`.

iOS/DuckDuckGo/NetworkProtectionTunnelController.swift
```diff
@@ -448,9 +454,10 @@ final class NetworkProtectionTunnelController: VPNConnectionContextProvidingTunn
             let tunnelManager = NETunnelProviderManager()
             try await setupAndSave(tunnelManager)
             internalManager = tunnelManager
+            configurationInstalledSubject.send()
```
alessandroboron, line 442: setupAndSave can still fail e.g. if the user does not give permission. do we consider that as configuration installed? Probably better to call the subject after `try await tunnelManager.saveToPreferences()`?

alessandroboron (approved):
@hanyutang-sandra I left some comments to the new code. I'll be on leave as of Tomorrow so I won't block your PR. Trusting you will address the comments I'm approving the PR

iOS/DuckDuckGo/Subscription/Onboarding/Models/SubscriptionOnboardingActivationRecorder.swift
hanyutang-sandra: they are leftover namings from previous iterations. Fixed

iOS/DuckDuckGo/NetworkProtectionTunnelController.swift
hanyutang-sandra: No in this case it’s not considered a config installed. If the user denied it `setupAndSave` will throw `StartError.configSystemPermissionsDenied` so `configurationInstalledSubject.send()` is not reached. 

If it’s moved into `try await tunnelManager.saveToPreferences()` inside setupAndSave, it will fire not only from first activation but also the following activations, but we only want to measure first time activation here.

iOS/DuckDuckGo/Subscription/ViewModel/SubscriptionFlowViewModel.swift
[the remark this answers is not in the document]
hanyutang-sandra: cool. Fixed all.

iOS/DuckDuckGo/MainViewController.swift
```diff
@@ -3948,12 +3948,26 @@ class MainViewController: UIViewController {
...
     }
 
+    /// Dismisses whatever's presented, then opens Duck.ai chat from the subscription onboarding flow.
+    /// - Returns: Always `true`; a `weak self` caller sees `false` only once `self` is deallocated.
+    func requestOnboardingDuckAIChat(modelID: String?) -> Bool {
+        dismiss(animated: true) {
+            self.openAIChat(source: .onboarding, flowType: .mobileAppOnboarding, modelId: modelID)
+        }
+        return true
+    }
```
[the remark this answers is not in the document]
hanyutang-sandra: The actual race here is return to settings vs. dismiss called on MainViewController. It’s working fine in practice but I removed the back to setting call just to be sure

iOS/PixelDefinitions/pixels/definitions/subscription.json5
```diff
@@ -943,5 +1036,21 @@
...
+        "parameters": [
+            "appVersion",
+            "pixelSource",
+            {
+                "key": "reason",
+                "type": "string",
+                "description": "Why the flow never launched",
+                "enum": ["missingPersistor", "emptyChecklist"]
+            }
+        ]
```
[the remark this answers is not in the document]
hanyutang-sandra: No this is not a temporary diagnostic pixel. Similar to error pixels, we did not include retry for the experiment, and this pixel provides observability into it so that we know that if there are more failures than expected, then we need to invest in retry mechanisms

PR #6334, opened 2026-08-13
CI: Print macOS version at the start of every macOS job

ayoy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203301625297703/task/1217455483209825?focus=true

### Description

Every CI job that runs on a macOS machine now prints the OS version of that machine, so the logs show which macOS version a build or test actually ran on. Until now you had to infer it from the runner label, which is not reliable — labels like `macos-latest` move over time, and the self-hosted and Bitrise machines can be updated without any change in this repository.

The step runs directly after the checkout step. It cannot run before it: most of these workflows set `defaults.run.working-directory` to `macOS` or `iOS`, and that directory does not exist until the repository is checked out, so bash cannot start there.

### Testing Steps
1. Verify that logs show macOS version being used by the runner.

### Impact

None — this is internal tooling. The new step only prints information. It does not change how the apps are built, signed or tested, and it does not touch app code.

### Quality Considerations

- The change is only insertions: 208 added lines, no deleted lines. Nothing that exists today was moved or re-indented.
- `actionlint -shellcheck=` — the same command the `Actionlint` workflow runs — passes on all workflow files.
- I checked every job of every workflow, not only the ones with a `macos-*` label. The runner is given indirectly in five jobs, through `matrix.runner`, `matrix.runs-on` or `inputs.runner`; every value those can take is a macOS runner.
- Three jobs check out twice under mutually exclusive conditions (`pull_request`/`push` against everything else). The step goes after the second checkout in those jobs, so it runs after whichever one applied.
- The Mattermost job of the variants workflow checks out only when the jobs it needs did not cancel. The step carries that same condition, like every other step of that job, so a cancelled run does not fail on it.
- One job has no checkout and sets `working-directory: .`, which always exists, so its step stays at the top.
- Seven jobs use a sparse checkout. Each sparse list includes paths below that job's working directory, so the directory is created.
- `sw_vers` needs no network and no credentials, and it adds well under a second per job.

---
###### Internal references:

### DIFF
--- a/.github/workflows/bsk_pr_checks.yml
+++ b/.github/workflows/bsk_pr_checks.yml
@@ -49,6 +49,9 @@ jobs:
 
       steps:
 
+      - name: Print macOS version
+        run: sw_vers
+
       - name: Register SSH keys for private repos
         uses: webfactory/ssh-agent@v0.10.0
         with:
@@ -207,6 +210,9 @@ jobs:
         WORKFLOW_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}/attempts/${{ github.run_attempt }}
 
       steps:
+
+      - name: Print macOS version
+        run: sw_vers
 
       - name: Register SSH keys for private repos
         uses: webfactory/ssh-agent@v0.10.0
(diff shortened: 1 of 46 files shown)


federicocappelli (approved):
sw_vers galore!
I haven't tested it but the code changes are CI-only and very low risk.
Claude also gave thumb up.
LGTM

PR #6397, opened 2026-08-18
Fix crash & keep the tab bar's item count in sync when an insert re-enters

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214294661819890/task/1217529313084325?focus=true

### Description

A page opening a tab while another one is still opening can crash the window: the tab bar animates a single insertion while the tab list has already grown by two. Notifying the tab bar before the new selection publishes closes the window where that happens.

### Testing Steps

1. Open around twenty tabs so the tab bar enters overflow mode.
2. Press Cmd-T.
   The new tab appears at the full selected width, with no flicker.
3. Click the "+" button.
   The new tab appears at the full selected width, with no flicker.
4. Right-click a link and choose Open Link in New Tab.
   The new tab appears at the full selected width and is selected.
5. Turn off Switch to new tab when opened in Settings, then Cmd-click a link.
   The previously selected tab keeps its width and the new tab is at the unselected width.
6. Right-click a bookmark folder and choose Open All in Current Window.
   The last opened tab ends up selected at the correct width.
7. Right-click a tab and choose Duplicate Tab.
   The copy appears next to the original at the selected width and is selected.
8. Drag a tab out of the window and drop it onto another window.
   The tab arrives selected at the correct width in the destination window.
9. Open https://webbrowsertools.com/popup-blocker/ and allow pop-ups for the page.
   Several tabs open in sequence and the app does not crash.
10. Right-click a tab, choose Pin Tab, then right-click it again and choose Unpin Tab.
    The tab moves between the pinned and unpinned areas at the correct widths.
11. Relaunch with a saved many-tab session.
    The restored selected tab is at the correct width.

### Impact

Medium. The change touches the ordering between the tab model and the tab bar, so a mistake would show up as a wrong tab width, a selection landing on the wrong tab, or the crash it is meant to fix.

#### What could go wrong?

- Another re-entrant path could still reach the tab bar mid-update → the tab bar now compares its own item count against the model before animating and falls back to a full reload, so the worst case is a lost animation and a logged error rather than a crash.
- The incoming tab could be measured against a selection the view model has not applied yet → the tab bar uses the position and selection state passed to it for the duration of the insert, and steps 2 to 8 cover the visible widths.
- Opening several tabs at once could paint the selection on the wrong tab for a frame → the existing selection subscription reconciles it on the next pass; step 6 covers it.

### Quality Considerations

A unit test reproduces the crash by inserting a tab from a selection subscriber and asserting that the counts reported to the tab bar always add up. It fails without this change with the same numbers seen in the crash reports. The tab widths cannot be pinned the same way, because the tab bar view controller is not instantiable in the unit test target, so steps 2 to 8 are the check for those.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes ordering between tab model selection and tab bar updates—a regression could cause wrong selection, tab widths, or reintroduce the collection view crash; mitigated by count checks and full reload fallback.
> 
> **Overview**
> Fixes crashes when a tab opens while another is still opening (e.g. queued pop-ups): the tab bar could animate one insert while the model had already grown by two, triggering `NSInternalInconsistencyException`.
> 
> **Model → tab bar ordering:** On insert, append, batch append, duplicate, and cross-window move, `TabCollectionViewModel` now calls the delegate **before** updating `selectionIndex`. Selection publishes `selectedTabViewModel`, and subscribers can synchronously add another tab; notifying the tab bar first keeps the collection view’s item count aligned with the model.
> 
> **Tab bar defenses:** `TabBarViewController` compares the collection view’s item count to the model before `insertItems`; on mismatch it logs and uses `reloadData()` instead of an incremental insert. **`incomingSelectionIndex`** lets `sizeForItemAt` treat the in-flight insert as selected during the synchronous layout pass before `selectionIndex` updates, avoiding width flicker.
> 
> Unit tests assert delegate-before-selection ordering and re-entrant insert via a count-mirroring delegate mock.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 073c77bd263a3fdf2c3fff08dbf63d73fd9d835e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
+++ b/macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
@@ -1545,6 +1545,10 @@ final class TabBarViewController: NSViewController, TabBarRemoteMessagePresentin
     }
 
     private var frozenLayout = false
+    /// Selection an in-flight insert is about to apply. `sizeForItemAt` runs synchronously
+    /// from `insertItems(at:)`, before the view model publishes the new `selectionIndex`, so
+    /// it consults this to size the inserted tab as selected and the outgoing one as not.
+    private var incomingSelectionIndex: TabIndex?
     @Published private var tabMode = TabMode.divided
 
     private func updateTabMode(for numberOfItems: Int? = nil, updateLayout: Bool? = nil) {
@@ -1951,8 +1955,22 @@ extension TabBarViewController: TabCollectionViewModelDelegate {
         let indexPathSet = Set(arrayLiteral: IndexPath(item: index.item))
         if selected {
             clearSelection(animated: true)
+            incomingSelectionIndex = index
         }
-        collectionView.animator().insertItems(at: indexPathSet)
+        defer { incomingSelectionIndex = nil }
+
+        // A mutation re-entering between the model update and this callback leaves the
+        // collection view's item count stale; an incremental insert would then raise
+        // NSInternalInconsistencyException, so reconcile with a full reload instead.
+        let modelCount = index.isPinnedTab
+            ? (tabCollectionViewModel.pinnedTabsCollection?.tabs.count ?? 0)
+            : tabCollectionViewModel.tabCollection.tabs.count
+        if collectionView.numberOfItems(inSection: 0) == modelCount - 1 {
+            collectionView.animator().insertItems(at: indexPathSet)
+        } else {
+            Logger.general.error("TabBarViewController: item count out of sync on insert, reloading")
+            collectionView.reloadData()
+        }
         if selected {
             collectionView.selectItems(at: indexPathSet, scrollPosition: .centeredHorizontally)
             collectionView.scrollToSelected()
@@ -2073,6 +2091,21 @@ extension TabBarViewController: TabCollectionViewModelDelegate {
 
         if selected {
             clearSelection()
+            incomingSelectionIndex = .unpinned(lastIndex)
+        }
+        defer { incomingSelectionIndex = nil }
+
+        // See `tabCollectionViewModelDidInsert(_:at:selected:)` on the count check.
+        guard collectionView.numberOfItems(inSection: 0) == tabCollectionViewModel.tabCollection.tabs.count - 1 else {
+            Logger.general.error("TabBarViewController: item count out of sync on append, reloading")
+            collectionView.reloadData()
+            if selected {
+                collectionView.selectItems(at: lastIndexPathSet, scrollPosition: .centeredHorizontally)
+            }
+            scrollCollectionViewToEnd()
+            updateEmptyTabArea()
+            hideTabPreview()
+            return
         }
 
         if tabMode == .divided {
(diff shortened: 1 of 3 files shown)


macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
```diff
@@ -2156,7 +2189,11 @@ extension TabBarViewController: NSCollectionViewDelegateFlowLayout {
         guard collectionView != pinnedTabsCollectionView else {
             return NSSize(width: pinnedTabWidth, height: pinnedTabHeight)
         }
-        let isItemSelected = tabCollectionViewModel.selectionIndex == .unpinned(indexPath.item)
+        // During an insert the view model's `selectionIndex` is not updated yet — the
+        // delegate is notified first so the item count stays in sync — so prefer the
+        // in-flight selection when there is one.
+        let selectionIndex = incomingSelectionIndex ?? tabCollectionViewModel.selectionIndex
```
diegoreymendez, line 2201: We're reverting most of the code, except we're adding this to fix the issue the previous change was trying to resolve.

macOS/DuckDuckGo/TabBar/View/TabBarViewController.swift
```diff
@@ -2073,6 +2091,21 @@ extension TabBarViewController: TabCollectionViewModelDelegate {
...
+        defer { incomingSelectionIndex = nil }
+
+        // See `tabCollectionViewModelDidInsert(_:at:selected:)` on the count check.
+        guard collectionView.numberOfItems(inSection: 0) == tabCollectionViewModel.tabCollection.tabs.count - 1 else {
+            Logger.general.error("TabBarViewController: item count out of sync on append, reloading")
+            collectionView.reloadData()
+            if selected {
+                collectionView.selectItems(at: lastIndexPathSet, scrollPosition: .centeredHorizontally)
+            }
+            scrollCollectionViewToEnd()
```
mallexxx: This scroll call behaves incorrectly, scrolling to the end of the overflown collection view even if the inserted tab is in the beginning of it – scrolling it out of sight. It also duplicates the scroll called for the selected item in the preceding `if`.
I guess we should call `scroll(to: IndexPath(item: index.item))` here and make it in an `else` branch of the `if`.
I was testing the reentrant insertion using your example:
```
        cancellable = tabCollectionViewModel.$selectedTabViewModel
            .sink { [weak tabCollectionViewModel] selectedTabViewModel in
                tabCollectionViewModel?.insert(Tab(content: .none, parentTab: nil), at: .unpinned(tabCollectionViewModel!.tabCollection.tabs.count), selected: false)
            }
```
- it was causing scrolling inserted/selected tabs out of sight on each tab insertion/selection

macOS/DuckDuckGo/TabBar/ViewModel/TabCollectionViewModel.swift
```diff
@@ -492,11 +493,12 @@ final class TabCollectionViewModel: NSObject {
         }
 
         tabCollection.append(tabs: tabsToAppend)
+        // Notify the delegate before updating selection — see `insert(_:at:selected:)`.
+        delegate?.tabCollectionViewModelDidMultipleChanges(self)
         if shouldSelectLastTab {
             let newSelectionIndex = tabCollection.tabs.count - 1
             selectUnpinnedTab(at: newSelectionIndex)
         }
-        delegate?.tabCollectionViewModelDidMultipleChanges(self)
```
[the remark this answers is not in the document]
mallexxx: @diegoreymendez, the comment seems correct: when opening multiple tabs (open all bookmarks in current window) - the selection blinks during the animation

mallexxx (approved):
Validated the crash is gone.
Although I've encountered couple of issues during testing (see my comments) - would be good to fix before merging

mallexxx (approved):
LGTM

PR #6398, opened 2026-08-18
iOS: Contextual Duck.ai surface polish and handover fixes

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215757651854873?focus=true
Tech Design URL:
CC:

### Description

Follow-up polish on the contextual Duck.ai surfaces introduced in #6283, all behind `aiChatContextualFloatingInput`.

The sheet's chrome is reworked to the newer design: glassy interactive header pills, a Duck.ai bubble-and-star title icon, and a hand-off button that opens the conversation in a Duck.ai tab. The chats menu gains an "Open Duck.ai" row and now opens even with no chat history, where it previously jumped straight to the history page and left both of its actions unreachable.

The input above the chat is now a plain single-line pill whenever there is no keyboard, and it grows only while composing. It collapses when the keyboard goes — including when a sheet drag takes the keyboard without releasing the field — but not for the phantom hides a hardware keyboard emits on every keystroke.

"New Chat" returns to the floating input rather than the sheet's own start surface, and submitting from the floating input reliably promotes to the sheet that owns the conversation: previously the promotion could be skipped entirely, or open a second sheet whose empty web view was not the one receiving the answer. Dictation opens from whichever surface is showing the input, where the microphone previously did nothing on the floating input. A tap on the page dismisses the floating input instead of also following the link it landed on, and the input keeps the app in portrait while it is up.

Not the full A2 design — the designs changed during the branch and this reflects the latest direction, not a completed slice.

### Testing Steps

Enable `aiChatContextualFloatingInput`.

1. On a web page, tap the Duck.ai icon in the address bar → menu opens and the address bar stays visible behind it.
2. Choose "Ask About Page" → floating input appears over the page with suggestions.
3. Tap the microphone without typing → dictation opens.
4. Tap a link on the page → the input dismisses and the link is *not* followed. Scroll the page → it still scrolls.
5. Rotate to landscape while the input is up → stays portrait. Dismiss it → landscape works again.
6. Rotate to landscape first, then "Ask About Page" → rotates to portrait and the input stays up.
7. Submit a prompt → the sheet opens onto the chat, with your prompt and the answer in it.
8. With no keyboard, confirm the input is a single-line pill. Tap it → it expands; drag the sheet down to half → it collapses again.
9. Open the chats menu → New Chat → you land back on the floating input; submit again → the sheet returns with that prompt.
10. Chats menu → "Open Duck.ai" → a Duck.ai tab opens, leaving this chat alone.
11. On a tab with no chat history, open the chats menu → it opens and shows "Open Duck.ai" rather than jumping to all chats.

### Impact

Medium — several contextual Duck.ai flows change, all behind `aiChatContextualFloatingInput`.

#### What could go wrong?

- A submission could be reported twice, once by the input and once by the frontend's acknowledgement → mitigated by a single one-shot gate both paths share.
- The address-bar button could stop matching what a tap does after a cold launch → mitigated by binding its state from the omnibar refresh every tab-change route passes through, rather than per-route.
- A hardware keyboard's phantom keyboard-hide could collapse the input mid-typing → mitigated by only collapsing when a keyboard was actually shown.

### Quality Considerations

The address bar blanking while its menu was open turned out to be a UIKit behaviour on iOS 26: `UITargetedPreview` makes UIKit reparent the preview view into the menu platter, and its default choice — the button — sits inside the field's glass group, so removing it stops the group rendering. The button now offers a throwaway stand-in outside the glass. Worth a Feedback to Apple; if they fix it, the stand-in becomes a harmless no-op.

Tests were not run locally: the Xcode MCP test runner failed on every invocation this session, including identifiers taken verbatim from its own test list, so this relies on CI.

### Known follow-ups

Deferred deliberately, none blocking:

1. **Keyboard visibility reaches only the omnibar's UTI coordinator.** The contextual surface works around it with its own `keyboardWillHide` observer, so three observers with three guard sets now exist on this path. The two paths' guards have since diverged on purpose — the omnibar bails while the field holds first responder (a hardware keyboard's phantom hide), the contextual input must not (a sheet drag). Unifying them wants a real discriminator, and touches the omnibar's live keyboard path.
2. **Contextual voice-search availability is read once at construction**, so the microphone goes stale if speech availability flips while the sheet is open.
3. **Drag-to-dismiss fires on upward scroll** (pre-existing, from #5676) — `abs(velocity.y)` means scrolling up to re-read dismisses the keyboard as readily as swiping down.

Three review findings were investigated and left alone as unreproduced or not worth the risk: a chips-row fade race needing an empty-then-populated resolve inside 200ms, a pose change during the floating input's exit slide, and a promoted *retained* sheet opening at half then animating to full (only reachable when that sheet was first opened directly rather than by promotion).

---

> [!NOTE]
> **Medium Risk**
> Touches contextual Duck.ai presentation, UTI state, and keyboard/focus handover behind `aiChatContextualFloatingInput`. Bugs here can drop chats, double-submit, or mis-dismiss the floating surface.
> 
> **Overview**
> Follow-up polish for contextual Duck.ai (behind `aiChatContextualFloatingInput`): the sheet header uses glass pills, swapped close/hand-off, and a native chats menu instead of the custom overlay popup. The menu always opens (even with no history) and adds **Open Duck.ai**, which leaves the current chat intact; expand/hand-off still moves the conversation to a tab and clears this tab’s session (`moved_to_tab`).
> 
> The unified input on this surface can now **collapse to a pill** when the keyboard goes away (not on hardware-keyboard phantom hides or attachment pickers). New Chat returns to the floating input; promotion reuses the retained sheet so the prompt lands in the web view that owns the chat. Dictation is hosted by whichever surface shows the input.
> 
> Floating-input taps on the page dismiss without following the link; the app stays portrait while it is up. Address-bar Duck.ai menus use a throwaway highlight target so iOS 26 glass chrome does not blank. A `chevronCircleDown` glyph replaces the page favicon on “Ask about page.”
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bea8d8dc22f414d79c322638f7e8ccee4ae79a45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualFloatingInputViewController.swift
```diff
@@ -507,9 +562,30 @@ extension AIChatContextualFloatingInputViewController: UIGestureRecognizerDelega
...
     }
 
+    /// A page tap is the user leaving, so the page's own tap has to wait for ours and lose — otherwise the
+    /// link it landed on opens behind the dismissal. Pans are left alone, so the page still scrolls.
+    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer,
+                           shouldBeRequiredToFailBy otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        guard gestureRecognizer === dismissOnPageTapRecognizer,
+              !(otherGestureRecognizer is UIPanGestureRecognizer) else { return false }
+        return isWithinWebContent(otherGestureRecognizer.view)
+    }
```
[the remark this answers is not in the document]
jaceklyp: Confirmed and fixed in df6ee4b — the diagnosis was right, and worse than "delayed": a `UITapGestureRecognizer` stays `.possible` while a finger is held still, so it never failed and WebKit's long-press was queued behind something that never resolved. Our own tap recognizer also fails on held fingers by design, so a long press did nothing at all — no preview, no selection, no dismissal.

The rule now only defers *taps*:

```swift
guard gestureRecognizer === dismissOnPageTapRecognizer,
      !(otherGestureRecognizer is UIPanGestureRecognizer),
      !(otherGestureRecognizer is UILongPressGestureRecognizer) else { return false }
return isWithinWebContent(otherGestureRecognizer.view)
```

I kept the long-press exclusion rather than narrowing to `is UITapGestureRecognizer`, because I can't confirm WebKit's synthetic click recognizer is a `UITapGestureRecognizer` subclass — if it isn't, that tighter rule silently restores the link-follows-tap bug this whole path exists to prevent. Verified on device: tap dismisses without navigating, long-press previews and text selection work, scrolling unaffected.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -978,8 +1022,29 @@ extension AIChatContextualSheetCoordinator: AIChatContextualSheetViewControllerD
...
+                await self?.presentFloatingInput(from: presenter)
+            }
+        }
+    }
+
+    /// Hands the conversation off to a Duck.ai tab, leaving this sheet's chat as it is — the user asked
+    /// to go to Duck.ai, not to start over here.
+    func aiChatContextualSheetViewControllerDidRequestOpenDuckAI(_ viewController: AIChatContextualSheetViewController) {
+        delegate?.aiChatContextualSheetCoordinator(self, didRequestExpandWithURL: aiChatSettings.aiChatURL)
+        viewController.dismiss(animated: true)
```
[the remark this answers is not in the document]
jaceklyp: Confirmed and fixed in df6ee4b — `sessionState.cancelManualAttach()` added, matching `didRequestExpandWithURL` and `DidRequestViewAllChats`.

While checking this I also found the comment above it was wrong: it claimed the action "hands the conversation off to a Duck.ai tab", but it passes `aiChatSettings.aiChatURL` (no chat ID) — a fresh, context-free Duck.ai tab. Handing *this* conversation off is the header button, which passes `sessionState.contextualChatURL`. Comment corrected in the same commit.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualFloatingInputViewController.swift
```diff
@@ -258,12 +281,37 @@ final class AIChatContextualFloatingInputViewController: UIViewController {
...
+    /// Clears the chips only once they are invisible: removing them collapses the stack, and that
+    /// collapse rides whatever layout animation the input is already running — which reads as the row
+    /// sliding down behind the bar rather than leaving.
+    func clearChipsFadingOut() {
+        fadeChipsContainer(to: 0) { [weak self] in
+            self?.chipsViewController.updateStartActions(suggestions: [], quickActions: [])
+        }
+    }
+
+    private func fadeChipsContainer(to alpha: CGFloat, completion: (() -> Void)? = nil) {
```
aataraxiaa, line 299: agent caught this, seems correct - fadeChipsContainer's completion ignores the finished flag; an interrupted fade-out's stale completion still clears chipsViewController after the fade-in has already repopulated it. The PR's own "Known follow-ups" mentions a similar race as "unreproduced or not worth the risk" — this finding gives it a concrete mechanism.

iOS/DuckDuckGo/AIChat/ContextualMode/MainViewController+DuckAIAddressBarMenu.swift
```diff
@@ -40,9 +40,18 @@ extension MainViewController {
...
+            (button as? BrowserChromeButton)?.menuHighlightTarget = nil
+            duckAIMenuAnchor?.removeFromSuperview()
+            duckAIMenuAnchor = nil
             return
         }
 
+        // UIKit reparents the preview into the menu platter. Hand it a stand-in that lives outside the
+        // field's glass group, so nothing is pulled out of the group and the bar keeps rendering.
+        if let button = button as? BrowserChromeButton {
+            button.menuHighlightTarget = { [weak self] in self?.duckAIMenuAnchorView(over: button) }
```
aataraxiaa: can capture `button` weakly here too

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualFloatingInputViewController.swift
```diff
@@ -507,9 +562,32 @@ extension AIChatContextualFloatingInputViewController: UIGestureRecognizerDelega
...
         guard gestureRecognizer === dismissOnPageTapRecognizer else { return true }
+        // Only the page is swallowed. Chrome acts on its own taps, and the address bar dismisses this
+        // surface by taking focus, which never happens if the tap that focuses it is cancelled.
+        gestureRecognizer.cancelsTouchesInView = isWithinWebContent(touch.view)
         return view.hitTest(touch.location(in: view), with: nil) == nil
     }
 
+    /// A page tap is the user leaving, so the page's own tap waits for ours and loses — the link it landed
+    /// on would otherwise open behind the dismissal. Only taps: this recognizer never fails while a finger
+    /// is held, so making long presses wait for it would leave a hold doing nothing at all.
```
aataraxiaa: think we can shrink/remove multiline doc comments. Claude over-comments.

aataraxiaa (changes requested):
Few minor comments. Also one test failied - tapping voice chat icon did nothing, no voice chat started.

aataraxiaa (approved):
Nice one @jaceklyp. Looks great. Voice works.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualFloatingInputViewController.swift
jaceklyp: fixed in d6b07da2b4 — completion bails unless `finished`, so an interrupted fade-out can't clear the row the next fade-in already repopulated.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualUTIHost.swift
```diff
@@ -119,6 +129,30 @@ final class AIChatContextualUTIHost: UnifiedToggleInputDelegate, AIChatContextua
...
+    /// Only a keyboard that was really shown can be lost: a hardware keyboard hides on every keystroke.
+    private func collapseForKeyboardDismissal() {
+        guard hasKeyboardAppeared else { return }
+        hasKeyboardAppeared = false
+        // Backgrounding takes the keyboard too, and offscreen hosts were never above one.
+        guard UIApplication.shared.applicationState == .active,
+              !coordinator.isPresentingAttachmentModal,
+              coordinator.isInputOnScreen,
+              !coordinator.isContextualChatCollapsed else { return }
+        deactivateInput()
```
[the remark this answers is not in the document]
jaceklyp: leaving this as is. every phantom hide we measured arrived with no preceding `keyboardWillShow` — that's exactly what the gate suppresses — and after the fix typing produced 7 shows and 0 collapses.

adding a first-responder check back is what caused the bug this replaced: dragging the sheet to half hides the keyboard but keeps first responder, so the input stayed expanded over nothing. from the notification alone the two cases are indistinguishable, so it's in `.pr-followups.md` with `GCKeyboard` / `hasKeyboardAppeared` as candidate discriminators. hardware keyboard case is untested.

PR #6405, opened 2026-08-19
PixelKit, Common and DDGError extracted from BSK

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1216780180127947?focus=true

### Description

PixelKit could not be moved out of BrowserServicesKit on its own. It depends on two BSK targets, Common and Persistence, and six BSK targets depend on it. Moving it alone creates a dependency loop.

So this splits four packages out of BSK instead of one, stacked in order: 
- DDGError
- Common
- Persistence
- PixelKit.

#### Minimal code changes:
- `PixelKitConfig` moved into the PixelKit package. It only exists to set PixelKit up, and it was sitting in Common.
- A dead copy of PixelKit under `macOS/LocalPackages` was deleted. Nothing referenced it.
- Two lines in the macOS URL extension used `idnaDecoded`, which belongs to the Punycode library, without importing it. That worked only because BSK happened to pull Punycode in for Common, and the symbol came along for the ride. Rather than have the macOS app depend on Punycode directly, Common now exposes a decode helper next to the encode helper it already had, and the two call sites use that. Punycode stays an internal detail of Common, which is the only place that actually needs it. No other code outside Common touches the Punycode API now.

#### PixelKit testing utilities

`PixelKitTestingUtilities` no longer exists as a separate module. Its three files now live in the PixelKit target under `TestingSupport/`, marked `@_spi(Testing)`. 

#### Two problems this surfaced rather than caused:
- The moved code left behind SwiftLint settings that BSK applies to everything beneath it. Those settings are now with the packages, so the same rules apply as before the move.
- The iOS Core target used URLPredictor without declaring it, and only worked because BSK happened to declare it. It now declares it directly, matching what the macOS project already did.
- Xcode emits a set of `'X' is missing a dependency on 'Y'` warnings under `build-for-testing`. They predate this branch: `main` has 117 unique ones and this branch has 116. They are not manifest defects, since `BrowserServicesKit` is reported as missing `PixelKit` and `UserScript` while declaring both, and adding the declaration explicitly was measured not to silence them. Left alone.

### Testing Steps

1. Green CI
2. macOS unit tests: 5420 tests, 29 skipped, 0 failures on this branch.
---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Large-scale package and Xcode project rewiring with no intended runtime change, but a wrong manifest or duplicate PixelKit linkage could break builds or silently drop pixels in tests; merge conflicts on project files are likely.
> 
> **Overview**
> **Extracts four Swift packages from BrowserServicesKit** in dependency order: `DDGError`, `Common`, `Persistence`, and `PixelKit`. BrowserServicesKit drops those products and targets and instead depends on the new local packages; the workspace and iOS/macOS Xcode projects add matching package references so apps and downstream packages link them directly.
> 
> **Dependency graph updates** sweep SharedPackages (AIChat, AttributedMetric, VPN, WebExtensions, EventHub, Onboarding, SERPSettings, DataBrokerProtection, etc.) and local macOS packages so `Common`, `Persistence`, and `PixelKit` come from the new paths instead of BrowserServicesKit. Networking and Subscription code that conforms to `DDGError` now `import DDGError` explicitly.
> 
> **PixelKit testing** removes the separate `PixelKitTestingUtilities` product; mocks and expectation helpers stay in the `PixelKit` target and are exposed only via `@_spi(Testing)` so test bundles do not load a second static copy of PixelKit (avoiding split `PixelKit.shared`). Tests switch from `import PixelKitTestingUtilities` to `@_spi(Testing) import PixelKit`.
> 
> **Small API/wiring fixes** bundled with the move: `Common` adds `punycodeDecodedHostname` so macOS URL code no longer relies on Punycode through BSK; iOS Core declares `URLPredictor` directly; `PixelKitConfig` imports `Common`; several `Package.resolved` pins bump (e.g. TrackerRadarKit 4.0, content-scope-scripts).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 405e0751bce84c99b445eb2b2dced48ccb670b58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/PixelKit/Sources/PixelKit/TestingSupport/PixelFireExpectations.swift
```diff
@@ -17,15 +17,15 @@
...
 
 import Foundation
-import PixelKit
 
 /// Structure containing information about a pixel fire event.
 ///
 /// This is useful for test validation for libraries that rely on PixelKit, to make sure the pixels contain
 /// all of the fields they are supposed to contain..
 ///
+@_spi(Testing)
```
federicocappelli, line 26: A plain `import PixelKit` does not see these classes, so production code cannot reach the mocks. Test code can opt in with `@_spi(Testing) import PixelKit`. 

The use of `@_spi` was necessary because once PixelKit became its own package, Xcode built its product as a dynamic framework, and the same-package edge from `PixelKitTestingUtilities` to the `PixelKit` target pulled a second, static copy of PixelKit into `Unit Tests.xctest`. Tests configured that copy while app code read the framework's, so `PixelKit.shared` was nil at fire time and 38 pixel assertions failed. 
Removing the second module removes the edge, so there is one copy of PixelKit in the process again.

diegoreymendez (approved):
I haven't found anything worth flagging here.  Nice job.

PR #6407, opened 2026-08-19
[iOS] Reduce authentication-related hangs on the passwords screen

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201462886803403/task/1217579031816069?focus=true
Tech Design URL:
CC:

### Description
Avoids repeatedly checking device authentication availability whenever the passwords screen updates. This removes the main-thread work seen in the hang report while preserving credential protection and re-authentication behaviour.

Hang report: 
image

### Testing Steps
1. Open Settings → Passwords with saved credentials → the credentials remain hidden and the authentication prompt appears.
2. Authenticate successfully → the saved credentials appear.
3. Search the credentials list → results update normally and the screen remains responsive.
4. Open Control Center, then return to the app → authentication is required again and credentials remain hidden until it succeeds.
5. Minimise the app, then return to foreground → authentication is required again and credentials remain hidden until it succeeds.

---

> [!NOTE]
> **Medium Risk**
> Touches autofill lock/unlock UI logic; wrong caching could briefly show the wrong screen before auth is checked, but credential protection still depends on the authenticator and vault.
> 
> **Overview**
> Reduces main-thread hangs on the passwords list by **not calling** `canAuthenticate()` every time `updateViewState()` runs (e.g. when sections refresh).
> 
> **`AutofillLoginListViewModel`** now keeps a cached `authenticationAvailable` flag: it is set once when `authenticate()` runs, and updated from `authenticator.$state` (`loggedIn` → available, `notAvailable` → unavailable). View state uses that cache for `.noAuthAvailable` instead of re-querying LocalAuthentication, and treats the lock screen when the user is not `.loggedIn` rather than only when `.loggedOut`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 857d7b71b969288e81b5142b2dd49fe796da16db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AutofillLoginListViewModel.swift
+++ b/iOS/DuckDuckGo/AutofillLoginListViewModel.swift
@@ -80,6 +80,7 @@ class AutofillLoginListViewModel: ObservableObject {
     private var showBreakageReporter: Bool = false
     private let extensionPromotionManager: AutofillExtensionPromotionManaging
     private let featureFlagger: FeatureFlagger
+    private var authenticationAvailable: Bool?
 
     private lazy var reporterDateFormatter = {
         let dateFormatter = DateFormatter()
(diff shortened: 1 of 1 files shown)


frosty (approved):
Works for me!

PR #6408, opened 2026-08-19
Sync: Self-heal current-device metadata after device-list fallback

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203822806345703/task/1217605006097875?focus=true
Tech Design URL:
CC:

### Description
Unified device lists use encrypted device_info so all clients can share a device’s name and type. When that metadata is missing or cannot be read, Apple still displays the device using its legacy name and type.

This PR makes that fallback self-healing for the current Apple device. When the device list falls back to legacy data for this device, the app recreates and uploads its unified metadata in the background so future refreshes can use device_info again.

Repair requires unified-device writes and is attempted at most once per app session to avoid repeatedly updating the server during device-list polling. It does not delay or block the device list, and it never attempts to repair another device. Work based on an outdated account or overlapping migration is safely ignored without consuming the session’s repair attempt.

### Testing Steps
Prerequisite: Test on iOS because it provides the device-list diagnostics in Debug > Sync. Start with Sync disabled and create a brand-new account. Keep the app running while changing the flags so launch-time migration does not populate the metadata first.

1. Ensure both feature flags `syncCanReadUnifiedDeviceList` & `syncCanWriteUnifiedDeviceList` are disabled
2. Enable Sync → account creation succeeds without unified device metadata.
3. Enable `syncCanReadUnifiedDeviceList`, leaving `syncCanWriteUnifiedDeviceList` disabled.
4. Open the Sync screen → the current device remains visible with its correct legacy name and type.
5. Open Debug > Sync and tap to refresh the device list → the current device reports the legacy source and a missing `device_info` issue.
6. In the Xcode console, filter for `Sync-UnifiedDevices`.
7. Enable `syncCanWriteUnifiedDeviceList`.
8. Return to the Sync settings screen → the device remains visible while the console reports that current-device repair was scheduled and completed.
9. Return to Debug > Sync and refresh the device list → the current device reports `device_info` as its source with no device-info issue.
10. Return to the Sync settings screen again → the device remains correct and no additional repair is scheduled in logs

---

> [!NOTE]
> **Medium Risk**
> Background server updates to device metadata and new concurrency guards around migration/rename; scoped to the current device and feature-flagged writes, but touches core sync device-list flows.
> 
> **Overview**
> Adds **self-healing for the current device’s unified `device_info`** when the device list had to fall back to legacy name/type for this device.
> 
> After `fetchDevices()`, if the mapper reports `needsCurrentDeviceInfoRepair` and launch migration wasn’t in flight, `DDGSync` schedules a **best-effort background repair** (unified writes enabled only, **once per app session**, non-blocking). Repair is skipped during rename, while migration/repair tasks run, or when the stored account no longer matches the fetch; rename now cancels **both** migration and repair via `cancelDeviceInfoUpdatesAndWait()`.
> 
> `DeviceInfoMigrationCoordinator` gains `repairCurrentDeviceInfo`, sharing upload logic with migration but **without** requiring migration to be incomplete. Account teardown clears repair state. Mapper adds a debug log when refreshing `account_info` after stale `device_info`. Tests cover scheduling edge cases (migration overlap, account swap, rename, write flag off).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fdb65500612d58c45bb74e8ef2ed489cd01d6441. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/DDGSync/DDGSync.swift
@@ -246,9 +246,10 @@ public class DDGSync: DDGSyncing {
             throw SyncError.accountNotFound
         }
 
+        let wasDeviceInfoMigrationRunning = deviceInfoMigrationTask != nil
         do {
             let result = try await dependencies.account.fetchDevicesForAccount(account)
-            if result.needsCurrentDeviceInfoRepair {
+            if result.needsCurrentDeviceInfoRepair && !wasDeviceInfoMigrationRunning {
                 scheduleCurrentDeviceInfoRepair(for: account)
             }
             return result.devices
@@ -731,6 +732,7 @@ public class DDGSync: DDGSyncing {
 
     private func scheduleCurrentDeviceInfoRepair(for account: SyncAccount) {
         guard dependencies.syncFeatureFlags.canWriteUnifiedDeviceList(),
+              deviceInfoMigrationTask == nil,
               currentDeviceInfoRepairTask == nil,
               !hasAttemptedCurrentDeviceInfoRepair else {
             return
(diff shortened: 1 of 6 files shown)


pballart (approved):
LGTM

PR #6409, opened 2026-08-19
[iOS] Purge app switcher snapshots on Fire

bkunat:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1217575998563977

### Description
Deletes OS-cached app-switcher snapshots after every Fire request so snapshots captured in another orientation cannot reveal burned content. It also clears snapshots after successful Duck.ai chat deletions that use the Fire Button animation—including deleting one chat, selected chats, or all chats—and when tabs are closed. This prevents a snapshot captured in one orientation from showing content that was later deleted in another orientation. The cleanup is default-on behind a remotely releasable kill switch.

### Testing Steps
1. Load a website in portrait.
2. Background the app to capture a portrait snapshot.
3. Reopen the app and rotate to landscape.
4. Tap the Fire Button.
5. Background the app to capture the post-fire landscape snapshot.
6. Rotate the device to portrait and open the app switcher → the burned website is not visible.
7. Reopen the app → the burned website does not flash during the launch animation.
8. Repeat with portrait and landscape swapped.
9. Repeat steps 1–3, close the tab normally instead of using Fire, then background and open the app switcher in the original orientation → the closed tab is not visible.
10. Repeat the orientation sequence in Duck.ai, delete one chat, selected chats, or all chats using the Fire Button flow, then open the app switcher in the original orientation → deleted chat content is not visible.

### Impact
Medium: stale app-switcher snapshots can expose content the user burned.

#### What could go wrong?
Snapshot cleanup could fail or OS storage behavior could change → failures are isolated per item, never fail the burn, and the remotely releasable feature flag can disable cleanup.

### Quality Considerations
The cleaner removes only the contents of Library/SplashBoard/Snapshots and preserves the OS-owned directory. File I/O runs off the main actor. Missing directories are a silent no-op. Physical-device validation is still required because SpringBoard snapshot behavior cannot be fully confirmed by unit tests or the simulator.

---

> [!NOTE]
> **Medium Risk**
> Touches privacy-sensitive data clearing and deletes files under Library/SplashBoard/Snapshots. Failures are isolated and gated by a remotely releasable default-on flag.
> 
> **Overview**
> Clears iOS app-switcher snapshots after Fire, successful Duck.ai chat deletion, and tab close so a snapshot captured in another orientation cannot show content the user already burned.
> 
> Adds `AppSwitcherSnapshotCleaner`, which removes items under `Library/SplashBoard/Snapshots` while keeping the OS directory. `FireExecutor` and `TabManager` invoke it behind the new default-on, remotely releasable `appSwitcherSnapshotClearing` flag. Chat burns only clear snapshots on success.
> 
> Also drops the Fire animation’s pre-burn window snapshot when the app resigns active, so that overlay cannot leak into a new switcher snapshot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c63b5f622b4c1aea45cd9fad9b03404a189b781. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/FireButtonAnimator.swift
+++ b/iOS/DuckDuckGo/FireButtonAnimator.swift
@@ -114,13 +114,11 @@ enum FireButtonAnimationType: String, CaseIterable, Identifiable, CustomStringCo
 class FireButtonAnimator {
     
     private let appSettings: AppSettings
-    private let notificationCenter: NotificationCenter
     private var preLoadedComposition: LottieAnimation?
     private weak var preBurnSnapshot: UIView?
 
     init(appSettings: AppSettings, notificationCenter: NotificationCenter = .default) {
         self.appSettings = appSettings
-        self.notificationCenter = notificationCenter
         reloadPreLoadedComposition()
                 
         notificationCenter.addObserver(self,
@@ -168,11 +166,8 @@ class FireButtonAnimator {
 
         let duration = Double(composition.duration) / speed
         let delay = duration * currentAnimation.transition
-        DispatchQueue.main.asyncAfter(deadline: .now() + delay) { [weak self] in
+        DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
             snapshot.removeFromSuperview()
-            if self?.preBurnSnapshot === snapshot {
-                self?.preBurnSnapshot = nil
-            }
             Task { @MainActor in
                 await onTransitionCompleted()
             }
--- a/iOS/DuckDuckGo/FireButtonAnimator.swift
+++ b/iOS/DuckDuckGo/FireButtonAnimator.swift
@@ -110,20 +110,27 @@ enum FireButtonAnimationType: String, CaseIterable, Identifiable, CustomStringCo
 
 }
 
+@MainActor
 class FireButtonAnimator {
     
     private let appSettings: AppSettings
+    private let notificationCenter: NotificationCenter
     private var preLoadedComposition: LottieAnimation?
     private weak var preBurnSnapshot: UIView?
 
-    init(appSettings: AppSettings) {
+    init(appSettings: AppSettings, notificationCenter: NotificationCenter = .default) {
         self.appSettings = appSettings
+        self.notificationCenter = notificationCenter
         reloadPreLoadedComposition()
                 
-        NotificationCenter.default.addObserver(self,
-                                               selector: #selector(onFireButtonAnimationChange),
-                                               name: AppUserDefaults.Notifications.currentFireButtonAnimationChange,
-                                               object: nil)
+        notificationCenter.addObserver(self,
+                                       selector: #selector(onFireButtonAnimationChange),
+                                       name: AppUserDefaults.Notifications.currentFireButtonAnimationChange,
+                                       object: nil)
+        notificationCenter.addObserver(self,
+                                       selector: #selector(onApplicationWillResignActive),
+                                       name: UIApplication.willResignActiveNotification,
+                                       object: nil)
     }
         
     func animate(onAnimationStart: @escaping () async -> Void, onTransitionCompleted: @escaping () async -> Void, completion: @escaping () async -> Void) {
@@ -185,13 +192,17 @@ class FireButtonAnimator {
         }
     }
 
-    func removePreBurnSnapshot() {
+    private func removePreBurnSnapshot() {
         preBurnSnapshot?.removeFromSuperview()
         preBurnSnapshot = nil
     }
     
     @objc func onFireButtonAnimationChange() {
         reloadPreLoadedComposition()
+    }
+
+    @objc private func onApplicationWillResignActive() {
+        removePreBurnSnapshot()
     }
     
     private func reloadPreLoadedComposition() {
--- a/iOS/DuckDuckGo/FireButtonAnimator.swift
+++ b/iOS/DuckDuckGo/FireButtonAnimator.swift
@@ -114,6 +114,7 @@ class FireButtonAnimator {
     
     private let appSettings: AppSettings
     private var preLoadedComposition: LottieAnimation?
+    private weak var preBurnSnapshot: UIView?
 
     init(appSettings: AppSettings) {
         self.appSettings = appSettings
@@ -126,6 +127,7 @@ class FireButtonAnimator {
     }
         
     func animate(onAnimationStart: @escaping () async -> Void, onTransitionCompleted: @escaping () async -> Void, completion: @escaping () async -> Void) {
+        removePreBurnSnapshot()
 
         guard let window = UIApplication.shared.firstKeyWindow,
               let snapshot = window.snapshotView(afterScreenUpdates: false) else {
@@ -147,6 +149,7 @@ class FireButtonAnimator {
         }
         
         window.addSubview(snapshot)
+        preBurnSnapshot = snapshot
         
         let animationView = LottieAnimationView(animation: composition)
         let currentAnimation = appSettings.currentFireButtonAnimation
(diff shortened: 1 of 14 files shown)


iOS/DuckDuckGo/SettingsViewModel.swift
```diff
@@ -67,7 +67,6 @@ final class SettingsViewModel: ObservableObject {
     private(set) lazy var appSettings = AppDependencyProvider.shared.appSettings
     private(set) var privacyStore = PrivacyUserDefaults()
     lazy var featureFlagger = AppDependencyProvider.shared.featureFlagger
-    private lazy var animator: FireButtonAnimator = FireButtonAnimator(appSettings: AppUserDefaults())
```
bkunat, line 70: This property was unused

iOS/DuckDuckGo/FireButtonAnimator.swift
```diff
@@ -110,22 +110,29 @@ enum FireButtonAnimationType: String, CaseIterable, Identifiable, CustomStringCo
 
 }
 
+@MainActor
 class FireButtonAnimator {
     
     private let appSettings: AppSettings
     private var preLoadedComposition: LottieAnimation?
+    private weak var preBurnSnapshot: UIView?
```
bkunat, line 118: We need this property so the animator can remove the temporary pre-burn view if the app resigns active before the Fire animation reaches its transition point.

Previously, the snapshot was only a local variable and remained visible until the delayed animation callback removed it. If the app was backgrounded during that window, iOS could capture the pre-burn website again, effectively undoing this branch’s app-switcher snapshot cleanup.

bkunat:
**01 — The actor provides no mutual exclusion because a new instance is built on every call**

Fresh actor instances allow clears to overlap, but the likely outcome is redundant work and benign `ENOENT` errors

1. Both cleaners enumerate the same snapshot files.
2. Cleaner A removes a file.
3. Cleaner B tries to remove the same file and receives `ENOENT`.

Still, sharing one cleaner from `MainCoordinator` makes sense. Introducing a protocol would add no value here tho

Fixed in [b5ff7249](https://github.com/duckduckgo/apple-browsers/pull/6409/commits/b5ff724988fda3bbed7ea4601f498274441c546d).

**02 — A total failure is indistinguishable from “nothing to clean” and silently disables the fix**

Because we want to merge the main change today and adding a pixel requires privacy review, I moved the observability change into a separate stacked PR:

[PR #6470](https://github.com/duckduckgo/apple-browsers/pull/6470)

The pixel intentionally reports missing-directory errors. A path change and a legitimately absent directory both produce `ENOENT`, so they cannot be distinguished per event. Reporting them does, however, establish a baseline and could reveal an iOS-version-correlated increase.

One limitation remains: if iOS leaves the old directory present but stops using it, enumeration succeeds and this signal cannot detect the change.

**03 — The kill switch is a partial revert because the animator change ships ungated**

Good callout on the scope of the flag. This is intentional: the kill switch specifically guards the risky filesystem behavior: deleting files from the OS-owned `Library/SplashBoard/Snapshots` location, whose implementation could change.

The animator change is separate lifecycle cleanup. It removes only the temporary snapshot view that `FireButtonAnimator` added when the app resigns active. The normal animation path removes that same view shortly afterward anyway. There isn’t a useful state where we want to keep that pre-Fire view around for iOS to capture, even when on-disk cleanup is disabled.

Gating the observer would unnecessarily couple this UI cleanup to the filesystem kill switch and deliberately retain stale content. For that reason, I don’t think we should change it.

**04 — The tab-close clear can be dropped on suspend in exactly the scenario the fix targets**

I don’t think the deterministic failure sequence described here applies.

`Task {}` creates an unstructured task, and discarding its handle does not cancel it. UIKit also doesn’t suspend the process immediately when the user starts leaving the app; it first performs the normal background transition. This cleanup is scheduled while the app is still active and performs a small local directory operation.

I agree that there is no strict executor-ordering guarantee that the cleanup finishes before UIKit takes its snapshot, so a theoretical scheduling window remains. However, the task itself is not dropped. Given that, I don’t think making the entire tab-close path asynchronous or adding another lifecycle flush is justified.

Swift explicitly says:

> “Tasks run to completion even if there are no remaining uses of their task handle.”

[Swift Evolution SE-0304](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#unstructured-tasks)

For `applicationDidEnterBackground(_:)`, Apple says:

> “That method has five seconds to perform any tasks and return. Shortly after that method returns, the system puts your app into the suspended state.”

[Apple Developer Documentation](https://developer.apple.com/documentation/uikit/extending-your-app-s-background-execution-time)

bkunat:
**05 — A residual window remains where snapshots are deleted but the pre-burn page is still on screen**

I agree with the timing analysis: the on-disk clear can finish before the animator’s pre-burn snapshot reaches its normal transition-removal point. That is exactly why the willResignActive cleanup is part of this change.

When UIKit begins the background transition, the notification removes the retained pre-burn view on the main actor before UIKit’s later snapshot phase. The underlying tabs UI has also already been updated before the on-disk cleaner runs. There is a focused unit test covering the notification-driven view removal: `FireButtonAnimatorTests.testWhenApplicationWillResignActiveThenRetainedPreBurnSnapshotIsRemoved()`

What we cannot unit-test is SpringBoard’s subsequent capture and the Core Animation commit timing. I agree this deserves a physical-device test using .waterSwirl, backgrounding during the animation and checking both orientations. I don’t think another code change or forced redraw would provide a stronger system guarantee.

**06 — Duck.ai chat-history burns play the Fire animation but never clear snapshots**

This is valid. Great spot! Fixed in https://github.com/duckduckgo/apple-browsers/pull/6409/commits/3c3e43614f5217f4fb56da146013c657b161899e

**07 — The clear runs for burns that close no tabs and lands unmeasured inside the data-clearing SLO**

Both are mechanically correct, but I don’t think either warrants a change.

Running the cleanup for every Fire request is intentional. Trying to infer from the selected options whether an existing snapshot is still safe adds branching for little benefit, while clearing conservatively keeps the privacy behavior predictable.

The overall data-clearing event includes this duration because the burn is not complete until the awaited cleanup finishes. A dedicated sub-step would improve attribution, but I’d add that only if rollout data shows a meaningful latency regression. Adding a timeout now would deliberately allow stale snapshots and would not reliably interrupt synchronous file deletion.

**08 — Cold-launch appearance now changes after every tab close**

This consequence is real, but “after every tab close” is broader than the actual behavior. The cleaner removes the existing variants, and UIKit creates a fresh current-state snapshot the next time the app backgrounds. The LaunchScreen appears only when no suitable post-clear snapshot has been regenerated.

That is an intentional privacy-for-polish trade: showing the LaunchScreen is preferable to flashing burned or closed content.

iOS/DuckDuckGo/TabManager.swift
```diff
@@ -894,6 +899,12 @@ class TabManager: TabManaging, TrackerAnimationSuppressing {
             removeTabHistory(for: tabIDs)
         }
 
+        if featureFlagger.isFeatureOn(.appSwitcherSnapshotClearing) {
+            Task {
+                await clearAppSwitcherSnapshots()
+            }
+        }
```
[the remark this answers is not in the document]
bkunat: Addressed in https://github.com/duckduckgo/apple-browsers/pull/6409#issuecomment-5369695832

> 04 — The tab-close clear can be dropped on suspend in exactly the scenario the fix targets

jaceklyp (approved):
LGTM!

PR #6415, opened 2026-08-19
Fix upgradedHttps false positives in iOS breakage reports

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1217273311409053

### Description

iOS breakage reports were setting `upgradedHttps: true` on pages that were never upgraded. In `TabViewController`'s `didCommit`, `httpsForced` was computed as `tld.domain(lastUpgradedURL?.host) == tld.domain(webView.url?.host)`. For hosts with no Public Suffix List match — IP literals, `.local`, `.lan`, single-label hostnames — `TLD.domain(...)` returns `nil` on both sides, so `nil == nil` evaluated to `true` and falsely flagged an upgrade. This affected ~6% of flagged iOS reports (May–Jul 2026); it's a data-quality issue that skews breakage analytics and is not user-facing.

The fix extracts the computation into a testable `isHTTPSForced(...)` helper that guards on `lastUpgradedURL != nil`, so an upgrade is only reported when one actually occurred. This mirrors macOS, which keys the field off `connectionUpgradedTo != nil`.

### Testing Steps

This is a data-quality fix with no visible UI change, so it's primarily verified by the new unit tests.

1. Run the `TabViewControllerHTTPSForcedTests` suite in the `UnitTests` target — all 4 tests should pass.
2. (Optional manual check) On an iOS build, load a page on a host with no Public Suffix List match — e.g. `http://printer.local/`, an IP literal like `http://192.168.1.10/`, or a single-label host — directly, without any HTTP→HTTPS upgrade occurring.
3. Submit a breakage report for that page (Privacy Dashboard → report broken site).
4. Confirm the report's params show `upgradedHttps=false` (previously it was incorrectly `true`).
5. As a regression check, load a site that genuinely upgrades from HTTP to HTTPS and confirm its report still shows `upgradedHttps=true`.

### Impact and Risks

**Low.** Only affects the `upgradedHttps` diagnostic flag in breakage reports — no user-facing behavior.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Only changes the breakage-report diagnostic flag; no navigation, upgrade, or UI behavior. Genuine upgrades should still report true via the new helper.
> 
> **Overview**
> Fixes **incorrect `upgradedHttps: true`** on iOS breakage reports when no HTTP→HTTPS upgrade occurred.
> 
> On navigation commit, **`httpsForced`** (which feeds **`upgradedHttps`**) used **`tld.domain(lastUpgradedURL?.host) == tld.domain(webView.url?.host)`**. For hosts without a Public Suffix List match (IPs, `.local`, `.lan`, single-label names), **`TLD.domain`** returns **`nil` on both sides**, so **`nil == nil`** was treated as an upgrade—about **~6%** of flagged reports (May–Jul 2026).
> 
> The logic is moved to **`TabViewController.isHTTPSForced(...)`**, which **returns false when `lastUpgradedURL` is nil** and only then compares registrable domains—aligned with macOS treating upgrades as tied to an actual upgrade URL.
> 
> **`TabViewControllerHTTPSForcedTests`** covers PSL-less hosts, normal hosts without a prior upgrade, same-domain upgrades, and cross-domain navigation after an upgrade.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f452e04699a42ce275419611d5aea5ec16a04640. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/TabViewController.swift
+++ b/iOS/DuckDuckGo/TabViewController.swift
@@ -2239,10 +2239,15 @@ extension TabViewController: WKNavigationDelegate {
 
     /// Whether the committed page was reached via an HTTPS upgrade.
     ///
-    /// Guards on `lastUpgradedURL` so an upgrade is only reported when one actually happened.
+    /// Needs both an upgrade on record and an HTTPS commit: `lastUpgradedURL` isn't reset across
+    /// same-domain navigations, so without the scheme check a later HTTP commit on the same domain
+    /// would be mis-flagged. Mirrors macOS's `connectionUpgradedTo != nil`.
     static func isHTTPSForced(lastUpgradedURL: URL?, currentURL: URL?, tld: TLD) -> Bool {
-        guard let lastUpgradedURL else { return false }
-        return tld.domain(lastUpgradedURL.host) == tld.domain(currentURL?.host)
+        guard let lastUpgradedURL, let currentURL, currentURL.isHttps else { return false }
+        guard let upgradedDomain = tld.domain(lastUpgradedURL.host) else {
+            return lastUpgradedURL.host == currentURL.host
+        }
+        return upgradedDomain == tld.domain(currentURL.host)
     }
 
     private func onWebpageDidStartLoading(httpsForced: Bool) {
--- a/iOS/DuckDuckGo/TabViewController.swift
+++ b/iOS/DuckDuckGo/TabViewController.swift
@@ -2228,13 +2228,21 @@ extension TabViewController: WKNavigationDelegate {
 
         url = webView.url
         let tld = storageCache.tld
-        let httpsForced = tld.domain(lastUpgradedURL?.host) == tld.domain(webView.url?.host)
+        let httpsForced = Self.isHTTPSForced(lastUpgradedURL: lastUpgradedURL, currentURL: webView.url, tld: tld)
         onWebpageDidStartLoading(httpsForced: httpsForced)
         textZoomCoordinator.onNavigationCommitted(applyToWebView: webView)
         
         // Check cache for instant logo display during back navigation
         checkDaxEasterEggCacheIfDuckDuckGoSearch(webView)
 
+    }
+
+    /// Whether the committed page was reached via an HTTPS upgrade.
+    ///
+    /// Guards on `lastUpgradedURL` so an upgrade is only reported when one actually happened.
+    static func isHTTPSForced(lastUpgradedURL: URL?, currentURL: URL?, tld: TLD) -> Bool {
+        guard let lastUpgradedURL else { return false }
+        return tld.domain(lastUpgradedURL.host) == tld.domain(currentURL?.host)
     }
 
     private func onWebpageDidStartLoading(httpsForced: Bool) {
--- a/iOS/DuckDuckGoTests/TabViewControllerHTTPSForcedTests.swift
+++ b/iOS/DuckDuckGoTests/TabViewControllerHTTPSForcedTests.swift
@@ -1,0 +1,57 @@
+//
+//  TabViewControllerHTTPSForcedTests.swift
+//  DuckDuckGo
+//
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -2237,6 +2237,14 @@ extension TabViewController: WKNavigationDelegate {
 
     }
 
+    /// Whether the committed page was reached via an HTTPS upgrade.
+    ///
+    /// Guards on `lastUpgradedURL` so an upgrade is only reported when one actually happened.
+    static func isHTTPSForced(lastUpgradedURL: URL?, currentURL: URL?, tld: TLD) -> Bool {
+        guard let lastUpgradedURL else { return false }
```
bkunat: This fixes the nil comparison path, but I think we should also handle the path where a page commits over HTTP after a genuine upgrade. isHTTPSForced may still report an upgrade there: lastUpgradedURL records that an upgrade was applied to some navigation, not that the page currently committed arrived over HTTPS, and the helper compares registrable domains without ever looking at the scheme.

1. The user opens `http://www.example.com`.
2. `www.example.com` is in the HTTPS upgrade list, so `upgradeToHttps` sets `lastUpgradedURL = https://www.example.com` and loads the HTTPS URL.
3. The page commits over HTTPS. `isHTTPSForced` compares `example.com` with `example.com` and returns `true`. This is correct; the page really was upgraded.
4. The user follows a link to `http://legacy.example.com`, a host that only serves HTTP. It isn't in the upgrade list, so no upgrade happens.
5. `decidePolicyFor` decides whether to clear `lastUpgradedURL`. It compares `tld.domain("legacy.example.com")` with `tld.domain("www.example.com")`, gets `example.com` on both sides, treats the two hosts as the same site, and skips the reset. `lastUpgradedURL` stays `https://www.example.com`.
6. The page commits over HTTP. `isHTTPSForced` compares `example.com` with `example.com` again and returns `true`: this time for a page that was never upgraded.
7. The user reports breakage. The report carries `url=http://legacy.example.com` together with `upgradedHttps=true`

Adding `currentURL.isHttps` to the existing guard should break the chain at step 6 and cover both cases.

```swift
static func isHTTPSForced(lastUpgradedURL: URL?, currentURL: URL?, tld: TLD) -> Bool {
    guard let lastUpgradedURL, let currentURL, currentURL.isHttps else { return false }
    guard let upgradedDomain = tld.domain(lastUpgradedURL.host) else {
        return lastUpgradedURL.host == currentURL.host
    }
    return upgradedDomain == tld.domain(currentURL.host)
}
```

iOS/DuckDuckGoTests/TabViewControllerHTTPSForcedTests.swift
```diff
@@ -0,0 +1,57 @@
...
+@testable import DuckDuckGo
+
+final class TabViewControllerHTTPSForcedTests: XCTestCase {
+
+    private let tld = TLD()
+
+    // Regression: hosts with no Public Suffix List match (IP literals, .local, .lan, single-label)
+    // must not report an upgrade when none happened. Previously tld.domain(...) returned nil on both
+    // sides so nil == nil falsely read as upgraded (~6% of flagged iOS reports, May–Jul 2026).
+    func test_noPriorUpgrade_returnsFalseForHostsWithoutPSLMatch() {
```
bkunat: Both noPriorUpgrade cases pass lastUpgradedURL: nil, so they bail out at the first guard and never actually hit the domain comparison. That means the regression this PR is fixing isn't really covered, since those four hosts with no PSL match don't end up changing the test outcome either way.

I think we're missing a case with a non-nil lastUpgradedURL and a committed http:// URL to actually test that path.

iOS/DuckDuckGo/TabViewController.swift
hassaanelgarem: @bkunat Good catch!  Applied your suggestion in 52747e99d1ae188e467e53c00c12547ec86bbc83 👍

iOS/DuckDuckGoTests/TabViewControllerHTTPSForcedTests.swift
hassaanelgarem: Renamed to test_noPriorUpgrade_returnsFalse with https:// URLs to actually isolate the nil guard, and added test_priorUpgrade_httpCommitOnNoPSLHost_returnsFalse for the non-nil + http:// path you flagged.

bkunat (approved):
LGTM :shipit:

PR #6418, opened 2026-08-19
[macOS] Fire duck_ai_new_chat retention metric on new-chat creation

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204186595873227/task/1216921686081266
Tech Design URL:
CC:

### Description
Wires up the `duck_ai_new_chat` retention experiment metric on macOS. The metric and its firing helper already existed in shared `PixelExperimentKit` but were never called from app code/

### Testing Steps
1. Force a config with an active experiment and enroll (there should be some ongoing experiment other wise you can put a breakpoint on PixelExperimentKit: fireNewAIChatExperimentPixels
2. Filter debug pixels for the experiment metrics.
3. Start a new chat via each surface (Duck.ai menu → New Chat, title-bar button, omnibar submit, address-bar AI chat, NTP submit, File → New AI Chat, New Voice Chat, New Image Chat, sidebar open) → each fires one `duck_ai_new_chat` (value=1, window per enrollment day).
4. Tap New Chat while already on an empty Duck.ai tab → no additional pixel.
5. Open a recent/existing chat and reopen a kept sidebar → no `duck_ai_new_chat`.

### Impact
None (adds metric instrumentation only; no user-facing behaviour change).

### Quality Considerations
Unit tests in `AIChatTabOpenerTests` cover: fires for new-chat/query/mode and image-mode URL; silent for the no-op path, existing-chat, and plain (non-mode) URL triggers.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit a78f3a88bdf6f36f676a5fbd07628540813e0586. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatTabOpener.swift
@@ -121,7 +121,6 @@ protocol AIChatTabOpening {
 struct AIChatTabOpener: AIChatTabOpening {
     private let promptHandler: AIChatPromptHandler
     private let aiChatTabManaging: AIChatTabManaging
-    /// Fires the `duck_ai_new_chat` retention experiment metric. Injected so it can be observed in tests.
     private let fireNewChatExperimentPixels: () -> Void
 
     let aiChatRemoteSettings = AIChatRemoteSettings()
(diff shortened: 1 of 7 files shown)


tomasstrba (commented):
TS1/ Submitting a new chat through the New Tab Page fires two pixels. Is that intended?

TS2/ We already have `m_mac_aichat_start_new_conversation`, which is likely more reliable than the current implementation. Is there any particular reason why aren't we reusing that, or at least making a common firing logic? The current approach makes it hard to maintain and makes the codebase a little bit messy. Also, I believe a couple of entry points are missing

In this PR https://github.com/duckduckgo/apple-browsers/pull/6494, I've just added a source parameter, but perhaps you could build this logic on top of it.

I don't have a full context, so if this approach is desired, we can move on once TS1/ is resolved

SabrinaTardio:
> TS1/ Submitting a new chat through the New Tab Page fires two pixels. Is that intended?
> 
> TS2/ We already have `m_mac_aichat_start_new_conversation`, which is likely more reliable than the current implementation. Is there any particular reason why aren't we reusing that, or at least making a common firing logic? The current approach makes it hard to maintain and makes the codebase a little bit messy. Also, I believe a couple of entry points are missing
> 
> In this PR #6494, I've just added a source parameter, but perhaps you could build this logic on top of it.
> 
> I don't have a full context, so if this approach is desired, we can move on once TS1/ is resolved

@tomasstrba I am trying to achieve a different thing (send relevant experiment pixels when a new AIChat is opened) not on prompt (we have one for all prompts already) and unfortunately the entry points are scattered, 
but… I am thinking that it would make more sense on new conversation (first prompt) I’ll chat with DS to confirm, so I’ll pause for now.

PR #6426, opened 2026-08-19
[iOS/macOS] Add a more detailed automatic sign out pixel

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1199333091098016/task/1217653717089857?focus=true
Tech Design URL:
CC:

### Description

This PR adds a new automatic sign out pixel that tells us the detailed reason that led to the sign out, and bucketed metadata about the account to tell us whether the sign out makes sense.

### Testing Steps

1. Sign into a subscription account
2. Search for `Uncomment this to simulate an automatic signout` and uncomment the line below
3. Rebuild the app and check that you get signed out
4. Search the console logs for `automatically.signed.out` and check that the pixel is present & the params make sense

### Quality Considerations

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches the automatic sign-out path in token retrieval, but does not change when users are signed out—only when/how telemetry is captured. Incorrect snapshots could misreport auth diagnostics, not alter auth itself.
> 
> **Overview**
> Adds a diagnostic pixel fired when Privacy Pro automatically signs the user out after an unknown-account or unrecoverable invalid-refresh-token error. Parameters are low-cardinality buckets of local token and cached-subscription state captured *before* recovery/sign-out, so later restore work cannot rewrite what the pixel reports.
> 
> `getTokenContainer` now snapshots local tokens around the failed request, then signs out through a shared helper that emits `.automaticSignOut` on iOS (`m_privacy-pro_auth_account_automatically_signed_out` with platform/form-factor suffixes) and macOS (store/direct names). Existing invalid-refresh-token pixels are unchanged. Successful recovery still does not fire this pixel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aad944978708d8c84fcbbeead70839c50734706d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -649,6 +684,207 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
         }
     }
 
+    private func automaticallySignOut(reason: SubscriptionAutomaticSignOutPixelData.Reason,
```
federicocappelli: Instead of creating a new SubscriptionAutomaticSignOutPixelData (or in addition), why don't we send the Error causing the logout directly? Are you concerned about the content for privacy reasons?
The triggering error should be already built nicely, with an underlying error chain, so it could give us more useful info

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -649,6 +684,207 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
+            guard let tokenContainer = try oAuthClient.currentTokenContainer() else {
+                return .missing
+            }
+            return .present(tokenContainer)
+        } catch {
+            return .readError
+        }
+    }
+
+    private func automaticSignOutTokenStatus(
```
federicocappelli: The SubscriptionManager is becoming huge, can we move all of this new code into a dedicated analytics file/extension?

federicocappelli (changes requested):
All good, just a couple of questions

samsymons:
Pixel validation is failing due to an unrelated issue, which I am fixing in https://github.com/duckduckgo/apple-browsers/pull/6455.

SharedPackages/BrowserServicesKit/Sources/Subscription/Managers/SubscriptionManager.swift
samsymons: Agree, that makes sense - done, I added the error params to the pixel and removed redundant values from the sign out pixel data that the error params would give us signal on.

federicocappelli (approved):
LGTM

PR #6437, opened 2026-08-20
[iOS] Replace TimedPixel with a PixelKit event for data clearing

federicocappelli:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205842942115003/task/1213803217990072
CC: @hassaanelgarem @dus7 

### Description

*This is a side-quest of https://app.asana.com/1/137249556945/project/1205842942115003/task/1208502294499469?focus=true*, where, rather than reimplementing a timing primitive inside PixelKit, I decided to migrate all 4 pixels using it to PixelKit.

`TimedPixel` (`iOS/Core/TimedPixel.swift`) was deprecated in its own doc comment and fired through the legacy `Pixel` singleton, so nothing downstream of it was injectable or mockable. Its only production caller was `FireExecutor`, for the four data clearing completion pixels.

Rather than reimplement a timing primitive inside PixelKit, the duration now travels as an associated value on a `PixelKit.Event` and is encoded in `var parameters`, following the `SiteLoadingPixel` precedent. That removes the need for a shared helper entirely: the "inject the duration parameter" half belongs to the event, and the "capture a start instant" half is one call.

- **New `DataClearingCompletionPixels`** (in `DataClearingPixels.swift`) covering `mf_dc`, `m_single-tab-data_cleared`, `m_fire-mode_data-cleared` and `m_normal-mode_data-cleared`.
- **No singletons.** Firing and timing both run through the `PixelFiring` and `timeProvider` seams `DataClearingPixelsReporter` already had, which `FireExecutor` already receives via its initialiser. No new DI wiring was required.
- **`FireExecutor` no longer knows about pixels, parameters or clocks.** `dataClearingTimedPixel(for:)` and `dataClearingPixelParams(for:domains:)` collapse into a single scope-to-event mapping.
- **Removed:** `TimedPixel.swift` and its `project.pbxproj` entries, `PixelParameters.duration`, and the four now-unused `Pixel.Event` cases and name mappings.

#### Why a separate enum rather than adding cases to `DataClearingPixels`

Both would have changed pixels this PR does not own:

- `DataClearingCompletionPixels` conforms to `PixelKitEventWithCustomPrefix`, which is what appends the `_ios_phone` / `_ios_tablet` suffix these four have always sent. `DataClearingPixels` has no such conformance, so retrofitting it there would have started suffixing `m_fire_retrigger_in_20s` and `m_fire_user_action_before_completion` too.
- `DataClearingPixels` reports `pixelSource`, which these four do not declare in `forget_all.json5`.

#### No behaviour change

Same names, same suffixes, same parameters, same `dur` key carrying floating-point seconds. **Pixel definitions are untouched**, so no dashboard or query changes are needed.

### Testing Steps

1. Build and run the iOS app.
2. Tap the Fire Button and confirm the burn completes as before.
3. In a debug build, check the pixel log (`Pixel fired …`) and confirm `mf_dc_ios_phone` (or `_ios_tablet` on iPad) is emitted with a `dur` parameter holding a floating-point seconds value, plus `tc`.
4. Repeat for the other three scopes: burn a single tab (`m_single-tab-data_cleared`, with `tabType`, `browsing_mode` and `domainsCount`), and a Fire Mode burn (`m_fire-mode_data-cleared`).
5. Confirm the names and parameters match what the build before this change emitted.

#### What could go wrong?

The single real risk is the **wire name**, since these pixels moved off `Pixel.fire` (which bakes the form factor into the URL path) onto PixelKit (which does not, unless the event conforms to `PixelKitEventWithCustomPrefix`). Losing the suffix would silently orphan the existing dashboards.
---

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 02aa071953213a3989f1f0312c03b4bfe229637e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hassaanelgarem (approved):
LGTM 👍

PR #6400, opened 2026-08-18
iOS: Duck.ai usage-warning footer behind the UTI

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201011656765697/task/1216355909736641?focus=true
Tech Design URL:
CC:

### Description
Adds a message card that sits behind the unified toggle input and slides out below it to warn Duck.ai users as they approach a usage limit, reading the privacy-safe usage snapshot the web app writes into native storage. It appears in the Duck.ai input and in the address bar when the toggle is on Duck.ai, and carries a primary action plus a dismiss. Behind the internal-only `utiDuckAIWarnings` flag; the actions are not wired to real destinations yet and the copy is not final.

### Testing Steps
1. Debug ▸ AI Chat ▸ Duck.ai Usage Warnings → tap **Weekly 90%**, then go back.
2. Open a Duck.ai tab and focus the input → the card grows out from behind it in one motion, showing "90% of weekly limit / Resets in 2 days"; dismiss it with ✕ and confirm it stays gone for that window but returns after seeding a different percentage.
3. Tap the address bar with the toggle on **Duck.ai** → same card; flip the toggle to **Search** → it hides, flip back → it returns.

### Impact
Low

#### What could go wrong?
The card shares the slot behind the input with the edit-replace disclaimer and changes the bar's height → mitigated by a single-occupant slot the disclaimer always wins, and by unit tests pinning the card to one line per label so a long string can never grow the bar.

---

> [!NOTE]
> **Medium Risk**
> Touches unified-input layout and height (shared slot with the edit disclaimer) and reads usage from native storage, but the feature is internal-only and primary actions are not wired yet.
> 
> **Overview**
> Adds a **Duck.ai usage-warning footer** that slides out behind the unified toggle input when daily/weekly usage hits 50/75/90% or the limit. It reads the privacy-safe `usageLimits` snapshot the web app already writes to native storage, and is gated by the internal-only `utiDuckAIWarnings` flag.
> 
> The card appears in Duck.ai input and in the omnibar when the toggle is on Duck.ai; it hides in Search mode, while editing, and when collapsed. Fire tabs use a separate storage snapshot so usage never leaks across modes. Dismiss is session-scoped per warning; **Reduce Usage** / **Switch** buttons are present but not connected to destinations yet. Copy is still `NotLocalizedString`.
> 
[...]

PR #6445, opened 2026-08-20
Resolve Duck.ai usage-limit messages on iOS and macOS

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1217578255373544?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1216355483179556

### Description

This PR is a follow-up of #6350 and adds the decision which message to show, at what severity, whether it can be dismissed, and what the copy says. The logic lives in the shared AIChat package so both platforms can use it, but **only macOS is wired here** — see the note on #6400 below.

Note: **Still no UI**: the output is a debug log

### Testing Steps

Warnings are behind `aiChatUsageWarnings` (internal-only), and there is no UI, so the log is the surface.

1. Become internal via `https://staging.duck.ai/setup-aichat`
2. Filter logs based on "Duck.ai usage warning"'`
3. On tab A, load Duck.ai → Settings → Usage and set to 0
4. On tab B → activate the address bar. In logs, expect `none — reason=belowVisibilityFloor`
5. Tab A → set usage to 50%. Tab B → reactivate address bar. Expect "50% of daily limit" message in logs
6. Repeat at 75%
7. Set the weekly window to 99% → expect the message switches to `window=weekly`, since the more severe window wins
8. Set to blocked → expect "Daily limit reached" dismissible=false`

### Impact

Low — internal-only flag, no UI, and the only behaviour change outside the new code is that the macOS surfaces now hold a view model where they held a raw snapshot.

#### What could go wrong?

- Blocked detection is **inferred** from `percentUsed == 100` rather than told to us. Verified against a real blocked account, but a web-side change to that value would silently stop "limit reached" firing.
- The CTA assumes `/models` labels are **per-tier**, like `entityHasAccess`. The anonymous payload only shows the free-tier split, so this is unconfirmed for Pro → the suggester logs `current=<id> label=<label>` on every resolution, which surfaces it on the first real Pro session. If labels turn out to be global, a Pro user on Opus gets no nudge.
- The backend could stop labelling models → the CTA silently stops offering, mitigated only in that the log gives `reason=noEverydayUseModelAvailable` rather than going quiet.
- A dismissal could outlive its window → mitigated by keying it to `resetsAt`; a stale record is ignored and the message reappears, which is the safe direction.

### Quality Considerations

Known gaps, all deliberate:

- **`didRecentlyStepDown` is stubbed** to `false` — neither platform tracks "user just stepped down on a new chat", and there is no UI yet to produce a step-down to observe.
- **Fire windows show nothing**, by decision — an isolated session must not surface the regular session's usage and has no usage of its own worth warning about. Expressed as an `isFireMode` predicate the view model re-reads on every refresh, rather than a storage choice fixed at construction, so a surface whose fire state changes at runtime cannot go stale.
- **Free tier will rarely see an approaching message** — the backend withholds usage until a window is already blocked (#6350 known gap 1). The tier gating is still correct; it just has no data to act on.

### Overlap with #6400

#6400 landed an iOS usage-warning footer while this was in review, with its own resolver, message mapping and card UI. Rather than run two rule engines over the same snapshot behind two flags, the iOS surfaces here revert to main and this PR keeps the shared logic plus the macOS wiring. The iOS `aiChatUsageWarnings` flag goes with them, since the footer gates on its own `utiDuckAIWarnings`.

The shared resolver still carries rules the footer does not — tier gating, dismissal persistence, blocked versus approaching, and the label-driven cheaper-model CTA — so the follow-up is to point `UTIFooterWarningResolver` at it rather than grow a second copy. Worth agreeing that with @Bunn before either side moves.

Tests: 63 across the rule matrix — threshold and rounding boundaries, the tier matrix, window precedence, both dismissal ladders, the label-driven model suggestion, and the rendered copy.

---

> [!NOTE]
> **Medium Risk**
> New limit-warning and model-switch decision logic plus persisted dismissals, but gated behind an internal-only flag with no UI yet. Blocked state is inferred from 100% usage.
> 
> **Overview**
> Adds a shared **usage-warning resolver** on top of the existing native usage snapshot, so Duck.ai can decide *which* daily/weekly message to show, at what severity, whether it is dismissible, and which CTA to offer. **No UI yet** — decisions go to the debug log. Only **macOS** is wired (address bar + NTP omnibar).
> 
> The resolver keys off displayed (rounded, 99-capped) percent, a 50% visibility floor, paid/internal-tier gating for approaching banners, and window precedence (more severe / closer to the cap wins; weekly wins when both are blocked). Dismissals persist per window keyed to `resetsAt`, with different redisplay ladders for daily vs weekly.
> 
> **Cheaper-model CTAs** use `AIChatModelLabel` (`usesLimitsFaster` → `everydayUse`, same provider preferred) rather than a name ladder, and respect draft capabilities (images, MIME types, tools). Reached advanced-model windows can suggest a free-tier model; free users get upsell copy instead.
> 
> macOS surfaces now hold a `DuckAiUsageWarningViewModel` instead of a raw `DuckAiUsageLimits` snapshot. Model-switch actions apply immediately; try-for-free and “start using weekly limit” are logged only. Fire/burner sessions stay silent.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74fdab3f2f88b328d4fdb89caafd25785950e52c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Bunn:
`setUpUsageWarnings` decides storage from the init-time `isFireTab` (`UnifiedToggleInputCoordinator.swift:470`), but the shared omnibar coordinator flips fire state later via `updateIsFireTab(_:)` on every tab switch.

So a coordinator created on a regular tab keeps the regular storage handler forever — switching to a fire tab and binding an AI chat resolves the *regular* session's usage in fire mode, the leak this PR is meant to avoid. The inverse also holds: created on a fire tab, regular tabs get no warnings all session.

Fix could be gating at refresh time on `viewController.handler.isFireTab`, or rebuilding the view model in `updateIsFireTab`.

(Claude)

Bunn (approved):
LGTM, claude did flag an issue (comment above).

PR #6451, opened 2026-08-20
Fix issue where rapidly-made tap events are discarded

felixlawrence:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1206302084112559?focus=true
Tech Design URL:
CC:

### Description

When making a rapid sequence of taps in the iOS browser (e.g. typing quickly in Wordle, or playing a game), every second tap gets ignored. This doesn't happen in Firefox or Safari.

Per Claude this appears to be because we had a gesture recognizer looking to detect a double click, _and_ this gesture recogniser was only actually needed in a very specific circumstance (search/address bar hidden, the first tap is in the region where the search/address bar would go, and it's a double tap). The fix was to restrict the gesture recogniser to apply in the scenario it's needed. I don't have enough context to verify any of this!

Someone with more context should consider whether this gesture recogniser is needed _at all_, because after this fix, it would still be annoying to someone playing a full-screen game if this behaviour still existed for taps along one edge of the screen.

N.B. the `project.pbxproj` file might be corrupted with hallucinated hex codes - if that's the case then revert it, and let XCode update it to include the new test file.

### Testing Steps

I don't have a mac so I wasn't able to build or test this.
1. Go to https://patrickhlauke.github.io/touch/tests/event-listener.html on DDG iOS
2. Double tap the button quickly -> two pairs of (touchstart, touchend) should be captured

### Impact

High/Medium - could break user interaction with the browser and I don't understand the breadth of potentially-affected circumstances.

#### What could go wrong?

Do we fully understand why the original behaviour was there? Does the fix preserve the original behaviour in the circumstances where it's needed? If "no" to either, then this "fix" could break other user interactions.

### Quality Considerations

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches core tab touch handling and chrome reveal behavior; mitigated by remote flag default-on with rollback path and unit tests for arbitration and delaysTouchesEnded.
> 
> **Overview**
> Fixes **dropped second taps** during fast tapping in web content (e.g. on-screen keyboards and games) by changing how the **show hidden chrome** tap recognizer interacts with **WKWebView** multi-tap recognizers.
> 
> When the new remote flag **`suppressShowBarsGestureRecogniserDelay`** is on (default **enabled**), the show-bars recognizer **no longer delays touch end** on every tap, and it **only claims failure priority** over other tap recognizers when the tap could actually reveal bars (hidden chrome + tap in the bottom strip). Taps in page content no longer wait behind that recognizer. **Bottom-strip taps** still wait so revealing the toolbar keeps working.
> 
> Adds **`TabViewControllerGestureArbitrationTests`** and flag wiring in **FeatureFlags-iOS** (plus a focused flag test).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42b1eb937204c175440d5be287f541cdb880f105. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsymons:
This is working well for me, I'm just doing some further testing and adding a kill-switch feature flag to be safe. I'll merge this on Monday, so it will go live to users on September 7. Thanks again!

samsymons (approved):

PR #6453, opened 2026-08-20
Updated Sync device renaming for unified device lists

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203822806345703/task/1216959197715042?focus=true
Tech Design URL:
CC:

### Description
Updates how the current device is renamed as Sync moves to the unified device list. When unified-device writes are enabled, renaming updates both the new encrypted device_info and the legacy name/type fields in a single authenticated request. This keeps new and older clients in agreement without refreshing the account token.

When unified-device writes are disabled, a new default-on rollback flag uses the same endpoint to update the legacy fields and clear potentially stale unified metadata. If that rollback flag is also disabled, the existing login-based rename behaviour remains available.

Renames are coordinated with background device-info migration and repair so older work cannot overwrite the new name. The local account is updated only after the server accepts the rename.

### Testing Steps
Prerequisite: Install on an iOS device and a macOS device, then connect both to the same Sync account. On both devices, enable `syncCanReadUnifiedDeviceList` and confirm `syncCanUsePatchEndpointForLegacyDeviceRename` is enabled by default. 

1. On iOS, enable `syncCanWriteUnifiedDeviceList` and confirm `syncCanUsePatchEndpointForLegacyDeviceRename` is already enabled.
2. Open Settings > Sync and rename the iOS device → the new name appears on iOS and updates on macOS; the macOS device’s name remains unchanged.
3. Close and reopen Sync settings on both devices → the renamed iOS device still has the new name and both devices remain connected.
4. On iOS, disable `syncCanWriteUnifiedDeviceList`. Leave `syncCanUsePatchEndpointForLegacyDeviceRename` and `syncCanReadUnifiedDeviceList` enabled on both devices.
5. Rename the iOS device again → both devices display the latest name after refresh, confirming that the legacy-only PATCH does not leave the previous unified name visible.
6. On iOS, disable `syncCanUsePatchEndpointForLegacyDeviceRename`. On both iOS and macOS, disable `syncCanReadUnifiedDeviceList`. At this point the iOS write, read, and legacy-PATCH flags are all disabled.
7. Rename the iOS device again → both devices display the new name from the legacy device list after refresh, and the Sync account remains connected.
8. Take the iOS device offline and attempt another rename → an error is shown; after restoring the connection and refreshing Sync settings, the last successful name remains unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes authenticated device-update and account persistence paths for rename; wrong flag routing or PATCH semantics could leave device lists inconsistent across clients, though behavior is gated and heavily tested.
> 
> **Overview**
> **Sync device rename** no longer always goes through `refreshToken`. `updateDeviceName` picks a path from feature flags at the start of the call: **unified writes** → PATCH with encrypted legacy fields plus unified `device_info`; **new default-on `canUsePatchEndpointForLegacyDeviceRename`** → PATCH legacy name/type only and omit `info` so stale unified metadata is cleared; otherwise the existing login/refresh rename remains.
> 
> Rename work lives in **`DeviceInfoMigrationCoordinator.renameCurrentDevice`** (`.unified` / `.legacyOnly`), with shared **`DeviceInfoUpdateBuilder`** for PATCH payloads. The local account name is updated only after the server accepts the rename; in-flight migration/repair is still cancelled and awaited first. The old “reset migration after legacy rename when unified write was on” path is removed.
> 
> **Rollout:** privacy subfeature + iOS/macOS `FeatureFlag.syncCanUsePatchEndpointForLegacyDeviceRename` (default enabled). Tests cover routing, flag snapshot behavior, migration/repair ordering, and 401 logout for both PATCH rename modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6e030c06a07fc25b18362ae91bc6beca1140ed3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/DeviceInfoUpdateBuilder.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/DeviceInfoUpdateBuilder.swift
@@ -1,0 +1,44 @@
+//
+//  DeviceInfoUpdateBuilder.swift
+//
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct DeviceInfoUpdateBuilder {
+
+    let crypter: CryptingInternal
+    let deviceInfoCodec: DeviceInfoCoding
+
+    func makeUpdate(deviceID: String,
+                    deviceName: String,
+                    deviceType: String,
+                    primaryKey: Data,
+                    protectedKey: ProtectedKey) throws -> UpdateDevices.Update {
+        let encryptedDeviceInfo = try deviceInfoCodec.encrypt(
+            DeviceInfo(name: deviceName, type: deviceType),
+            using: protectedKey)
+        guard encryptedDeviceInfo.utf8.count <= DeviceInfo.maximumEncryptedLength else {
+            throw DeviceInfoMigrationError.encryptedDeviceInfoTooLarge
+        }
+
+        return UpdateDevices.Update(
+            id: deviceID,
+            name: try crypter.encryptAndBase64Encode(deviceName, using: primaryKey),
+            type: try crypter.encryptAndBase64Encode(deviceType, using: primaryKey),
+            info: encryptedDeviceInfo)
+    }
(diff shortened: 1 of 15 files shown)


SharedPackages/BrowserServicesKit/Sources/DDGSync/internal/DeviceInfoUpdateBuilder.swift
```diff
@@ -0,0 +1,67 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct DeviceInfoUpdateBuilder {
```
pballart, line 21: Could we add tests for this? We could mock the crypter and assess that info is always nil for `makeUpdateWithoutUnifiedInfo`.

pballart (approved):
LGTM, just left one small comment but non blocking ✅

PR #6473, opened 2026-08-21
iOS: Contextual Duck.ai follow-ups — chats menu rename, translations, glyph rule, scrim/pill polish

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215757651854873?focus=true

### Description

Follow-ups to the A2 contextual Duck.ai branch (#6398), plus two changes from design/product review.

- Renamed the popup-era chats-menu types — there is no popup any more, only a native `UIMenu`. The pixel name and localisation keys are deliberately unchanged.
- Added the missing `Open Duck.ai` translations for all 25 non-English locales, reusing each locale's existing translation of the same English string.
- Removing the page-context chip no longer leaves a re-attach placeholder behind; re-attaching is the attachment menu's job.
- The address-bar Duck.ai button now shows its contextual glyph whenever the tab has a contextual session — a surface on screen, or a chat to return to — instead of only while a surface is open. Dismissing without prompting leaves neither, so it reverts.
- Darkened the contextual scrim and the suggestion pills' glass, per design review. The pill tint is theme-aware.

### Testing Steps

1. Open a page → tap the Duck.ai icon → the floating input appears and the icon shows the arrow glyph
2. Dismiss without typing → icon returns to the plain Duck.ai glyph
3. Send a prompt, then dismiss the surface → icon keeps the arrow glyph
4. Tap the chats icon in the sheet header → Open Duck.ai / New Chat / Recent Chats / View all chats, header pill shadows fade while open
5. Expand the chat to its own tab (⤢), return to the original tab, tap Duck.ai → fresh floating input, plain glyph
6. Force-quit, relaunch, tap Duck.ai on that tab → still fresh
7. With the page context chip attached, tap its X → chip disappears with nothing in its place; re-attach from the attachment menu
8. Check the scrim and pills in both light and dark appearance

### Impact

Low — the contextual Duck.ai surface is behind `aiChatContextualFloatingInput`. Nothing here changes what a tap does; the glyph, scrim and pill tint are presentational, and the chip change removes an affordance rather than adding one.

#### What could go wrong?

- The glyph could disagree with the tap action → the rule now lives in `DuckAIAddressBarEntry` beside `resolve`, over the same inputs, with tests covering the home-tab and no-chat cases.
- The pill tint could be wrong in one appearance → tint is resolved per `userInterfaceStyle` and the glass effect is rebuilt on a light/dark flip, since `UIGlassEffect`'s tint is fixed at init.

### Quality Considerations

- **The pill tint values have not been reviewed on screen.** They were picked by judgement: the Figma shows the stock `T-Surface-Glass/iOS/Regular-Large` token, so it does not encode the requested darkening. Light 0.06 / dark 0.14, both single constants in `AIChatQuickActionChipView.Constants`, trivial to retune.
- The pill tint is iOS 26 only — the earlier `UIBlurEffect` fallback has no tint to carry it, and was left on `.systemThinMaterial` to match every other pre-26 glass fallback in the app. The darker scrim still applies there.
- `AIChatQuickActionChipViewTests` (SPM target) was not run locally: no `.xctestplan` includes it, and `swift test` on the package fails to build for macOS on a pre-existing `DesignResourcesKitIcons` issue unrelated to this branch.
- `AIChatContextualSheetCoordinatorTests.testExpandRequestRetainsActiveChat` fails on `main` today; the fix is #6467. Rebase once that lands for a clean run.

### Known follow-up (not in this PR)

The address-bar Duck.ai menu fires its shown pixel twice per presentation — `fireAddressBarMenuShown()` sits inside a `UIDeferredMenuElement.uncached` provider, which UIKit resolves more than once per open (observed 136ms apart on one tap). The chats menu in this diff already avoids that by firing from `onMenuWillDisplay`; the address-bar menu should move to the same mechanism.

---

> [!NOTE]
> **Low Risk**
> Presentational and affordance tweaks behind the contextual floating-input flag. No auth, data, or tap-action changes.
> 
> **Overview**
> Follow-up polish for contextual Duck.ai: the address-bar glyph now tracks a **session** (open surface *or* a chat to reopen), not only whether a surface is on screen. Dismissing without a prompt reverts to the plain icon.
> 
> Removing page context no longer leaves a re-attach placeholder; re-attach is only via the attachment menu. “Ask about page” still returns after a new chat.
> 
> The floating input always resigns with the dismiss slide so the keyboard travels with the surface. Sheet and floating input share a darker `ContextualSurfaceScrim` (0.3). Suggestion pills get a theme-aware glass tint on iOS 26.
> 
> Renames chats “popup” types/pixels APIs to “menu” (pixel names and loc keys unchanged) and adds missing **Open Duck.ai** strings for non-English locales.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e91e10f97bba3451a35a098cf2c1728d19c42415. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/AIChat/Sources/AIChat/iOS/Views/QuickActions/AIChatQuickActionChipView.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/iOS/Views/QuickActions/AIChatQuickActionChipView.swift
@@ -42,6 +42,9 @@ public final class AIChatQuickActionChipView: UIView {
         static let glassShadowOpacity: Float = 0.02
         static let glassShadowRadius: CGFloat = 15
         static let glassShadowOffsetY: CGFloat = 8
+        /// Darkens the glass so the pills separate from light page content behind them.
+        static let glassTintAlphaLight: CGFloat = 0.06
+        static let glassTintAlphaDark: CGFloat = 0.14
         /// Outer 8 plus the label group's inner 6, per the design.
         static let glassIconLeadingPadding: CGFloat = 14
         static let glassIconLabelSpacing: CGFloat = 8
@@ -141,13 +144,14 @@ private extension AIChatQuickActionChipView {
                                    iconToLabel: Constants.iconLabelSpacing,
                                    trailing: Constants.trailingPadding)
         case .glass:
-            // The glass is the background: nothing of ours behind it, and no tint holding it lighter than the
-            // page it sits over. It adapts to that page, and its `contentView` keeps the label legible as it does.
+            // The glass is the background: nothing of ours behind it. It adapts to the page it sits over,
+            // darkened by its own tint so the pills read against light page content.
             backgroundColor = .clear
             // No border: it lives on this layer, so it can't scale with the glass's interactive
             // expansion and would sit inside the enlarged pill on touch.
             layer.borderWidth = 0
             installGlassBackgroundIfNeeded()
+            applyGlassEffect()
             hostContent(in: glassBackgroundView?.contentView ?? self)
             applyGlassShadow()
             applyCornerRadius(Constants.height / 2)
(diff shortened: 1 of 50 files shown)


aataraxiaa (approved):
Functionally looks good. Two minor things, more design feedback:
1. in dark mode mode pills are kinda too similar to scrim IMO
2. on dismiaal, the way the input disappears first, then the keyboard, is slightly strange/slow

jaceklyp:
> Functionally looks good. Two minor things, more design feedback:
> 
> 1. in dark mode mode pills are kinda too similar to scrim IMO
> 2. on dismiaal, the way the input disappears first, then the keyboard, is slightly strange/slow

1. leaving as is
2. fixed

PR #6491, opened 2026-08-21
[iOS] Move install logic into a new InstallStatistics package

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1217731844162488?focus=true
Tech Design URL:
CC:

### Description

This PR moves Core's install logic into an InstallStatistics package.

### Testing Steps

1. Check that ATB works as expected
2. Check that the Share and Open extensions work as expected

### Quality Considerations

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes how ATB/install data is stored and which targets/extensions link statistics code; misaligned app groups or entitlements could break pixels/ATB or extension behavior until verified on app and extensions.
> 
> **Overview**
> Introduces a new **`InstallStatistics`** local Swift package and moves **`StatisticsUserDefaults`** out of **Core**, with the app wiring it to **`BrowserServicesKit.StatisticsStore`** via **`InstallStatisticsCompatibility`**. Statistics UserDefaults now resolve their suite through **`BundleGroup.statistics`** / **`STATISTICS_APP_GROUP`** in xcconfig, plists, and entitlements (main app, Open Action, Autofill credential provider) instead of hardcoding `$(GROUP_ID_PREFIX).statistics`.
> 
> **Core** is slimmed: website-data, history capture, bookmarks import/export/caching, and data-import types compile in the **DuckDuckGo** app target (including synced **History** / **WebsiteData** groups) rather than the Core framework. Shared autofill presentation helpers stay in **BrowserServicesKit** only—duplicate copies and Core-local tests are removed, with email truncator tests living under BSK.
> 
> Call sites switch from **`isTargetingMainFrame()`** to **`isTargetingMainFrame`**. UI-test launch overrides use the same statistics app-group helper when clearing or backdating install stats.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78a06cf524498a73579974797cd13b37d7cff403. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jaceklyp:
## Agentic review of this PR

This review came from an automated multi-agent pass — three parallel static-analysis agents, plus a local build and an on-simulator run. Every claim was re-verified against the source or measured on the built artifact before posting.

**This note is the only place flagging the agentic origin** — the 4 inline comments on this PR all come from here, so please read them as part of this. **Dismiss any that are invalid**, no justification needed. I've deliberately limited them to things *this PR causes*; I dropped several notes that turned out to be pre-existing.

### Verification: the app-group migration is safe (measured, not inferred)

The main risk here is that the resolved statistics app group changes and every existing user's ATB / install date / variant is orphaned. **It doesn't.** I built this branch, installed it on an iOS 26.4 simulator, and measured both resolution paths on the same artifact:

| | resolved value |
|---|---|
| `STATISTICS_APP_GROUP` (new path) | `group.com.duckduckgo.statistics` |
| `DuckDuckGoGroupIdentifierPrefix` (old path) | `group.com.duckduckgo` → `+ ".statistics"` → `group.com.duckduckgo.statistics` |

Byte-identical, so a pre-#6491 build and this one read the same container — **the upgrade path is safe by string identity on the built product**, no migration needed. Confirmed at runtime too: after first launch the app had written the following into that exact suite, with the legacy key names unchanged:

```
com.duckduckgo.statistics.atb.key             => "v551-1"
com.duckduckgo.statistics.appretentionatb.key => "v551-1"
com.duckduckgo.statistics.installdate.key     => 1788971334.60327
```

All four rewritten `$(STATISTICS_APP_GROUP)` entitlements resolve to that same group — read from the generated `-Simulated.xcent` files, since `codesign` can't surface entitlements on simulator builds. The Share extension, Open action and Autofill credential provider each launched clean in-process, with no crash reports and no `Info.plist is missing` anywhere in the system log.

**Scope limits, so this isn't over-read:** Debug configuration only. **Alpha and Experimental were not verified** — and those are precisely the configs where the entitlement went from a hard-coded literal to the variable, so they're worth a manual check before merge. I also didn't exercise the VPN extension or a home-screen widget.

Given the above I don't think a regression test pinning the suite name is needed. If you want one anyway, the cheap version belongs in the app-hosted `DuckDuckGoTests` (where `Bundle.main` is `DuckDuckGo.app`) and passes as-is:

```swift
XCTAssertEqual(Bundle.main.appGroup(bundle: .statistics), "\(Global.groupIdPrefix).statistics")
```

iOS/OpenAction/Info.plist
```diff
@@ -22,6 +22,8 @@
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>DuckDuckGoGroupIdentifierPrefix</key>
 	<string>$(GROUP_ID_PREFIX)</string>
+	<key>STATISTICS_APP_GROUP</key>
```
jaceklyp, line 25: **Major — introduced by this PR.** This PR adds `STATISTICS_APP_GROUP` to 5 Info.plists, but **7** targets link `Core.framework`. The two that didn't get it are `iOS/PacketTunnelProvider/Info.plist` and `iOS/Widgets/Info.plist`.

To be clear about causation, since those two files aren't in this diff: the problem is created *here*, by changing which key the store reads. Before this PR `StatisticsUserDefaults.init()` read `DuckDuckGoGroupIdentifierPrefix` (`iOS/Core/global.swift:29-35`) — a key present in **all 7** of those bundles, so construction always succeeded. After this PR it reads `STATISTICS_APP_GROUP`, and `Bundle.appGroup(bundle:)` traps on a miss rather than degrading:

```swift
// SharedPackages/.../FoundationExtensions/BundleExtension.swift:111-117
guard let appGroup = object(forInfoDictionaryKey: appGroupName) as? String else {
    fatalError("Info.plist is missing \(appGroupName)")
}
```

I confirmed this on the built product, not just from the source — from the compiled `.app`:

```
DuckDuckGo.app                    -> group.com.duckduckgo.statistics
ShareExtension.appex              -> group.com.duckduckgo.statistics
OpenAction.appex                  -> group.com.duckduckgo.statistics
AutofillCredentialProvider.appex  -> group.com.duckduckgo.statistics
PacketTunnelProvider.appex        -> ABSENT
WidgetsExtension.appex            -> ABSENT
```

**Not a live crash today** — I verified neither target currently reaches a `StatisticsUserDefaults()` construction, so this is latent. Recording one dead end so nobody repeats it: the tunnel *looks* live via `NetworkProtectionPacketTunnelProvider.swift:43,61` → `PersistentPixel.sendQueuedPixels` → `Pixel.fire` → `URL.makePixelURL` → `AppURLs.swift:179`/`:233` → `StatisticsUserDefaults()`. It isn't reachable: `sendQueuedPixels` early-returns on an empty queue (`PersistentPixel.swift:213-217`), the queue file lives in the process's *own* Application Support dir (`PersistentPixelStoring.swift:84-86`, returning `[]` when absent at `:134-136`), and the tunnel never enqueues — no `persistentPixel.fire`, no legacy `Pixel.fire`/`DailyPixel.fire` in that target.

It's latent but sharp, because Core constructs this store through *invisible default arguments* on hot singletons — `AppURLs.swift:179`/`:233`, `ContentBlocking.swift:50`, `DefaultVariantManager.swift:118`, `PixelKitParameterProvider.swift:36`, `StatisticsLoader.swift:49`, `UserAgentManaging.swift:155`. The next `import Core` usage added to a widget or the tunnel crashes that process at launch, with nothing in the build or CI to catch it.

Cheapest fix — two lines each, matching how PTP already surfaces `VPN_APP_GROUP`/`SUBSCRIPTION_APP_GROUP`:
```xml
<key>STATISTICS_APP_GROUP</key>
<string>$(STATISTICS_APP_GROUP)</string>
```
Worth knowing: neither bundle is *entitled* to the statistics group (PTP's entitlements list only `$(GROUP_ID_PREFIX).netp` + `app-configuration`), so the plist key alone restores today's silent behaviour rather than making shared reads correct. If you'd rather kill the class of bug outright, a non-trapping `appGroupIfPresent(bundle:) -> String?` lets a bundle with no statistics group degrade instead of crashing.

iOS/Core/InstallStatisticsCompatibility.swift
```diff
@@ -0,0 +1,25 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import BrowserServicesKit
+import InstallStatistics
+
+public typealias StatisticsUserDefaults = InstallStatistics.StatisticsUserDefaults
```
jaceklyp, line 23: **Introduced by this PR** — a new Swift 6 diagnostic. Straight from a local build of this branch:

```
iOS/Core/PixelKitParameterProvider.swift:36:52: warning: initializer 'init()' cannot be used in
a default argument value because 'InstallStatistics' was not imported by this file;
this is an error in the Swift 6 language mode
```

That file isn't in this diff, so to be explicit about causation: the warning didn't exist before this PR. `StatisticsUserDefaults` used to live in Core, so a Core file needed no import to name it. Moving it into `InstallStatistics` and re-exposing it through the typealias on this line means a **public** default argument now references an initializer from a module the use site doesn't import:

```swift
// iOS/Core/PixelKitParameterProvider.swift:36 — imports Foundation, BrowserServicesKit, PixelKit
public init(statisticsStore: StatisticsStore = StatisticsUserDefaults()) {
```

Worth noting *why this is the only warning site*, since it looks arbitrary: it's the only **public** default argument, and public default arguments are emitted into the module interface. The siblings are internal and therefore exempt (`AppURLs.swift:233`, `StatisticsLoader.swift:49`, `UserAgentManaging.swift:155`), and `ContentBlocking.swift:50` / `DefaultVariantManager.swift:118` construct in a body rather than a default argument. So it's one line to fix today — but it becomes a hard error for every public site once this target moves to the Swift 6 language mode.

```swift
import Foundation
import BrowserServicesKit
import InstallStatistics   // + add
import PixelKit
```

iOS/LocalPackages/InstallStatistics/Sources/InstallStatistics/StatisticsUserDefaults.swift
```diff
@@ -102,9 +101,13 @@ public class StatisticsUserDefaults: StatisticsStore {
         get {
             return userDefaults?.string(forKey: Keys.duckAIRetentionAtb)
         }
-
         set {
             userDefaults?.setValue(newValue, forKey: Keys.duckAIRetentionAtb)
         }
     }
+
+    public var atbWithVariant: String? {
```
jaceklyp: **Introduced by this PR** — `atbWithVariant` is new here (it wasn't on the old Core type), and it now duplicates the pre-existing protocol extension at `iOS/Core/StatisticsStore.swift:23-27`.

No behaviour change today: I compared both bodies across all four permutations (atb nil; variant nil; variant `""`; both set) and they agree — `guard let atb = atb` vs the `guard let atb` shorthand, same `atb + (variant ?? "")`. **No ATB wire drift.**

The issue is that `atbWithVariant` is *not* a `StatisticsStore` requirement (`SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/Statistics/StatisticsStore.swift:21-30`), so selection is static — and the copy production runs is not the copy the new tests cover:

| receiver static type | implementation that runs |
|---|---|
| `StatisticsStore` — `AppURLs.swift:237,258,270,283,316`, `PixelKitParameterProvider.swift:41`, `FeedbackSender.swift:94` | the **Core** extension |
| concrete `StatisticsUserDefaults` — `InstallStatisticsTests/StatisticsUserDefaultsTests.swift:177,183,190` | **this** new copy |

So every production call site runs Core's copy, while all three new tests exercise this one. Edit either and they diverge with a green build.

Simplest fix is to drop the new member, leaving one definition:
```swift
// StatisticsUserDefaults.swift:109-112 — remove
public var atbWithVariant: String? {
    guard let atb else { return nil }
    return atb + (variant ?? "")
}
```
If the package should instead be self-contained, the complete version is to promote `atbWithVariant` to a requirement on BSK's `StatisticsStore` and drop the duplicates — but leaving two copies is the option I'd avoid.

SharedPackages/Infrastructure/SystemFrameworksExtensions/Tests/FoundationExtensionsTests/BundleExtensionTests.swift
```diff
@@ -0,0 +1,27 @@
...
+//  limitations under the License.
+//
+
+import XCTest
+@testable import FoundationExtensions
+
+final class BundleExtensionTests: XCTestCase {
+
+    func testWhenRequestingStatisticsAppGroupThenReturnsStatisticsInfoPlistKey() {
+        XCTAssertEqual(BundleGroup.statistics.appGroupKey, "STATISTICS_APP_GROUP")
```
jaceklyp, line 25: **Minor**, on the new test file added here. This assertion restates the implementation line-for-line — `BundleGroup.statistics.appGroupKey` is literally `case .statistics: return "STATISTICS_APP_GROUP"` at `BundleExtension.swift:170-171` — so it can't fail unless someone edits both sides in the same direction.

Since you're adding the file anyway, a table gets real value for the same line count, and would catch a typo in any case rather than just the new one:

```swift
func testAppGroupKeysMatchInfoPlistKeys() {
    let expected: [(BundleGroup, String)] = [
        (.netP, "NETP_APP_GROUP"), (.ipc, "IPC_APP_GROUP"), (.dbp, "DBP_APP_GROUP"),
        (.subs, "SUBSCRIPTION_APP_GROUP"), (.appConfiguration, "APP_CONFIGURATION_APP_GROUP"),
        (.statistics, "STATISTICS_APP_GROUP")
    ]
    for (group, key) in expected { XCTAssertEqual(group.appGroupKey, key) }
}
```

The test with actual teeth would cover `appGroup(bundle:)` itself — the hit and the miss — but that's not writable while the miss path is a `fatalError`, which is pre-existing and not yours to fix here. Flagging it only because it's the same root as the Major comment I left on `iOS/OpenAction/Info.plist`.

jaceklyp (approved):

samsymons:
@jaceklyp Thanks for the comments! I've addressed everything and done another review pass myself, which turned up no issues via both manual testing and getting Astra to review. I also ran an E2E run [here](https://github.com/duckduckgo/apple-browsers/actions/runs/34401253308) which was green - given all of that, I'll merge this.

PR #6492, opened 2026-08-23
Hand PDF tabs to Duck.ai as page context

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1217576322698640?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1217621296618559
CC:

### Description

Duck.ai can now answer about a PDF you have open, the same way it answers about a web page.

Page context has only ever had one carrier — the page as markdown — so a PDF tab had nothing to hand over and was marked non-attachable. Per the tech design the same payload now carries the document itself instead: its bytes, plus a `mimeType` on every page-context payload saying which carrier to read. A context is one or the other, never both.

Native reads the bytes straight out of the web view, so there's no second download and it works for background tabs. This covers every route a page context takes to Duck.ai: the sidebar's initial pull, the push on navigation, the sidebar tab picker, and the omnibar's "Add Page Content".

Duck.ai keeps ownership of validation — page count, tier limits, encrypted files, model support — and its existing error messaging, as agreed on the tech design. The one thing native refuses is a PDF over 30 MB, which it reports as non-attachable without sending anything; that's a guard on the JS bridge, not a product limit.

### Testing Steps

1. In Debug → Custom Configuration → set the Duck.ai URL to `https://euw-serp-dev-testing20.duck.ai/` (the FE support for the new payload isn't in production yet).
2. Open a PDF in a tab  → open the Duck.ai sidebar.
3. With "Automatically add page content" ON: the page-context chip appears → ask something only answerable from the PDF → the answer comes from the document.
4. Turn "Automatically add page content" OFF, reload the PDF tab, open the sidebar → the chip shows but nothing is attached → click "Ask about page" → the PDF is attached and answers.
5. Navigate PDF → a normal web page → back to a PDF, asking a question each time → each answer uses the right source; the web page still attaches as text.
6. In the sidebar, use the `@` tab picker to attach a *different* PDF tab → it attaches and answers.
7. From the address bar omnibar, use "Add Page Content" to attach a PDF tab → same.
8. Open a PDF larger than 30 MB → no attachment, no error state, chat still usable.
9. Debug → Feature Flag Overrides → disable `aiChatPdfPageContext`, reload the PDF tab → previous behaviour: no page context for PDFs.

### Impact

Medium — new behaviour on a page type that previously offered no context at all, and it ships on.

`aiChatPdfPageContext` is `defaultValue: .enabled` and `pdfPageContext` is not in privacy-configuration yet. A subfeature that's missing from config falls back to the default, so until that lands the flag is on for everyone and can't be turned off remotely except by killing the whole `aiChat` feature. That's deliberate for this release; adding the subfeature is a follow-up.

#### What could go wrong?

- The document bytes come from a private WebKit API. If a future OS stops vending them, the PDF simply isn't attached (no crash, no error UI) → mitigated by a `responds(to:)` guard and a `document_unavailable` pixel so we'd see it.
- A very large PDF crossing the bridge as base64 could hurt memory → mitigated by the 30 MB ceiling, which drops the bytes rather than sending them. Note the bytes are still read out of the web view before being measured, so a very large PDF is a transient allocation either way.
- A tab that navigates mid-read could attach the previous document → mitigated by the pre-read document-match guard and a post-read check that the tab is still on the same URL.

### Quality Considerations

`mimeType` is required on the wire but decodes with a `text/html` default, because the payload Content-Scope-Scripts returns doesn't carry one — without that, every HTML page-context collection would fail to decode. Unit tests cover the carrier invariant, the ceiling, the magic-byte check and the decode fallback.

Follow-ups, neither blocking: add the `pdfPageContext` subfeature to privacy-configuration so the flag becomes remotely controllable, and consider refusing on `expectedContentLength` at navigation-response time so an oversized PDF is never read into memory at all.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> New data path sends large base64 payloads across the JS bridge and depends on a private WebKit API; mitigations include size limits, navigation checks, and failure pixels, but memory spikes and OS API changes remain plausible.
> 
> **Overview**
> **PDF tabs can now be handed to Duck.ai as page context** instead of being blocked or empty. Page context gains a second carrier: **`mimeType`** (always on the wire) plus optional **`data`** (base64 document bytes), separate from markdown in **`content`** so truncation/chunking logic does not touch PDFs.
> 
> **Native reads PDF bytes from the web view** via **`WKWebView+MainResourceData`** (private WebKit API) and **`DocumentPageContextProvider`**, with a **30 MB ceiling**, PDF magic-byte validation, and navigation guards so the wrong document is not attached after back/forward or mid-read navigation. **`PageContextTabExtension`** bypasses the page-context user script for PDFs (sidebar auto-attach, signals-only metadata when auto-attach is off, omnibar/tab picker via **`extractPageContext`**).
> 
> **Attachability and delivery** use **`hasAttachedPage`** / **`isDocument`**; PDFs are no longer treated as blocklisted **`pdf`** when the feature is on. New extraction outcomes **`document_unavailable`** and **`documentTooLarge`** are wired into pixels. Gated by **`aiChatPdfPageContext`** (default enabled).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbc6333f31f081e549e8a08d023b4e4c3fadc8d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/LocalPackages/FeatureFlags-macOS/Sources/FeatureFlags/FeatureFlag.swift
```diff
@@ -723,6 +727,8 @@ extension FeatureFlag: FeatureFlagDescribing {
             Config(source: .remoteReleasable(AIChatSubfeature.supportsSyncChatsDeletion))
         case .aiChatMultiplePageContexts:
             Config(source: .remoteReleasable(AIChatSubfeature.multiplePageContexts), category: .duckAI)
+        case .aiChatPdfPageContext:
+            Config(defaultValue: .enabled, source: .remoteReleasable(AIChatSubfeature.pdfPageContext), category: .duckAI)
```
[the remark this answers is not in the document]
tomasstrba: Intentional for this release — we're shipping PDF page context on by default.

You're right about the mechanics: a subfeature missing from privacy config falls back to `defaultValue`, so until `pdfPageContext` lands there the only remote off-switch is the `aiChat` parent feature. Adding the subfeature is a tracked follow-up, and the PR description now states the rollout accurately rather than claiming a remote-releasable flag.

macOS/DuckDuckGo/AIChat/DocumentPageContextProvider.swift
```diff
@@ -0,0 +1,98 @@
...
+import Foundation
+
+/// Builds page context for tabs the page-context user script can't read: documents rendered by
+/// WebKit itself (PDFs today). Instead of markdown in `content`, these contexts carry the document's
+/// bytes in `data`, which Duck.ai treats as an attachment.
+///
+/// Validation stays on the Duck.ai side — page count, tier limits, encryption, model support — with
+/// one exception: `maxDocumentBytes`, a safety ceiling so a huge document can't be base64'd across
+/// the JS bridge. Over it the context still says what the tab is, but hands over nothing.
+enum DocumentPageContextProvider {
```
SabrinaTardio, line 29: This could live in the shared AIChat module.
I can also move it in the iOS PR

macOS/DuckDuckGo/Common/Extensions/WKWebView+MainResourceData.swift
```diff
@@ -0,0 +1,81 @@
...
+        await mainResourceData(timeout: 5)
+    }
+}
+
+@objc private protocol WKWebViewMainResourcePrivate {
+    @objc(_getMainResourceDataWithCompletionHandler:)
+    func getMainResourceData(completionHandler: @escaping (NSData?, NSError?) -> Void)
+}
+
+extension WKWebView: MainResourceDataProviding {
```
SabrinaTardio, line 42: This could probably go in a new package within SystemFrameworkExtensions to be shared with iOS, again I could do it in the iOS PR

macOS/DuckDuckGo/AIChat/DocumentPageContextProvider.swift
```diff
@@ -0,0 +1,98 @@
...
+    /// restore doesn't re-fire the response delegate).
+    static func isSupportedDocument(mimeType: String?, url: URL) -> Bool {
+        if let mimeType, !mimeType.isEmpty {
+            return mimeType.lowercased() == AIChatPageContextData.pdfMIMEType
+        }
+        return url.pathExtension.lowercased() == "pdf"
+    }
+
+    @MainActor
+    static func makeDocumentContext(webView: MainResourceDataProviding,
```
SabrinaTardio, line 57: Is it worth reusing AIChatPDFInspector.inspect(data:mimeType:)

SabrinaTardio (commented):
I tested and found one issue (which is probably on the FE side)
but it seems like auto-attach = off is not respected for PDF.

macOS/DuckDuckGo/AIChat/DocumentPageContextProvider.swift
tomasstrba: Good suggestion, but in [tech design](https://app.asana.com/1/137249556945/project/481882893211075/task/1217621296618559?focus=true) we specifically agreed on skipping local validation so the user gets a real error message from Duck.ai instead of silently missing the attached chip.

tomasstrba:
@SabrinaTardio, as discussed on MM, attaching the PDF with the setting off is expected, and the chip only drops on subsequent navigations.

Regarding the refactoring, I'm keeping the logic where is since you've already refactored it in your PR

SabrinaTardio (approved):

PR #6514, opened 2026-08-25
[iOS] Drive the Duck.ai usage-limit footer from the shared resolver

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201188456623839/task/1217605036256290?focus=true
Tech Design URL:
CC:

### Description
Use the shared DuckAiUsageWarningViewModel for iOS duck.ai token usage messages

### Testing Steps
1. Enable the `utiDuckAIWarnings` flag, then Settings → Debug → AI Chat → Duck.ai Usage Warnings → seed "Weekly 90%". Open the Duck.ai input → the card shows "90% of weekly limit / Resets in …" with the ring filled to 90%, and no web usage banner appears above the chat.
2. Tap the CTA → the suggested model is applied; tap the chevron instead → the model picker opens anchored to the chevron. Tap ✕ → the card goes; reopen the input and it stays hidden, then seed "Weekly limit reached" and confirm it returns.
3. On a free-tier account seed "Weekly 90%" → no card. Seed "Weekly limit reached" → the card appears with a "Try for free" / "Subscribe" CTA that opens the subscription flow. In a fire tab, no card in either case.

### Impact
Low

#### What could go wrong?
The card could show the wrong copy, offer a model switch that the current draft's attachments can't support, or fail to reappear after a dismissal — all behind an internal-only feature flag.

### Notes to Reviewer
`supportsNativeUsageWarnings` is gated on `utiDuckAIWarnings && supportsNativeChatInput && isNativeStorageBridgeAvailable` — the same three conditions the footer itself needs, mirroring how `supportsPromoCards` requires the native chat input. A looser gate would tell the FE to hide its banner on builds where the native card has nowhere to appear, so a user near their limit would see no warning at all.

---

> [!NOTE]
> **Medium Risk**
> Changes subscription funnel entry points and model-switch CTAs driven by live usage snapshots; mis-gating `supportsNativeUsageWarnings` could hide both native and web warnings.
> 
> **Overview**
> iOS **replaces** its local usage-footer stack (`UTIFooterWarning`, resolver, and warning providers) with the shared **`DuckAiUsageWarningViewModel`**, so limit messaging, dismissal ladders, and CTAs match macOS and the web logic.
> 
> **`DuckAiUsageLimitsStore`** now builds that view model (flag, native storage, dismissal store, tier/trial/fire-mode inputs) instead of only exposing raw limits. **`UTIFooterController`** refreshes, dismisses, and runs actions through the view model; **`UTIFooterMessageMapper`** maps **`DuckAiUsageWarning`** into localized card copy and action labels (model switch, free model, subscribe / try for free).
> 
> The footer card gains **`UTIFooterActionButton`** with an optional **model-picker chevron** tied to the toolbar menu; **`UnifiedToggleInputCoordinator`** routes CTAs to model selection (with attachment-aware suggesting) or **`presentPurchaseFlow(origin:)`** via new **`addressBarUsageLimit`** / **`duckAIUsageLimit`** funnel origins. Fire tabs no longer get a usage store; contextual mode passes **`usageLimitsStore`** instead of a footer warning provider.
> 
> Shared config adds **`supportsNativeUsageWarnings`**, advertised to the FE only when **`utiDuckAIWarnings`**, native chat input, and the storage bridge are all available so the web usage banner is suppressed when the native card can show.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f931a54c37345682e1b55748700cde00537f79b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tomasstrba (commented):
Free user:
TS1/ Free user can only get the daily usage limit. I'd recommend to add debug option for this case.

TS2/ Just raising that the "Try for free" button is aligned with model picker, which looks slightly off. Maybe it's the intention.
Image

TS3/ I got this error when clicking on "Try for free"
Image

Paid user:
TS4/ There was no button to switch to a model at 90 percent of the weekly usage. I believe this is the current implementation of the view model, until we integrate latest changes on the frontend. Testing steps are likely wrong, but raising it in case not
Image

tomasstrba (approved):
LGTM! ✔️

PR #6520, opened 2026-08-25
[iOS] Attempt to fix the link cleaner crash

samsymons:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214294661819890/task/1217844347479204?focus=true
Tech Design URL:
CC:

### Description

This PR reworks the way link cleaning is handled to avoid issues when processing extremely large URLs.

### Testing Steps

1. Run the tests in https://privacy-test-pages.site/privacy-protections/query-parameters/ and check that they look correct
2. Smoke test URL navigation in general
3. Check that the added test suite looks good

### Quality Considerations

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes privacy-sensitive navigation URL cleaning; incorrect stripping could leak tracking params or break links, though reference tests and new edge-case tests mitigate regression risk.
> 
> **Overview**
> **Reworks tracking-parameter removal in `LinkCleaner.cleanTrackingParameters`** so it no longer builds `URLComponents` / `URLQueryItem` for every query key—addressing crashes and heavy work on extremely large URLs.
> 
> The new **`PercentEncodedQueryFilter`** walks the query on `url.relativeString` as UTF-8, splits on encoded `&`, removes only configured tracking names, and copies surviving items unchanged so percent-encoding and odd delimiter shapes stay intact. Fragment handling ignores `#` in the query portion; relative URLs and `baseURL` are preserved when rebuilding via `URL(string:relativeTo:)`. Parameter-name matching is capped by the longest configured tracking name so huge page-controlled names are not fully decoded for comparison.
> 
> **Tests** add coverage for encoded values, empty/query-only relative URLs, million-character values and names, and percent-encoded parameter names that should not match `utm_source`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35638d77dbb6df9202ddb616eab4bca0a41b6cdd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jaceklyp (approved):

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/LinkProtection/LinkCleaner.swift
```diff
@@ -93,28 +93,150 @@ public class LinkCleaner {
             return url
         }
 
-        guard var urlsComps = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
+        let trackingParams = TrackingLinkSettings(fromConfig: privacyConfig).trackingParameters
+        guard let cleanedURL = PercentEncodedQueryFilter(parameterNames: trackingParams).filter(url: url) else {
```
jaceklyp, line 97: **Major:** The crash may relocate rather than disappear — the largest allocation site on this navigation is untouched and runs *first*.

`requestTrackingLinkRewrite` is an `else if` chain, and `cleanTrackingParameters` is the **last** branch (`LinkProtection.swift:109-127`):

```
:109  extractCanonicalFromAMPLink(...)   <- runs first, calls ampFormat(matching:) at LinkCleaner.swift:36
:114  ampExtractor.urlContainsAMPKeyword(...)
:127  cleanTrackingParameters(...)       <- what this PR fixes
```

Measured on the *same* 5 MB URL (non-AMP host, so nothing short-circuits; peak live-heap high-water above a baseline taken after the URL exists):

| site | peak | wall |
|---|---|---|
| `ampFormat(matching:)` — `LinkCleaner.swift:36` | **3.20x input** (15.27 MB) | 31.2 ms |
| `urlContainsAMPKeyword` body — `AMPCanonicalExtractor.swift:132` | 0.00x | 3.8 ms |
| OLD `cleanTrackingParameters` | 2.01x | 9.1 ms |
| NEW `cleanTrackingParameters` | 0.00-1.00x | 14.6 ms |

So `ampFormat` peaks **higher than the 2.01x site this PR removes** (4.88x on a `google.*` host, where the first pattern engages past its literal prefix) — that's ICU transcoding the whole string to UTF-16 for `NSRegularExpression`. Same URL, same main thread, same navigation, and **no length bound**: the only guard in the entire path is `AMPCanonicalExtractor.Constants.maxURLLength = 80_000`, which sits in the branch *after* `ampFormat` has already allocated.

Under the same memory pressure the OOM abort still fires — it just moves off `__StringStorage.create` to a CF/ICU allocation. **That changes the Sentry grouping, so this will look like it worked even if it didn't.** Worth stating in the PR description so the fix isn't read as complete.

Suggested follow-up (not this PR), reusing the 80k precedent already in the codebase:

```swift
public func ampFormat(matching url: URL) -> String? {
    let ampStr = url.absoluteString
    // ICU transcodes the whole string to UTF-16; measured 3.2x peak heap on a 5 MB URL.
    guard ampStr.count < AMPCanonicalExtractor.Constants.maxURLLength else { return nil }
    let ampFormats = TrackingLinkSettings(fromConfig: privacyConfig).ampLinkFormats
    ...
```

Deliberately **not** a cap on `cleanTrackingParameters`: after this PR it's the cheapest site in the chain and the only one with a privacy benefit, so bailing out here would mean navigating with the parameters intact — the wrong trade.

Two things I checked that turned out fine, so you don't re-verify them: `url.absoluteString` re-evaluated per loop iteration at `LinkCleaner.swift:36` is free (0.000 ms, 0.00x — it returns shared storage), and `isDuckDuckGoSearch` measures 4.51x peak on a 5 MB URL but never sees the same URL, since `TabViewController.swift:3223` excludes DDG URLs from reaching `requestTrackingLinkRewrite`.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/LinkProtection/LinkCleaner.swift
```diff
@@ -93,28 +93,150 @@ public class LinkCleaner {
...
+        action(itemStart..<range.upperBound)
+    }
+
+    private func isTrackingParameter(
+        in bytes: String.UTF8View,
+        itemRange: Range<String.Index>
+    ) -> Bool {
+        let nameEnd = bytes[itemRange].firstIndex(of: ASCII.equalsSign) ?? itemRange.upperBound
+        let name = bytes[itemRange.lowerBound..<nameEnd]
+        return parameterNames.contains { name.elementsEqual($0.utf8) }
```
jaceklyp: **Major:** Main-thread CPU regression — this makes the new path **1.15-1.6x slower than the code it replaces**, and it's free to fix.

`parameterNames` is a 25-element `[String]` (verified in both `iOS/Core/ios-config.json` and `macOS/DuckDuckGo/ContentBlocker/Resources/macos-config.json`, `features.trackingParameters.settings.parameters`, both `state: enabled`), linearly scanned — and `isTrackingParameter` is called **twice per query item**, once from `inspectQuery` (line 160) and once from `rebuildURLString` (line 199). A 100k-parameter URL costs 5,000,000 byte-wise comparisons.

Measured (Swift 6.3, `-O`, arm64):

| shape | OLD | NEW as written | NEW + `Set` |
|---|---|---|---|
| 100k params (1.42 MB) | 27.9 ms | **43.3 ms** | **18.9 ms** |
| 5 MB value on a removed param | 9.1 ms | 14.6 ms | 13.5 ms |

Current:

```swift
private let parameterNames: [String]

init(parameterNames: [String]) {
    self.parameterNames = parameterNames
}

// ...
return parameterNames.contains { name.elementsEqual($0.utf8) }
```

Suggested:

```swift
private let parameterNames: Set<String>

init(parameterNames: some Sequence<String>) {
    self.parameterNames = Set(parameterNames)
}

// ...
private func isExcludedParameter(in source: String, bytes: String.UTF8View, itemRange: Range<String.Index>) -> Bool {
    let nameEnd = bytes[itemRange].firstIndex(of: ASCII.equalsSign) ?? itemRange.upperBound
    return parameterNames.contains(String(source[itemRange.lowerBound..<nameEnd]))
}
```

That's ~3 lines, keeps peak heap at 0.88x, produces byte-identical output, and makes the new path **faster than the old code** rather than slower. (The rename also drops the one genuinely new introduction of "tracking" in this diff — the type is `PercentEncodedQueryFilter(parameterNames:)` and doesn't need to know *why* a name is on the list. `trackingParameters` / `TrackingLinkSettings` / `cleanTrackingParameters` are config-schema and public API, so those stay.)

**Important — do not "simplify" this into a single pass.** The obvious next step is to delete `QueryInspection` and reserve `bytes.count` up front. I measured that variant: it's faster still, but it reintroduces a **full 1.00x (5 MB) single allocation on the crash shape**, i.e. it undoes the fix. The exact `reserveCapacity` computed by `inspectQuery` is what makes the crash-shape peak 0.00x instead of 1.00x — it's the load-bearing part of this design. **Keep the two passes; change only the lookup.**

One caveat: constructing the 25-element `Set` costs 0.51 us, versus 0.15 us for a linear scan over a typical 6-parameter URL — so building it per navigation is a small net loss at the typical end. `TrackingLinkSettings(fromConfig:)` is already rebuilt on **every** call (line 96, and again at line 35), so caching it on `LinkCleaner` keyed off the config version fixes both, and also drops the per-navigation `config.settings(for:)` dictionary lookup + `as? [String]` bridge that old and new code both pay.

jaceklyp (commented):
Reviewed the rework against the old implementation with a differential harness (~1.1M cases incl. all 20 privacy-reference fixtures) and a memory/CPU benchmark. **Parameter-stripping behaviour is preserved to the byte** and the memory win is real (peak 2-8x input -> 0-1x), so the core of this is sound. Two majors below: one on whether the crash actually goes away, one on a main-thread CPU regression that's a ~3-line fix. Minors (silent fail-open at line 151, the empty-port normalisation change, and the fact that 4 of the 6 new tests pass against the old crashing implementation) I can add separately if useful.

PR #6528, opened 2026-08-26
iOS: Lower the collapsed bottom floating capsule

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217377416813355
Tech Design URL: N/A
CC: N/A

### Description
- Sit the collapsed bottom floating capsule 12pt lower on the device by reducing its bottom rest padding, rather than adding space above the pill
- Drive the chrome-collapse morph to that lower rest so the floating bar animates down into the capsule
- Keep the web-view footer inset tracked to the pill so page-fixed content does not gain an extra gap above it
- Leave the top address-bar capsule unchanged; clamp the reduction so the pill cannot leave the screen on devices with no home indicator

### Testing Steps
1. Enable floating UI and use the bottom address bar on a home-indicator iPhone
2. Load a site with a domain, then scroll to collapse the chrome into the capsule pill
3. Confirm the collapsed pill sits lower toward the home indicator, not with extra empty space above it
4. Confirm the collapse animation moves the bar down into that lower rest
5. Confirm a page with a `position: fixed` footer still clears the pill (no overlap, no new 12pt gap)
6. Switch to the top address bar, collapse chrome, and confirm the top capsule rest is unchanged
7. Repeat on a device/simulator without a home indicator and confirm the pill stays on-screen

### Impact
**Impact Level: Low**

#### What could go wrong?
The pill could sit too close to the home indicator or overlap a page-fixed footer → mitigated by clamping rest padding to the screen edge and keeping the web-view inset tracked to the pill (plus `fixedElementClearance`).

### Quality Considerations
- Bottom-only change; top address-bar rest padding is unchanged
- On devices with no home-indicator inset, the 12pt reduction clamps at the screen edge
- Unit tests cover the 12pt lower rest, the clamp, and that obscured height tracks the pill instead of adding top padding

### Notes to Reviewer
This is the base PR for a stacked floating-UI series. Later PRs should target this branch, not `main`.

### DIFF
--- a/iOS/DuckDuckGo/FloatingUI/FloatingDomainCapsuleController.swift
+++ b/iOS/DuckDuckGo/FloatingUI/FloatingDomainCapsuleController.swift
@@ -26,12 +26,37 @@ final class FloatingDomainCapsuleController {
 
     static let handoffBandHalfWidth: CGFloat = 0.02
 
-    /// Gap between the pill and the edge of the safe area at its resting position.
+    /// Gap between the pill and the edge of the safe area at its resting position. Top address bar
+    /// uses this as-is; the collapsed bottom capsule sits `restBottomInsetReduction` closer to the
+    /// device bottom so the chrome morphs down into the pill rather than gaining space above it.
     static let restEdgePadding: CGFloat = 8
+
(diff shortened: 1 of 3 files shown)


brindy (approved):
LGTM

PR #6531, opened 2026-08-26
Add separate pixel for the Add to Home Screen share sheet action

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1217866996011787
Tech Design URL: N/A
CC:

### Description
Picking "Add to Home Screen" in the share sheet was counted in the catch-all `m_sharesheet_activity_other` pixel, so we had no way to tell how often people use it. It now reports its own pixel, which is what the task asks for before we decide whether to do anything about those shortcuts opening in a Safari web view.

The action is only offered to non-Safari apps on iOS 16.4 and later, so on older versions nothing changes and the catch-all pixel still applies.

### Testing Steps
Note: Run on iOS 16.4+ or later with pixel logging visible.
1. Open any website, tap the share icon in the browsing menu.
2. Select "Add to Home Screen". 
3. Ensure that `m_sharesheet_activity_addtohomescreen_ios_phone` fires (or `_ios_tablet` for iPad), and the `m_sharesheet_activity_other` does not.
4. Repeat steps 1-2 but now picking "Copy". Ensure that `m_sharesheet_activity_copy` still fires, confirming no other branch regressed.
5. Repeat steps 1-2 but now picking an optionw with no dedicated pixel (e.g. Messages). Ensure `m_sharesheet_activity_other` fires as before.

### Impact
None — telemetry only, no user-visible behaviour change.

#### What could go wrong?
The share sheet reports the picked action as an opaque `UIActivity.ActivityType`, so matching the wrong value would silently keep counting these as `_other` → mitigated by matching UIKit's own `UIActivity.ActivityType.addToHomeScreen` constant instead of a hardcoded identifier string, so a rename would be a compile error rather than silent data loss.

### Quality Considerations
`addToHomeScreen` is only available from iOS 16.4, and the deployment target is 15.6, so the constant is resolved through an `#available` check that yields `nil` on older versions; the switch subject is non-optional, so the pattern cannot match there and those users keep falling through to `m_sharesheet_activity_other`.

The pixel carries no parameters beyond the standard ones and is declared in `browsing_menu.json5` alongside the other browsing-menu pixels. None of the pre-existing `m_sharesheet_*` pixels have JSON5 definitions; backfilling them is out of scope here.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Telemetry-only change with no user-facing behavior; incorrect activity matching would under-report the new pixel but would not affect app functionality.
> 
> **Overview**
> **Add to Home Screen** in the browsing-menu share sheet now fires **`m_sharesheet_activity_addtohomescreen`** instead of being lumped into **`m_sharesheet_activity_other`**, so usage can be measured before any product decisions about home-screen shortcuts.
> 
> The new **`PixelEvent`** case and **`browsing_menu.json5`** definition match the existing share-sheet activity pixels. In **`firePixelForActivityType`**, **`UIActivity.ActivityType.addToHomeScreen`** is only matched on **iOS 16.4+** (via `#available`); older OS versions keep reporting **`_other`** because that activity type does not exist there.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f1a1ac31fdae1fdab36775f4f5e178315e98f43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andremartingo (approved):
LGTM 👍

miasma13:
> Warnings
> ⚠️	
> The `m_` (or `m-`) pixel name prefix is no longer recommended. Please group pixels by app feature name instead.
> 
> Found these new pixel definitions:
> 
> * `return "m_sharesheet_activity_addtohomescreen"` in `iOS/Core/PixelEvent.swift`
> 
> Note: for iOS the platform suffix is added automatically by the pixel pipeline, so you do not need to include `ios` (or `m_`/`m-`) in the pixel name. The same convention applies to macOS.
> 
> ⚠️	
> Legacy iOS pixel system is deprecated – use `PixelKit` instead. (`Pixel`, `DailyPixel`, `UniquePixel`, `TimedPixel`, and `PersistentPixel` are deprecated). See https://app.asana.com/1/137249556945/project/1208546505108826/task/1216768405353137?focus=true
> 
> Found these new uses:
> 
> * `Pixel` in `iOS/DuckDuckGo/TabViewControllerMenuBuilderExtension.swift`: `Pixel.fire(pixel: .shareSheetActivityAddToHomeScreen)`
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [4f7f7cf](https://github.com/duckduckgo/apple-browsers/commit/4f7f7cfb3ec4b76084cd542eee21e8cf6a997c5e)

Chatted with @federicocappelli about the above as he is doing some work around migrating existing iOS pixel firing code. He is aware of the legacy approach here and it will be addressed as a whole in the upcoming pixel migration changes.

PR #6444, opened 2026-08-20
[iOS] Remove obsolete SwitchBar and Duck.ai full-mode code

aataraxiaa:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217637451822545

### Description

Third and final UTI cleanup PR. Removes the superseded SwitchBar toggle experience, its funnel and retention metrics, and the old Duck.ai full-mode wrapper. Shared UTI components and the live iPad omnibar path remain.

Depends on #6434.

### Testing Steps

1. On iPhone, use the input from a new tab and a webpage: submit a search, switch to Duck.ai, and submit a prompt → both paths work normally.
2. Open a Duck.ai tab and use its native input → the browser chrome stays hidden and no old full-mode branding appears.
3. With a hardware keyboard connected, focus the input on an iPhone Duck.ai tab → the native input opens without falling back to the deleted editing screen.
4. Open **Settings → AI Features**, switch between **Search** and **Search and toggle**, and relaunch → the picker remains available and the selection persists.
5. On iPad, edit the omnibar on a new tab and webpage, then open Duck.ai → the classic omnibar remains and Duck.ai uses the normal address bar.
6. On iPad, type `https://duck.ai` into the omnibar → Duck.ai navigates normally.

### Automated Testing

- [x] Diff whitespace, deleted-symbol, resource, and project-file checks
- [x] Focused affected unit tests
- [x] iPad simulator build, launch, and direct Duck.ai URL regression check
- [ ] Full unit-test suite
- [ ] iPhone simulator build and manual regression steps above
- [ ] Unified Toggle Input, iPad Duck.ai controls, and AI settings Maestro flows

### Impact

Medium: this removes several legacy input and Duck.ai navigation paths. A missed live caller could break iPhone AI input or the retained iPad omnibar.

#### What could go wrong?

iPhone could fall back to a deleted editing path, or iPad could inherit iPhone-only behavior → mitigated by device-specific tests and the manual regression matrix above.

### Quality Considerations

No storage, privacy, or network behavior changes are intended. Existing UTI settings and live subfeature flags remain unchanged.

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Medium Risk**
[...]

PR #6537, opened 2026-08-26
[iOS] Remove the unreachable Duck.ai modal and container chat surface

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201011656765697/task/1217881001402096?focus=true
Tech Design URL:
CC:

### Description

Duck.ai has opened in a tab since #6444 removed the last caller of `AIChatViewControllerManager`, leaving both of its presentation modes — the `.modal` sheet and the `.container` embed — reachable only from tests. This deletes the manager and the `AIChatViewController` stack it exclusively owned, along with the title bar, view model, web view controller, request-auth handler, user-agent protocol and the `webViewBackgroundColor` asset that had no other consumers. The live in-tab path (`openAIChatInTab`, `AIChatContentHandler`, the contextual sheet) is untouched, and it still fires `m_aichat_open` and sets the `.aiChat` feature-discovery flag, so no measurement changes.

### Testing Steps

1. Open Duck.ai from several entry points — the address bar, the tab switcher button, the NTP and a Duck.ai widget/lock-screen deep link — and confirm each lands in a tab with no title bar or sheet.
2. Start voice mode from the NTP and from the voice deep link; confirm both open in a tab.
3. Submit a prompt, then burn with the fire button and switch in and out of fire mode; confirm chat history clears and the app stays responsive.

### Impact and Risks

Low

#### What could go wrong?

If some entry point still reached the sheet through a path not covered by the call-graph sweep, it would now do nothing instead of presenting.

---

> [!NOTE]
> **Low Risk**
> Large deletion of unreachable UI with the live in-tab path unchanged; main regression risk is a missed entry point that still expected the sheet.
> 
> **Overview**
> Removes the **dead modal sheet and embedded-container Duck.ai surface** that nothing in the app called anymore after chat moved into tabs. The deletion spans `AIChatViewControllerManager`, the shared `AIChatViewController` / web view stack (view model, title bar, request-auth handler, user-agent protocol), related assets and strings, and the manager integration tests.
> 
> **Deep links** no longer skip opening when a modal sheet was already up; they always dismiss and route through the tab-based `openAIChat` path. **Fire / subscription wiring** is trimmed: `FireExecutor` no longer notifies delegates around AI history burns (history cleaning still runs), `MainViewController` drops manager delegate hooks and browsing-mode session kills, and `SubscriptionAIChatStateHandler` only fires `onSubscriptionStateChanged` without a sticky refresh flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70bf80cd2893cb06ed379461e5b87de8d168dd80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aataraxiaa (approved):
👍🏼

PR #6559, opened 2026-08-28
Retry Keychain read on startup instead of crashing immediately

diegoreymendez:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217908690610540

### Description

When launched as a login item, the app can start before `loginwindow` has finished unlocking the Keychain. When this happens, the app crashes immediately with a `fatalError`.

This crash matches our top crasher by event count.  According to Sentry data 57% of affected launches have `launchd` as the parent process, consistent with a login-item timing race.

### Testing Steps

Reproduce with the released app:

1. Quit DuckDuckGo
2. Lock the login keychain: `security lock-keychain ~/Library/Keychains/login.keychain-db`
3. Launch the released app
4. Cancel the Keychain prompt
   - Another prompt may appear
5. Cancel any further prompts
   - The app crashes

Confirm with a build from this branch:

1. Unlock the login keychain: `security unlock-keychain ~/Library/Keychains/login.keychain-db`
2. Launch the released app
3. Open a few tabs
4. Quit the app
5. Launch the build from this branch
   - The tabs opened in the released app are restored, so the encryption key is unchanged
6. Quit the app
7. Lock the login keychain: `security lock-keychain ~/Library/Keychains/login.keychain-db`
8. Launch the build from this branch
9. Cancel the Keychain prompt
   - The app quits, with no further prompt and no crash report

The login-item race itself is not reproducible by hand. A Keychain prompt blocks until it
is answered, so the key read either succeeds or is declined. The retry window only opens
when the Keychain is unavailable without a prompt, which is what happens during login.

### Impact

High: Resolves our top crasher by event count — a startup crash affecting users who have the app configured as a login item.

### Quality Considerations

The retry blocks the main thread, so a launch that never gets the key shows no window for a minute before terminating. That is invisible for a login item, and costs nothing on a normal launch, where the first read succeeds.

Startup metrics are discarded for a launch that retried, so the sleeps don't inflate the timings.

Cancelling is the only failure treated as final; everything else is retried, so an unreadable key waits out the full minute.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Startup now blocks the main thread for up to a minute on persistent Keychain failures and changes quit-vs-crash behavior on user-cancelled Keychain prompts, though successful launches should still read the key on the first attempt.
> 
> **Overview**
> Addresses login-item launches where the Keychain is not unlocked yet by **retrying the app encryption key read** for up to **60 seconds** (6-second sleeps on the main thread) instead of failing immediately when `keyStore.readKey()` errors.
> 
> **Behavior changes:** user cancel on the Keychain prompt (`errSecUserCanceled`) now **quits cleanly** with `exit(0)` instead of looping or crashing. If the deadline passes, startup still proceeds with a nil key (same as before); `Database()` is expected to fail shortly afterward. Retries call **`startupProfiler.invalidate()`** so delayed reads do not skew startup metrics.
> 
> **Telemetry:** launch pixel firing moves to `fireLaunchPixel()`, which attaches **`keychainErrorCode`** from the **first** failed read when a later retry succeeds. `EncryptionKeyStoreError` now exposes a **`status`** property for that mapping; pixel definitions document optional `keychainErrorCode` on `ml_mac_app-launch`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34a60df15e368d6dcd4e7405762ac71cef993a9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -543,7 +543,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         bookmarkDatabase = BookmarkDatabase()
 
         if AppVersion.runType.requiresEnvironment {
-            let commonDatabase = Database()
+            let commonDatabase = Database(onRetry: {
+                // Keychain retry sleeps contaminate startup timings.
+                startupProfiler.invalidate()
```
diegoreymendez: We don't want delayed startups to taint our startup profiling data.

macOS/DuckDuckGo/Common/FileSystem/EncryptionKeys/EncryptionKeyStore.swift
```diff
@@ -30,20 +30,20 @@ enum EncryptionKeyStoreError: Error, ErrorWithPixelParameters {
...
+             .deletionFailed(let status),
+             .cannotTransformDataToString(let status),
+             .cannotTransfrotmStringToBase64Data(let status):
+            return status
         }
     }
+
+    var errorParameters: [String: String] {
+        [PixelKit.Parameters.keychainErrorCode: "\(status)"]
+    }
```
diegoreymendez, line 46: Kinda unrelated but it was ugly.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -34,6 +34,7 @@ import Configuration
 import ContentScopeScripts
 import CoreData
 import Crashes
+import CryptoKit
```
diegoreymendez, line 37: For `SymmetricKey`.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -2532,6 +2533,46 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+                // rather than the number of attempts.
+                guard Date() < deadline else {
+                    // Carry on without a key, as this has always done. This is not a
+                    // supported degraded mode: `Database()` reads the same item moments
+                    // later and terminates there, reporting the failure.
+                    Logger.general.error("App Encryption Key could not be read: \(error.localizedDescription)")
+                    return nil
+                }
+                startupProfiler.invalidate()
+                Thread.sleep(forTimeInterval: retryInterval)
```
ayoy, line 2596: @diegoreymendez this is looking great, but I'd like to propose adding pixels:
1. When entering the catch clause for the first time (meaning first time `readKey` failed)
2. On "user canceled" event (that wouldn't probably happen IRL, but would be good to be sure about that)
3. After a 60s deadline was hit.

ayoy (commented):
The code is great! I'm leaving a comment about observability and will now test the builds.

macOS/DuckDuckGo/Application/AppDelegate.swift
diegoreymendez: I had an initial version doing this, but it means I need to initialize PixelKit earlier, and then initialize it again later (twice).

I was trying to keep this as simple as possible, are you sure you want me to add more here, or can we follow up?

diegoreymendez: For future context the proposed follow up is [here](https://app.asana.com/1/137249556945/task/1218006736119770).

ayoy: For the happy path, you could just store metadata in this code and send the pixel after it's initialized. Would that make sense? Maybe even add a parameter to the launch pixel instead of making a new pixel.

For the sad path of crashing, we can surely follow up. WDYT?

diegoreymendez: I don't think we should make pixel changes here.  In bug fixes I'd really prefer to prioritize a small surface to avoid breakage and keeps things simple, and would prefer to follow-up with what's not urgent.

That said, I'm pushing the requested changes, as I want to get this out ASAP and the pixels will require privacy triage, retesting and re-review.

diegoreymendez: Converting this back to draft until it's ready again.

ayoy: To clarify, let's just add the parameter (keychain error code is fine) to the launch pixel and ship it. Thanks @diegoreymendez!

diegoreymendez: Ok I just pushed that now.  Opening triage.

pballart (approved):
Tested the build and LGTM. I am only wondering if it makes sense that the `readEncryptionKeyRetryingKeychainAccess` could be moved outside of AppDelegate and properly tested.

diegoreymendez:
@pballart  - I'll go ahead to merge so I can make it in the internal right away and have it tested, but will [follow up](https://app.asana.com/1/137249556945/project/1201037661562251/task/1218006736119770?focus=true), as I'm already planning follow-up work to improve the UI.  Thanks!

PR #6561, opened 2026-08-28
iOS: Keep split bottom toolbar icons in lockstep with the glass

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217904396397155
Tech Design URL: N/A
CC: N/A

### Description
- Keep the split (top address bar) bottom pill and its icons moving as one item when chrome hides. Nested Liquid Glass on the buttons and Auto Layout on the effect view were interpolating the glyphs separately from the capsule.
- Slide that pill with a transform on a plain container instead of moving the toolbar’s layout slot. Combined bottom chrome is unchanged.
- Size the top-address-bar domain-capsule morph to the visible glass field so it matches the address bar, not the full bar including gutters.
- Restore the resting bottom-field fill after switching address-bar position from top to bottom.

### Testing Steps
1. On iPhone, set the address bar to **Top** and open a long page such as apple.com.
2. Scroll down to hide chrome → the bottom pill and its five icons should slide off together, staying centered in the capsule.
3. Scroll up to reveal chrome → the same pill should come back as one piece.
4. Switch the address bar to **Bottom** and confirm combined chrome spacing and the hide morph are unchanged.
5. Switch back to **Top**, then to **Bottom** again without focusing the field → the address field should keep its resting grey fill.
6. With the address bar on **Top**, load a page and scroll until the domain capsule appears → it should match the address field’s width and height, not sit wider than the bar.

### Impact
**Impact Level: Low**

#### What could go wrong?
- Combined bottom chrome hide could regress if the standalone transform path is applied there → gated on top-address-bar / not-in-toolbar.
- Toolbar buttons lose their individual glass highlight → they are now glyphs on the shared capsule; highlight fill still applies on press.
- Reduce Motion still uses the layout-slot slide, so that path is unchanged.

### Quality Considerations
- Verified on device simulator with a mid-hide scrub: icons stayed inside the capsule as it left the screen.
- Combined portrait-bottom metrics (12pt spacing, 48pt field, embedded buttons) were left alone.
- Privacy/security: none.


brindy (approved):
LGTM but note with the floating ui flag off the effect is a little strange - the items look like they are in a postbox :) That said, it's probably better than it is now.  Have a quick look and see what you think.  Thanks!

kemiljk:
Yeah I feel like that’s okay personally given that most people will swipe quickly anyway.

PR #6563, opened 2026-08-28
iOS: Match tab manager spacing to floating browsing chrome

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1216033800174503/task/1216167808982814
Tech Design URL: N/A
CC: N/A

Stacked on https://github.com/duckduckgo/apple-browsers/pull/6561

### Description
- Keep Fire Tabs empty state 10pt below the top navbar (and 24pt above the pictogram) by pinning it to the chrome's laid-out nav bottom instead of `safeArea + 60`
- Apply matching collection top insets to both Normal and Fire pages during the swipe so cards do not jump when the transition settles
- Pin the tab switcher toolbar and split-mode bottom capsule to the same iOS 26 concentric safe-area guides, including UIToolbar's extra 20pt glass inset
- Keep combined bottom chrome at an equal 20pt inset on every edge and preserve its native concentric corner configuration through focus transitions

### Testing Steps
- On iPhone running iOS 26, open Tab Manager with no Fire tabs and confirm the empty state sits below the navbar buttons and above the fire pictogram
- Swipe Normal ↔ Fire and confirm tab cards do not jump when the page settles
- With the address bar at the top, compare the webview buttons-only bar and tab switcher bar and confirm their side and bottom insets match
- With the address bar at the bottom, confirm the combined chrome has equal 20pt leading, trailing, and bottom insets
- Focus and dismiss the bottom address bar and confirm its outer frame and corner shape do not change

### Impact and Risks
**Impact Level: Low**

#### What could go wrong?
- Chrome hide/show morph could drift if `restingCapsuleFrame` and the live glass disagree; split mode shares the guide-based geometry, while combined bottom mode shares one 20pt edge inset
- Scroll-edge interactions were removed from the bottom bars so they do not apply a second concentric inset; if a future OS change relies on that interaction, glass could drift from the icons

### Quality Considerations
- iPad / large-size tab switcher still hides the toolbar and skips the concentric pin
- Pre-iOS 26 keeps the previous safe-area pin and 24/16pt fallbacks
- No privacy or network changes

### Notes to Reviewer
- The 20pt `floatingConcentricGlassInset` matches the extra inset `UIToolbar` applies to the split-mode buttons-only capsule
- Combined bottom chrome uses its own equal 20pt edge inset and `containerConcentric(minimum: 40)` corners
- Combined portrait-bottom internal metrics (12pt spacing, 48pt field, embedded buttons) are unchanged

### DIFF
--- a/iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
+++ b/iOS/DuckDuckGo-iOS.xcodeproj/project.pbxproj
@@ -1690,6 +1690,7 @@
 		A8440B0101C0DE0000000004 /* FloatingOmnibarTransitionMetricsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8440B0101C0DE0000000003 /* FloatingOmnibarTransitionMetricsTests.swift */; };
 		A8452A0101C0DE0000000002 /* HomeScreenTransitionGeometry.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8452A0101C0DE0000000001 /* HomeScreenTransitionGeometry.swift */; };
 		A8452A0101C0DE0000000004 /* HomeScreenTransitionGeometryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8452A0101C0DE0000000003 /* HomeScreenTransitionGeometryTests.swift */; };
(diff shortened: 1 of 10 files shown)


brindy (changes requested):
Just one bug bot comment to address, looks ok otherwise.

brindy (approved):
LGTM

PR #6565, opened 2026-08-28
iOS: Match focused UTI inset to the resting address bar

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217904396397158
Tech Design URL: N/A
CC: N/A

### Description
- Keep the focused unified input in split mode (top address bar) at the same 16pt horizontal inset as the resting address bar
- Stop the focused card shrinking to 8pt when editing begins
- Leave bottom-bar and flanked AI-tab layouts unchanged

Stacked on https://github.com/duckduckgo/apple-browsers/pull/6563

### Testing Steps
1. Enable floating UI and set the address bar to Top (split mode)
2. Note the resting address bar’s left and right inset
3. Tap the address bar to focus the unified input
4. Confirm the focused card lines up with that same horizontal inset, rather than sitting closer to the screen edges
5. Dismiss the input and confirm the resting address bar is unchanged
6. Switch the address bar to Bottom, focus it, and confirm that layout is unaffected

### Impact
**Impact Level: Low**

#### What could go wrong?
The focused top/split card stays 8pt wider than before. If surrounding chrome or suggestions assumed the narrower width, they could look slightly tighter → mitigated by keeping the 16pt value that the resting address bar and the hand-off pose already use.

### Quality Considerations
- Bottom-bar focused UTI already used the 16pt expanded inset; this only changes the top/split omnibar session
- Landscape / iPad still inherit the same 16pt token as the portrait address bar
- No privacy or analytics impact

### Notes to Reviewer
Parent of this stack is #6563.

### DIFF
--- a/iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputView.swift
+++ b/iOS/DuckDuckGo/UnifiedToggleInput/UnifiedToggleInputView.swift
@@ -99,13 +99,12 @@ final class UnifiedToggleInputView: UIView {
         /// (see `flankedHorizontalInset` in `setupConstraints`). Keeps the flanked input's left/right
         /// padding consistent end-to-end.
         static let flankedCardHorizontalMargin: CGFloat = 16
-        /// Card container's outer horizontal margin in the non-flanked layouts.
-        static let cardHorizontalMargin: CGFloat = 8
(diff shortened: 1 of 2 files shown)


brindy (approved):
LGTM

PR #6571, opened 2026-08-28
Add Maestro coverage for the Duck.ai usage-warning footer

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1217605036256293?focus=true
Tech Design URL:
CC:

### Description
Adds Maestro flows covering the Duck.ai usage-warning footer behind `utiDuckAIWarnings`: the approaching-limit card, both limit-reached cards, and a flag-off control. The states are seeded through the AI Chat debug menu so the real read path drives the UI, since a free-tier account never reaches these cards on its own. The only production change is a set of `AIChat.Footer.*` accessibility identifiers, so the flows keep working once the placeholder copy is localised.

### Testing Steps
1. Check if https://github.com/duckduckgo/apple-browsers/actions/runs/33192059824 is passing

### Impact
None — test-only, apart from accessibility identifiers that add no behaviour.

#### What could go wrong?
The three flows are tagged `unified_toggle_input`, which no CI leg currently runs, so they only execute when someone runs that tag by hand.

### DIFF
--- a/.maestro/shared/seed_duckai_usage_warning.yaml
+++ b/.maestro/shared/seed_duckai_usage_warning.yaml
@@ -1,0 +1,39 @@
+appId: com.duckduckgo.mobile.ios
+
+---
+
+# Seeds a synthetic usage-limit snapshot via Debug ▸ AI Chat ▸ Duck.ai Usage Warnings, which writes
+# the reserved `usageLimits` entry the web app writes, so the real read path drives the footer.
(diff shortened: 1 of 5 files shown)


Bunn:
[Passed] test_uti_duckai_warnings_approaching (1m 56s)
[Passed] test_uti_duckai_warnings_limit_reached (2m 13s)
[Passed] test_uti_duckai_warnings_disabled (1m 33s)
🎉

jaceklyp (approved):

PR #6576, opened 2026-08-31
Fix attributed.metric.subscribed month=0/1/2+ over-reporting (iOS & Mac)

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1218089571925205?focus=true
Tech Design URL: N/A
CC:

### Description
Two follow-up fixes to the earlier `attributed.metric.subscribed` work, addressing the remaining inconsistencies flagged on the task:

**1. `month=2+` (ongoing month) was over-reporting.** It fired on every active day a subscription had been running for more than a month, instead of once per month. Now dedups per month, mirroring the retention pixels.

**2. `month=1` (paid start) is potentially over-reporting at the expense of `month=0` (trial start).** The purchase notification that drives the trial-vs-paid classification was posted right after the store transaction, before the subscription was confirmed. The notification now fires once the confirmed subscription (with its trial offer) is available, so trial starts are classified correctly.

### Testing Steps
Use an alpha build (StoreKit sandbox). Watch the Xcode console (subsystem `PixelKit`) for `attributed_metric_subscribed` and check `month`; relaunch between date changes so the startup check runs.

Trial → paid → renewals:
1. In Debug -> Attributed Metrics click Reset stored data
1. Subscribe with a **free trial** → fires once with `month=0`.
2. Move the device date past the trial so it converts, relaunch → fires once with `month=1`.
3. Set current time in  Debug -> Attributed Metrics one day later, relaunch → nothing new (no duplicate `month=1`).
4. Set current time in  Debug -> Attributed Metrics ~a month later, relaunch → fires once with `month=2`.
5. Set current time in  Debug -> Attributed Metrics forward one day, relaunch → **no** duplicate `month=2`.
6. Set current time in  Debug -> Attributed Metrics forward another month, relaunch → fires once with `month=2`.

Direct paid (no trial):
1. Subscribe **without** a trial → fires once with `month=1` (no `month=0`).

### Impact
Low — telemetry only. No user-facing behaviour and no change to what data is collected. If wrong, the affected subscription-start metrics stay inconsistent.

#### What could go wrong?
- `month=2+` guard too strict → never advances → mitigated by a test asserting it fires once per month across launches and advances on the next month boundary.
- Moving the purchase notification later → a purchase whose confirmation fails no longer counts as a start → intended: an unconfirmed purchase is not a real start. Notification-timing covered by tests (posts on success, not on missing-entitlements failure).

### Quality Considerations
- Shared package — one change covers iOS and macOS.
- Persists the highest reported month so `month=2+` dedup survives relaunches.
- `.userDidPurchaseSubscription` is consumed only by the attributed-metric managers, so moving when it's posted has no other listeners to affect.

---

> [!NOTE]
> **Low Risk**
> Telemetry and notification timing only; purchase flows now signal subscription start after confirmation, which is intentional and covered by tests.
> 
> **Overview**
> Fixes **over-reporting** of `attributed_metric_subscribed` for ongoing subscriptions and improves **trial vs paid** classification at purchase time.
> 
> **Month 2+:** Startup checks used to emit the subscribed pixel on every launch once the subscription was past the first month. The manager now compares the computed month bucket to persisted `subscriptionLastMonthFired` and only fires when the month advances, then stores the new high water mark (survives relaunches).
> 
> **Purchase timing:** For App Store, `.userDidPurchaseSubscription` is no longer posted right after the StoreKit transaction; it posts only after `completeSubscriptionPurchase` succeeds with active subscription and entitlements, so attributed metrics can read confirmed trial state. Stripe completion refreshes subscription via `getSubscription(forceRefresh:)` instead of only clearing cache before posting the same notification.
> 
> Tests cover once-per-month month=2+ behavior and notification posting on success vs missing entitlements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e3477641aef4ba3203002df076dffda7f76e35f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
@@ -268,7 +268,6 @@ public final class DefaultAppStorePurchaseFlow: AppStorePurchaseFlow {
                     return .failure(.missingEntitlements)
                 } else {
                     pendingTransactionHandler?.handleSubscriptionActivated()
-                    // Posted here, after confirmation, so listeners read the confirmed subscription (incl. its trial offer) rather than pre-confirmation cache.
                     NotificationCenter.default.post(name: .userDidPurchaseSubscription, object: self)
                     return .success(.completed)
                 }
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
@@ -199,7 +199,6 @@ public final class DefaultAppStorePurchaseFlow: AppStorePurchaseFlow {
         // Make the purchase
         switch await storePurchaseManager.purchaseSubscription(with: subscriptionIdentifier, externalID: externalID, includeProTier: includeProTier) {
         case .success(let transactionJWS):
-            NotificationCenter.default.post(name: .userDidPurchaseSubscription, object: self)
             return .success((transactionJWS: transactionJWS, accountCreationDuration: accountCreationDuration))
         case .failure(let error):
             Logger.subscriptionAppStorePurchaseFlow.error("purchaseSubscription error: \(String(describing: error), privacy: .public)")
@@ -269,6 +268,8 @@ public final class DefaultAppStorePurchaseFlow: AppStorePurchaseFlow {
                     return .failure(.missingEntitlements)
                 } else {
                     pendingTransactionHandler?.handleSubscriptionActivated()
+                    // Posted here, after confirmation, so listeners read the confirmed subscription (incl. its trial offer) rather than pre-confirmation cache.
+                    NotificationCenter.default.post(name: .userDidPurchaseSubscription, object: self)
                     return .success(.completed)
                 }
             } else {
--- a/SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
@@ -138,6 +138,8 @@ public final class DefaultStripePurchaseFlow: StripePurchaseFlow {
         Logger.subscriptionStripePurchaseFlow.log("Completing subscription purchase")
         subscriptionManager.clearSubscriptionCache()
         _ = try? await subscriptionManager.getTokenContainer(policy: .localForceRefresh)
+        // Fetch the confirmed subscription before notifying, so listeners read its trial offer rather than the just-cleared cache.
+        _ = try? await subscriptionManager.getSubscription(forceRefresh: true)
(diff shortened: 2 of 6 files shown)


SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
```diff
@@ -138,6 +138,7 @@ public final class DefaultStripePurchaseFlow: StripePurchaseFlow {
         Logger.subscriptionStripePurchaseFlow.log("Completing subscription purchase")
         subscriptionManager.clearSubscriptionCache()
         _ = try? await subscriptionManager.getTokenContainer(policy: .localForceRefresh)
+        _ = try? await subscriptionManager.getSubscription(forceRefresh: true)
```
federicocappelli, line 140: Why is this needed? 
The BE already sees a higher number of API calls from Apple subscription flows compared to Android and windows. If we do this, we need to be absolutely certain it's needed.

SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -269,6 +268,7 @@ public final class DefaultAppStorePurchaseFlow: AppStorePurchaseFlow {
                     return .failure(.missingEntitlements)
                 } else {
                     pendingTransactionHandler?.handleSubscriptionActivated()
+                    NotificationCenter.default.post(name: .userDidPurchaseSubscription, object: self)
```
federicocappelli, line 271: .userDidPurchaseSubscription is used only for attributed metrics, and this change uses the same approach as the Stripe purchase. I like it but we need to be sure we communicate what the change can mean:
- the user can purchase the subscription, then something goes wrong in the next 2 API calls and we don't track the purchase

federicocappelli (commented):
A couple of questions/points for you

SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
federicocappelli: Having investigated further, I believe this call is necessary but we can then remove `subscriptionManager.clearSubscriptionCache()`.
This is especially important with a slow connection as the steps of 
1. removing the subscription
2. refreshing the token
3. retrieving the subscription
can cause concurrency issues. For example, if another piece of code branches and asks “do we have a subscription” it might return false if the request happens between 1 and 3

federicocappelli (changes requested):
just a small change

SharedPackages/BrowserServicesKit/Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
SabrinaTardio: As discussed in MM those cases should be reported on the next app start up

federicocappelli (approved):
LGTM

PR #6592, opened 2026-09-01
iOS: Loading chip while a PDF's bytes are read

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204186595873227/task/1218023815227728?focus=true

### Description
On iOS, attaching a PDF to Duck.ai re-fetches the document over the network, so there is a silent pause between tapping "Ask about page" (or auto-attach) and the attachment appearing. This adds a loading state: the page-context attachment chip shows an animated dots indicator while the bytes are read, and the start surface (suggestions and the "Ask about page" action) is hidden during the read so the chip is the only loading indicator. The loading chip matches the size/shape of the attached chip it becomes.

### Testing Steps

1. In Debug → Custom Configuration → set the Duck.ai URL to https://euw-serp-dev-testing20.duck.ai/ (the FE support for the new payload isn't in production yet). 
2. Load full Duck.ai in order to go through the authentication
3. **Manual attach**: open Duck.ai with auto-add-page-content off → tap "Ask about page" → the attachment chip shows animated dots, and "Ask about page"/suggestions disappear while it loads → the attached PDF chip replaces the dots once ready.
4. **Auto attach**: with auto-add-page-content on, open Duck.ai on a PDF → the attachment chip shows the dots, then the attached chip.
5. Attach a normal web page → no loading chip (bytes aren't re-fetched); behaves as before.

### Impact

Low — a loading affordance on the contextual Duck.ai attachment; no change to what's attached or sent.

---

> [!NOTE]
> **Medium Risk**
> Subscription notification timing shifts for App Store purchases (post-confirmation only), which could affect listeners that relied on the earlier post; PDF loading is localized UI with no payload changes.
> 
> **Overview**
> Adds a **loading** state to the Duck.ai page-context attachment chip while document bytes are read from the web view (e.g. PDF re-fetch). The chip shows the existing animated-dots loader, stays non-interactive, and session UI hides suggestions and quick actions (including “Ask about page”) so only the chip signals progress until attach completes or fails.
> 
> **Wiring:** `AIChatPageContextHandler` publishes `documentReadInProgress` around async document collection; the contextual sheet coordinator drives the chip view model and `setDocumentChipLoading` on session state. Regular HTML pages skip this path.
> 
> **Also in this PR:** `userDidPurchaseSubscription` is no longer posted immediately after a successful App Store `purchaseSubscription` call; it fires only after `completeSubscriptionPurchase` succeeds (with entitlements). Stripe `completeSubscriptionPurchase` now force-refreshes subscription before posting so listeners see confirmed trial/offer data instead of a cleared cache. Tests cover notification posting on success vs missing entitlements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20862acdabb146f4a0f3a4181c24b34dc52f2f0a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/AIChat/Sources/AIChat/iOS/Views/ContextChip/AIChatContextChipView.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/iOS/Views/ContextChip/AIChatContextChipView.swift
@@ -61,9 +61,12 @@ public final class AIChatContextChipView: UIView {
     public enum State {
         case placeholder
         case attached(title: String, favicon: UIImage?)
+        case loading
     }
 
     private var currentState: State = .placeholder
+
+    private var loadingView: AIChatSuggestionsLoadingView?
 
     /// Only live in the placeholder state. Left enabled in the attached state it would recognise taps on
     /// the remove button and cancel them, since it spans the whole chip.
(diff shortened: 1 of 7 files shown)


aataraxiaa (approved):
PDF loading looks good. But looks like there are some subscription changes in the PR which should be removed.

SabrinaTardio:
> PDF loading looks good. But looks like there are some subscription changes in the PR which should be removed.

Sorry I had just pushed to the wrong branch by mistake!

tomasstrba:
One tiny issue found with Claude, just making sure the current implementation is intended:

The loading chip isn't the same width as the chip it turns into. `.loading` deactivates `fixedWidthConstraint`, so its width comes from `AIChatSuggestionsLoadingView`'s intrinsic size (~59pt), while the attached chip is a fixed 240pt. Heights match at 44pt, so it's only the width — but the strip jumps ~180pt when the read finishes.

Intended? The description says the loading chip matches the size/shape of the attached chip it becomes, so I assumed not.

tomasstrba (approved):
LGMT! ✔️ The loading is really nice

Image

SabrinaTardio:
> One tiny issue found with Claude, just making sure the current implementation is intended:
> 
> The loading chip isn't the same width as the chip it turns into. `.loading` deactivates `fixedWidthConstraint`, so its width comes from `AIChatSuggestionsLoadingView`'s intrinsic size (~59pt), while the attached chip is a fixed 240pt. Heights match at 44pt, so it's only the width — but the strip jumps ~180pt when the read finishes.
> 
> Intended? The description says the loading chip matches the size/shape of the attached chip it becomes, so I assumed not.

Yeah it seemed it looked better using the AIChatSuggestionsLoadingView size

PR #6606, opened 2026-09-01
EventHub metrics #1: hub-level de-duplication

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218062429029803
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1216911265362552
CC:

### Description
First in the `michal/eventhub-metrics` stack. Reworks how the hub decides whether a web event is a repeat, ahead of the config-driven experiment metrics handler that lands next.

Until now each pixel counter made its own de-duplication decision, and immediate-fire pixels made none at all. The hub now makes one decision when the event arrives, before anything is told about it: a repeat reaches no handler, a first occurrence reaches every handler.

Additional behavioural changes:
- event's payload is also part of the de-duplication key, so a detector that reports different values is no longer collapsed into a single occurrence
- immediate type trigger is renamed to `immediate_v2`, change is meant to reflect the expectations that it is now subject to de-duplication

Fixes to update behaviour in line with the cross-platform specification: 
- immediate type pixel whose data parameters all resolve to nothing no longer fires
- nil data parameter now clears preexisting value

Cross-platform tech designs: [event-driven metrics](https://app.asana.com/1/137249556945/project/72649045549333/task/1217052832008956) and [de-duplication with data payload](https://app.asana.com/1/137249556945/project/481882893211075/task/1217874991729731).

### Testing Steps
Use the setup in https://app.asana.com/1/137249556945/project/414235014887631/task/1218049910864102?focus=true and use Xcode debugger to see EventHub logs

**Test case 1:**
1. Navigate to https://www.wikihow.com/Disable-Your-Ad-Blocker  (detector needs to find "disable your ad blocker" text on the page)
- [x] Event is properly recorded on the initial page load in a given tab
- [x] Event is blocked by de-duplication logic when reloading the same tab
- [x] After navigating to some to some other page and returning back makes the event being recorded again (navigation clears de-dup state)
- [x] Opening new multiple new tabs with the test page causes each event to be properly recorded

**Test case 2:**
1.  Navigate to https://privacy-test-pages.site/features/web-interference-detection/youtube-detection-events.html
2. Click "Show Ad-Blocker Modal" button
- [x] Triggering the event properly records is with `unknown` login state
- [x] Repeated re-triggering (without page reload) is blocked by de-duplication logic (including the immediate type event)

**Test case 3:**
1. Navigate to https://privacy-test-pages.site/features/web-interference-detection/youtube-detection-events.html
2. Select "Login state" from the drop down
3. Click "Reload with this state" (causes page reload navigation event)
4. Click "Show Ad-Blocker Modal" button
- [x] Triggering the event properly records is with preselected login state

**Test case 4:**
1. Navigate to https://privacy-test-pages.site/features/web-interference-detection/youtube-detection-events.html?loginState=premium&loginStateDelayMs=3000 (page will load with unknown login state but will dynamically update to premium after 3s)
2. Click "Show Ad-Blocker Modal" button when the page loads (needs to happen within first 3s) - this will record event with unknown login state.
3. Click "Show Ad-Blocker Modal" button twice to disable and re-enable it (do it after more than 3s from the page load) - this will record event with premium login state.
- [x] Triggering the event properly records is first with the `unknown` login state and then with `premium` (no de-duplication happens)
- [x] Without page reload click "Show Ad-Blocker Modal" button twice causes the `premium` event to be re-triggered but it is now de-duplicated since the `data` param is unchanged.

### Impact
High — changes what already-released telemetry reports, and the trigger rename stops shipped pixels until remote config catches up.

#### What could go wrong?
- The five shipped `webTelemetry_captcha_*` immediate pixels stop firing until privacy config ships `immediate_v2` entries → the config change must land before this reaches users. Embedded configs pick it up via `scripts/update_embedded.sh`, so no in-app change is needed.
- Those same captcha pixels drop in volume once they do return, since immediate pixels now de-duplicate per page → intended, but worth flagging to Data Science as a step change rather than a regression.
- The five `webTelemetry_youtube_*_day` counters can over-count: each pairs a counter with a `loginState` data parameter on one source, and `loginState` starts `unknown` before resolving, so an early detection and a later one now count separately where they previously counted once → accepted trade-off per the de-duplication tech design, but it lands on the pixel carrying the public disclaimer.
- An immediate pixel that resolves none of its data parameters now stays silent → required by the specification, covered by tests, but it is a behaviour change for any config relying on the old unconditional fire.

### Quality Considerations
- Covers the cross-platform specification suite: D-DEL-1…6, D-NAV-1…4, D-NAT-1, plus T-IMM-1, T-IMM-4 and T-DAT-2 from the telemetry suite. Each case ID leads its test name, so coverage greps cleanly and a failure names the spec case.
- Not covered here: the week-counter leg of the three telemetry cases, which needs a decision on modelling "each period ends exactly once" when a seven-day period passes seven daily boundaries. It belongs with the rest of the telemetry suite.
- Payload identity uses canonical JSON with sorted members, so member order does not matter and an absent payload equals an empty one. A payload that cannot be serialised is delivered rather than dropped — under-de-duplicating is the safer failure.
- The per-tab key set is unbounded, cleared on navigation and tab close. With the payload in the key a page emitting many distinct payloads grows it until it navigates away; left uncapped deliberately, since a cap would be an undeclared divergence from the other platforms.
- Privacy: the received-event log line now includes the payload. It is `.private` and `debug`-only, so redacted unless private-data logging is deliberately enabled. The logger's documentation previously stated payloads were never logged and has been updated to match.
- De-duplication state remains in-memory only, so a restart lets a page be counted once more. Unchanged, but it now governs metric conversions too.

---
###### Internal references:

---

> [!NOTE]
> **High Risk**
> Changes shipped telemetry semantics (hub dedup, payload-aware keys, immediate_v2 rename) and can silence or reshape pixel volume until remote config updates; requires coordinated privacy-config rollout.
> 
> **Overview**
> Moves **web event de-duplication** from per-counter logic into `EventHub` at ingestion: each `(event type, canonical payload, tab)` is decided once per page before immediate and period handlers run, so repeats never fan out and multi-frame duplicates no longer inflate every consumer.
> 
> **`DedupStore`** keys on sorted-key JSON (member order irrelevant; missing vs empty `data` match), clears on real URL change, tab close, or feature disable—not on period boundaries. Native ingress still bypasses dedup. **`immediate`** config is no longer parsed; **`immediate_v2`** is the trigger for one pixel per *delivered* event.
> 
> **Telemetry behavior** aligned with spec: immediate pixels with data parameters fire only when at least one resolves; **`DataParameter`** clears when a later event omits its key instead of keeping a stale value. **`EventHubTabID.empty`** is removed for native paths.
> 
> Adds cross-platform **spec test suites** (`DeduplicationSpecTests`, `TelemetrySpecTests`, `SpecFixture`) and wires **EventHub** into the workspace.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d1b6240b62e266aa7fc0869d949e7d0f70900f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/EventHub/Sources/EventHub/Core/EventHub.swift
+++ b/SharedPackages/EventHub/Sources/EventHub/Core/EventHub.swift
@@ -26,15 +26,19 @@ import os.log
 public protocol EventHubManaging: AnyObject {
     /// Processes an incoming `webEvent` envelope (`{ "type": ..., "data": ... }`) from the tab
     /// identified by `tabID` against the active telemetry configs.
+    ///
+    /// De-duplicated once at ingestion, keyed `type × data × tabID` against the tab's current page: a
+    /// repeat reaches no handler, a first occurrence reaches every handler. See `DedupStore`.
     func handleWebEvent(_ webEventData: [String: Any], tabID: EventHubTabID)
 
     /// Fires any enabled immediate-trigger telemetry whose `trigger.source` equals `type`. For
-    /// browser-native events (not content-scope-scripts); there is no tab context.
+    /// browser-native events (not content-scope-scripts): there is no tab context, so no page to
+    /// de-duplicate against, and every call is a genuine occurrence.
     func handleImmediateEvent(_ type: String, data: Encodable?)
 
     /// Counts a browser-native event toward any enabled period/aggregated telemetry whose parameter
-    /// `source` equals `type`. Unlike web events there is no per-tab dedup: each call is a genuine
-    /// occurrence.
+    /// `source` equals `type`. Like `handleImmediateEvent`, and unlike web events, this never passes
+    /// through de-duplication: each call is a genuine occurrence.
     func handleAggregatedEvent(_ type: String, data: Encodable?)
 
     /// Signals that the given tab has navigated to `url` (clears per-tab dedup on URL change).
@@ -91,8 +95,9 @@ public final class EventHub: EventHubManaging {
     private var telemetries: [String: Telemetry] = [:]
     private var dirtyNames: Set<String> = []
     private var tabURLs: [EventHubTabID: String] = [:]
-    /// Hub-lifetime so per-tab dedup outlives the `Telemetry` objects that consult it — a period
-    /// rollover, and a period firing while backgrounded, both replace those. See `DedupStore`.
+    /// The one de-duplication decision, taken in `handleWebEvent` before any handler runs. Hub-owned
+    /// and hub-lifetime, so it outlives the `Telemetry` objects a period rollover replaces — dedup is
+    /// page-scoped, not period-scoped. See `DedupStore`.
     private let dedupStore = DedupStore()
     private var latestConfigs: [TelemetryPixelConfig] = []
     private var latestEnabled = false
@@ -171,8 +176,8 @@ public final class EventHub: EventHubManaging {
             Logger.eventHub.debug("[EventHub] web event ignored, `type` was missing or empty")
             return
         }
-        Logger.eventHub.debug("[EventHub] web event received: \(type, privacy: .private), tab \(tabID.rawValue.uuidString, privacy: .private)")
         let data = webEventData["data"] as? [String: Any]
+        Logger.eventHub.debug("[EventHub] web event received: \(type, privacy: .private), tab \(tabID.rawValue.uuidString, privacy: .private), data \(Self.payloadDescription(data), privacy: .private)")
         // Sampled here, not in the block: `now` decides whether the event still belongs to the running
         // period, so it has to be the arrival time. Read on the queue it would be the drain time, and an
         // event arriving just before `periodEnd` could be dropped for landing in no period at all.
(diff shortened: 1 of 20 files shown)


SharedPackages/EventHub/Sources/EventHub/Core/EventHub.swift
```diff
@@ -244,7 +264,7 @@ public final class EventHub: EventHubManaging {
     ///   onto the queue. The timer is best-effort, so it can legitimately be past `periodEnd` before the
     ///   sweep has run (notably on macOS, where a period can end while the app runs unfocused). An event
     ///   arriving in that window belongs to no period and must not be counted into the elapsed one.
-    private func countPeriodLocked(source: String, data: [String: Any]?, tabID: EventHubTabID, nowMillis now: Int64) {
+    private func countPeriodLocked(source: String, data: [String: Any]?, nowMillis now: Int64) {
```
SabrinaTardio, line 267: In the cases we hit continue and we don't handle the event "markSeen" would have already been run, which means the event will be marked as delivered even if it wasn't. Should we worry about any of these cases?

SabrinaTardio (approved):
Looks Good! I left one question for consideration

SharedPackages/EventHub/Sources/EventHub/Core/EventHub.swift
miasma13: Good catch — that's a real change. Before, `markSeen` sat inside `CounterParameter.handle`, so it only ran when a live period actually counted. Now it runs at ingestion, so every `continue` path marks.

Three of the four are harmless:

- **Source not handled** — that pixel was never fed by this event.
- **Counter capped** — already at its final bucket, nothing left to count.
- **Period elapsed, not yet swept** — a repeat after rollover was already suppressed on purpose, so this just makes it consistent.

The fourth is real: **no period running**. On macOS `applicationDidResignActive` sets `isForeground = false`, which stops new periods. Immediate pixels still fire. So an event can mark its key with no counter listening, and the repeat after you return is dropped.

The window is narrow, though. It needs the app to lose focus *and* an event to arrive while it's unfocused. Most events don't work that way — webDetection auto triggers are one-shot `setTimeout`s at 500/1000/5000ms after page load, so they fire while the user is actively browsing. The exception is the YouTube sweep, which runs every 2s regardless of focus, so a background video could hit it.

I'd still leave it. D-DEL-P1 puts the decision before fan-out — delivery is the hub's job, and whether a handler has a period is the handler's business. Gating the mark on a handler acting would re-couple them and break D-DEL-4.

The actual gap is that measurement pauses while the app is inactive — already flagged as unaddressed in #6054. This narrows the workaround, it doesn't create the problem.

PR #6609, opened 2026-09-01
Remove feature flag: duckAINativeStorageMigrationLockedLaunchFix

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211834678943996/task/1215920422347500
Tech Design URL:
CC:

### Description
Feature flag `duckAINativeStorageMigrationLockedLaunchFix` approved, removing conditional code. The Duck.ai native-storage container migration's protected-data gate now always guards only the relocation step: a completed or not-needed migration proceeds on a locked launch instead of deferring up front. This was already the flagged-on behavior for rollout users; it is now the only behavior.

Removed the `duckAINativeStorageMigrationLockedLaunchFix` case from `FeatureFlag` and its `AIChatSubfeature.nativeStorageMigrationLockedLaunchFix` mapping, the `lockedLaunchFixEnabled` parameter from `DuckAiNativeStorageContainerMigration` (and both of its call sites in `Launching.swift` and `FireModeNativeStorageController.swift`), and the unit tests that exercised the now-removed disabled/kill-switch path.

`iOS/Core/ios-config.json` was intentionally left untouched (server-managed).

### Testing Steps
1. Fresh-install the app from the main branch, do some prompts on duck.ai, create images.
2. Run this branch on top of the same app, validate if the prompts and images are still there

### Impact
Low — this is a fully-approved flag removal; the enabled code path was already live for rollout users, so there is no user-facing behavior change.

### Quality Considerations
Removed the three unit tests (`testWhenFixDisabled...`) that only existed to verify the now-deleted disabled/kill-switch path; the remaining migration test coverage (locked-launch gate behavior, retries, protection) is unchanged.

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/Features/PrivacyFeature.swift
@@ -480,12 +480,6 @@ public enum AIChatSubfeature: String, Equatable, PrivacySubfeature {
     /// legacy App Group path.
     case nativeStoragePathMigration
 
-    /// Once the native-storage path migration is complete (or not needed), opens
-    /// the store on locked / background launches instead of deferring on the
-    /// protected-data gate. Off keeps the legacy behavior where any locked launch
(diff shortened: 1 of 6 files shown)


jotaemepereira (approved):
:shipit:

PR #6622, opened 2026-09-01
iOS: Hide Duck.ai usage warning after a manual model switch

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201011656765697/task/1218070823421574?focus=true
Tech Design URL:
CC:

### Description
The Duck.ai usage-limit card asks the user to switch to a lighter model, but switching from the bar's own model picker left the card up, still asking for the switch. A manual switch now settles the card the same way its Switch button does, but only when the picked model is one the card would have offered as a step down from the model the user was on; a heavier or sideways switch leaves it up. Separately, after taking the Switch button the card could never come back in the same app session, even after the debug menu cleared its record, because the footer kept its own copy of the acted-on message; that copy is now released with the record. iOS only; macOS is untouched.

### Testing Steps
1. On an internal build, open Settings → Debug → AI Chat, and under "Duck.ai Usage Warnings — Free" seed an approaching-limit case that offers a model switch → open a Duck.ai tab and focus the input; the usage card shows with a Switch button.
2. Open the model chip and pick the model the card suggests → the card hides without tapping its button.
3. Re-seed the case, then pick a heavier model instead → the card stays up.
4. Tap the card's Switch button → the card hides. In the debug menu tap "Clear dismissals", then collapse and re-expand the input → the card is back.
5. Back in the debug menu tap "Re-seed the last case (simulates a republish)" and focus the input again → the card returns.

### Impact
Low. Small behaviour fix to a card that is behind the internal-only `utiDuckAIWarnings` flag; the card's button, close and upsell paths are unchanged, and the model-switched pixel still fires for every manual switch.

#### What could go wrong?
Web's offer table names a model the picker does not list → the card stays up as today, since only a model actually picked from the list can match.

---

> [!NOTE]
> **Low Risk**
> iOS-only UX fix behind internal usage-warnings flag; core submit, upsell, and dismiss paths unchanged, with added guardrails around CTA target matching.
> 
> **Overview**
> The Duck.ai usage-limit footer could stay visible after the user switched models from the **bar’s model chip**, even when they picked the lighter model the card was pushing. **Manual picker switches** now retire the warning the same way as the card’s Switch CTA, but only when the CTA is a model-switch type and the chosen model is among the step-down candidates for the model the user left (including per-model CTA tables from web).
> 
> **UTIModelSelector** reports selection changes as `(previousModelId, newModelId)` and drops the pre-apply `onModelSelectionWillChange` hook; the coordinator forwards that to the footer, which always fires the model-switched pixel while the view model decides whether to record “acted on” state.
> 
> **Footer hiding after acting** no longer depends only on a local `actedOnMessage` copy: **`hasActedOnCurrentNotice`** ties visibility to the shared dismissal store, so the card stays hidden across snapshot republishes with the same text and **reappears** when debug (or similar) clears the acted record.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf5de05c4ac2f3a3099e98830aa80e05ca3f6f8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
+++ b/iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
@@ -141,10 +141,11 @@ final class UTIFooterController {
         measurement.promptSubmitted()
     }
 
-    /// A switch the user made from the bar's picker; the card's own CTA reports and retires itself.
-    func userSwitchedModel() {
+    /// A switch from the bar's picker: always reported, but it only retires the message when it is the
+    /// step down the message asked for. The card's own CTA reports and retires itself.
+    func userSwitchedModel(from previousModelId: String?, to modelId: String) {
         measurement.modelSwitched()
-        viewModel.userSwitchedModel()
+        viewModel.userSwitchedModel(from: previousModelId, to: modelId)
         applyCurrentState()
     }
 
--- a/iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
+++ b/iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
@@ -245,9 +245,11 @@ final class UTIFooterController {
         return nil
     }
 
-    /// Releases as soon as the resolver produces a different message, so the next rung still shows.
+    /// Releases as soon as the resolver produces a different message, so the next rung still shows,
+    /// and once the acted-on record is gone, so clearing it is not undone by this copy.
     private func unlessActedOn(_ message: UTIFooterMessage) -> UTIFooterMessage? {
-        message == actedOnMessage ? nil : message
+        guard viewModel.hasActedOnCurrentNotice, message == actedOnMessage else { return message }
+        return nil
     }
 
     static let springAnimator: Animator = { changes in
--- a/iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
+++ b/iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
@@ -149,13 +149,6 @@ final class UTIFooterController {
         applyCurrentState()
     }
 
-    /// Picking the model the card suggests is its CTA by another route, so it retires the card too.
-    /// Runs before the switch lands, while the suggestion still points at what the user picked.
-    func noteModelSelectionWillChange(to modelId: String) {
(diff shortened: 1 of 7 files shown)


iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
```diff
@@ -118,9 +118,12 @@ final class UTIFooterController {
         measurement.promptSubmitted()
     }
 
-    /// A switch the user made themselves; the card's own switch CTA reports its own tap.
-    func recordModelSwitched() {
+    /// A switch from the bar's picker: always reported, but it only retires the message when it is the
+    /// step down the message asked for. The card's own CTA reports and retires itself.
+    func userSwitchedModel(from previousModelId: String?, to modelId: String) {
```
tomasstrba, line 146: `applyCurrentState()` here without a matching `highUsageNotice?.refresh()` can swap the usage card for a stale high-usage notice.

On `claude-opus-4-8` with an approaching-limit card up, the high-usage notice is already resolved — it's just outranked by the warning in `resolveCard()`. Pick the suggested lighter model, the warning stands down, and `applyCurrentState()` falls through to `highUsageNotice?.notice`, which still names Opus. So the footer starts talking about the model the user just left, and stays there until the next `refresh()` — the coordinator doesn't refresh the footer on a plain model selection.

`performPrimaryAction` only avoids this by accident: by the time it calls `retireCurrent()`, `performAction()` has already nil'd the warning, so it lands in the `else` branch and dismisses the notice.

```swift
func userSwitchedModel(from previousModelId: String?, to modelId: String) {
    measurement.modelSwitched()
    viewModel.userSwitchedModel(from: previousModelId, to: modelId)
    // The notice is keyed to the selected model, which is exactly what just changed.
    highUsageNotice?.refresh()
    applyCurrentState()
}
```

That also covers the other direction: switching *to* Opus with no usage card up currently shows nothing until the next refresh.

tomasstrba (approved):
Claude found one issue, reported above. Otherwise, LGTM ✔️

iOS/DuckDuckGo/UnifiedToggleInput/Footer/UTIFooterController.swift
Bunn: updateSelectedModel already calls onModelsUpdated → handleModelsUpdated → footerController.refresh(), which re-resolves highUsageNotice against the new model

PR #6639, opened 2026-09-02
[macOS] Migrate Cookie Pop-ups Blocked promo to the Promo Queue

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1218060281710245?focus=true
CC:

### Description

The Cookie Pop-ups Blocked promo now goes through the Promo Queue instead of presenting itself directly. This moves the logic that shows the promo from the (now deleted) `AutoconsentStatsPopoverCoordinator` to a new promo delegate.

For users, the promo behaves mostly as it did before. What changes is that it now takes its turn alongside other promo rather than appearing on top of whatever else is on screen, and it can be turned off remotely. One intentional change in behavior:

* The auto-dismiss timer is now **5 seconds** (the promo queue's standard timing for this kind of tip) instead of the old popover's hardcoded **8 seconds**.

Two additional notes:

* **The promo now depends on the Promo Queue being enabled.** There is no fallback to the old standalone path — if the queue feature flag is disabled, the promo simply doesn't appear.
* **Anyone who already saw the promo won't see it again.** The legacy "already shown" setting is still read to retire the promo for those users, so migrating doesn't re-ask people who already answered.

Also adds a per-promo remote kill switch (`promoQueueCookiePopupsBlockedPromo`, enabled by default) so this promo can be disabled on its own without taking down the whole queue.

### Testing Steps

Prerequisites:

- An internal build with `promoQueue` and `promoQueueCookiePopupsBlockedPromo` both on (both default to enabled)
- The build was installed at least 2 days ago. Simulate this with Debug menu → Reset Data → Set Launch Date a Week in the Past.
- If you have already seen this promo (on any earlier build), first clear it with Debug menu → CPM → NTP widget → clear blockedCookiesPopoverSeen flag — otherwise the promo will correctly refuse to appear.

Testing steps:
1. Debug menu → CPM → NTP widget → "Increment Autoconsent Stats" 5 times (or block 5 real cookie pop-ups)
2. Bring a different app to the foreground and then go back to a DuckDuckGo window → the "N cookie pop-ups blocked" popover appears near the "+" (new tab) button.
3. Click on the popover → a new tab opens scrolled to the Protections Report, and the popover doesn't appear again on subsequent launches.
4. Reset with Debug menu → Promo Queue → cookie-popups-blocked → Undismiss + Clear History
5. Repeat steps 1-2 to trigger the promo, then click the "+" new-tab button while the popover is showing → the popover dismisses immediately and does not reappear later.
6. Reset with Debug menu → Promo Queue → Cookie Popups Blocked → Undismiss + Clear History
7. Repeat steps 1-2 to trigger the promo, then close the popover with the "x" button → the popover dismisses immediately and does not reappear later.
8. Reset with Debug menu → Promo Queue → Cookie Popups Blocked → Undismiss + Clear History
9. Repeat steps 1-2 to trigger the promo, then wait 5 seconds → the popover auto-dismisses and does not reappear later.

### Impact
Medium — this changes how the promo is triggered, and a mistake here could either re-ask people who already answered or stop the promo appearing at all.

#### What could go wrong?

- Users who already dismissed the old dialog could be asked again → mitigated by reading the existing "already shown" setting and retiring the promo, covered by unit tests.
- The promo could be suppressed more than expected → mitigated by monitoring the promo's existing pixels in a dashboard after the promo ships in the queue.
- A problem with this promo specifically would previously have meant disabling the entire Promo Queue → mitigated by the per-promo kill switch.
- With the Promo Queue off, the promo never appears → accepted (similar to other promos already in the queue).

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes promo orchestration and dismissal timing; mistakes could suppress the tip, double-prompt users, or leave promos pending if presentation fails—mitigated by legacy seen-flag handling, presenter success/failure signaling, and unit tests.
> 
> **Overview**
> Moves the **“N cookie pop-ups blocked”** feature tip from the removed `AutoconsentStatsPopoverCoordinator` into the **Promo Queue** via new `CookiePopupsBlockedPromoDelegate`, registered as `cookie-popups-blocked` with trigger **`appBecameActive`** (replacing the direct `applicationDidBecomeActive` coordinator call).
> 
> **Eligibility and rollout:** Gated by remote flags `promoQueue` (existing) and new **`promoQueueCookiePopupsBlockedPromo`** / `PromoQueueSubfeature.cookiePopupsBlockedPromo`. Same rules as before (autoconsent on, protections report on NTP, onboarding done, ≥2 days since install, ≥5 blocked pop-ups), plus the legacy **`blockedCookiesPopoverSeen`** key so users who already saw the old popover are not re-prompted.
> 
> **UX / timing:** Popover auto-dismiss is owned by the queue (**~5s** feature-tip timeout) instead of the popover’s former **8s** timer (`autoDismissDuration` is now `nil`). `AutoconsentStatsPopoverPresenter.showPopover` returns **`Bool`** so failed anchor presentation yields `.noChange` instead of a stuck pending promo.
> 
> **Wiring:** `AppDelegate` and `TabBarViewController` use the delegate for new-tab dismiss; debug menu drops the old “show feature awareness” action and clears the seen flag via the delegate storage key. Coordinator tests are replaced by `CookiePopupsBlockedPromoDelegateTests`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80f12350249d846cd0784931dcb61a9c6ea37650. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


miasma13 (approved):
Tested and the promo works as expected before, thanks for migrating it to Promo Queue.

PR #6643, opened 2026-09-02
macOS: Add debug Permissions inspector at duck://permissions

andremartingo:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671677432066/task/1218107429302561?focus=true
Tech Design URL:
CC:

### Description

Adds an internal-user-only debug page that dumps every persisted website permission, at `duck://permissions`, reachable via **Debug ▸ Permissions ▸ Inspect**. It's modelled on the existing Favicons inspector (`duck://favicons`).

It's read-only apart from row deletion, and can be filtered by domain text or by permission type.

Two columns are not stored columns, are marked italic behind a divider, and carry tooltips saying so:

- **effective** — what the app actually acts on. It differs from the stored decision only while a decision override is active for that domain and type (duck.ai microphone today), which is precisely the case the raw columns can't reveal. Such rows are highlighted.
- **fireproof** — whether the domain is fireproof, since those keep their permissions through a burn.

Reading the real columns needed a small store addition. The existing entity type only exposes the already-decoded decision, so deriving `allow`/`isRemoved` back from it would misreport any row the app didn't write itself. The new read returns the columns as stored, with a protocol-extension default that re-encodes them from the decoded entities so test doubles need no changes.

The page's markup and script ship as bundle resources rather than Swift string literals, so they can be edited and linted as real HTML and JS.

CleanShot 2026-09-02 at 18 13 27@2x

### Testing Steps

1. Run a Debug build and enable **Debug ▸ Set Internal User State** (the page is gated on internal user regardless of build type)
2. Visit a site that asks for a permission (camera, location, notifications) and allow or deny it
3. **Debug ▸ Permissions ▸ Inspect** → new tab lists the persisted permissions, one row per domain + type, with the stored `allow` and `isRemoved` values
4. Type in the filter field → rows narrow by domain or type; pick a type from the dropdown → only that type remains; both together narrow further
5. Click a column header → sorts by it; click again → reverses
6. Tick some rows → **Delete selected** → those rows go, the rest stay
7. **Delete all** → click once to arm ("Click again to delete ALL"), click again → list empties
8. **Debug ▸ Remove Internal User State**, then reload the tab → blank page instead of the inspector

### Impact

None — internal tooling. The page is unreachable for external users, and nothing in the change alters permission behaviour for anyone.

#### What could go wrong?

- The page ships as bundle resources, so a packaging mistake would serve a blank page instead of failing to build → mitigated by a unit test asserting both resources are present in the bundle.
- A row the app didn't write (e.g. `allow:true, isRemoved:true`) could be displayed as a decision it isn't → mitigated by dumping the raw columns instead of deriving them, plus a test covering the encoding.
- Deleting from the page could leave open tabs acting on stale state → mitigated by routing deletes through the same manager API the rest of the app uses, so the existing publisher fires.

### Quality Considerations

- **Privacy**: the page lists the user's own local permission data, in their own browser, gated on internal-user state. Nothing leaves the device, and it makes no network requests.
- Requests are served synchronously (the permission store reads are synchronous), so there's no in-flight scheme-task handling to get wrong.
- Tests cover the internal-user gate in both directions, the JSON API, the stored-column encoding, the override flag, and an end-to-end load of the page in a `WKWebView` through the real scheme handler asserting the rendered rows and the type filter.

---

> [!NOTE]
> **Low Risk**
> Debug-only internal tooling with no change to production permission behavior for non-internal users; delete paths reuse existing manager APIs.
> 
> **Overview**
> Adds an **internal-user-only** debug permissions inspector at `duck://permissions`, opened from **Debug ▸ Permissions ▸ Inspect**, following the same `duck://` scheme pattern as the favicons debug page.
> 
> `PermissionsDebugInspector` serves bundled HTML/JS and JSON APIs (`/api/list`, `/api/remove`, `/api/removeAll`) that dump **raw stored** permission rows (`allow`, `isRemoved`, types) plus runtime **effective** decisions and **fireproof** status, including highlighting when a `PermissionDecisionOverriding` disagrees with storage. Deletes go through `PermissionManager` so in-memory state and publishers stay consistent.
> 
> Supporting changes wire `duck://permissions` in `DuckURLSchemeHandler` (gated on internal user), extend `PermissionStore` with `loadRawPermissions()` for true Core Data columns, and add `PermissionManagerDebugging` on `PermissionManager` and mocks. Unit tests cover the API, override reporting, internal-user gating, and end-to-end rendering in a `WKWebView`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e82f8bec8967009c73699dade96fe1dad24ad3e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ayoy (approved):
LGTM, this is working great and will help reason about and debug permissions going forward. Great improvement, @andremartingo!

You'll have to fix SwiftLint violation before merging, and here are a few suggestions you may implement to make the inspector more eye-catching:
1. I'd add color-coding to the page, i.e. use green, yellow and red for allow, ask and deny.
2. autoplay_policy has a special mapping of database state to the effective value: _allow_ -> _video and audio_, _ask_ -> _stop videos with sound_,  _deny_ -> _never_. Would be great to have it reflected in the inspector.
3. You could consider add emojis to permissions, e.g. "🎥 camera", "📍 geolocation", etc. Claude should come up with appropriate emojis.
4. Consider making domains clickable and opening a website in a new tab.

No need for another review unless you want an extra pair of eyes, I have verified that this website is visible to internal users only so we should be good to go from here :)

Thanks!

andremartingo:
> LGTM, this is working great and will help reason about and debug permissions going forward. Great improvement, @andremartingo!
> 
> You'll have to fix SwiftLint violation before merging, and here are a few suggestions you may implement to make the inspector more eye-catching:
> 
> 1. I'd add color-coding to the page, i.e. use green, yellow and red for allow, ask and deny.
> 2. autoplay_policy has a special mapping of database state to the effective value: _allow_ -> _video and audio_, _ask_ -> _stop videos with sound_,  _deny_ -> _never_. Would be great to have it reflected in the inspector.
> 3. You could consider add emojis to permissions, e.g. "🎥 camera", "📍 geolocation", etc. Claude should come up with appropriate emojis.
> 4. Consider making domains clickable and opening a website in a new tab.
> 
> No need for another review unless you want an extra pair of eyes, I have verified that this website is visible to internal users only so we should be good to go from here :)
> 
> Thanks!

Sounds good 👍There you go. 
image

PR #6664, opened 2026-09-03
iOS: Grow the unified input out of the floating omnibar pill

jaceklyp:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217951612193820?focus=true

### Description

iPhone, floating omnibar, bottom address bar. Focusing the omnibar made the expanded input slide up from behind the floating bar. The morph out of the search pill was already correct — it was just drawn underneath the bar, so it was invisible until it had grown past it.

- The input now grows out of the search pill while the floating bar collapses to its button row.
- The collapse no longer snaps at the end. The keyboard is released only once the collapse animation owns the container's position, enforced in one place for every dismiss route (back button, query submit, suggestion, voice).

Two pre-existing bugs found while testing:

- The address bar could disappear: focus + dismiss, then move it to top or rotate to landscape, left it invisible. Its visibility is now part of the "hosted in the nav container" pose instead of a leftover of the focus transition.
- Rotating mid-focus made the collapse replay the other orientation's geometry, landing the card as a narrow stub in the wrong place. That geometry is now stamped with the window size it describes.

### Testing Steps

iPhone only, floating omnibar on.

1. Bottom address bar → tap it. The input grows out of the pill; it should not appear from behind the bar.
2. Dismiss via back button → collapses onto the pill, no jump at the end.
3. Same for submitting a query, tapping a suggestion, and voice.
4. Focus + dismiss, then long-press → Move to Top → address bar still visible.
5. Focus + dismiss, then rotate to landscape → address bar still visible.
6. Open the input and rotate while open → no second omnibar behind it.
7. Focus landscape → rotate → dismiss portrait, and mirrored. Sane landing (see below).
8. Top address bar, and a Duck.ai tab rotated → unchanged.

### Known limitation

Dismissing after rotating mid-focus lands on standard margins rather than exactly on the pill, so it snaps ~83pt. Landing exactly needs the pill's frame derived while the omnibar is detached, which touches hand-off math used by every dismiss — not worth the regression risk for an edge case. Filed as a follow-up.

### Impact

Low — iPhone only, behind the floating-UI flag. The visibility fix touches the shared address-bar hosting path; `verifyRuntimeAddressBarRelocation` now asserts the bar is visible after a focus/dismiss-then-relocate cycle, the case that previously slipped through.

---

> [!NOTE]
> **Medium Risk**
> Touches shared address-bar layout, UTI dismiss sequencing, and z-ordering behind a feature flag; incorrect timing or alpha restoration could affect omnibar visibility and keyboard dismiss on iPhone floating UI.
> 
> **Overview**
> **Floating bottom omnibar:** The unified toggle input is layered so it **grows out of the search pill** instead of appearing behind the bar. `bringFloatingTopNavigationBarToFrontIfNeeded` is renamed to **`bringFloatingNavigationBarToFrontIfNeeded`** and now runs whenever the omnibar is **not** hosted in the toolbar (top or bottom), keeping the pill above full-screen UTI layers for gestures after focus/dismiss.
> 
> **Dismiss animation:** Keyboard/input resign is deferred via a new **`resigningInput`** hook on `hideUnifiedToggleInputOmnibar`, invoked when the dismiss animator starts so the collapse owns layout before `deactivateInput()` (all dismiss paths). Coordinator dismiss cleanup no longer calls `deactivateInput()` immediately.
> 
> **Address bar visibility after UTI:** Re-hosting in the nav container goes through **`applyOmnibarHostedInNavigationContainerPose()`**, which restores nav container visibility and sets **`navigationBarCollectionView.alpha`** back to 1 when chrome isn’t hidden and UTI isn’t visible—fixing invisible bars after focus/dismiss then relocate or rotate.
> 
> **Rotation-safe hand-off geometry:** Omnibar placeholder X and bottom pill-matched insets are cached **with window size** and ignored when the window size changes, avoiding wrong collapse geometry mid-focus.
> 
> **Test:** `verifyRuntimeAddressBarRelocation` asserts the search field stays hittable after relocating the bar following a focus/dismiss cycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e805386eddb34b1925c7fad1457e9b04cd7d39bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


brindy (approved):
LGTM

PR #6666, opened 2026-09-03
EventHub metrics #2: Add config-driven experiment metrics to the EventHub

miasma13:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218062429029804
Tech Design URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1216911265362552
CC:

### Description

Adds experiment metrics to the EventHub. Metrics are declared in the `settings` of Content Scope Scripts and TDS experiment subfeatures in the remote config, and a matching event now reports a conversion to the Native Apps experiment framework — so a new metric can be hooked up to an experiment entirely through config, with no client release.

Metrics consume the same de-duplicated event stream the telemetry handlers do: the hub takes one de-duplication decision per web event, before fan-out, so a repeat on a page reaches no handler at all. Enrollment, conversion windows, threshold counting and the pixel itself remain the experiment framework's job — this side only reports.

Also folds the three native-event entry points into one, so a native occurrence reaches every handler exactly as a web event does, and fixes data parameter values arriving at the endpoint percent-encoded twice.

This is the second branch in a stack and targets `michal/eventhub-metrics-1-dedup`.

### Testing Steps

Use the setup in https://app.asana.com/1/137249556945/task/1218049910864102 and the Xcode debugger to see EventHub logs, then **add** the following override alongside the `webDetection` and `eventHub` ones from that task. It declares two metrics on one auto-enrolling CSS experiment: `adwallSeen` on the `test` event with two windows and two thresholds (four conversion requests), and `ytAdBlocker` on the `youtube_adBlocker` event (one request).

```json
"contentScopeExperiments": {
    "state": "enabled",
    "exceptions": [],
    "features": {
        "contentScopeExperimentMetricsTest": {
            "state": "enabled",
            "settings": {
                "metrics": {
                    "adwallSeen": {
                        "event": "test",
                        "conversions": [
                            { "windows": [[0, 0], [0, 7]], "thresholds": [1, 2] }
                        ]
                    },
                    "ytAdBlocker": {
                        "event": "youtube_adBlocker",
                        "conversions": [
                            { "windows": [[0, 7]], "thresholds": [1] }
                        ]
                    }
                }
            },
            "cohorts": [
                { "name": "control", "weight": 1 },
                { "name": "treatment", "weight": 1 }
            ]
        }
    }
}
```

Conversions are reported as `experiment_metrics_contentScopeExperimentMetricsTest_<cohort>` pixels carrying `metric`, `conversionWindowDays` and `value` (the threshold). Those fire unique by name and parameters, so each `(metric, window, threshold)` combination converts at most once.

**Test case 0:**

1. Load the config and launch the app
2. Filter the Xcode console on `EventHub`

- [x] A `parsed 5 metric conversion request(s) from N experiment(s)` line appears (the request count is the number to check — `N` counts every experiment subfeature carrying any settings, so it will be larger than one)

**Test case 1:**

1. Navigate to https://www.wikihow.com/Disable-Your-Ad-Blocker (detector needs to find "disable your ad blocker" text on the page)

- [x] `metric routing for test →` lists **four** requests for `adwallSeen` — both windows (`0-0`, `0-7`) against both thresholds (`1`, `2`). Order is not significant
- [x] Conversion pixels fire for `value=1` in both windows; the `value=2` combinations do not convert yet

2. Reload the same tab

- [x] No new `metric routing` line at all — the repeat is dropped at the hub before any handler runs, the same decision the counters and immediate pixels see

3. Navigate to some other page, then back to the wikihow page

- [x] `metric routing for test →` appears again (navigation cleared the de-dup state for that tab)
- [x] The `value=2` pixels now fire — a threshold of 2 means two distinct page loads, not two events on one page

4. Open the wikihow page in two more new tabs

- [x] Each tab produces its own `metric routing` line (de-dup is per tab)

**Test case 2:**

1. Navigate to https://privacy-test-pages.site/features/web-interference-detection/youtube-detection-events.html
2. Click "Show Ad-Blocker Modal" button

- [x] `test_ytVideoAd_immediate` fires with the `loginState` parameter, the `test_ytVideoAd_60s` counter increments, **and** `metric routing for youtube_adBlocker →` lists `ytAdBlocker 0-7×1` — one delivered event reaching all three handlers
- [x] The `ytAdBlocker` conversion pixel fires

3. Click "Show Ad-Blocker Modal" button again without reloading

- [x] No immediate pixel, no counter increment and no `metric routing` line — all three handlers share the one de-duplication decision

**Test case 3:**

1. On the same page, click "Show Ad-Blocker Modal" button

- [x] The EventHub log shows the data parameter as compact JSON with no percent escapes, e.g. `loginState: "unknown"`
- [ ] Inspecting the pixel request (network proxy) shows `loginState=%22unknown%22` — **not** `%2522unknown%2522`. Percent-decoding it once yields `"unknown"`, and JSON-decoding that yields `unknown`

**Test case 4:**

1. Trigger the `test` event on a fresh page and confirm `adwallSeen` converts
2. Remove `adwallSeen` from the experiment's `settings.metrics`, leaving `ytAdBlocker`, and reload the config

- [x] A new `parsed 1 metric conversion request(s)` line appears
- [x] A `test` event on a fresh page reports nothing for `adwallSeen`; `youtube_adBlocker` still reports `ytAdBlocker`

3. Restore `adwallSeen`, set the experiment's `state` to `disabled`, reload the config, trigger the `test` event on a fresh page

- [ ] No conversion pixels fire — a disabled experiment leaves the framework's active set

4. Set the `eventHub` feature's `state` to `disabled`, reload the config, trigger the `test` event on a fresh page

- [ ] No `metric routing` line at all — events reach metrics only through the hub, so the feature gate stops them

**Test case 5:**

1. Trigger an event type that no metric declares (any `webDetection` detector firing a different `fireEvent` type)

- [ ] `metric routing for <type> → no metric is declared for it (N request(s) loaded)`

### Impact

Medium. Metrics are inert until a config declares them, so nothing changes for users on current config. The de-duplication and telemetry paths from the previous branch in the stack are unchanged in behaviour.

The data parameter encoding fix does change the value of telemetry pixels that carry a `data` parameter — they were arriving double-encoded and unusable, and now arrive decodable. Worth flagging to Data Science, since any dashboard reading those values sees the format change.

#### What could go wrong?

- A misconfigured metric could report conversions for the wrong experiment → mitigated by requests being built per (experiment, metric) from the declaring experiment's own settings, covered by the M-SEL spec cases.
- A config typo could produce a runaway pixel → mitigated by the parser dropping thresholds below 1 and rejecting inverted or negative conversion windows.
- A period persisted before this change holds an already-encoded data value, so one pixel per data parameter can still go out double-encoded after upgrade, until the next matching event overwrites it.

### Quality Considerations

Conformance with the cross-platform behavioural specification is now complete on Apple: every case ID in the de-duplication, telemetry and metrics suites is covered by a named test, and `grep -o '[DMT]-[A-Z]\{3\}-[0-9]\+'` over `SharedPackages/EventHub/Sources/EventHubTests/Spec` returns the same set as the documents. 192 tests pass via both the iOS and macOS schemes.

Metrics hold no state and persist nothing — the registry is a derived view of the current config, rebuilt on every config change, and the event-time path is a filter over a flat list.

`platform-support.md` in `ddg-workflow` needs updating after this lands: it records five Apple divergences that this stack resolves, and its Apple column should become de-duplication-at-ingestion `Yes`, experiment metrics `Yes`, and `data` parameters `Yes`.

---

> [!NOTE]
> **Medium Risk**
> Changes telemetry wire format for data parameters and adds a new experiment conversion reporting path tied to remote config parsing; behavior is gated by `eventHub` and config but affects pixels Data Science may consume.
> 
> **Overview**
> Adds **config-driven experiment metrics** to EventHub: metrics declared in CSS/TDS experiment subfeature `settings` are parsed into flat `MetricRequest`s (windows × thresholds), kept in memory from a dedicated `experimentSettings` publisher, and on each delivered event reported through `EventHubConversionReporting` to **PixelExperimentKit** on iOS and macOS. **`PrivacyConfiguration.allSubfeatureSettings`** supports remote-only experiment IDs without compile-time subfeature enums; **`EventHubExperimentSettings`** merges settings from `contentScopeExperiments` and `contentBlocking`.
> 
> **Native ingress** replaces separate immediate and aggregated handlers with **`handleNativeEvent`**, which uses the same **`deliverLocked`** path as web events (immediate telemetry, period counters, and metrics). **`DataParameter`** now emits compact JSON only—percent-encoding is left to the pixel transport, fixing double-encoded telemetry `data` parameters.
> 
> Includes cross-platform **metrics spec tests**, expanded telemetry spec coverage, and mock/protocol updates across PrivacyConfig and app integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d91e134bdb454be3ea41e1d55cce2b92aaef8b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/PrivacyConfiguration.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/PrivacyConfiguration.swift
@@ -109,6 +109,13 @@ public protocol PrivacyConfiguration {
     /// Returns settings for a specified subfeature.
     func settings(for subfeature: any PrivacySubfeature) -> PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings?
 
+    /// Returns the raw `settings` JSON of every subfeature of `feature`, keyed by subfeature ID.
+    ///
+    /// Unlike `settings(for: any PrivacySubfeature)` this needs no compile-time subfeature enum, so it
+    /// reaches subfeatures whose IDs only exist in remote config — experiment subfeatures such as
+    /// `contentScopeExperiment1` or `tdsNextExperiment007`. Subfeatures with no `settings` are omitted.
+    func subfeatureSettings(for feature: PrivacyFeature) -> [SubfeatureID: PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings]
+
     /// Removes given domain from locally unprotected list.
     func userEnabledProtection(forDomain: String)
     /// Adds given domain to locally unprotected list.
@@ -142,5 +149,10 @@ public extension PrivacyConfiguration {
     func stateFor(subfeatureID: SubfeatureID, parentFeatureID: ParentFeatureID, randomizer: (Range<Double>) -> Double = Double.random(in:)) -> PrivacyConfigurationFeatureState {
         return stateFor(subfeatureID: subfeatureID, parentFeatureID: parentFeatureID, versionProvider: AppVersionProvider(), randomizer: randomizer)
     }
+
+    /// Defaulted so the many test doubles conforming to this protocol need not each implement it. It is
+    /// still a protocol requirement, so `AppPrivacyConfiguration`'s real implementation is reached
+    /// through the witness table rather than shadowed by this.
+    func subfeatureSettings(for feature: PrivacyFeature) -> [SubfeatureID: PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings] { [:] }
 
 }
--- a/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/PrivacyConfiguration.swift
+++ b/SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/PrivacyConfiguration.swift
@@ -114,7 +114,7 @@ public protocol PrivacyConfiguration {
     /// Unlike `settings(for: any PrivacySubfeature)` this needs no compile-time subfeature enum, so it
     /// reaches subfeatures whose IDs only exist in remote config — experiment subfeatures such as
     /// `contentScopeExperiment1` or `tdsNextExperiment007`. Subfeatures with no `settings` are omitted.
-    func subfeatureSettings(for feature: PrivacyFeature) -> [SubfeatureID: PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings]
+    func allSubfeatureSettings(for feature: PrivacyFeature) -> [SubfeatureID: PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings]
 
     /// Removes given domain from locally unprotected list.
     func userEnabledProtection(forDomain: String)
@@ -149,10 +149,5 @@ public extension PrivacyConfiguration {
     func stateFor(subfeatureID: SubfeatureID, parentFeatureID: ParentFeatureID, randomizer: (Range<Double>) -> Double = Double.random(in:)) -> PrivacyConfigurationFeatureState {
         return stateFor(subfeatureID: subfeatureID, parentFeatureID: parentFeatureID, versionProvider: AppVersionProvider(), randomizer: randomizer)
     }
-
-    /// Defaulted so the many test doubles conforming to this protocol need not each implement it. It is
-    /// still a protocol requirement, so `AppPrivacyConfiguration`'s real implementation is reached
-    /// through the witness table rather than shadowed by this.
-    func subfeatureSettings(for feature: PrivacyFeature) -> [SubfeatureID: PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings] { [:] }
 
 }
--- a/SharedPackages/EventHub/Sources/EventHubTests/Spec/MetricsSpecTests.swift
+++ b/SharedPackages/EventHub/Sources/EventHubTests/Spec/MetricsSpecTests.swift
@@ -1,0 +1,391 @@
+//
+//  MetricsSpecTests.swift
+//
+//  Copyright © 2026 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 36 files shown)


SharedPackages/BrowserServicesKit/Sources/PrivacyConfig/PrivacyConfiguration.swift
```diff
@@ -143,4 +150,9 @@ public extension PrivacyConfiguration {
         return stateFor(subfeatureID: subfeatureID, parentFeatureID: parentFeatureID, versionProvider: AppVersionProvider(), randomizer: randomizer)
     }
 
+    /// Defaulted so the many test doubles conforming to this protocol need not each implement it. It is
+    /// still a protocol requirement, so `AppPrivacyConfiguration`'s real implementation is reached
+    /// through the witness table rather than shadowed by this.
+    func subfeatureSettings(for feature: PrivacyFeature) -> [SubfeatureID: PrivacyConfigurationData.PrivacyFeature.SubfeatureSettings] { [:] }
```
SabrinaTardio: I am not sure about this default... can we implement it on the shared mocks?

SharedPackages/EventHub/Sources/EventHubTests/Spec/MetricsSpecTests.swift
```diff
@@ -0,0 +1,391 @@
...
+        let page = f.openPage()
+        f.send("captchaDetected", reason: "any", on: page)
+        f.send("reloadLoopDetected", reason: "any", on: page)
+        f.send("searchPerformed", reason: "any", on: page)
+        f.sendNative("appLaunch", payload: ["launchType": "cold"])
+
+        #expect(f.requested.isEmpty)
+    }
+
+    @Test("M-SEL-8: metrics attach under both experiment parent features")
```
SabrinaTardio, line 223: this injects the settings dictionary directly, so removing .contentBlocking from parentFeatures leaves the test green

SabrinaTardio (approved):
LGTM

PR #6673, opened 2026-09-03
iOS: Inset the address field content and fix the trackers-blocked shield burst

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1216033800174500
Tech Design URL: N/A
CC: N/A

Stacked on `kk/remove-fire-button-select-mode`.

### Description
- Insets the address field's content 2pt from both capsule ends. The 44pt icon slots were flush with the field's edges, so the shield sat 2pt short of the leading arc's centre and the trailing buttons ran up to the edge.
- Fixes the trackers-blocked shield animation only ever showing confetti to the bottom-right. The confetti is one dot copied eight times by a Lottie Repeater, which the main-thread rendering engine skips. The shield views now use the automatic engine, which renders it through Core Animation and falls back if a file isn't supported.

### Testing Steps
1. Enable the floating UI and set the address bar to the bottom.
2. Load a page → the shield sits centred on the field's leading arc, and the trailing buttons keep 2pt of air from the trailing end.
3. Load a tracker-heavy page (e.g. cnn.com) → after the "N Trackers Blocked" notification, the shield bursts with a full ring of dots rather than a single dot to the bottom-right.
4. Toggle dark mode and repeat step 3 → the shield renders as before at rest and during the burst.

### Impact
Low: visual change to the address field and its animation.

#### What could go wrong?
- A Lottie file unsupported by the Core Animation engine could render differently → `.automatic` falls back to the main-thread engine for such files; the three shield files were checked on device.

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Visual layout and Lottie rendering only; `.automatic` falls back to main-thread for unsupported files, which were verified on device.
> 
> **Overview**
> **Insets embedded address-field content** by 2pt on the leading and trailing edges so icon slots and trailing controls sit inside the capsule rather than flush with the ends. On a 48pt-tall field that also centers the privacy shield on the leading arc for a symmetrical trackers-blocked burst. When the leading icon area is hidden, leading inset stays at 16pt via shared `Metrics` constants.
> 
[...]

PR #6674, opened 2026-09-03
iOS: Match the split toolbar to the bottom chrome and flatten the embedded address field

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414235014887631/task/1216033800174500
Tech Design URL: N/A
CC: N/A

Stacked on https://github.com/duckduckgo/apple-browsers/pull/6673.

### Description
- Lays out the split (top-address-bar) toolbar pill with the same geometry as the combined bottom chrome: same edge inset on every side, same button-row padding, so its outer buttons land exactly where the bottom chrome's icons do.
- Derives that edge inset from the system's corner-adapted safe-area guide plus a 4pt tuck, instead of a fixed 20pt. On iPhone 17 that's 22pt, matching the system tab bar, and it adapts per device. The fixed value remains only as a fallback while the guide is unresolved.
- Gives the address field embedded in the bottom chrome a flat, bright fill instead of a second layer of glass with its own highlights. The top-placement and minimal-chrome fields keep their glass.

### Testing Steps
1. Enable the floating UI on an iPhone running iOS 26.
2. Set the address bar to the bottom and load a page → the address field is a flat white fill inside the chrome, with no glass highlights.
3. Scroll the page so the chrome collapses to the button row → note the pill's edges and button positions.
4. Switch the address bar to the top → the standalone toolbar pill has the same edges and button positions as in step 3, sitting about 22pt from the screen edges.
5. Compare against a system tab bar (e.g. any first-party app) → the side inset matches.

### Impact
Low: visual layout change to the floating chrome.

#### What could go wrong?
- The corner-adapted guide reports an empty frame before the view is in a window → the previous fixed 20pt inset is used until it resolves, covered by unit tests.
- Landscape with the Dynamic Island widens one guide → the narrower side sets the inset and the wider side stays on its guide, covered by the asymmetric-guide test.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Visual and layout-only changes to iOS 26 floating toolbar/omnibar; pre-iOS 26 paths are largely preserved and behavior is covered by updated unit tests.
> 
> **Overview**
> Aligns **iOS 26 floating chrome** so the split (top address bar) toolbar pill uses the same layout as the combined bottom bar: shared button-row padding, content insets, and capsule framing via one `floatingRestStateOuterInsets` path instead of separate standalone vs embedded geometry.
> 
> **Edge inset** is no longer a fixed 20pt on iOS 26. It comes from the corner-adapted safe-area guide plus a **4pt tuck** (`floatingPhysicalInset`), with the previous concentric value as fallback when the guide is unresolved. Bottom glass offset and `restingCapsuleFrame` use that same physical inset on every edge.
> 
> Adds **`floatingEmbeddedAddressBarBackground`** and changes the **bottom embedded omnibar** on iOS 26 to a flat fill (no second `UIGlassEffect` layer); top and minimal-chrome fields still use glass.
> 
> Unit tests are updated for the new inset math and to assert standalone outer buttons line up with combined chrome.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2338098237e2308b821643480d767bd5d951ed06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/Infrastructure/DesignResourcesKit/Sources/DesignResourcesKit/Colors/ColorPalettes/DefaultColorPalette.swift
+++ b/SharedPackages/Infrastructure/DesignResourcesKit/Sources/DesignResourcesKit/Colors/ColorPalettes/DefaultColorPalette.swift
@@ -340,6 +340,8 @@ struct DefaultColorPalette: ColorPaletteDefinition {
             return DynamicColor(lightColor: .shade(0.06), darkColor: .tint(0.12))
         case .floatingAddressBarBackground:
             return DynamicColor(lightColor: .shade(0.05), darkColor: .tint(0.08))
+        case .floatingEmbeddedAddressBarBackground:
+            return DynamicColor(lightColor: .tint(0.8), darkColor: .tint(0.08))
         case .unifiedToggleInputAttachmentErrorBannerBackground:
             return DynamicColor(lightColor: xF6CDD1, darkColor: x5A2A2A)
         case .unifiedToggleInputAttachmentErrorText:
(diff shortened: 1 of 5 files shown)


brindy (approved):
Bit wordy on the comments again :) but LGTM

PR #6675, opened 2026-09-03
macOS: New Create Image flow telemetry

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1218029146860297?focus=true
Tech Design URL:
CC:

### Description
Adds pixels for the new Create Image flow to mimic iOS.

### Testing Steps
Use a macOS debug build with `updatedCreateImage` enabled and inspect outgoing PixelKit events.

1. Select a model without image-generation support, then choose Tools → Create Image → `m_mac_aichat_addressbar_create_image_model_switched` fires with the previous/new model IDs, privacy flag, and `entry_point=tools_menu`.
2. Dismiss the resulting model-switch notice → `m_mac_aichat_addressbar_create_image_model_switch_notice_dismissed` fires.
3. Stub `/models` with an unsupported selected model and no accessible image-capable model, then choose Create Image → `m_mac_aichat_addressbar_create_image_unavailable` fires.
4. Keep Create Image active while stubbing the selected model as unsupported, then submit a prompt → `m_mac_aichat_addressbar_create_image_submitted_with_unsupported_model` fires.

### Impact
Low. Telemetry-only changes behind the existing Create Image flow.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Telemetry-only changes on the existing Create Image path; no auth, data handling, or submission logic beyond firing pixels.
> 
> **Overview**
> Adds **macOS analytics for the updated Create Image flow**, aligned with iOS, for both the AI chat address bar and the Prompt Bar.
> 
> **Instrumentation:** Fires when Create Image auto-switches from an unsupported model (with `from_model_id`, `to_model_id`, privacy flag, `entry_point=tools_menu`), when the user dismisses the model-switch notice, when no image-capable model is available, and as an error monitor when a Create Image prompt is submitted while the selected model does not support image generation. Submit-time image/web-search pixels are centralized in `firePromptSubmissionPixels()` on `activeToolMode`.
> 
> **Plumbing:** New `DuckAIPromptPixelEvent` cases map through `AddressBarPromptPixelHandler` and `PromptBarPixelHandler` to `AIChatPixel` / `PromptBarPixel`, with `create_image_unavailable` using **daily** frequency (others **daily_count**). JSON pixel definitions and unit tests cover mappings, parameters, and controller behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c99728bd8f143518e849066466d538e0d9e1c53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
@@ -868,13 +868,23 @@ final class AIChatOmnibarController {
     private func switchToImageGenerationModelIfNeeded() -> AIChatCreateImageModelSwitchNotice? {
         guard isUpdatedCreateImageEnabled,
               let previousModel = selectedModel,
-              !previousModel.supportsTool(.imageGeneration),
-              let fallbackModel = imageGenerationModel else {
+              !previousModel.supportsTool(.imageGeneration) else {
+            return nil
+        }
+
+        guard let fallbackModel = imageGenerationModel else {
+            pixelHandler.fire(.createImageUnavailable)
             return nil
         }
 
         updateSelectedModel(fallbackModel.id)
-        return AIChatCreateImageModelSwitchNotice(previousModel: previousModel, newModel: fallbackModel)
+        let notice = AIChatCreateImageModelSwitchNotice(previousModel: previousModel, newModel: fallbackModel)
+        pixelHandler.fire(.createImageModelSwitched(
+            fromModelId: previousModel.id,
+            toModelId: fallbackModel.id,
+            fromModelPrivacyPreserving: notice.previousModelHasExtraPrivacyProtections
+        ))
+        return notice
     }
 
     /// The model ID to use for the current submission. In image-generation mode an
@@ -1370,6 +1380,9 @@ final class AIChatOmnibarController {
         usageWarningMeasurement.promptSubmitted()
 
         if isImageGenerationMode {
+            if !selectedModelSupportsImageGeneration {
+                pixelHandler.fire(.createImageSubmittedWithUnsupportedModel)
+            }
             pixelHandler.fire(.imageGenerationSubmitted)
         } else if isWebSearchMode {
             pixelHandler.fire(.webSearchSubmitted)
--- a/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
@@ -1375,18 +1375,9 @@ final class AIChatOmnibarController {
             return
         }
 
-        pixelHandler.fire(.promptSubmitted)
+        firePromptSubmissionPixels()
         // After the URL branch: navigating away is not a prompt spent against the allowance.
         usageWarningMeasurement.promptSubmitted()
-
-        if isImageGenerationMode {
-            if !selectedModelSupportsImageGeneration {
-                pixelHandler.fire(.createImageSubmittedWithUnsupportedModel)
-            }
-            pixelHandler.fire(.imageGenerationSubmitted)
-        } else if isWebSearchMode {
-            pixelHandler.fire(.webSearchSubmitted)
-        }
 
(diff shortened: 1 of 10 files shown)


tomasstrba (changes requested):
@jotaemepereira, I'd recommend to fire these pixels for the prompt bar as well. (Ideally on NTP too, but I assume it's not scope of this PR)

macOS/UnitTests/AIChat/AIChatOmnibarControllerTests.swift
```diff
@@ -1604,6 +1622,27 @@ final class AIChatOmnibarControllerTests: XCTestCase {
         XCTAssertEqual(query.modelId, "img-model")
         XCTAssertEqual(query.toolChoice, [AIChatRAGTool.imageGeneration.rawValue])
         XCTAssertNil(query.mode)
+        XCTAssertFalse(pixelHandler.events.contains(.createImageSubmittedWithUnsupportedModel))
+    }
+
+    func testWhenImageGenerationIsSubmittedWithKnownUnsupportedSelectedModel_ThenErrorPixelFires() async {
```
tomasstrba: This test asserts a state that can't be reached through the UI, and it would pass even if the guard were wrong.

With `updatedCreateImage = false` and an unsupported selected model, `AIChatCreateImagePresentationPolicy.isImageGenerationItemVisible` is `false` — the tools-menu Create Image item is hidden. The test only gets into image-generation mode by calling `toggleImageGenerationMode()` directly.

Separately, the stub includes a `fallback` model with `GenerateImage`, so `effectiveModelId` resolves to it and the submission actually succeeds. We're asserting an error-monitor pixel for a healthy submission.

The case that can actually regress is: flag **on**, no image-capable model in the list at all — so `imageGenerationModel == nil`, `effectiveModelId` is nil and submission degrades to `effectiveMode`. That's what testing step 4 in the description describes, and it isn't covered. As written, the test passes whether the production guard is `!selectedModelSupportsImageGeneration` or `imageGenerationModel == nil`.

The production guard itself is fine — with the flag on, a successful switch makes `selectedModelSupportsImageGeneration` true, so it reduces to "no image-capable model". Only the test setup is off.

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
```diff
@@ -860,13 +860,23 @@ final class AIChatOmnibarController {
...
         guard isUpdatedCreateImageEnabled,
               let previousModel = selectedModel,
-              !previousModel.supportsTool(.imageGeneration),
-              let fallbackModel = imageGenerationModel else {
+              !previousModel.supportsTool(.imageGeneration) else {
+            return nil
+        }
+
+        guard let fallbackModel = imageGenerationModel else {
+            pixelHandler.fire(.createImageUnavailable)
```
tomasstrba, line 876: Worth knowing when reading the dashboard: `switchToImageGenerationModelIfNeeded()` has a second caller inside the `fetchModels()` completion, which runs on every omnibar activation while image mode is active. So `create_image_unavailable` can fire without the user touching Create Image in that session — e.g. tool mode restored from the draft store on a tab switch.

`.daily` caps the volume and the json5 wording is loose enough that I wouldn't change the behaviour. Just flagging it so nobody reads this as a pure user-action count.

tomasstrba (commented):
Two minor finds by Claude:

macOS/UnitTests/AIChat/AIChatOmnibarControllerTests.swift
jotaemepereira: Fixed.

macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift
jotaemepereira: Makes sense.

tomasstrba (approved):
LGTM! ✔️ Thanks for covering the prompt bar too

PR #6681, opened 2026-09-04
iOS: Keep deleted Duck.ai chats deleted after reinstall

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414709148257752/task/1218160695252279?focus=true

### Description

Duck.ai chats that a user deleted could come back after the app was reinstalled, including chats deleted with the Fire Button.

The sync account is stored in the keychain, which survives app deletion, and auto-restore keeps it on purpose so a user can restore sync without their recovery code. In that state sync is off but an account still exists, and the read and delete paths treated it differently. Duck.ai could still read chat history from the sync server, while the app was not allowed to delete chats from it, so every local deletion came back the next time Duck.ai loaded.

The app now reads and writes chat history on the sync server only when there is a live sync session. That is the same condition already used for server-side deletion.

Nothing changes when sync is on, or when there is no account. Users with a preserved account will no longer see server-side chats pulled into Duck.ai until they restore sync. macOS does not enable auto-restore, so it never reaches this state and is not affected.

### Testing Steps

**Set up the affected state**

1. On a device running the current App Store build, set up Sync (Settings → Sync & Backup). Completing setup also opts you into auto-restore.
2. Open Duck.ai and start two or three chats so there is history on the server.
3. Delete the DuckDuckGo app from the device. The sync account stays in the keychain.
4. Install the build from this branch.
5. Launch the app and skip the prompt to restore sync → sync is now off, but the old account is still in the keychain.

**Verify the fix**

6. Open Duck.ai → the chat list is empty. The chats from step 2 are not pulled down from the server. (Before this change they would appear.)
7. Start a new chat, then delete it from the recent chats list → it disappears from the list.
8. Force quit, relaunch, and open Duck.ai → the deleted chat does not come back.
9. Start another chat and burn it with the Fire Button → the chat is cleared.
10. Force quit, relaunch, and open Duck.ai → the burned chat does not come back.

**Check sync still works**

11. Go to Settings → Sync & Backup and restore or enable sync.
12. Open Duck.ai → chat history from the server appears as normal.
13. Delete a chat, then check a second synced device → the chat is gone there too.

### Impact and Risks

**Medium.**

This changes when Duck.ai is allowed to talk to the sync server. If the new condition were too strict, chat history sync could stop working for users who do have sync on. The condition covers both an active session and the mid-pairing state, and the unit tests cover each auth state, so that is unlikely.

The change lives in a shared package, so macOS compiles it too. macOS does not enable auto-restore, so it never reaches the affected state and behaviour there is unchanged.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes when Duck.ai may read from the sync server and mint tokens; overly strict gating could break chat sync for users with sync on, though active and adding-new-device paths are explicitly preserved and tested.
> 
> **Overview**
> Fixes Duck.ai chat history reappearing after reinstall or when sync is off but a **keychain-preserved** sync account still exists (`authState` `.inactive`).
> 
> **`AIChatSyncHandler`** now uses an `activeSyncAccount` helper that only treats `.active` and `.addingNewDevice` as a live session—not a bare `sync.account != nil`. **`isSyncTurnedOn()`**, **`getScopedToken()`**, and **`getSyncStatus()`** all follow that rule: inactive + preserved account reports sync off (no scoped token, no user/device IDs), while still advertising sync as available-but-signed-out so the web app does not retry on errors.
> 
> Unit tests cover inactive preserved account, mid-pairing **`.addingNewDevice`**, and the signed-out status shape. **`content-scope-scripts`** is bumped **16.9.0 → 17.0.0** in `Package.resolved`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit caca0bd14a2e56dc233d6aefb6dde6e3cd8f642a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SharedPackages/AIChat/Package.resolved
+++ b/SharedPackages/AIChat/Package.resolved
@@ -1,13 +1,13 @@
 {
-  "originHash" : "d61e9ef500900a5d720f556507bc377c5db9d8c657dbfb0d578f7c3233fcced1",
+  "originHash" : "8ec7a7948f4320f24c9b3ccca107b34bd425edcbfd3f0e6d48521dd5cfea73bf",
   "pins" : [
     {
       "identity" : "content-scope-scripts",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/duckduckgo/content-scope-scripts.git",
       "state" : {
-        "revision" : "21aec4d902338cb10c422e091eb6a8c9f38a9be1",
-        "version" : "16.9.0"
+        "revision" : "e27d61495b3e8793e507a483d885b758e75c0732",
+        "version" : "17.0.0"
       }
     },
     {
--- a/SharedPackages/AIChat/Sources/AIChat/Shared/Sync/AIChatSyncHandler.swift
+++ b/SharedPackages/AIChat/Sources/AIChat/Shared/Sync/AIChatSyncHandler.swift
@@ -85,8 +85,22 @@ public class AIChatSyncHandler: AIChatSyncHandling {
         }
     }
 
+    /// The account backing a live sync session, or `nil` when sync is off.
+    ///
+    /// Deliberately not a plain `account != nil` check: the account lives in the keychain, which
+    /// survives app deletion, and `syncAutoRestore` preserves it while leaving `authState` as
+    /// `.inactive`.
+    private var activeSyncAccount: SyncAccount? {
+        switch sync.authState {
+        case .active, .addingNewDevice:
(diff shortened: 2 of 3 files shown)


amddg44 (approved):
LGTM! Nice work 👏

PR #6695, opened 2026-09-04
[macOS] Migrate Broken Site promo to Promo Queue

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1209825025475019/task/1218155791703504?focus=true
CC:

### Description
Moves the "Site not working?" prompt onto the macOS Promo Queue. This prevents collisions with other promos (two promos appearing at the same time) and ensures a cooldown between this and a following promo.

Notable changes:

* If another promo with a `webPage` or `global` context is already visible, this popover will be suppressed until it is triggered again.
* Once this promo has been dismissed, a 1-hour global cooldown will be applied, suppressing any other user-initiated promos. This promo bypasses the global cooldown, so it can appear immediately after another promo was dismissed.

Otherwise, the promo behavior remains the same, e.g. triggered by 3 consecutive refreshes, a 7-day cooldown, a longer 30-day cooldown after a 3-dismiss streak.

A note about scope:
Because this is a recurring promo (never permanently dismissed), it doesn't fit the current Promo Queue framework for recording promo history. Instead of migrating that history/eligibility logic entirely to the Promo Queue, which would require framework updates out of scope for this project, this promo continues to use `BrokenSitePromptLimiter` to gate its eligibility and record its history. (If future promos require similar logic, the Promo Queue can be extended to support that.)

### Testing Steps
1. Debug menu → Promo Queue → Reset Broken Site Prompt State
2. Open a non-DuckDuckGo site and reload it three times in quick succession →
   the "Site not working?" popover appears from the privacy dashboard button.
3. Click the "Let us know" button → the Privacy Dashboard breakage report opens and the popover closes.
4. Reload three times again → nothing appears (7-day cooldown).
5. Debug menu → "Advance Simulated Date" → by 1 week, then reload three times →
   the popover appears again.
6. Repeat step 5, dismissing with the close button three times in a row → on the
   fourth cycle nothing appears (dismiss streak).
7. Debug menu → "Advance Simulated Date" → by 1 month, then reload three times →
   the popover appears again.

### Impact

Medium — this changes the mechanism for triggering the promo (handling when it is triggered, how it is presented, and the side effects when it is clicked or dismissed)

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes when and whether the breakage prompt appears relative to other promos and cooldowns, though eligibility rules stay on the limiter. Persistence migration must preserve cooldown state for upgrading users.
> 
> **Overview**
> Moves the **"Site not working?"** popover off the navigation bar’s direct `pageRefreshMonitor` handler and into the macOS **Promo Queue**, triggered by a new **`pageRefreshPatternDetected`** event after repeated reloads.
> 
> **`BrokenSitePromoDelegate`** gates on onboarding completion, the privacy-config feature flag, and existing **`BrokenSitePromptLimiter`** rules (7-day cooldown, dismiss streak). Show/dismiss/report still update the limiter and pixels; promo history uses **`.ignored(cooldown:)`** instead of permanent retirement so the prompt can recur. The delegate does **not** respect the global promo cooldown; other user-initiated promos still get the usual suppression after this one dismisses.
> 
> **`BrokenSitePromptLimiterStore`** now uses **`KeyedStorage`** with migration from legacy UserDefaults keys. macOS-only **`coolDownInterval`**, **`reset()`**, and **`debugAdvanceDate`** support Promo Queue debug menus.
> 
> Inline UI logic was removed from **`NavigationBarViewController`**; **`PromoDependencies`** / **`AppDelegate`** wire the limiter and config into promo factory registration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c23129fac10308442a1ce5918a980e4c36b59ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jaceklyp:
General comment: I don’t think this feature should be treated as a “promo” category. It’s not really a promo - something isn’t working, and we want to know about it as soon as possible rather than waiting for another opportunity for it to appear to the user.

I feel like we’re effectively degrading this feature by treating it this way. IMO, this popup should be an exception to the other popups that might currently be active. Did we actually consider that approach?

jaceklyp (changes requested):
if we decide to include it in the promo queue, here's the result of the agentic review combined with some of my input: https://claude.ai/code/artifact/98c972f6-2539-4665-976b-2e90fcb92947?via=banner_open

rachelmcr:
Thanks, I was putting together an overview of what could suppress this promo if it's in the queue and I also just discovered the retraction risk ("A queue retraction spends the 7-day cooldown on a prompt nobody saw"), which I hadn't caught before. That's a downside of retaining the limiter separate from the promo queue framework, and it's likely to be messy to work around.

It's starting to look like we should add this as an "external promo" in the queue. In that case, the popover's behavior wouldn't change at all (it would still present itself as before) but the queue would observe it to suppress other promos so they don't overlap. I'll revisit that decision before moving forward.

rachelmcr:
I reviewed the project requirements as agreed with stakeholders, and the concerns and other issues can be addressed by migrating this as an external promo (influences promos in the queue but retains control of its own presentation). Closing this PR in favor of https://github.com/duckduckgo/apple-browsers/pull/6757.

PR #6698, opened 2026-09-04
macOS: Improve Create Image flow on New Tab Page

jotaemepereira:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204006570077678/task/1216436095926133
Tech Design URL:
CC:

### Description

Adds the updated Create Image flow to the macOS New Tab Page. Native code selects an accessible image-capable model and provides the localized model-switch notice, while the frontend renders and dismisses it.

This depends on the corresponding `content-scope-scripts` changes in branch `pr/3007`.

### Testing Steps

1. Clone `content-scope-scripts`, check out branch `feature/anpete/ntp-omnibar-ai-mode-usage-drawer`, and drag the repository's root folder into `DuckDuckGo.xcworkspace` in Xcode.
2. Run macOS with the updated Create Image and NTP image-generation feature flags enabled, then open a New Tab Page in Duck.ai mode.
3. Select a model that does not support image generation, then choose **Tools → Create Image**. Verify that the model switches and the native-localized notice appears.
4. Dismiss the notice and submit an image prompt. Verify that the notice closes and Create Image remains active.

### Impact

Medium. The native and frontend contracts must remain aligned; the behavior is gated by the existing feature flags.

---

###### Internal references:
[Definition of Done](https://app.asana.com/0/1204186595873227/1204186896266138/f) | [Engineering Expectations](https://app.asana.com/0/481882893211075/1202500774821704/f) | [Tech Design Template](https://app.asana.com/0/481882893211075/1202500774821700/f)

---

> [!NOTE]
> **Medium Risk**
> Changes model selection, omnibar bridge payloads, and chat submission parameters behind feature flags, but native and web must stay in sync and wrong model resolution could send prompts on an unexpected model.
> 
> **Overview**
> **Adds a native-driven “updated Create Image” path for the macOS New Tab Page Duck.ai omnibar**, gated by `aiChatNtpImageGeneration` plus `updatedCreateImage`. When the web enables Create Image, native can pick an accessible image-generation model, persist the selection (and clear reasoning effort the new model does not support), and push a **localized model-switch notice** back through omnibar config for the frontend to show and dismiss.
> 
> The **native ↔ web contract** grows with `enableUpdatedCreateImage`, `createImageModelSwitch`, `omnibar_setImageGenerationActive`, and `omnibar_dismissCreateImageModelSwitch`. **Submissions** in image-generation mode under the new flow use the natively resolved model id and `imageGeneration` tool choice instead of legacy `mode`, with a fallback to the old behavior when no model resolves. **Analytics** count image-generation submits when `toolChoice` includes image generation, not only when `mode` is set.
> 
> Copy for the switch notice is centralized on `AIChatCreateImageModelSwitchNotice` (`localizedTitle` / `localizedSubtitle`), reused by the usage warning card and the NTP config payload. The **NewTabPage** package now depends on **AIChat** for those submission types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8104c2841fd9ad81751a2e544c9d48e39967205c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
+++ b/macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
@@ -266,6 +266,47 @@ final class NewTabPageOmnibarConfigProvider: NewTabPageOmnibarConfigProviding {
         featureFlagger.isFeatureOn(.aiChatNtpImageGeneration)
     }
 
+    var isUpdatedCreateImageEnabled: Bool {
+        isImageGenerationEnabled && featureFlagger.isFeatureOn(.updatedCreateImage)
+    }
+
+    @MainActor
+    func activateImageGeneration() -> NewTabPageDataModel.OmnibarCreateImageModelSwitch? {
+        guard isUpdatedCreateImageEnabled else { return nil }
+
+        let models = availableModelsProvider()
+        let previousModel = models.first(where: { $0.id == aiChatPreferencesPersistor.selectedModelId })
+        guard let previousModel,
+              !previousModel.supportsTool(.imageGeneration) else {
+            return nil
+        }
+
+        guard let imageModel = AIChatModel.preferredImageGenerationModel(in: models) else {
+            return nil
+        }
+
+        aiChatPreferencesPersistor.selectedModelId = imageModel.id
+        aiChatPreferencesPersistor.selectedModelShortName = imageModel.shortName
+        clearReasoningEffortIfUnsupported(by: imageModel)
+
+        let secondaryText = previousModel.provider == .oss
+            ? UserText.aiChatCreateImageModelSwitchPrivacySubtitle(previousModel.shortName)
+            : UserText.aiChatCreateImageModelSwitchSubtitle(previousModel.shortName)
+        return NewTabPageDataModel.OmnibarCreateImageModelSwitch(
+            message: UserText.aiChatCreateImageModelSwitchTitle(imageModel.shortName),
+            secondaryText: secondaryText
+        )
+    }
+
+    private func clearReasoningEffortIfUnsupported(by model: AIChatModel) {
+        guard let rawValue = aiChatPreferencesPersistor.selectedReasoningEffort,
+              let effort = AIChatReasoningEffort(rawValue: rawValue),
+              !model.isAccessible(effort) else {
+            return
+        }
+        aiChatPreferencesPersistor.selectedReasoningEffort = nil
+    }
+
     var isWebSearchEnabled: Bool {
         featureFlagger.isFeatureOn(.aiChatNtpWebSearch)
     }
--- a/macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
+++ b/macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
@@ -271,17 +271,19 @@ final class NewTabPageOmnibarConfigProvider: NewTabPageOmnibarConfigProviding {
     }
 
     @MainActor
+    var imageGenerationModelId: String? {
+        guard isUpdatedCreateImageEnabled else { return nil }
+        return imageGenerationModel(in: availableModelsProvider())?.id
+    }
+
+    @MainActor
     func activateImageGeneration() -> NewTabPageDataModel.OmnibarCreateImageModelSwitch? {
         guard isUpdatedCreateImageEnabled else { return nil }
 
         let models = availableModelsProvider()
         let previousModel = models.first(where: { $0.id == aiChatPreferencesPersistor.selectedModelId })
-        guard let previousModel,
-              !previousModel.supportsTool(.imageGeneration) else {
-            return nil
-        }
-
-        guard let imageModel = AIChatModel.preferredImageGenerationModel(in: models) else {
+        guard let imageModel = imageGenerationModel(in: models),
+              previousModel?.id != imageModel.id else {
             return nil
         }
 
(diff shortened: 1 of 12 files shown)


macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
```diff
@@ -266,6 +266,47 @@ final class NewTabPageOmnibarConfigProvider: NewTabPageOmnibarConfigProviding {
...
+
+        guard let imageModel = AIChatModel.preferredImageGenerationModel(in: models) else {
+            return nil
+        }
+
+        aiChatPreferencesPersistor.selectedModelId = imageModel.id
+        aiChatPreferencesPersistor.selectedModelShortName = imageModel.shortName
+        clearReasoningEffortIfUnsupported(by: imageModel)
+
+        let secondaryText = previousModel.provider == .oss
```
tomasstrba: **Found by Claude**

This duplicates copy logic that already exists twice over. `previousModel.provider == .oss` is exactly what `AIChatCreateImageModelSwitchNotice.init` derives into `previousModelHasExtraPrivacyProtections`, and these three `UserText` calls are the same ones `AIChatUsageWarningCardView.update(with:)` makes for the address bar.

Both types live in the app target, so this could build an `AIChatCreateImageModelSwitchNotice` from the two models and map it to `OmnibarCreateImageModelSwitch`. That keeps one definition of "has extra privacy protections" and one place where the copy is chosen, so the NTP and address-bar notices can't drift apart later.

tomasstrba (commented):
A couple of issues found by Claude. Three non-blocking notes below. Also Bugbot's comment above seems legit

macOS/LocalPackages/NewTabPage/Sources/NewTabPage/Omnibar/NewTabPageOmnibarConfigProviding.swift
```diff
@@ -104,3 +111,10 @@ public protocol NewTabPageOmnibarConfigProviding: AnyObject {
     var isSearchSuggestionDeletionEnabled: Bool { get }
     var isSearchSuggestionDeletionEnabledPublisher: AnyPublisher<Bool, Never> { get }
 }
+
+public extension NewTabPageOmnibarConfigProviding {
```
tomasstrba: **Found by Claude**

These defaults make the failure mode silent: a conformer that doesn't implement `isUpdatedCreateImageEnabled` / `activateImageGeneration()` gets the feature switched off rather than a compile error. That's a real risk for a native↔web contract — a rename or a new conformer loses the flow with nothing failing.

It's also why `MockNewTabPageOmnibarConfigProvider` needed no edit in this PR, and so why nothing forces coverage of the new paths. `NewTabPageOmnibarClientTests.swift` (1081 lines) and `NewTabPageOmnibarConfigProviderTests.swift` both already target exactly these two units, so `setImageGenerationActive`, `dismissCreateImageModelSwitch` and `activateImageGeneration` are cheap to cover — worth it on something marked medium risk.

macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
```diff
@@ -266,6 +266,47 @@ final class NewTabPageOmnibarConfigProvider: NewTabPageOmnibarConfigProviding {
...
+            ? UserText.aiChatCreateImageModelSwitchPrivacySubtitle(previousModel.shortName)
+            : UserText.aiChatCreateImageModelSwitchSubtitle(previousModel.shortName)
+        return NewTabPageDataModel.OmnibarCreateImageModelSwitch(
+            message: UserText.aiChatCreateImageModelSwitchTitle(imageModel.shortName),
+            secondaryText: secondaryText
+        )
+    }
+
+    private func clearReasoningEffortIfUnsupported(by model: AIChatModel) {
```
tomasstrba, line 311: **Found by Claude**

This diverges from the address bar's equivalent in two ways, and the two should agree.

`clearStaleReasoningEffortIfNeeded()` ([AIChatOmnibarController.swift#L598](https://github.com/duckduckgo/apple-browsers/blob/b36aa86752f2c25592f840ccc15df3da9ca4475c/macOS/DuckDuckGo/AIChat/AIChatOmnibarController.swift#L598)) checks `supportedReasoningEffort.contains(effort)`, whereas `isAccessible(effort)` is stricter — it returns false for a supported-but-tier-gated effort too, so this clears in cases the address bar keeps.

Also, when the stored raw value can't be decoded into `AIChatReasoningEffort`, the `let effort =` binding fails and this returns early, leaving the undecodable value in place. The address-bar version explicitly wipes it.

Stricter may well be the behaviour you want — but then the address bar should match, rather than the two surfaces disagreeing about when an effort is stale.

tomasstrba (changes requested):
It works great, I'd only recommend to polish the UI a little, but feel free to consult with design folks before:
1. The line where the warning begins should be straight:
Image

2. The warning doesn't have a stroke:

- This is how it's rendered in the omnibar
Image

3. Not a blocker: The warning collides with elements below:
Image

jotaemepereira:
> It works great, I'd only recommend to polish the UI a little, but feel free to consult with design folks before:
> 
> 1. The line where the warning begins should be straight:
> 
> Image
> 2. The warning doesn't have a stroke:
> 
> * This is how it's rendered in the omnibar
> 
> Image
> 3. Not a blocker: The warning collides with elements below:
> 
> Image

@tomasstrba Thanks for the review! The design request changes are part of the c-s-s parent task (which is the AI limits NTP changes), so it problably makes sense to request the changes in the main PR. I can do it.

macOS/DuckDuckGo/NewTabPage/NewTabPageOmnibarConfigProvider.swift
jotaemepereira: Resolved

jotaemepereira:
@tomasstrba I've fixed the UI issues. Please, can you re-review?

Screenshot 2026-09-08 at 1 08 06 PM

tomasstrba (approved):
LGTM! ✔️ The UI looks really polished now 🏅

PR #6706, opened 2026-09-04
Fix the Duck.ai usage card's collapsing title and block a spent allowance

Bunn:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201011656765697/task/1218185901618524?focus=true
Tech Design URL:
CC:

### Description
The usage-limit card's title sometimes rendered as a column of one or two characters a line, because the card's width follows the input card and its text stack left the title at whatever width it was last measured at — including the much narrower flanked AI-tab pose. It now fills the room it is given, so the wrap always matches the space actually available. Separately, a spent allowance now makes the whole input inert on iOS as it already does on macOS: submit and voice grey out, the tools/attach/reasoning/model controls disable, and the keyboard's own routes into a prompt close, leaving the card's own CTA as the only thing to act on.

### Testing Steps
1. Enable `utiDuckAIWarnings`, then seed a blocked state via Debug ▸ AI Chat ▸ Duck.ai Usage Warnings ▸ daily limit reached.
2. Open a Duck.ai tab and focus the input → the card reads "Daily limit reached" on one or two lines, never a vertical column of characters. Rotate and collapse/expand the input a few times → it stays readable.
3. With the card up: the send button is grey and does nothing, the voice button is grey and does nothing, the attach/tools/reasoning/model controls are disabled, and typing in the field does nothing. Tap "Start Using Weekly Limit" → the card retires and all of the above become live again.

### Impact
Low — behind `utiDuckAIWarnings`, internal only, and iOS-only.

#### What could go wrong?
A block could outlive the card that explains it and leave the user with a permanently dead input → it is reported off the same suppression the card is, lifted by `showCollapsed`/`hide`, and covered by tests for the suppression, pose-reset and CTA paths.

---

> [!NOTE]
> **Low Risk**
> iOS-only Duck.ai usage UI behind `utiDuckAIWarnings`; input blocking is tied to the same footer suppression and CTA paths covered by new tests.
> 
> **Overview**
> Fixes the Duck.ai usage footer so long titles use the full card width (including the narrow flanked AI-tab pose) instead of wrapping one character per line, via stack **fill** alignment and tighter button compression/hugging on the CTA pill.
> 
> When a usage warning **`blocksInput`** (spent allowance), **`UTIFooterController`** now reports **`onInputBlockChanged`** and the unified input goes inert like macOS: typing is rejected without dismissing the keyboard, submit/voice grey out, toolbar controls disable, and **`submitText`** / attachment-only paths are no-ops. The block clears with suppression, pose reset, or taking the weekly hand-off CTA. Expanded input chrome also shows the border when only the toolbar is visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee2049bf6f1f399b9dfc140385e6e6f20e6118ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/SwitchBarHandling.swift
+++ b/iOS/DuckDuckGo/AIChat/InputBox/SwitchBar/SwitchBarHandling.swift
@@ -56,6 +56,9 @@ protocol SwitchBarHandling: AnyObject {
     /// Suppresses the in-pill voice button — used when an external flank already provides one.
     var hidesVoiceButton: Bool { get set }
 
+    /// A spent Duck.ai allowance: the field takes no more text and no prompt can be sent.
+    var isInputBlockedByUsageLimit: Bool { get }
+
     var hasSubmittedPrompt: Bool { get set }
     var hasSubmittedPromptPublisher: AnyPublisher<Bool, Never> { get }
     var submitsAIChatOnKeyboardReturn: Bool { get }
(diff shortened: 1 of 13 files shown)


aataraxiaa (approved):
latest LGTM

PR #6709, opened 2026-09-05
[iOS] Coordinate the App Store rating prompt with the Promo Queue

hassaanelgarem:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218171233893204

### Description

Brings the iOS App Store rating prompt into the Promo Queue, so it can no longer land alongside or immediately after another promo, and so it starts the shared cooldown when it fires.

The rating prompt cannot be an ordinary `ModalPromptProvider`. The system rating dialog is not ours to present, and it triggers on search navigation rather than at the foreground checkpoint where providers are evaluated. So this adds `ModalPromptPresentationKind` to the provider protocol. A provider can now declare itself `.deferred`, meaning selection *holds* the promo slot instead of presenting anything. `ModalPromptCoordinationManager` gains a matching attempt state plus `redeemDeferredModal()` and `releaseDeferredModal()`. Existing providers default to `.modal` and do not change behaviour.

`AppRatingPromptCoordinator` is the first deferred provider. When it is eligible at foreground it takes the slot and holds it. The user's next DuckDuckGo search redeems it, which writes the shared appearance timestamp and requests the dialog. Backgrounding without searching releases the slot and starts no cooldown. iOS gives no signal about whether it drew the dialog, so redemption is best effort: request it, assume it appeared, apply the cooldown.

The prompt's eligibility rules are untouched. Still two requests per install, at 3 and 4 unique usage days. `AppRatingPrompt` now has a single owner instead of one instance per tab, which matters because its Core Data storage builds a context per instance. `TabViewController` reaches it through three new `TabDelegate` methods.

This sits behind a new `appRatingPromptCoordination` flag, off by default. `AppRatingPromptCoordinationPolicy` requires both that flag and the queue's own `promoPresentationCoordination`, so the prompt can never opt into a queue that is not coordinating. With either off, behaviour matches today exactly.

A remotely configurable `maxUnredeemedSlots` (default 3, zero disables it) stops a user who never searches from taking the slot every foreground and starving the rest of the queue.

Pixels are not included here. They follow in a separate PR.

### Testing Steps

**Setup**

1. Use a development build. The rating dialog never appears in TestFlight builds and always appears in development builds, so this cannot be verified on TestFlight.
2. To reach eligibility without waiting days, temporarily hardcode `AppRatingPrompt.uniqueAccessDays` to return `4`. That satisfies the threshold for both the first request (3) and the second (4).
3. Open Debug > Prompt Coordination. It shows Mode, Owner, the cooldowns, and Unredeemed Rating Slots.
4. The coordination mode is latched at launch, so force-quit and relaunch after any flag change.

**Coordination off, regression check**

5. Leave `appRatingPromptCoordination` off. Relaunch.
6. Run a DuckDuckGo search. The rating dialog appears, exactly as it does on main.

**Coordination on, happy path**

7. Enable `promoPresentationCoordination` and `appRatingPromptCoordination` in Debug > Feature Flags. Relaunch.
8. From Debug > Prompt Coordination, reset both cooldowns and the app rating prompt.
9. Foreground the app. Confirm no promo appears and Owner shows a modal owner. The rating prompt is holding the slot.
10. Run a DuckDuckGo search. Confirm the rating dialog appears, Owner returns to None, and Last Modal Appearance updates.
11. Foreground again within 24 hours. Confirm no other promo appears, because the rating prompt started the shared cooldown.

**Unredeemed slot release**

12. Reset the app rating prompt.
13. Foreground, then background without searching.
14. Confirm Owner is None, Last Modal Appearance did not change, and Unredeemed Rating Slots incremented by one.
15. Repeat until the count reaches 3. On the next foreground, confirm a different promo takes the slot instead.

### Impact and Risks

**Medium.** The new behaviour sits behind `appRatingPromptCoordination`, off by default and gated on the queue's own flag as well, so with either off nothing changes.

The risk is `ModalPromptCoordinationManager`, which every promo flows through and which gained a `.deferred` state and two entry points. `presentationKind` defaults to `.modal`, so no existing provider changes behaviour, and the existing manager suites pass alongside 20 new tests.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Touches `ModalPromptCoordinationManager`, which all promos use, but deferred behavior is opt-in per provider and the feature is flag-gated with legacy fallback.
> 
> **Overview**
> Routes the **App Store rating prompt** through the shared promo queue so it no longer stacks with other launch promos and applies the **shared modal cooldown** when it fires.
> 
> Because the system review dialog is triggered on **search**, not at foreground, the PR adds **`ModalPromptPresentationKind.deferred`** to `ModalPromptProvider` and extends **`ModalPromptCoordinationManager`** with a deferred attempt state plus **`redeemDeferredModal()`** / **`releaseDeferredModal()`**. **`AppRatingPromptCoordinator`** is the first deferred provider: eligible at foreground it **holds the lease**; the next DuckDuckGo search **redeems** it (cooldown + `SKStoreReviewController` path via new **`TabDelegate`** hooks and **`AppRatingPromptGating`** on **`PromoCoordinationService`**). Background without search **releases** the slot without cooldown. An **`maxUnredeemedSlots`** remote cap (default 3) limits queue starvation when users never search.
> 
> Gated by new **`appRatingPromptCoordination`** (off by default) together with **`promoPresentationCoordination`**; legacy behavior is unchanged when either is off. Debug **Prompt Coordination** UI exposes unredeemed slot count and a rating reset. Broad unit test coverage for deferred coordination, policy, and service wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8654b612ba51d9a47308f4761e5cfed62e88c85f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/AppServices/PromoCoordinationService.swift
```diff
@@ -166,6 +178,42 @@ extension PromoCoordinationService: PromoGating {
...
+        appRatingPromptCoordinator.registerUsage()
+    }
+
+    func shouldRequestAppRatingPrompt() -> Bool {
+        guard appRatingPromptCoordinator.isCoordinationEnabled else {
+            return appRatingPromptCoordinator.shouldRequestUncoordinated()
+        }
+
+        return modalPromptCoordinationManager.redeemDeferredModal()
+    }
```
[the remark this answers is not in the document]
hassaanelgarem: Addressed in 2f4eb322456e325c704ba1bf76ee0f7baa331a0e

iOS/DuckDuckGo/ModalPromptCoordination/AppRatingPrompt/AppRatingPromptCoordinator.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Never called: the manager short-circuits `.deferred` providers after eligibility.
+    func provideModalPrompt() -> ModalPromptConfiguration? {
+        nil
+    }
+
+    /// Deliberately empty. This fires only when a slot was taken *and* redeemed, never in the legacy path.
+    /// Eligibility is consumed at the request site, the one point common to every path.
+    func didPresentModal() {}
```
bkunat: nit: we can drop this empty override; ModalPromptProvider already supplies the same implementation.

iOS/DuckDuckGo/ModalPromptCoordination/AppRatingPrompt/AppRatingPromptCoordinator.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    // MARK: - ModalPromptProvider
+
+    var presentationKind: ModalPromptPresentationKind { .deferred }
+
+    func isEligibleToPresent(isOnboardingComplete: Bool) -> Bool {
+        guard isCoordinationEnabled else { return false }
+
+        guard !isUnredeemedSlotCapReached else {
```
bkunat, line 102: Could we let the rating prompt become eligible again when the user resumes searches?

With coordination enabled, we reserve the promo slot at foreground and wait for a search-results page before calling StoreKit. If the user only visits websites and then backgrounds the app, we release the reservation and increment `unredeemedSlotCount`

After three such sessions, this guard prevents every later reservation. The only normal reset is in `didRequestRating()`, which runs after we call StoreKit, not after the user submits a rating. But calling StoreKit requires a reservation, so that reset becomes unreachable once we hit the cap. Later searches and app restarts do not clear the counter.

If I understand correctly, under the same configuration, this appears to disable all remaining rating requests permanently. Could a later search clear the counter so the next eligible foreground can reserve a slot again? That would still let other promos proceed during sessions without searches.

iOS/DuckDuckGo/ModalPromptCoordination/AppRatingPrompt/AppRatingPromptCoordinator.swift
```diff
@@ -0,0 +1,158 @@
...
+            Logger.modalPrompt.debug("[App Rating Prompt] - Unredeemed slot cap reached; not taking the slot.")
+            return false
+        }
+
+        return appRatingPrompt.shouldPrompt()
```
bkunat: Could we account for the usage day that the next search will register when deciding whether to reserve the rating prompt’s slot?

The existing flow registers a page load before checking rating eligibility. That lets the first search on the third usage day trigger the first rating request.

Here, eligibility is checked at foreground, potentially before any page loads that day. Consider someone who opens the app once daily and starts from a new tab. On their third usage day, this check still sees two recorded days, so it does not reserve a slot. Their first search then increments the count to three, but there is no reservation to use. Further searches cannot reserve a slot either.

That moves the earliest request to day four, even with no competing promo or cooldown. 

Could we preserve the existing timing and add a test that starts with two recorded usage days, enters the foreground, and then loads a search? The same boundary issue applies to the second request, which needs four more usage days.

iOS/DuckDuckGo/TabDelegate.swift
```diff
@@ -162,6 +154,17 @@ protocol TabDelegate: AnyObject {
 
     /// User activated an in-page link in this tab.
     func tabDidEngageWithPage(_ tab: TabViewController)
+
+    /// A page started loading. Advances the App Store rating prompt's usage-day counter, which is
+    /// what makes the prompt become due. Call for every page, not only searches.
+    func tabDidLoadPageForAppRatingPrompt(_ tab: TabViewController)
+
+    /// Whether to request the App Store rating dialog now. Only ask on a search page, the prompt's
```
bkunat, line 170: nit: could we combine the shouldRequest and didRequest callbacks into one requestIfNeeded(in:) operation? That would keep redemption, the StoreKit request, and eligibility consumption together, and remove some forwarding methods.

currently those two callbacks always surround the same immediate StoreKit call.

bkunat (approved):
The implementation looks great!

I left a few comments, but only the one about the rating prompt becoming eligible again when the user resumes searches requires discussion. 

Other than that :shipit:

iOS/DuckDuckGo/ModalPromptCoordination/AppRatingPrompt/AppRatingPromptCoordinator.swift
hassaanelgarem: I like this idea! I moved the discussion over to Asana 👍

iOS/DuckDuckGo/TabDelegate.swift
hassaanelgarem: I agree that this would be more concise, but it will also make the code harder to test IMO.

iOS/DuckDuckGo/ModalPromptCoordination/AppRatingPrompt/AppRatingPromptCoordinator.swift
hassaanelgarem: Applied in https://github.com/duckduckgo/apple-browsers/pull/6709/commits/f52bdd0dbd645a17203f1cdd9a847f18972144ad 👍

iOS/DuckDuckGo/TabDelegate.swift
bkunat: I only added it as a suggestion. If you think that testing will be problematic if we remove it, then I'm all for keeping it 👍

PR #6671, opened 2026-09-03
[macOS] New success screen with recovery code for simplified Sync setup 

amddg44:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1217193469998965?focus=true
Tech Design URL:
CC:

### Description
Adds the simplified Sync setup success screen on macOS. The screen displays the device name and recovery code, with actions to copy the code or download it as a PDF.

The screen appears when the Mac first enables or joins Sync. When an already-synced Mac adds another device, pairing completes without showing a redundant success screen. Legacy behaviour remains unchanged when the feature flag is disabled.

### Testing Steps

1. Enable the `simplifiedSyncSetupV2` feature flag and ensure Sync is disabled.
2. Enable Sync and select **Sync This Device Only** → the V2 success screen shows the Mac’s name and recovery code.
3. Select **Copy Code** → the complete recovery code is copied to the clipboard.
4. Select **Download as PDF** → a recovery-code PDF can be saved and contains the correct code.
5. Select **Done** → the dialog closes and Sync remains enabled.
6. With Sync disabled on the Mac and enabled on another device, pair the devices → the Mac shows the V2 success screen after joining.
7. Pair two devices with Sync initially disabled, with the Mac acting as the new host (paste code into macOS)  → the success screen appears only after pairing completes.
8. With Sync already enabled on the Mac, add another device → pairing completes without showing the success screen on the Mac.
9. Disable `simplifiedSyncSetupV2` and repeat setup → the legacy success flow remains unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes sync setup completion and when recovery codes are shown, but only behind a feature flag with legacy paths preserved; incorrect pairing-state handling could skip or duplicate success UI.
> 
> **Overview**
> When **`simplifiedSyncSetupV2`** is on, macOS shows a new **`SyncSuccessViewV2`** (device name, recovery code, copy/PDF, Done) instead of the legacy save-PDF / “now syncing” steps. **`ManagementDialog`** still uses the `.saveRecoveryCode` case but branches to the V2 view; a **`QR-Download-Hero`** asset and V2 copy strings support the layout.
> 
[...]

PR #6725, opened 2026-09-07
[macOS] Wait for the joining device before ending V2 pairing on the host

pballart:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214200115953388/task/1218231939766475?focus=true

### Description
In the simplified Sync setup (V2), both devices show a "Sync Now" confirmation. If you confirmed on a Mac that already had Sync turned on, the Mac closed the setup dialog immediately instead of waiting for the other device to confirm, and cancelled its side of the connection. The Mac now stays on the connecting screen until the other device has joined. This regressed in #6671.

It also update the fastlane-plugin-ddg_apple_automation dependency so that I can make a macOS build and get the proper link in MM.

### Testing Steps
1. Enable the `simplifiedSyncSetupV2` feature flag on a Mac and an iOS device. Turn Sync on on the Mac, and leave Sync off on the iOS device.
2. On the Mac, go to Settings → Sync & Backup → Sync With Another Device, and scan the code with the iOS device → both devices show a "Sync Now" confirmation.
3. Tap **Sync Now** on the Mac first, and leave the iOS confirmation untouched for a few seconds → the Mac stays on the connecting screen (before this change it closed immediately).
4. Tap **Sync Now** on the iOS device → the Mac closes the dialog and the iOS device appears in the Mac's synced devices list.
5. Regression check: repeat with Sync turned **off** on the Mac and the Mac acting as the new host → the Mac still shows the recovery-code success screen once pairing completes.

### Impact
Low

#### What could go wrong?
The Mac now waits on the synced-devices list refresh (a 3s timer that only ticks while the Sync settings pane is visible) before closing, so the dialog could linger if that refresh doesn't fire → the pairing session still has its own 5 minute timeout, and this is the same mechanism the pre-#6671 behaviour relied on.

---
###### Internal references:

---

> [!NOTE]
> **Low Risk**
> Scoped macOS sync setup UX fix; dialog may linger if device list refresh does not fire, with existing pairing timeout as fallback.
> 
> **Overview**
> Fixes a regression where a Mac that **already had Sync on** closed the simplified V2 pairing dialog as soon as the host tapped “Sync Now,” which cancelled the connection before the other device could confirm.
> 
> **`completeV2HostFlow`** now always honors `shouldWaitForDevicesToChange`: it waits for the synced-devices list to update, then either shows the recovery-code success UI (new host) or **`endFlow()`** (existing host). The wait helper is generalized from “then show success” to **`waitForDevicesToChange(then:)`** so the same device-change signal drives both paths; the legacy (non-V2) finish path uses it to call **`showSyncSuccess()`** unchanged.
> 
> Unit tests split the existing-host V2 case into “wait for device change then dismiss” vs “no wait, dismiss immediately.” iOS/macOS **fastlane-plugin-ddg_apple_automation** is bumped **4.2.1 → 4.2.2** in Pluginfiles and lockfiles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 415cc2261f356a1d3162916519de9b899892331a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/Gemfile.lock
+++ b/iOS/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/duckduckgo/fastlane-plugin-ddg_apple_automation
-  revision: c7e9bd112cb2771fe5c2bc2fb0a19f8cd2d7d5e9
-  tag: 4.2.1
+  revision: 1fa03a406801995d6e7c5068935d88184df1eda7
+  tag: 4.2.2
   specs:
-    fastlane-plugin-ddg_apple_automation (4.2.1)
+    fastlane-plugin-ddg_apple_automation (4.2.2)
       asana
       climate_control
       httpparty
--- a/iOS/fastlane/Pluginfile
+++ b/iOS/fastlane/Pluginfile
@@ -2,4 +2,4 @@
 #
 # Ensure this file is checked in to source control!
 
-gem 'fastlane-plugin-ddg_apple_automation', git: 'https://github.com/duckduckgo/fastlane-plugin-ddg_apple_automation', tag: '4.2.1'
+gem 'fastlane-plugin-ddg_apple_automation', git: 'https://github.com/duckduckgo/fastlane-plugin-ddg_apple_automation', tag: '4.2.2'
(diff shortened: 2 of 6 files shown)


amddg44 (approved):
Nice catch!

PR #6728, opened 2026-09-07
iOS bug: Only discard a restored contextual chat when it was really deleted

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204186595873227/task/1218232234241878?focus=true

### Description

Reopening the Duck.ai contextual sheet could silently throw away the conversation, leaving the empty state on a tab that still had a chat.
Once someone has five or more pinned chats there is no room for any recent one, so an active chat could never be found and every reopen discarded it.
Deletion is now established by asking the store for that specific chat rather than inferring it from a list, and only for a chat we know reached the store: the storage bridge reports each chat it writes, so a conversation the frontend has named but not yet saved is no longer mistaken for a deleted one. A store that is missing, unmigrated, still opening or unreadable declines to answer instead of guessing.

### Testing Steps

1. Pin five or more chats in Duck.ai,
2. Open a web page → open the Duck.ai contextual sheet → send a prompt and let it answer
3. Close the sheet → reopen it on the same page → the conversation is still there 
4. Open "View all chats" → delete that conversation → return to the tab and open the sheet → a new chat, not the deleted one
5. Close the sheet check the glyph on the aiChat bauble disappears and clicking on it shows the menu
6. Start a fresh chat on a page, force-quit and relaunch, then open the sheet on that tab → the conversation is restored
7. Start a chat, delete it from "View all chats", force-quit and relaunch, then open the sheet on that tab → a new chat 

### Impact

High — the failure this fixes loses a user's conversation with no warning and no way back.

#### What could go wrong?

- The check wrongly reports a chat as deleted and we are back to losing chats → only a successful keyed read returning nothing counts as deletion, and only for a chat the storage bridge has confirmed writing. Absent, unmigrated, still-opening and unreadable stores all answer "not deleted", each covered by a test.
- A deleted chat is restored anyway → the saved URL is vetted before the sheet opens, and a live session is reconciled on reopen. Testing steps 5 and 7.
- The lookup stalls the UI → it runs on a background queue, and nothing on the address bar's path reads the store at all.

### Quality Considerations

- The address bar is driven by session state alone, exactly as before this change. The store is consulted only when a surface is about to open, and off the main thread — `refreshOmniBar` never reads it.
- Absence counts as deletion only for a chat the storage bridge has reported writing, or one restored from an earlier launch. Treating an unwritten chat's absence as deletion is what destroyed conversations in the first place, whether the evidence came from a capped suggestions list or a keyed read.
- Write confirmation arrives only while a contextual sheet's web view is alive, so a chat deleted from a different tab's history is caught when that tab is restored rather than immediately.
- Reconciling is skipped while the sheet is on screen. `presentSheet` also serves "attach this selection to the sheet already open", where there is no reopen to guard and clearing the chat would strip a sheet the user is looking at.
- The recent-chats list is still wrong for its own purpose — five pinned chats hide every recent chat from the menu. Nothing infers deletion from it any more, so it is left to a separate fix.
- Fourteen coordinator tests, including the failed-read case that caused this bug and the not-yet-written case.

---

> [!NOTE]
> **Medium Risk**
> Touches contextual session restore and tab URL persistence with async native storage reads; behavior is conservative on errors but wrong keyed reads could still clear valid chats.
> 
> **Overview**
> Fixes contextual Duck.ai reopening **silently dropping** conversations when the active chat was not visible in the capped recent-chats suggestions (e.g. five+ pinned chats).
> 
> **Deletion detection** now uses a keyed `getChat` on native storage (off the main thread, gated by `aiChatNativeDataAccess` and store readiness). `presentSheet` **vets** the tab’s restore URL and **clears** a background active session only when that lookup says the chat is gone; unmigrated, unreadable, or not-yet-ready stores **do not** treat absence as deletion.
> 
> A chat counts as “really on disk” only after the **storage bridge** reports a write (`didPersistChatWithID`) or after a successful restore—so a frontend-assigned ID that has **not** been saved yet is not discarded. Stale tab URLs are cleared via `didUpdateContextualChatURL: nil` when restore is rejected; reconciliation is **skipped** while the sheet is already on screen.
> 
> The sheet no longer runs **`prefetchRecentChatsVisibility`** / `didDetectActiveChatRemoved`; persistence is signaled from **`AIChatContextualWebViewController`** via `chatUpdatesPublisher`. Coordinator tests cover restore, deletion, not-yet-written chats, and store failure paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af43e224487aaa315e269f11b29725b7e3f00168. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iOS/DuckDuckGo/TabViewControllerContextualAIChatExtension.swift
```diff
@@ -65,7 +65,9 @@ extension TabViewController {
     /// A conversation this tab can return to: either still live in the coordinator, or persisted by an
     /// earlier launch and waiting to be restored.
     var hasContextualChatToReopen: Bool {
-        aiChatContextualSheetCoordinator.sessionState.hasActiveChat || tabModel.contextualChatURL != nil
+        aiChatContextualSheetCoordinator.hasChatToReopen(
```
aataraxiaa: This introduces a bug - start a chat in the sheet, View all chats, delete it, come back to the tab, tap the Duck.ai button. Nothing happens. Added a breakpoint on `MainViewController.swift:5908` to confirm it. There it returns `.menu` only when `hasChatToReopen == false`, so  after the delete the entry computes `.menu`, but nothing re-ran the refresh.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -149,6 +149,53 @@ final class AIChatContextualSheetCoordinator {
...
+    }
+
+    /// Whether a chat has been deleted, or `false` when that cannot be established.
+    private func isChatDeleted(chatID: String?) -> Bool {
+        guard let chatID,
+              isNativeDataAccessEnabled,
+              let storage = chatStorage,
+              (try? storage.isMigrationDone()) == true else { return false }
+        do {
+            return try storage.getChat(chatId: chatID) == nil
```
aataraxiaa: Think this goes through `DuckAiNativeDataStore.dbQueue()`, whose doc says "Blocks the caller until background setup completes". This is driven by `refreshOmniBar()`,  so every navigation, tab switch, would block.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetViewController.swift
```diff
@@ -72,7 +72,6 @@ protocol AIChatContextualSheetViewControllerDelegate: AnyObject {
     /// Called when the user taps the "New Chat" button to start a fresh conversation
     func aiChatContextualSheetViewControllerDidRequestNewChat(_ viewController: AIChatContextualSheetViewController)
     /// Housekeeping, not a user request: reset in place, never move the user to another surface.
```
aataraxiaa: Minor - can remove this comment as method is gone

aataraxiaa (changes requested):
Testing steps don't pass - can't open the sheet after deleting a chat.

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
```diff
@@ -149,6 +149,68 @@ final class AIChatContextualSheetCoordinator {
...
+              sessionState.hasActiveChat,
+              let chatID = sessionState.contextualChatURL?.duckAIChatID,
+              persistedChatIDs.contains(chatID) else { return }
+        guard await isChatDeleted(chatID: chatID) else { return }
+        Logger.aiChat.debug("[Contextual] Active chat was deleted, clearing it")
+        persistedChatIDs.remove(chatID)
+        clearActiveChat()
+    }
+
+    private func vettedRestoreURL(_ restoreURL: URL?) async -> URL? {
```
aataraxiaa, line 182: agent flagged:

```
Severity: High
Category: correctness / state
File:Line: AIChatContextualSheetCoordinator.swift:182-189
Description: vettedRestoreURL inserts into persistedChatIDs whenever isChatDeleted returns false — but that's false for five "couldn't tell" cases (flag off, nil storage, setupSucceeded != true, migration not
done, read threw) as well as for "confirmed present". A never-written chat gets recorded as persisted, defeating the gate commit 95a5340b50 added.
Fix: Tri-state lookup (present/absent/unknown); insert only on .present.
```

aataraxiaa (approved):
Tested and looks good. One comment that agent-flagged. See what you think

iOS/DuckDuckGo/AIChat/ContextualMode/AIChatContextualSheetCoordinator.swift
SabrinaTardio: The insert on its own does nothing, we still only clear when isChatDeleted is true, which means the chat is really gone from the store. A chat with real content is never gone from the store, so it's never cleared. The only thing that can get cleared is a chat that has no record at all, and that's a chat that would've loaded empty anyway, so resetting would still be right.

PR #6731, opened 2026-09-07
Fix contrast regression on Duck.ai fire tabs from #6729

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218141258886143
Tech Design URL:
CC:

Follow-up to #6729 (merged) — fixes two review findings on that PR's `fireModeContentSubviews` change.

### Description
**Medium:** Including `aiTabCollapsedFireButton`/`aiTabCollapsedMenuButton` in `fireModeContentSubviews` forced a fire-mode dark trait onto buttons that keep their own regular glass and were originally excluded so they'd follow the OS trait. On Duck.ai fire tabs in light mode, that left their icons resolving light against a light `.glass()`, dropping contrast — the opposite problem from the one #6729 fixed. Restored the exclusion.

**Low:** The comment left behind ("no longer excluded") was PR-situational changelog phrasing rather than an ongoing why. Replaced with one that states the actual invariant.

### Testing Steps
1. Open a Duck.ai fire tab in light mode.
2. Confirm the fire/menu accessory buttons are visible against their own glass (not washed out).
3. Re-check the toolbar contrast fix from #6729 still holds (bookmarks/passwords/fire/tabs/menu icons visible on the dark bar).

### Impact
Low: visual contrast fix, no functional/behavioral change.

---

> [!NOTE]
> **Low Risk**
> Visual contrast-only tweak to which subviews inherit fire-mode trait overrides; no logic or API changes.
> 
> **Overview**
> Reverts part of #6729 by **excluding** `aiTabCollapsedFireButton` and `aiTabCollapsedMenuButton` from `fireModeContentSubviews` again. Those views were incorrectly receiving fire mode’s forced `.dark` `overrideUserInterfaceStyle`, which made their icons render for a dark surface while they still use regular `.glass()` in light mode—washed-out fire/menu controls on Duck.ai fire tabs in light mode.
> 
> The inline comment now documents the invariant: collapsed accessories keep their own glass and must follow the **OS** trait; only the other direct subviews (toggle, text entry, toolbar, etc.) still get the dark override for contrast on the fire card.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a755c98432fcd82f7c8d4b618f063257f65ee14c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6729, opened 2026-09-07
Fix Fire Tab toolbar icon contrast in floating UI

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218141258886143
Tech Design URL:
CC:

(Recreated from #6726, which GitHub auto-closed when the source branch was renamed to match our `karlk/` convention.)

### Description
In Fire Tabs, the floating bottom bar's glass background is forced dark for fire mode, but the toolbar buttons (bookmarks, passwords, fire, tabs, menu) live in a sibling view that never inherited that dark trait. Their icons kept resolving against the light OS trait while sitting on a now-dark glass, so they were nearly invisible.

Also fixed two related instances of the same pattern (a container needing the fire-mode dark trait but not being reached by the existing per-subview override) in `UnifiedToggleInputView` and `DefaultOmniBarView`, and a missing `super.traitCollectionDidChange` call in `BrowserChromeButton` that was silently preventing icon tints from ever re-resolving on a trait change — without it, none of the above overrides would have any visible effect.

### Testing Steps
1. Open a new Fire Tab (long-press the tab count → New Fire Tab).
2. Look at the floating bottom bar → bookmarks/passwords/fire/tabs/menu icons are clearly visible (white) against the dark bar.
3. Switch to a regular (non-fire) tab → toolbar icons still render normally, no regression.

### Impact
Low: visual contrast fix, no functional/behavioral change.

### Quality Considerations
Verified live in the iOS Simulator (iPhone 17 Pro Max, iOS 26.5) across multiple Fire Tabs, and confirmed no change to normal-tab appearance.

---

> [!NOTE]
> **Low Risk**
> Visual contrast and trait-propagation fixes in browser chrome only; no auth, data, or navigation behavior changes.
> 
> **Overview**
> Fixes **nearly invisible toolbar icons** on Fire Tabs when the floating bottom bar forces a dark glass background but sibling chrome views still resolved glyphs against the light system trait.
> 
> **`BrowserToolbarView`** now applies the same `overrideUserInterfaceStyle` to **`chromeContentHost`** (button row sibling of the glass) during material refresh. **`DefaultOmniBarView`** extends fire-mode dark trait overrides to **`leadingButtonsContainer`** and **`trailingButtonsContainer`**, which sit outside the search-area subview loop. **`UnifiedToggleInputView`** stops excluding collapsed AI-tab fire/menu buttons from fire-mode styling so they match the dark glass too.
> 
[...]

PR #6554, opened 2026-08-27
[iOS] Don't use website theme color for edge gradients

brindy:
Task/Issue URL: https://app.asana.com/1/137249556945/project/392891325557410/task/1217636674570180?focus=true
Tech Design URL:
CC:

### Description
Don't apply website theme color to webkit's under page background color in order to disable the coloured edges.

### Testing Steps
1. Floating UI flag on (restart required/expected)
2. Check the fixed elements pages on privacy-test-pages.site > view port test
3. Browse a few pages 
4. Floating UI flag off (restart required/expected) - validate no leakage

### Impact
Low: Minor visual changes, small bug fixes, improvement to existing features

#### What could go wrong?
Could break colour on some other UI elements that not expected -> Limited to floatingUI internal users.

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

---

> [!NOTE]
> **Low Risk**
> Visual-only change behind the floating UI flag; no auth, data, or security impact. Residual risk is minor color mismatches on other web view chrome for internal floating UI users.
> 
> **Overview**
> When **floating UI** is enabled, site theme colors no longer drive the web view’s `underPageBackgroundColor` and scroll view background. Those layers now use the design-system **surface canvas** color instead of the adjusted site theme color, which stops WebKit from showing colored edge gradients while status bar and pull-to-refresh styling still follow the site theme.
> 
> `SiteThemeColorManager` takes a new `isFloatingUIEnabled` flag; `MainViewController` passes its existing floating UI state when constructing the manager. With floating UI off, behavior is unchanged (theme color applies everywhere as before).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e81bc03b663d7fcff22acd9c53aad43932df7eaf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6732, opened 2026-09-07
Restore dark glass response on dark websites in light mode

kemiljk:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218233582452852
Tech Design URL:
CC:

Stacked on #6731. Follows #6729, which made the toolbar icons track the glass's dark trait — this fixes why the glass never went dark on a dark *website* in the first place.

### Description
On a dark website with the app in light mode (most visible with a top address bar), the bottom toolbar glass stayed light-trait, so its icons had no contrast against the dark page showing through it.

The floating glass decides light/dark from the page's background colour, which WebKit derives automatically from the page. #6554 started writing a fixed canvas colour over that value (to keep websites' theme-colour off the bar edges) — a side effect of writing it is that WebKit stops deriving it, so the glass could no longer see a dark page at all. This resets the value instead, which restores WebKit's derivation while still never applying the site theme-colour, so #6554's fix holds. It also observes that value per tab so the glass re-evaluates on navigation; before, it only did so after a tab-switcher transition or a system appearance change.

### Testing Steps
1. Floating UI on, app in light mode, address bar at the top.
2. Open a dark site with no theme-colour meta tag (e.g. musio.com) → bottom toolbar glass and icons go dark/light-on-dark and stay legible; the top address bar adapts too.
3. Navigate to a light site (e.g. apple.com) in the same tab → glass and icons return to light.
4. Repeat with the address bar at the bottom → same result.
5. Open a Fire Tab → still dark as before (#6729 unaffected).
6. Pull down to overscroll on the dark site → the bounce area is the page's own colour, not a brand theme-colour.

### Impact
Low: visual contrast fix, no functional change.

#### What could go wrong?
- Overscroll bounce area under floating UI is now the page-derived colour instead of the canvas colour → intentional; it is the page's actual background, not the site theme-colour #6554 removed, and it is what makes the glass response coherent.
- WebKit may update the derived colour a few times while a page paints, each re-running the glass refresh → cosmetic; an "unchanged → skip" guard was deliberately not added because the tab-switcher path relies on forcing a re-snapshot.

### Quality Considerations
Verified live on iPhone 17 Pro Max (iOS 26.5) for steps 1–3. `FloatingGlassAppearancePolicyTests` pass. `BrowserToolbarViewTests` has two inset assertions (`testWhenEmbeddedFloatingThenOuterInsetIsEqualOnEveryEdge`, `testWhenBottomOmnibarDetachesForFocusThenOuterInsetsStayUnchanged`) that fail identically on the base branch without these changes — pre-existing and geometry-related, not introduced here.

---

> [!NOTE]
> **Low Risk**
> Visual/contrast-only changes around WebKit background colour and glass refresh; no auth, data, or core navigation logic touched.
> 
> **Overview**
> Fixes floating toolbar/omnibar glass staying **light-trait** on dark websites when the app is in light mode, so icons lacked contrast against the page showing through the bar.
> 
> **`SiteThemeColorManager`** no longer forces `underPageBackgroundColor` to the canvas colour when floating UI is on—it sets it to **`nil`** so WebKit keeps a **page-derived** background (what `FloatingGlassAppearancePolicy` reads), while the scroll view still uses the canvas colour so site theme colours don’t bleed at the edges (#6554 behaviour preserved).
> 
> **`MainViewController`** **KVO-observes** `underPageBackgroundColor` on the active tab and calls **`refreshSettledFloatingGlassAppearance()`** when it changes, so glass re-evaluates on navigation—not only on tab switch or trait changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59d6c4d6cf368177651521c7f41e7e64cfb1d23c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/iOS/DuckDuckGo/MainViewController.swift
+++ b/iOS/DuckDuckGo/MainViewController.swift
@@ -441,6 +441,15 @@ class MainViewController: UIViewController {
                               isFloatingUIEnabled: isFloatingUIEnabled)
     }()
 
+    // Re-run the glass policy when WebKit's page-derived color changes; otherwise it only ran on tab-switch/trait changes.
+    private var pageBackgroundColorObservation: NSKeyValueObservation?
+
+    private func observePageBackgroundColor(for tab: TabViewController) {
+        pageBackgroundColorObservation = tab.webView.observe(\.underPageBackgroundColor, options: [.initial, .new]) { [weak self] _, _ in
+            self?.refreshSettledFloatingGlassAppearance()
+        }
+    }
+
     private lazy var browsingMenuSheetCapability = BrowsingMenuSheetCapability.create()
 
     let themeManager: ThemeManaging
@@ -2843,6 +2852,7 @@ class MainViewController: UIViewController {
         chromeManager.attach(to: tab.webView.scrollView)
         chromeManager.reset(animated: false)
         themeColorManager.attach(to: tab)
+        observePageBackgroundColor(for: tab)
         tab.chromeDelegate = self
         tab.updateWebViewBottomAnchor(for: currentBarsVisibility)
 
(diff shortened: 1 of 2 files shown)


brindy (approved):
LGTM

PR #6733, opened 2026-09-07
Add on-site geolocation bridge

bkunat:
Task/Issue URL: https://app.asana.com/1/137249556945/project/715106103902962/task/1217863452475662?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1216242282222619

### Description

Adds the JavaScript-to-native bridge for location requests, location watches, and permission queries, with shared script-loading fixes.

### Testing Steps

This PR adds the geolocation bridge without connecting it to browser tabs. These checks cover regressions in existing camera and microphone permissions.

Use the [Permissions API test page](https://privacy-test-pages.site/features/permissions-api.html). Enable `sitePermissions` ff, clear saved decisions for the test host, and set global camera/microphone defaults to **Ask Each Time**. Allow DuckDuckGo camera and microphone access in iOS Settings.

1. Tap **Request camera** and choose **Allow Once** → capture succeeds. Tap **Stop media streams** → capture stops and the camera indicator clears.
2. Tap **Request microphone** and choose **Never Allow** → the request is denied. Repeat the request → it remains denied without another site prompt.
3. Reload the page and request microphone access again → the saved denial still applies.
4. Disable `sitePermissions` and open the [test page on a fresh host](https://good.third-party.site/features/permissions-api.html) → camera and microphone requests use the existing WebKit permission flow. Site Permissions is absent from the browser menu and app Settings.

### Impact

Medium: shared script loading and permission-bridge behavior can affect website functionality and privacy.

### Quality Considerations

- Existing camera and microphone behavior should remain unchanged with the feature flag enabled or disabled.
- This PR should not introduce geolocation prompts or location access in browser tabs.
- Automated tests cover geolocation request validation, callback cleanup after navigation, and shared script-loading behavior.

---
###### Internal references:


iOS/DuckDuckGo/TabViewController+SitePermissions.swift
```diff
@@ -627,20 +728,273 @@ extension TabViewController {
...
+        // RFC 8941 parameters do not change the allowlist. Validate them before discarding them.
+        let parameterValue = [
+            #""(?:[\x20-\x21\x23-\x5B\x5D-\x7E]|\\["\\])*""#,
+            #"[A-Za-z*][A-Za-z0-9!#$%&'*+.^_`|~:/-]*"#,
+            #"-?(?:[0-9]{1,15}|[0-9]{1,12}\.[0-9]{1,3})"#,
+            #":(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}={0,2}|[A-Za-z0-9+/]{3}=?)?:"#,
+            #"\?[01]"#
+        ].joined(separator: "|")
+        let parameter = #"^ *[a-z*][a-z0-9_.*-]*(?:=(?:"# + parameterValue + #"))?$"#
+        guard components.dropFirst().allSatisfy({ $0.range(of: parameter, options: .regularExpression) != nil }) else { return false }
```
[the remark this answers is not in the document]
bkunat: RFC 8941 permits spaces after ;, not before it: parameters = *( ";" *SP parameter ). The parser at 81e77313fb, TabViewController+SitePermissions.swift:929, already accepts post-semicolon spaces through ^ *. Forms such as * ;report-to=… are malformed and follow the existing malformed-policy rejection behavior. Trimming the allowlist would change that behavior rather than fix valid parameter parsing. [RFC 8941 §3.1.2](https://www.rfc-editor.org/rfc/rfc8941.html#section-3.1.2)

iOS/DuckDuckGo/TabViewController+SitePermissions.swift
```diff
@@ -627,20 +728,273 @@ extension TabViewController {
...
+        return host == "localhost"
+            || host.hasSuffix(".localhost")
+            || host == "::1"
+            || host == "[::1]"
+            || isIPv4Loopback
+    }
+
+    /// Cross-origin delegation is intentionally unsupported in v1 because the shim cannot reliably
+    /// evaluate subframe response headers and `allow="geolocation"`. Revisit only with breakage evidence
+    /// or an availability-gated OS-managed API; until then, native attribution denies every cross-origin frame.
```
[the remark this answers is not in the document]
bkunat: fcfdc8300e

SharedPackages/BrowserServicesKit/Sources/UserScript/JSFileCache.swift
```diff
@@ -37,7 +37,7 @@ enum JSFileCache {
         }
 
         do {
-            let content = try String(contentsOfFile: path)
+            let content = try String(contentsOfFile: path, encoding: .utf8)
```
bkunat, line 41: This was flagged by Cursor bot. 

Our bundled JavaScript files are UTF-8, but the previous initializer tried to detect their encoding. Stale file metadata could make it choose incorrectly. 

Specifying UTF-8 avoids that, and there’s a regression test for it. Any future script using another encoding would need conversion first.

SharedPackages/BrowserServicesKit/Sources/UserScript/UserScript.swift
```diff
@@ -131,6 +131,13 @@ extension StaticUserScript {
 
 }
 
-public enum UserScriptError: Error {
+public enum UserScriptError: LocalizedError {
     case failedToLoadJS(jsFile: String, error: Error)
+
+    public var errorDescription: String? {
```
bkunat, line 137: When a script fails to load, the existing code logs error.localizedDescription. Without this change, that can be a generic error that doesn’t tell us which script failed or why. This adds the filename and underlying error to help with debugging; failure handling stays the same.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/Geolocation/GeolocationUserScript.swift
```diff
@@ -0,0 +1,682 @@
...
+    @MainActor private var watchRegistry = GeolocationWatchRegistry()
+    @MainActor private var permissionStatusRegistry = GeolocationWatchRegistry()
+    @MainActor private var frameRegistrations = GeolocationFrameRegistrationStore()
+
+    @MainActor
+    /// - Parameter installImmediately: Hardens the page API synchronously at document start. Use only when the
+    ///   embedding app scopes this script to its tab web views; exact `duck.ai` origins remain untouched.
+    public init(delegate: GeolocationUserScriptDelegate? = nil, installImmediately: Bool = false) {
+        self.delegate = delegate
+        self.installImmediately = installImmediately
```
bkunat, line 227: This option replaces the geolocation APIs before website code can save the originals. Once installed, those replacements stay for the lifetime of the document. When we connect this to tabs, the feature flag needs to control whether the script is installed, not just whether individual requests are accepted.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/jsSources/geolocation.js
```diff
@@ -0,0 +1,702 @@
...
+    const signHMAC = subtleCrypto?.sign.bind(subtleCrypto);
+    const NativeUint8Array = globalThis.Uint8Array;
+    const hmacKey = importHMACKey && encodeText
+        ? callThen(importHMACKey("raw", encodeText(capability), { name: "HMAC", hash: "SHA-256" }, false, ["sign"]), undefined, () => null)
+        : Promise.resolve(null);
+    const numberToString = globalThis.Number.prototype.toString;
+    const noncePattern = /^[0-9a-f]{32}$/;
+    const testPattern = globalThis.RegExp.prototype.test;
+
+    // Consume security values inside captured callbacks. Even await can call a
```
bkunat, line 112: Websites can replace JavaScript methods that our script also uses. We save the originals before the page runs and use them for these checks. The callback style is intentional: ordinary await can end up calling the website’s replacements, letting page code interfere with the sandbox result or access the signing key.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/jsSources/geolocation.js
```diff
@@ -0,0 +1,702 @@
...
+        } catch (_) {
+            return false;
+        }
+    })();
+    const hasNetworkOrigin = globalThis.origin !== "null" &&
+        globalThis.location.hostname.length > 0 &&
+        (globalThis.location.protocol === "https:" || globalThis.location.protocol === "http:");
+    const nativePermissionsPolicy = document.permissionsPolicy ?? document.featurePolicy;
+    const nativePolicyAllowsFeature = nativePermissionsPolicy?.allowsFeature;
+    const policyAllowsGeolocation = () => {
```
bkunat, line 154: An iframe can be blocked from using location even when it comes from the same origin as the main page. Some WebKit versions don’t expose enough information to check that. In those cases, we deny iframe requests; the main page can still continue through Swift’s permission checks.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/Geolocation/GeolocationUserScript.swift
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    private static func javaScriptObject(from payload: [String: Any]) -> String? {
+        guard JSONSerialization.isValidJSONObject(payload),
+              let data = try? JSONSerialization.data(withJSONObject: payload),
+              let value = String(data: data, encoding: .utf8) else { return nil }
+        return value
+    }
+
+    private static func callbackScript(_ method: String, requestID: String, payload: String) -> String {
```
bkunat, line 543: An iframe can load a different page while Swift still holds the old frame reference. Before sending a location, we check that the original document is still there. The receiving script also confirms that the callback still exists; otherwise, we cancel the subscription instead of continuing to send updates.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/jsSources/geolocation.js
```diff
@@ -0,0 +1,702 @@
...
+        ? callThen(importHMACKey("raw", encodeText(capability), { name: "HMAC", hash: "SHA-256" }, false, ["sign"]), undefined, () => null)
+        : Promise.resolve(null);
+    const numberToString = globalThis.Number.prototype.toString;
+    const noncePattern = /^[0-9a-f]{32}$/;
+    const testPattern = globalThis.RegExp.prototype.test;
+
+    // Consume security values inside captured callbacks. Even await can call a
+    // page-replaced then when the page changes Promise.prototype.constructor.
+    const signSandboxProbe = (value, completion) => {
+        callThen(hmacKey, (key) => {
```
dus7: Native `Promise.prototype.then` reads the receiver’s constructor through `SpeciesConstructor`.  A page can replace that property with a getter, even though we captured the original `.then`:
```
  const OriginalPromise = Promise;
  const capturedPromises = [];

  Object.defineProperty(Promise.prototype, "constructor", {
      configurable: true,
      get() {
          capturedPromises.push(this);
          return OriginalPromise;
      }
  });
```

When the next sandbox signing operation calls `callThen(hmacKey, ...)`, this getter captures the private `hmacKey` promise itself. The page can then restore the original constructor and attach its own callback to the captured promise to obtain the `CryptoKey`. `extractable: false` prevents exporting the key’s bytes, but code which captured the key can still use it to sign forged sandbox verdicts.

The website could obtain the signing key and create a valid-looking message saying _this frame is not sandboxed_, even when it is. Combined with the forged message issue, that could let a sandboxed iframe, coming from the same origin, receive location when the bridge intends to deny it.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/jsSources/geolocation.js
```diff
@@ -0,0 +1,702 @@
...
+        }
+        return null;
+    };
+
+    let finishSandboxProbe;
+    let sandboxProbeTimeout;
+    // Missing or unverifiable parent replies deny access after the timeout.
+    const sandboxVerdict = parentWindow === globalThis
+        ? Promise.resolve(hasOpaqueOrigin)
+        : new Promise((resolve) => {
```
dus7: Sandboxing can be also applied through a CSP response header. For example, `Content-Security-Policy: sandbox allow-scripts allow-same-origin` leaves the document sandboxed but preserves its origin, so `hasOpaqueOrigin` is `false` here.

That would let a top-level page pass this sandbox check and potentially receive location using an existing site grant, despite our intended denial for sandboxed documents. I think this will need to be checked on native side, not sure what's your plan here.

dus7 (approved):
Tested for regressions and it looks fine. Just two comments, which I leave to your judgement.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/jsSources/geolocation.js
bkunat: That's valid 
Fixed in 75390dc5d0

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/jsSources/geolocation.js
bkunat: WebKit’s public response API cannot reliably attribute CSP headers to a specific subframe document. I’m keeping the current iframe support and leaving this open for follow-up.

PR #6749, opened 2026-09-08
iOS: Offer the page you navigated to inside a contextual chat

SabrinaTardio:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204186595873227/task/1218089383264832?focus=true

### Description
With auto-attach off, the chip now offers the page you just navigated to, drawn as a dashed outline to read as an offer rather than an attachment: tap it to attach that page, or dismiss it to keep the conversation as it is. Dismissing opts out of that page only, and the next navigation offers again the same shape auto-attach already had. The offer appears when the sheet opens as well as on navigation, so a chat restored after relaunch gets one too.

Behind `contextualPagePlaceholder`.

### Testing Steps

1. Enable the `contextualPagePlaceholder` feature flag, and turn Duck.ai auto-attach off in settings
2. Open a web page → open the Duck.ai contextual sheet → send a prompt and let it answer
3. Navigate the page underneath to a different one → re open the sheet -> the chip offers "Ask About <new page title>" with a dashed outline
4. Tap the offer → the page attaches and the chip becomes a normal attached chip
5. Send a prompt → the answer takes the newly attached page into account
6. Navigate again → a fresh offer appears → dismiss it with the X → the chip returns to the placeholder and the conversation is unchanged
7. Navigate once more → the offer appears again 
8. With a chat in progress, force-quit and relaunch → open the sheet on that tab → the conversation is restored and the current page is offered
9. Turn auto-attach back on → navigate → the page attaches on its own and no offer is shown

### Impact

Medium — a new affordance inside an existing chat, behind a flag. Nothing is attached without a tap, and the conversation itself is untouched.

#### What could go wrong?

- The offer attaches the page without being asked → tapping is the only path to attachment; dismissing and ignoring both leave the chat as it was.
- One dismissal silences offers for the rest of the session → the opt-out is cleared on any navigation, covered by a session-state test.
- The offer appears when auto-attach is on, duplicating it → the trigger is shared with auto-attach and the offer only exists when it is off.

### Quality Considerations

- The offer and the attached chip share one view with a state on it, so the two can't drift apart visually or in size.
- The restore now happens before context is collected, since the offer depends on knowing a chat exists — a chat restored after relaunch previously got no offer at all.
- Nothing is persisted: the offer is session state, so it does not survive a tab restore beyond being re-offered for the current page.

---

> [!NOTE]
> **Medium Risk**
> Changes contextual chat page-context delivery and UTI chip state during navigation; gated by a remote flag and requires explicit tap to attach, but touches session/coordinator paths used on every contextual navigation.
> 
> **Overview**
> Adds a **flagged** (`contextualPagePlaceholder`) flow for contextual Duck.ai when **auto-attach is off** and a chat is already active: after page collection, the session **offers** the navigated page instead of attaching it, delivered to the UTI via `.utiSuggestedContext` (replacing the old hidden attach-affordance path).
> 
> The page-context chip gains a **`.suggested`** state (dashed outline, favicon, **Ask About %@** copy via `UserText.askAboutPage(title:)`), with tap-to-accept and X-to-dismiss (per-URL decline cleared on navigation). The chip view model now uses **optional state** (`nil` = hidden) rather than a visible placeholder; accept/dismiss wire through session state and coordinator restore order is adjusted so cold restores establish `hasActiveChat` before collection can offer the current page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dee30f77c807c9cf2e2d6a578a05b08d0fa0b4bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jaceklyp:
Agentic review pass (Claude Code) plus my own reading and checks on the simulator.

### Design — can we iterate on the offer pill before this ships?

It reads as *attached*, not *offered*: same size, shape and slot as the attached chip, only a dashed stroke apart (in code, `accentPrimary` at 16% over a 40% fill). And an `X` on something not yet attached isn't a real action — ignoring the offer does the same thing, and it invites the mis-tap below. If it can't land in this PR that's fair, but I'd rather flag it now than at ship review — where I will raise it if it ships as-is. Happy to pair on alternatives.

### Confirmed on the simulator

1. **A mis-tap ~4pt right of the `X` attaches the page.** `AIChatContextChipView.swift:203-205` — the exclusion is the button's exact 32×32 bounds while the accept region is the whole pill (and 32pt is under the 44pt HIG minimum). Aiming for "no thanks" gets "yes please", which matters given the feature's premise. Fix: `point(inside:)` claiming 44pt, with the chip's exclusion rect derived from the same constant.
2. **Dismiss doesn't stick** — dismiss, close the sheet, reopen with no navigation, and the same page is offered again. `AIChatContextualChatSessionState.swift:335-339` — `dismissSuggestedContext` records nothing, so `shouldCollectPage` still passes. `userDowngradedToPlaceholder` is the existing pattern; a `declinedOfferURL` cleared in `notifyPageChanged` would give the "that page only" behaviour the PR describes.

### Worth a look (source only, not reproduced)

- **Nil-collect detach**: `:613-627` is byte-identical to base, so `4b7c1270fc`'s guard was reverted by `f957586f4a`. Deliberate? An empty collect on navigation looks like it detaches a page the user chose, which contradicts the sticky-context comment at `:673`.
- **The flag gates only the trigger** — `shouldOfferPageContext` (`:533-538`) and the delivery (`:971-977`) don't check it, so turning auto-attach off mid-collect renders the offer with the flag off.
- `shouldOfferPageContext()` at `:637` is called without a URL, so the same-page suppression the triggers apply is absent at delivery.
- **Restore hoist**: `shouldCollectSignalsOnly` needs `!hasActiveChat`, so a cold restore no longer pushes `signalsOnlyPayload` to the frontend — flag-off too.

### Tests

`6c323f1eef` / `922f01736e` deleted five. Gone with them: VoiceOver activate (nothing else calls `accessibilityActivate()`), the navigation chip-clear emit, and the nil-probe-preserves-attachment case — the exact behaviour the last two commits flip-flopped on. Two others cover `deliverToUTIChip`, which this PR doesn't touch.

Bugbot's "failed offer probe skips navigation cleanup" and the VoiceOver fix both check out, as does the "one view, one state" claim.

jaceklyp (approved):
Approved, assuming the things I flagged are fixed :) Thanks, Sabrina!

PR #1065, opened 2025-06-17
PIR iOS: UI support: Various UIViewModel helper functions

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205142657210376/task/1210571788470656?focus=true
Tech Design URL:
CC:

### Description
Adds various helper functions to be used by the iOS view model (and later the macOS view model)

### Testing Steps

1. Smoke test PIR on iOS
2. Smoke test PIR on macOS

### Notes to Reviewer

---
###### Internal references:
[Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f) | [Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552) | [Tech Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #6768, opened 2026-09-09
Integrate on-site geolocation with browser tabs

bkunat:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218210418373760
Tech Design URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1216242282222619

### Description

Connects on-site geolocation to browser tabs, including script installation, frame attribution, navigation, background activity, and teardown. Adds location prompts and recovery, feature-flag transition handling, and request-routing tests, while preserving denial of combined camera/microphone requests when either permission is blocked.

### Testing Steps

This PR connects geolocation to browser tabs. Location controls in the permission sheet and Settings arrive in the next PR, so use page requests and permission prompts for these checks.

Enable `sitePermissions` and use a physical device with Location Services enabled. Allow DuckDuckGo location access in iOS Settings, or grant it during the first request if it has not been requested yet. Start with no saved decisions for the test hosts. Use the [Permissions API test page](https://privacy-test-pages.site/features/permissions-api.html), [a second host](https://good.third-party.site/features/permissions-api.html), and [a third host](https://allowlisted.third-party.site/features/permissions-api.html) for separate site decisions.

1. On the first host, tap **Query** in the geolocation row → the state is `prompt`, without a permission dialog. Tap **Request geolocation** and choose **Allow Once**. If iOS asks for location access, allow it → the page reports a position.
2. Reload and request location again → the site prompt returns. Choose **Allow While Using Site**, then reload and request again → the request succeeds without another site prompt, and querying geolocation reports `granted`.
3. On the second host, request location and choose **Never Allow** → the request is denied. Reload and repeat → it remains denied without another site prompt, and querying geolocation reports `denied`.
4. For the allowed host, disable DuckDuckGo's location access in iOS Settings, then return and request location → access is denied and a reminder appears. Tap **Change Permissions** → DuckDuckGo's iOS Settings opens. Enable location access, return, and make a fresh request → it succeeds.
5. Open the [repeated-location page](https://privacy-test-pages.site/features/geolocation.html) on the allowed host. Switch tabs or background the app for a few seconds, then return → the page shows a gap in location updates and resumes. Navigate away → no old location prompt or result appears on the new page.
6. Open the [iframe test page](https://privacy-test-pages.site/features/iframe-permissions.html) and tap **Geolocation** in the Cross-origin panel → the request is denied without a site prompt, even though the top-level host is allowed.
7. If Fire Tabs are available, request location on the third host in a Fire tab and choose **Allow While Using Site**. Close the Fire tab, visit that host in ordinary browsing, and request again → the site prompts; the Fire-tab choice has not been saved.
8. Disable `sitePermissions` and open a fresh page → location, camera, and microphone requests use the existing WebKit permission flow, and permission-management entries are hidden. Re-enable the flag and reload a host without a saved decision → app-managed prompts return.

### Impact

High — browser request routing determines which page or frame can receive permission results.

### Quality Considerations

- Location requests and replies must stay with the requesting page; navigation or tab closure must not leave old prompts or callbacks active.
- A saved site allowance must not override an iOS denial, and Fire-tab choices must not become persistent grants.
- Disabling the feature must preserve the existing WebKit permission flow on newly loaded pages.

---
###### Internal references:


iOS/LocalPackages/SitePermissions/Package.swift
```diff
@@ -47,6 +48,7 @@ let package = Package(
         .target(
             name: "SitePermissions",
             dependencies: [
+                .product(name: "RawStructuredFieldValues", package: "swift-http-structured-headers"),
```
bkunat, line 51: I used Apple’s package for parsing the HTTP Permissions-Policy header because maintaining the structured-header grammar ourselves was becoming complicated. We depend only on its low-level RawStructuredFieldValues product, while SitePermissions still interprets the permission allowlists. Keeping the dependency also avoids maintaining a copied parser and manually applying upstream fixes.

I'm not sure if we have an official process for adding new dependencies. This SPM package has a Apache-2.0 license. 

I've noticed that previously we've added swift-algorithms with no additional approval steps: https://github.com/duckduckgo/apple-browsers/pull/1065

iOS/DuckDuckGo/AppLifecycle/AppDependencyProvider.swift
```diff
@@ -181,16 +185,18 @@ final class AppDependencyProvider: DependencyProvider {
...
             self.contentScopeExperimentsManager = defaultFeatureFlagger
-            featureFlagger = defaultFeatureFlagger
 
             // Applied after DefaultFeatureFlagger.init, which clears local overrides for non-internal users.
             // Writing overrides afterwards keeps them intact for UI test mode where allowOverrides also returns true.
             LaunchOptionsHandler().applyUITestOverrides(
                 featureFlagOverrideStore: featureFlagOverrideStore,
                 configRolloutStore: .standard
             )
+            let sessionFeatureFlagger = SessionFeatureFlagger(base: defaultFeatureFlagger)
```
bkunat: I separated the app’s effective flags from the live values used by debug editors. Site permissions stays fixed for the launch because removing an injected script does not remove it from documents WebKit has already loaded or cached. This keeps existing documents and newly opened tabs consistent, while debug settings can still display changes intended for the next launch. Other feature flags continue updating normally.

iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -766,6 +766,7 @@ class TabViewController: UIViewController {
     var sitePermissionsDependenciesProvider: @MainActor () -> SitePermissionsDependencies?
 
     let sitePermissionsState = SitePermissionsState()
+    var sitePermissionsNavigationTimeout: TimeInterval = 10
```
bkunat, line 773: I added this 10-second limit because the geolocation bridge needs to be installed before the first document starts running, but an installation problem should not leave navigation waiting forever. If the deadline expires, we cancel the navigation and show the existing load-error page for a main-frame request, allowing the user to retry with Refresh. The limit applies only when site permissions is enabled, and unrelated asset updates cannot extend it. Stopping or replacing navigation also cancels the wait so an old timeout cannot overwrite the next page.

SharedPackages/BrowserServicesKit/Sources/BrowserServicesKit/ContentScopeScript/UserContentController.swift
```diff
@@ -128,14 +128,17 @@ final public class UserContentController: WKUserContentController {
...
     @MainActor
-    private let scriptMessageHandler = PermanentScriptMessageHandler()
+    private let scriptMessageHandler: PermanentScriptMessageHandler
 
     /// if earlyAccessHandlers (WKScriptMessageHandlers) are provided they are installed without waiting for contentBlockingAssets to be loaded if.
+    /// Opt into `replyToUnavailableHandlers` to reject late replies after handler teardown; existing callers retain their behavior.
     @MainActor
-    public init<Pub, Content>(assetsPublisher: Pub, privacyConfigurationManager: PrivacyConfigurationManaging, earlyAccessHandlers: [UserScript] = [])
+    public init<Pub, Content>(assetsPublisher: Pub, privacyConfigurationManager: PrivacyConfigurationManaging, earlyAccessHandlers: [UserScript] = [],
+                              replyToUnavailableHandlers: Bool = false)
```
bkunat, line 137: I made this behavior opt-in because this controller is shared across platforms. If a message reaches the permanent router after its script handler has disappeared, we now return an explicit bridge error rather than taking the assertion path without a reply. Only iOS tabs with site permissions enabled opt in; existing callers keep their behavior. Requests already dispatched to the geolocation provider still use its own cancellation handling.

iOS/DuckDuckGo/UserScripts.swift
```diff
@@ -234,18 +240,23 @@ final class UserScripts: UserScriptsProvider {
     
     @MainActor
     func loadWKUserScripts() async -> [WKUserScript] {
-        return await withTaskGroup(of: WKUserScriptBox.self) { @MainActor group in
-            var wkUserScripts = [WKUserScript]()
-            userScripts.forEach { userScript in
+        return await withTaskGroup(of: (Int, WKUserScriptBox).self) { @MainActor group in
```
bkunat, line 243: I added the original array index because the task group returns scripts as they finish preparing, which can change their installation order. That matters for geolocation: the page adapter starts registering its frame immediately, and registration needs the isolated-world policy script to be available. Keeping the index lets us restore the intended order without making script preparation sequential. We only apply that ordering when geolocation is present, preserving the existing behavior otherwise.

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/UI/SitePermissionDialogView.swift
```diff
@@ -60,6 +60,13 @@ public struct SitePermissionDialogView: View {
                         }
                     }
                     title
+                    if let body = viewModel.body {
+                        Text(body)
+                            .font(.body)
```
dus7: Please use font style modifiers from DRK whenever possible: `daxBodyRegular`

iOS/LocalPackages/SitePermissions/Sources/SitePermissions/PermissionsPolicy.swift
```diff
@@ -0,0 +1,110 @@
...
+    public var blockedMediaTypes: Set<SitePermissionType> {
+        Set([SitePermissionType.camera, .microphone].filter { type in
+            guard case .innerList(let list) = directives[type.rawValue] else { return false }
+            return list.bareInnerList.isEmpty
+        })
+    }
+
+    // https://www.w3.org/TR/permissions-policy/#allowlists uses CSP source-expression matching.
+    // This evaluates HTTP(S) response origins, not arbitrary resource URLs or inherited frame policies.
+    private static func sourceExpression(_ expression: String, matches pageURL: URL?) -> Bool {
```
dus7: This function looks quite complex to read and understand. There's a lot of static inline values and no vertical spacing. I think it could benefit from splitting into smaller chunks as it's basically doing 5 jobs: parse the origin, handle scheme form, split expression, validate and match.

From the top of my head:
* extract constants and name them,
* prepare indexed parts like parts[i], hostAndPath[n] so they don't need to be repeated
* separate into smaller functions or at least add some space between logical parts

iOS/DuckDuckGo/TabViewController+SitePermissions.swift
```diff
@@ -627,20 +723,190 @@ extension TabViewController {
...
+    func makeGeolocationSitePermissionContext(for frame: GeolocationFrame) -> SitePermissionRequestContext? {
+        guard !sitePermissionsState.isClosed,
+              frame.isAssociated(with: webView),
+              !isLinkPreview,
+              !isError,
+              !sitePermissionsState.isCommittedGeolocationPolicyBlocked,
+              let committedURL = sitePermissionsState.committedMainFrameURL,
+              let topLevelSite = currentSitePermissionKey(),
+              SitePermissionSecurityOrigin(frame.securityOrigin).isPotentiallyTrustworthy,
+              Self.isSameOrigin(frame.securityOrigin, as: committedURL) else {
```
dus7: This looks almost the same as the content of `isMediaCaptureAllowed(for:)`. Maybe let's rename it to e.g. `isSameOriginAsCommittedMainFrame` and use it in both places?

I think you could use an agent to reduce duplication in general as I'm seeing a lot of similar code operating on repeated information.

iOS/DuckDuckGo/TabViewController.swift
```diff
@@ -3534,26 +3558,85 @@ extension TabViewController: WKNavigationDelegate {
...
+                await userContentController.awaitContentBlockingAssetsInstalled()
+                rulesCompilationMonitor.reportTabFinishedWaitingForRules(tabModel.uid)
+                completion(true)
+            }
+            return true
+        }
 
-            await MainActor.run(body: completion)
+        // Geolocation must be installed before the first document, including SERP and content-blocking-off loads.
+        // Track these additional waits so tab teardown resolves WebKit's outstanding decisions.
```
dus7: We should limit the usage of words like "track". Maybe use "monitor" instead?

iOS/LocalPackages/SitePermissions/Package.swift
dus7: I wouldn't treat `swift-algorithms` as a precedence. Asked on MM about that. This looks like a blocker unless we agree that there's no additional review required to add this.

dus7:
When doing the first step, query on geolocation before and after requesting permission shows an error: `The PermissionStatus.name getter can only be used on instances of PermissionStatus`.
In that case permissions for site are showing the following:
iPhone 24-09-2026 at 12 18 24

Next steps failed for me as well.

PR #6772, opened 2026-09-10
Add Duck.ai browser tools bridge: session, discovery, and invocation

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1218275233507705
Tech Design URL: https://app.asana.com/1/137249556945/project/1148564399326804/task/1218275233507701
CC:

### Description

First of four PRs implementing the Duck.ai browser tools. This PR covers the session handshake, tool discovery, and invocation, which is enough for one working tool. 
Screenshot 2026-09-10 at 13 10 35

See [the tech design](https://app.asana.com/1/137249556945/project/1148564399326804/task/1218275233507701) for more info

### Testing Steps

1. Open a couple of tabs in one window
2. Open browser tools testing panel: Debug ▸ AI Chat ▸ Browser Tools Panel…
3. Click `tools/list` → fails with `not_initialized`, since no session exists yet
4. Click `initialize` → returns the protocol version and server info; the header now reads `awaiting notifications/initialized`
5. Click `notifications/initialized` → header in the testing panel now shows `initialized`
6. Click `tools/list` → returns the `switchToTab` descriptor
7. Click `Show tab IDs`, 
8. Copy one tab (not the current one) and paste it into the arguments field (`{"tabId": "4CDD4DC3-3F95-41B4-8D65-1FD18A470B3A"}`)
9. Click `Call` → **the browser switches to that tab,** and the reply echoes your call id with `isError: false`

### Impact

Low. No user-facing change: the feature is internal-only and off by default, and no front end can reach it.

#### What could go wrong?

- The BrowserServicesKit push change affects every subscription event on both platforms → mitigated by tests that execute the generated JavaScript for identifier, slashed, and quote-containing names; bracket notation is equivalent for every name shipping today.
- A tool could act on the wrong window, or reach across the Fire boundary → mitigated by scoping to the owner tab's window and refusing burner windows, both tested.

### Quality Considerations

Nothing leaves the device. The only tool switches between tabs the user already has open, and reports no page content.

Fire windows refuse every tool before it runs, with the same neutral `unavailable` a disabled feature gives — no error names Fire. Note that `tools/list` still advertises the catalogue there, matching the Windows browser: listing everything and then refusing every call is a pair no other state produces, so the situation remains inferable. That gap is deliberately kept identical across platforms rather than fixed on one, and is being raised with the Windows team to close on both.

---

> [!NOTE]
> **Medium Risk**
> Adds a new native↔web path that can change browser UI (tab focus) and changes subscription JS generation for all platforms; mitigated by internal-only flags, session gating, Fire/window scoping, and tests.
> 
> **Overview**
> Introduces a **Duck.ai browser tools bridge** on macOS: MCP-shaped messages (`initialize`, `notifications/initialized`, `tools/list`, `tools/call`) over the existing `aiChat` user-script channel, with per–owner-tab sessions, remote-config gating, and a pluggable tool catalog/invoker in shared **AIChat**.
> 
> Native config gains **`supportsBrowserTools`** so the web app only opens a session when the bridge exists. Privacy config and macOS feature flags add **`aiChatBrowserTools`** plus per-tool subfeatures; content-scope injection **excludes** that feature so it does not break the shared bridge. The first shipped tool is **`switchToTab`**, scoped to the chat window via an owner-tab/window token and refused in Fire windows as generic `unavailable`.
> 
> Wiring includes app-wide **`AIChatBrowserToolsService`**, handler routing in **`AIChatUserScript`**, and a DEBUG **Browser Tools Panel** for manual handshake/testing. **`SubscriptionEvent.toJS`** now uses bracket notation and JSON-escaped handler names so names like `notifications/initialized` work; tests execute the generated JS.
> 
> Unit tests cover session lifecycle, catalog/invoker gating, wire JSON contracts, and message routing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2618ccd11dd399518d52cdb5b28ae0b559cb8eb1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/macOS/DuckDuckGo/AIChat/AIChatDebugMenu.swift
+++ b/macOS/DuckDuckGo/AIChat/AIChatDebugMenu.swift
@@ -60,8 +60,33 @@ final class AIChatDebugMenu: NSMenu {
             NSMenuItem.separator()
 
             storageServerMenuItem
+
+#if DEBUG
+            NSMenuItem.separator()
+
+            NSMenuItem(title: "Browser Tools Panel…", action: #selector(openBrowserToolsPanel))
+                .targetting(self)
+#endif
         }
     }
+
+#if DEBUG
+
+    // MARK: - Browser Tools
+
+    private var browserToolsPanel: BrowserToolsDebugPanel?
+
+    /// Stands in for the Duck.ai front end, which cannot drive the browser tools bridge yet.
+    @MainActor
+    @objc func openBrowserToolsPanel() {
+        let panel = browserToolsPanel
+            ?? BrowserToolsDebugPanel(windowControllersManager: NSApp.delegateTyped.windowControllersManager)
+        browserToolsPanel = panel
+        panel.showWindow(nil)
+        panel.window?.makeKeyAndOrderFront(nil)
+    }
+
+#endif
 
     // MARK: - Duck.ai Usage Warnings
 
--- a/macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
+++ b/macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
@@ -1218,8 +1218,8 @@ extension AIChatUserScriptHandler {
 
     /// MCP `tools/list`.
     ///
-    /// A listing failure is reported as a top-level `error` token, unlike `tools/call`, which
-    /// carries failures inside the MCP result envelope.
+    /// A listing failure is a top-level `error` token, unlike `tools/call`, which carries failures
+    /// inside the MCP result envelope.
     @MainActor
     func mcpToolsList(params: Any, message: UserScriptMessage) async -> Encodable? {
         guard let ownerTabID = ownerTabID(for: message),
@@ -1227,6 +1227,13 @@ extension AIChatUserScriptHandler {
               session.isInitialized else {
             return BrowserToolsListResponse(failure: .notInitialized)
         }
+
+        // A Fire window advertises nothing, exactly as a disabled feature does. Listing the tools
+        // and then refusing every call would be a pair no other state produces, telling the front
+        // end it is in a Fire window.
+        let isBurner = AIChatTabPickerSource.originTabCollectionViewModel(for: message.messageWebView,
+                                                                          in: windowControllersManager)?.isBurner ?? true
+        guard !isBurner else { return BrowserToolsListResponse(tools: []) }
 
         return BrowserToolsListResponse(tools: browserTools.catalog.enabledTools.map { $0.descriptor() })
     }
@@ -1250,9 +1257,12 @@ extension AIChatUserScriptHandler {
             return InvokeBrowserToolResponse(callId: request.callId, result: .failure(.notInitialized))
         }
 
+        let originCollection = AIChatTabPickerSource.originTabCollectionViewModel(for: message.messageWebView,
+                                                                                  in: windowControllersManager)
         let context = BrowserToolCallContext(
             ownerTabID: ownerTabID,
-            isBurner: AIChatTabPickerSource.isBurner(ownerTabID: ownerTabID, in: windowControllersManager),
+            ownerWindowToken: originCollection.map(AIChatTabPickerSource.windowToken(forCollection:)),
+            isBurner: originCollection?.isBurner ?? true,
             supportsElicitationForm: session.supportsElicitationForm
         )
 
--- a/macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
+++ b/macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
@@ -1228,13 +1228,10 @@ extension AIChatUserScriptHandler {
             return BrowserToolsListResponse(failure: .notInitialized)
         }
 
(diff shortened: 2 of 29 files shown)


macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -885,6 +889,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         pinnedTabsManagerProvider.tabsPreferences = tabsPreferences
         pinnedTabsManagerProvider.windowControllersManager = windowControllersManager
 
+        aiChatBrowserToolsService = AIChatBrowserToolsService(featureFlagger: featureFlagger,
+                                                             windowControllersManager: windowControllersManager)
```
jotaemepereira: 🔍 nit: the `windowControllersManager: windowControllersManager` line is not aligned.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -279,6 +279,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     let aiChatConversationSourceHandler = AIChatConversationSourceHandler()
     let aiChatMenuConfiguration: AIChatMenuVisibilityConfigurable
     let aiChatSessionStore: AIChatSessionStoring
+
+    /// App-wide because a sidebar and its host tab share one session, while each web view gets
```
jotaemepereira: 🔍 This comment could be removed. Thoughts?

macOS/DuckDuckGo/AIChat/AIChatDebugMenu.swift
```diff
@@ -60,9 +60,34 @@ final class AIChatDebugMenu: NSMenu {
...
         }
     }
 
+#if DEBUG
+
+    // MARK: - Browser Tools
+
+    private var browserToolsPanel: BrowserToolsDebugPanel?
+
+    /// Stands in for the Duck.ai front end, which cannot drive the browser tools bridge yet.
```
jotaemepereira: 🔍 This message can be removed.

macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
```diff
@@ -1170,3 +1180,91 @@ extension AIChatUserScriptHandler: AIChatMetricReportingHandling {
...
+            return InvokeBrowserToolResponse(callId: Self.callID(fromRawParams: params), result: .failure(.invalidRequest))
+        }
+
+        guard let ownerTabID = ownerTabID(for: message),
+              let session = browserTools.sessions.session(forOwnerTabID: ownerTabID),
+              session.isInitialized else {
+            return InvokeBrowserToolResponse(callId: request.callId, result: .failure(.notInitialized))
+        }
+
+        let originCollection = AIChatTabPickerSource.originTabCollectionViewModel(for: message.messageWebView,
```
jotaemepereira: 🔴 This carries an important bug that we should fix right now. It will not trigger now with the debug tools, but it will fire once the FE implementation is done.

Duck.ai can run in the background while the user does something else in another window or tab. Inside `originTabCollectionViewModel` we fall back to `lastKeyMainWindowController` when the `webView.window` is nil, and that could happen if you have the following:
- Window A with two tabs: Duck.ai and another thing, Duck.ai not selected.
- Window B, currently focused.

In that scenario, Duck.ai still runs JS, so it can send messages. It will send the tool-call action, and instead of sending Window A's originCollection, it will send Window B's collection, which, if it is a Fire Window, could cause privacy issues.

A solution is not to use the `originTabCollectionViewModel` and use a new method that uses the already calculated `ownerTabId` to get the view models.

```swift
// AIChatTabPickerSource
/// The window a tool call is scoped to. Resolved from the owner tab rather than from the key
/// window: a backgrounded chat's web view has no window, and guessing lands on someone else's.
/// A shared pinned tab is in every window's collection, so it has no owner window — refuse.
static func collection(containingTabID tabID: TabIdentifier,
                       in windowControllersManager: WindowControllersManagerProtocol) -> TabCollectionViewModel? {
    let matches = windowControllersManager.allTabCollectionViewModels.filter {
        $0.indexInAllTabs(where: { $0.uuid == tabID }) != nil
    }
    return matches.count == 1 ? matches[0] : nil
}
```

Sorry for the long post; it was **not** AI-generated, and I couldn't make it shorter, ha.

jotaemepereira (changes requested):
@tomasstrba I did a first pass today and I found a blocker that we should fix. I'm going to do more testing tomorrow.

macOS/DuckDuckGo/AIChat/UserScript/AIChatUserScriptHandling.swift
tomasstrba: The post was very helpful, thank you @jotaemepereira!

tomasstrba:
@jotaemepereira, thanks again! This is ready for re-review:
- Blocker fixed
- Nits done

jotaemepereira (approved):
Looks great 👍🏼

PR #6775, opened 2026-09-10
[macOS] Migrate First-Time Quit Survey to the promo queue

rachelmcr:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218311516923100?focus=true

### Description

Moves the First-Time Quit Survey onto the macOS Promo Queue as an external promo. By migrating this as an external promo, it retains full control over its presentation, prevents collisions with other promos (two promos appearing at the same time), and ensures a cooldown between this and a following promo.

Notable changes:

- If another promo is already visible, the survey will be shown and the other promo will be retracted by the Promo Queue (preventing a collision).
- Once the survey has been dismissed, a 1-hour global cooldown will be applied, suppressing any other user-initiated promos when the app is relaunched. (Only applies if the app is restarted within that 1-hour cooldown.)

### Testing Steps

1. Debug → UI Triggers → enable Always Show First-Time Quit Survey
2. Debug → Promo Queue → `default-browser-and-dock-banner` → Force Show (this will show the persistent "Set as Default/Add to Dock" banner promo)
3. Quit the app and confirm the banner is auto-dismissed and the survey is shown.
4. Close the survey and confirm the app quits.
5. Relaunch the app.
6. Debug → Promo Queue → confirm the promo history was recorded and cooldown applied in the promo queue. You may need to open the debug menu twice for the menu to refresh with the promo history:
   * `first-time-quit-survey` shows as `dismissed`
   * user-initiated promos such as `autofill-toolbar-pinning` and `bookmark-toolbar` show as `on cooldown (global)`

### Impact

#### What could go wrong?
* Other user-initiated promos could be suppressed more than expected. Mitigation: Promo metrics are being monitored for anomalies during/after migrations.

---
###### Internal references:

---

> [!NOTE]
> **Medium Risk**
> Changes app termination and promo-queue coordination; mis-timed visibility or cooldown rules could suppress other user-initiated promos or affect quit-survey presentation when the promo queue flag is on.
> 
> **Overview**
> Registers the **First-Time Quit Survey** on the macOS promo queue as an external user-initiated app modal (`first-time-quit-survey`), so quit-time presentation participates in collision handling and promo history/cooldowns.
> 
> A new **`QuitSurveyPromoObserver`** publishes visibility to `PromoService`; **`AppDelegate`** calls `reportVisible()` / `reportHidden()` around `QuitSurveyPresenter` when the promo queue is enabled (otherwise behavior is unchanged). **`PromoService`** now waits for conflicting internal promos to finish hiding before invoking **`promoServiceDidApplyVisibility`**, and **`recordResultAndCleanup`** returns the hide task so external visibility reporting can await retraction. Debug **Reset Quit Survey** also **`undismiss`**es the quit-survey promo with history cleared.
> 
> Tests cover the observer handshake and the case where conflicting promos are hidden before visibility reporting completes; **`MockPromoHistoryProvider`** is moved to shared promo test mocks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc43dea1ba7e48aa05339da5156dac58f5eca7f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/Menus/MainMenuActions.swift
```diff
@@ -891,6 +891,7 @@ extension AppDelegate {
     @objc func resetQuitSurveyWasShown(_ sender: Any?) {
         let persistor = QuitSurveyUserDefaultsPersistor(keyValueStore: NSApp.delegateTyped.keyValueStore)
         persistor.hasQuitAppBefore = false
+        NSApp.delegateTyped.promoService?.undismiss(promoId: PromoServiceFactory.quitSurveyPromoID, clearHistory: true)
```
jotaemepereira: 🔍 The `NSApp.delegateTyped` call can be removed, we already have the `promoService` injected here.

macOS/DuckDuckGo/Application/AppDelegate.swift
```diff
@@ -1877,8 +1881,23 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+                        historyCoordinating: historyCoordinator,
+                        faviconManaging: faviconManager
+                    )
+
+                    guard let promoService else {
+                        await presenter.showSurvey()
+                        return
+                    }
+
+                    quitSurveyPromoObserver.reportVisible()
```
jotaemepereira: I'm having second thoughts about this approach. This reports that the survey is visible and then presents it, but it doesn't wait for the other promos to hide (it waits 0.5 seconds, but what happens if it needs more?).

⚠️ AI-Generated (I gave it a look as well): [codex/quit-survey-promo-handshake](https://github.com/duckduckgo/apple-browsers/tree/codex/quit-survey-promo-handshake)

This branch makes `reportVisible` and `reportHidden` async, so when showing the survey, we make sure other UIs are closed.

jotaemepereira (approved):
@rachelmcr I've tested it and looks good. I made suggestion that I would like your eyes on. I've worked on it with AI after seeing that `reportVisible` and `reportHidden` are fired without waiting for any return parameter.

The solution removes the need for the 0.5 magic number of waiting for other UIs to be closed.

macOS/DuckDuckGo/Application/AppDelegate.swift
rachelmcr: That's a really nice improvement, thanks! I reviewed the changes and added that commit to this branch. 👍

macOS/DuckDuckGo/QuitSurvey/QuitSurveyPromoObserver.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @MainActor
+    private func reportVisibility(_ isVisible: Bool) async {
+        guard visibilitySubject.value != isVisible else { return }
+        precondition(pendingReport == nil)
+        await withCheckedContinuation { continuation in
+            pendingReport = (isVisible, continuation)
+            visibilitySubject.send(isVisible)
+        }
```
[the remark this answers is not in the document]
rachelmcr: > If promoService exists but is not subscribed yet, or the callback never arrives, quit waits forever and the survey never appears.

This is not really accurate; when `promoService` subscribes it applies the current visibility, so it will proceed at that point. The only real issue would be if it doesn't subscribe, which it does when the app becomes active and the service starts.

macOS/DuckDuckGo/Promotions/PromoService.swift
```diff
@@ -435,24 +435,38 @@ final class PromoService: @unchecked Sendable, PromoHistoryProviding {
...
             // External visibility is not gated by checkRules; retract conflicting internal promos shown earlier.
-            hideInternalPromosConflictingWithCurrentVisibility()
-        } else if externalVisiblePromoIds.remove(promoId) != nil {
-            applyResult(delegate.resultWhenHidden, toRecordFor: promoId)
+            let hideTasks = hideInternalPromosConflictingWithCurrentVisibility()
+            Task { @MainActor in
+                for task in hideTasks {
+                    await task.value
+                }
+                delegate.promoServiceDidApplyVisibility(true)
```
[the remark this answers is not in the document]
rachelmcr: This is accurate but not entirely concerning at this point:
* Only the Set as Default/Add to Dock promos do this
* The popover and banner dismiss synchronously, so on MainActor their dismissal will occur before `promoServiceDidApplyVisibility(true)` is called
* The inactive modal suspends but also blocks app quit, so it can't conflict with the survey (the user has to explicitly dismiss the modal before quitting)

An improvement here could be to make `hide()` async so the promos are definitely hidden before proceeding, but that seems like an excessive change given that it isn't an issue at this point.

PR #6816, opened 2026-09-15
[macOS] Restore debug menu icons on macOS 27 with Xcode 26.6

andremartingo:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1216562341335231/task/1217998637419259?focus=true

### Description

Restore the Feature Flag Overrides legend, category markers, and flag status icons on macOS 27 when the app is built with Xcode 26.6. The visibility opt-in now works with the older build SDK, which previously left these icons hidden.

### Testing Steps

1. Build and run the macOS Browser scheme in Debug using Xcode 26.6 on macOS 27.
2. Enable internal-user mode using Debug → Set Internal User State if needed.
3. Open Debug → Feature Flag Overrides → all five legend icons appear, and categories with overrides show a flag marker.
4. Expand a feature-flag category → each flag displays its enabled or disabled status icon.

Validated locally: the Xcode 26.6 build succeeded, and the legend, category marker, and expanded Ad Blocking status icons were visually verified on macOS 27 and captured in screenshots. Unit tests were not run.

### Impact

Low: menu icon visibility only. The shared visibility fix also applies to other menus that already explicitly request visible icons on macOS 27.

#### What could go wrong?

The macOS 27 visibility API is unavailable on older systems → the runtime call is guarded by both the macOS version and setter availability.

Replaces #6815 after renaming the source branch to use the `andremartingo/` prefix.

---

> [!NOTE]
> **Low Risk**
> Debug-only menu presentation; the macOS 27 visibility workaround is version- and selector-guarded with no data or auth impact.
> 
> **Overview**
> Restores **Feature Flag Overrides** debug menu icons on **macOS 27** when the app is built with **Xcode 26.6**, where `preferredImageVisibility` could not be set at compile time.
> 
> `FeatureFlagOverridesMenu` now routes legend, category, and flag status icons through **`withImage(..., visibleOnMacOS27: true)`** and drops duplicated `#if compiler(>=6.4)` / `preferredImageVisibility` blocks at item creation.
> 
> The shared **`NSMenuItem.withImage`** helper gains an **Xcode &lt; 6.4** path: when visibility is requested on macOS 27, it uses a runtime **`setPreferredImageVisibility:`** / KVC workaround (raw value `1`) so icons opt in to display without referencing the macOS 27 API directly; the previous behavior of clearing the image when visibility is false is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64275c1a3ca8e1432be4ffd701e2a4f11d571a7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macOS/DuckDuckGo/InternalUserDecider/FeatureFlagOverridesMenu.swift
federicocappelli, line 151: Possible pre-existing bug here, `update(_ items: [NSMenuItem])` snapshots `self.items` into a local parameter before rebuildItemsIfInternalUserStateChanged() can replace self.items, so the update loop that applies withImage() walks the stale pre-rebuild array.
This could cause to display a menu item with correct titles/submenus but no icons until the menu is closed and reopened a second time.

macOS/LocalPackages/AppKitExtensions/Sources/AppKitExtensions/NSMenuItemExtension.swift
```diff
@@ -120,8 +120,17 @@ public extension NSMenuItem {
             preferredImageVisibility = visibleOnMacOS27 ? .visible : .hidden
         }
 #else
-        if #available(macOS 27.0, *), visibleOnMacOS27 == false {
-            effectiveImage = nil
+        if #available(macOS 27.0, *) {
```
federicocappelli, line 123: The shared `withImage(...)` helper that this PR changes is called roughly 164 times across the app, but there's no unit test for this new behaviour (or the old), so this change is not only impacting a debug menu.
What about checking how the function is used across the app and creating unit tests for all cases?

macOS/LocalPackages/AppKitExtensions/Sources/AppKitExtensions/NSMenuItemExtension.swift
```diff
@@ -120,8 +120,17 @@ public extension NSMenuItem {
...
+            if visibleOnMacOS27 {
+                // Temporary workaround until we can compile with Xcode 27.
+                // Xcode 26 cannot reference this macOS 27 API directly.
+                // NSMenuItem.ImageVisibility.visible has a raw value of 1.
+                if responds(to: NSSelectorFromString("setPreferredImageVisibility:")) {
+                    setValue(1, forKey: "preferredImageVisibility")
+                }
```
federicocappelli, line 130: This makes the image visible even if it's nil. Is this the desired behaviour?

federicocappelli (changes requested):
Just a couple of questions and a possible pre-existing bug
The bug is only in the debug-menu and quite minor,  but it may be worth investigating anyway

PR #6828, opened 2026-09-16
Stop a burn from hanging on a tab that never finishes loading blank

tomasstrba:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1204912272578138/task/1213104708128639?focus=true
Tech Design URL:
CC:

### Description

Users on 1.204 reported burns that never finish: the "Deleting browsing data" dialog stays up, every page goes blank, history isn't cleared, and the fire button then does nothing for the rest of the session. With Auto-Clear on, the app re-ran the stuck burn at every launch.

Cause: before clearing anything, a burn loads `about:blank` in every open tab and waits for all of them to report back. That wait had no time limit, and it only counted a load that finished or failed normally. A tab whose page process died, or whose load failed early, was never counted — and one uncounted tab was enough to stop the burn right there. The more tabs open, the more likely it was, which is why it hit users and not developer machines with a few tabs. Closing a tab mid-burn triggered it every time, and the fire overlay let clicks through, so that was easy to do by accident.

- The wait now gives up after 5s and the burn carries on. The blank load is only hygiene ahead of the real clearing, so nothing is lost; a debug pixel records when this happens.
- A load that fails early, or a page process that dies, now counts as done.
- Each tab is counted once, by identity, so a tab reporting twice can't stand in for one that is still loading.
- The fire overlay swallows clicks instead of passing them to the page and tab bar.

Independent of #6823, which fixes a different hang that needs non-default settings to trigger. This is the default-settings case and should ship first.

### Testing Steps

Open 10–20 tabs with real pages. Temporarily edit `prepareForDataClearing(caller:)` in `macOS/DuckDuckGo/Tab/Model/Tab.swift`, replacing the line `webView.load(URLRequest(url: .blankPage))` as listed for each case, and revert between cases:

1. **Tab never reports back** — delete the `webView.load(...)` line entirely. Fire button → Delete all data → the burn finishes after ~5s and the fire button works again. (Console, subsystem `Fire`, logs "Gave up waiting for N tab(s)…".)
2. **Navigation fails before committing** — replace the line with `caller.webView(webView, didFailProvisionalNavigation: nil, withError: URLError(.cancelled))`. Burn completes immediately, no "Gave up waiting" log.
3. **Web content process dies** — replace the line with `caller.webViewWebContentProcessDidTerminate(webView)`. Same result as step 2.
4. **One tab reports twice** — replace the line with both `caller.webViewWebContentProcessDidTerminate(webView)` and then `caller.webView(webView, didFailProvisionalNavigation: nil, withError: URLError(.cancelled))`. Burn completes immediately.
5. Revert all changes and test the Fire button again. During the burn, clicks on the page and tab bar should do nothing.

### Impact

High — data clearing; the bug left users unable to clear anything for the rest of a session.

#### What could go wrong?

- 5s isn't enough on a slow machine with many tabs → the burn proceeds anyway; those tabs are torn down moments later and clearing runs regardless. The pixel shows how often this happens.
- A tab's previous navigation reports to us after the delegate swap → it's counted once, so at worst that one tab skips the blank load; it can't finish the wait early for the others.
- The overlay now blocks clicks → intended; the dialog's own controls are unaffected. ⌘W still works during a burn (menu items stay enabled) — out of scope here, and it no longer hangs anything.

---

> [!NOTE]
> **Medium Risk**
> Changes core Fire/data-clearing synchronization; a 5s timeout could proceed before every tab finishes blank navigation, though clearing still runs afterward.
> 
> **Overview**
> Fixes **Fire burns that could hang forever** while waiting for every tab to finish loading `about:blank` before data clearing.
> 
> **`TabCleanupPreparer`** now caps that wait at **5 seconds**, then continues the burn and fires **`blank_navigation_on_burn_timed_out`**. Tabs that never report back (closed mid-burn, dead web content process) no longer block the session. **Pre-commit navigation failures** and **`webViewWebContentProcessDidTerminate`** are treated as done. Completion is tracked **per `WKWebView` identity** so double callbacks from one tab cannot satisfy another tab’s slot. Starting a new preparation **releases any stuck prior wait** instead of asserting. **`FireViewController`** enables **`interceptClickEvents`** on the burn overlay so clicks no longer reach the page/tab bar during a burn.
> 
> Adds **`TabCleanupPreparerTests`** and the debug pixel definition in **`debug_fire.json5`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97388485f0504925887ba908966c417950a803df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mallexxx (approved):
LGTM
Works as described!

